PR #1, opened 2015-06-19
Adds LICENSE

snostorm:
This was copied from the https://github.com/nodejs/iojs.org repo as-is (with a slight modification to the first line to mention "original contents".)
## Open Question(s):
- [ ] MIT as previously determined for the last project? See https://github.com/nodejs/iojs.org/issues/1
- [ ] Given we do not have ownership of some of the assets which may be bundled (namely logos), do we need to modify the license to reflect this? For now this is why I used the phrasing "original contents" in the opening line.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,0 +1,23 @@
+The original contents of the new.nodejs.org repo are licensed for use as follows:
+
+"""
+Copyright io.js website contributors. All rights reserved.
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to
+deal in the Software without restriction, including without limitation the
+rights to use, copy, modify, merge, publish, distribute, sublicense, and/or
+sell copies of the Software, and to permit persons to whom the Software is
+furnished to do so, subject to the following conditions:
+
+The above copyright notice and this permission notice shall be included in
+all copies or substantial portions of the Software.
+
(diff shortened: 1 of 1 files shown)


snostorm:
_I wonder how long we can keep the trend going of our Issue #s in this project matching the same Issues in the iojs.org project ... So far we're 1 for 1!_

bnb:
I think we need to talk to whatever lawyers the Node Foundation has about the licensing/logos deal. I assume the repo will bit MIT, but the logos will probably have to live with a WG. Website/Evangelism are the obvious ones. However, I know jack sh*t about law, so we need to consult someone who knows. Is there someone from the foundation that can direct us to the right place?

bnb:
@therebelrobot I hate to keep bothering @rvagg with this stuff, but he seems to know what's going on with the Foundation and where to go/who to talk to about this stuff. 

Rod: Do you have an idea of where we can go for this? Also, is there someone else that we can/should have as a point of contact/information about the Foundation?

rvagg:
Yes, I've started discussing some of this with them but will not likely connect the dots until next week. I'm going to try and connect some of the website & evangelism teams with appropriate people in LF so they can be used as a resource, at the moment there's a bit of a barrier between the community and the formal side of the foundation that we need to work on breaking down simply by connecting people together.

Issue #9, opened 2015-06-19
Manual line breaks in README.md


bnb:
I was just writing in README.md, and I noticed that some or all of the line breaks in the file are manually done. Is this intentional? If not, I'll send a PR to fix it.

therebelrobot:
Yeah, that's weird.

PR #10, opened 2015-06-19
Fix line breaks

bnb:
Realized I can't make updates to the README and then fix line breaks afterward. /cc #9


snostorm:
I'm in a habit of manually wrapping my markdown files' text after lines get long. But it looks funny when the rest of the document isn't doing that.

Thanks @bnb

PR #14, opened 2015-06-20
Adds WG membership section to the README. For #13

snostorm:
Other members, please add yourself to the list :)

PR #15, opened 2015-06-20
Readme adds WG collaborator Tierney/@bnb

bnb:
Added myself to the README.md.


snostorm:
@bnb you can directly clone and push back to branches hosted on the repos we control (such as the one setup for #14.) Alternatively, you can target a PR against a specific branch that is not master, although in this case the first option was preferred.

Regardless, thanks for getting your information pushed up. I've closed this PR because I merged the commits in this PR in to the existing one, #14.

PR #26, opened 2015-08-07
style: make header fit on one line

evanlucas:
Reduce padding for header links to make them fit on a single line

Before:
screen shot 2015-08-07 at 8 10 20 am

After:

screen shot 2015-08-07 at 8 09 36 am

### DIFF
--- a/layouts/css/page-modules/_header.styl
+++ b/layouts/css/page-modules/_header.styl
@@ -31,7 +31,7 @@ header
     a,
     a:link,
     a:active
-        padding 0 12px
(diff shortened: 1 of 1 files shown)


fhemberger:
Thanks!

mikeal:
Is there a tool that we could use here to make this work better on mobile and for other languages that vary quite a bit in size?

fhemberger:
I'll do this manually as soon as the desktop layout is set. It doesn't make much sense at the moment as the desktop view is still changing.

PR #33, opened 2015-08-16
A small set of design changes

joarwilk:
Here's a small group of changes that improves readability across the site, I'll wait with doing more until I know it's needed right now. Would love to help out.

### DIFF
--- a/layouts/css/base.styl
+++ b/layouts/css/base.styl
@@ -20,7 +20,16 @@ h3,
 h4,
 h5
     font-weight normal
+    margin: 1em 0 10px 0
 
+h1
+    font-size 30px
+    line-height 36px
+    margin 15px 0 11px
+
+h2
+    font-size 25px
+    letter-spacing 1px
 
 a,
 a:link,
--- a/layouts/css/base.styl
+++ b/layouts/css/base.styl
@@ -3,9 +3,11 @@
 // Base styles
 body
     box-sizing border-box
-    font 300 20px/1.5 'Source Sans Pro', Arial, sans-serif
-    color #000
+    font 400 15px/1.5 "source-sans-pro", sans-serif
+    color #333333
     margin 0
+    -webkit-font-smoothing antialiased
+    text-rendering optimizeLegibility
 
 header,
 main,
--- a/layouts/partials/html-head.hbs
+++ b/layouts/partials/html-head.hbs
@@ -19,8 +19,6 @@
     <link rel="stylesheet" href="/{{site.locale}}/styles.css" media="all">
     <link rel="stylesheet" href="/static/css/prism-tomorrow.css" media="all">
 
-    <link rel="stylesheet" href="https://use.typekit.net/c/5d5ccd/1w;din-condensed-web,2,ZwM:P:n4;source-sans-pro,2,Xw0:P:n2,Xw5:P:n4,Xw6:P:n6,Xvy:P:n7/d?3bb2a6e53c9684ffdc9a98f41a5b2a62f97c6e2d19ae8171762f36a47b76e2dab9866720c764a2dd276e81e45db51eba20e3d580df710cfe124f8cdf282019abff20cf1bb6b3dcb663d7e0fb8f74c8a1a5e32f78f888e5fc5a927ddf7928f759d4f2f06ecf71e5cae0215f59a63aa05c7c4265cd6a519d74de9dcaddb62d9ddd5e6d496c703544dc3b5e575470c25983eb933efb3fadbc9412a4d6a260a007066fc23ee1d196e3c11f562ab5f8417f84fae2d5f90659f0d40a66d841865a9cf025ec205388768613d859b008d2d814834463397b7a161b555acdc04d">
-
     <script src="//use.typekit.net/mse5tqx.js"></script>
     <script>try{Typekit.load();}catch(e){}</script>
 </head>
--- a/layouts/partials/html-head.hbs
+++ b/layouts/partials/html-head.hbs
@@ -6,6 +6,8 @@
     <link rel="dns-prefetch" href="http://fonts.gstatic.com">
     <link rel="dns-prefetch" href="http://www.google-analytics.com">
 
+    <script src="//use.typekit.net/mse5tqx.js"></script>
+
     <meta name="author" content="{{ site.author }}">
     <meta name="robots" content="{{#if robots}}{{ robots }}{{else}}index, follow{{/if}}">
     <meta name="viewport" content="width=device-width, initial-scale=1.0">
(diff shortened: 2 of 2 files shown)


layouts/css/base.styl
```diff
@@ -3,9 +3,11 @@
 // Base styles
 body
     box-sizing border-box
-    font 300 20px/1.5 'Source Sans Pro', Arial, sans-serif
-    color #000
+    font 400 15px/1.5 "source-sans-pro", sans-serif
+    color #333333
     margin 0
+    -webkit-font-smoothing antialiased
+    text-rendering optimizeLegibility
```
Fishrock123: Hmmmm, I think some base styles were lost from io.js here.

I suggest:

```styl
html
  -webkit-font-smoothing antialiased
  -moz-osx-font-smoothing grayscale
  -webkit-font-variant-ligatures none
          font-variant-ligatures none
```

Please remove `text-rendering optimizeLegibility` for text that isn't fairly large. It is very slow and can negatively impact page load times.

layouts/css/base.styl
```diff
@@ -3,9 +3,11 @@
 // Base styles
 body
     box-sizing border-box
-    font 300 20px/1.5 'Source Sans Pro', Arial, sans-serif
-    color #000
+    font 400 15px/1.5 "source-sans-pro", sans-serif
+    color #333333
```
Fishrock123: I feel this *may* be taking too much contrast away for those who have issues regarding less contrasting text. See: https://github.com/nodejs/iojs.org/pull/366

Perhaps `#222`?

layouts/css/base.styl
```diff
@@ -3,9 +3,11 @@
 // Base styles
 body
     box-sizing border-box
-    font 300 20px/1.5 'Source Sans Pro', Arial, sans-serif
-    color #000
+    font 400 15px/1.5 "source-sans-pro", sans-serif
```
Fishrock123: Please keep `Arial,` here as a fallback.

layouts/partials/html-head.hbs
```diff
@@ -18,4 +18,7 @@
     <link rel="alternate" href="/{{site.locale}}/tsc-minutes.xml" title="Node.js TSC meeting minutes" type="application/rss+xml">
     <link rel="stylesheet" href="/{{site.locale}}/styles.css" media="all">
     <link rel="stylesheet" href="/static/css/prism-tomorrow.css" media="all">
+
+    <script src="//use.typekit.net/mse5tqx.js"></script>
+    <script>try{Typekit.load();}catch(e){}</script>
```
Fishrock123: What's the exact need for this? How large is the file?

joarwilk: It's for the "source-sans-pro" font

layouts/css/base.styl
joarwilk: Yeah I didnt copy everything, was just a couple of base additions. I'll move the text rendering to headers only.

layouts/css/base.styl
joarwilk: Agreed, will change.

layouts/partials/html-head.hbs
Fishrock123: Hmm, wouldn't it be better to just include the font from link into css? That would make for less potential font resizing, and less javascript. :)

joarwilk: 9.8kb size, should be fine

Fishrock123: Right but this is two html requests compared to one. :)

joarwilk: I doubt we have the license to do that, I think typekit is necessary,

Fishrock123: Naw, we should be able to, I've done it before. https://github.com/Fishrock123/Searchbeam/blob/master/src/less/fonts.less

joarwilk: Ah yeah you're right it's free to use, I'll embed it instead!

fhemberger: It should only be embedded if the file size is < 2k, otherwise two requests are more performant than embedding.
You can also load an optimized version for the user's browser from Google Fonts:
https://www.google.com/fonts#QuickUsePlace:quickUse

For mobile use, it would be best to use the font-loading API and caching the fonts to localStorage to avoid the text being invisible until the web-font has finally loaded.

layouts/css/base.styl
fhemberger: Why go below 16px as base font size?

layouts/css/base.styl
```diff
@@ -3,9 +3,11 @@
...
+    color #333333
     margin 0
+    -webkit-font-smoothing antialiased
```
fhemberger: This should only be used for light text on dark backgrounds. Otherwise this may cause blurry text on Chrome/Windows. (So you can test it on the main navigation an footer for example)

layouts/css/base.styl
Fishrock123: Hmmm, was there anything wrong with 20px as base size? Medium uses about 20px for most of everything, makes it easy to read?

fhemberger:
Closing this:
- `-webkit-font-smoothing: antialiased` decreases legibility on OS X because of the thin font.
- We don't want to decrease the base font size
- Web fonts were added via Google web fonts in the mean time.

Fishrock123:
> `-webkit-font-smoothing: antialiased` decreases legibility on OS X because of the thin font.

I think you will find a good amount people who disagree with that... is the text we are using currently too small?

fhemberger:
**Without** -webkit-font-smoothing: antialiased on Chrome/OS X:



**With** -webkit-font-smoothing: antialiased on Chrome/OS X:

layouts/partials/html-head.hbs
mikeal: We only use that font in one place on the front page which is why it's in a CSS include to google fonts on the front page. I don't know what the benefit of typekit is but if we do include it for this font we should only do so on the front page.

mikeal:
what if we increase the default weight and do `-webkit-font-smoothing: antialiased`?

Fishrock123:
> what if we increase the default weight and do `-webkit-font-smoothing: antialiased`?

Yeah I was going to suggest something like this so that we can use the below, like io.js:

Generally you'll end up with more balanced text after that from my understanding.

```styl
html
  -webkit-font-smoothing antialiased
  -moz-osx-font-smoothing grayscale
  -webkit-font-variant-ligatures none
          font-variant-ligatures none
```

fhemberger:
I'm fine with trying a different weight, let's see how it turns out. With autoprefixer in the build pipeline, we should be able to drop any custom prefixes. Is there a reason to drop ligatures?

PR #34, opened 2015-08-17
Changed package.json to work in Windows

AustinWinstanley:
In Windows, running `npm run serve` results in the error:

```
'.' is not recognized as an internal or external command, operable program or batch file.
```

Explicitly using `iojs` instead of the dot current path works in all operating systems. 

Tested in Windows, OSX, and Ubuntu.

### DIFF
--- a/layouts/css/page-modules/_footer.styl
+++ b/layouts/css/page-modules/_footer.styl
@@ -1,8 +1,7 @@
 footer
     // display flex
     margin-top 2em
-    padding 0 20px
-    background-color #404042
+    background-color $node-gray
 
     a,
     a:link,
--- a/layouts/css/page-modules/_linuxfoundation.styl
+++ b/layouts/css/page-modules/_linuxfoundation.styl
@@ -1,6 +1,6 @@
 .linuxfoundation-footer
     width 100%
-    padding-bottom 1em
+    padding 1em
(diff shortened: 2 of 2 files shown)


layouts/css/page-modules/_footer.styl
```diff
@@ -1,7 +1,8 @@
 footer
     // display flex
     margin-top 2em
-    background-color $node-gray
+    padding 0 20px
+    background-color #404042
```
fhemberger: a) Please stick to use variables for colors, this makes it easier to tune them later.
b) Footer color differs on purpose, the language selection will go there later and have the `$node-gray` background color.

package.json
```diff
@@ -4,9 +4,9 @@
   "description": "New Nodejs.org website",
   "homepage": "https://new.nodejs.org",
   "scripts": {
-    "build": "./build.js",
-    "gulp": "./node_modules/gulp/bin/gulp.js",
-    "serve": "./build.js serve"
+    "build": "iojs build.js",
```
fhemberger: Good catch, thanks!

AustinWinstanley:
Ok, added the variable back and set the padding in the Linux Foundation style

footer

fhemberger:
Changes merged through #35

PR #38, opened 2015-08-18
Add linting script and dependencies

phillipj:
Hi!

I noticed there was an `.eslintrc` here, but couldnt see any linting script defined?

This enables eslinting of source JS files by running:

```bash
$ npm run lint
```

Also fixed some minor linting issues.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -17,6 +17,6 @@
     "no-trailing-spaces": 1,
     "no-underscore-dangle": 0,
     "quotes": [1, "single"],
-    "space-before-function-parentheses": [1, "always"]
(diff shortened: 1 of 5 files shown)


.eslintrc
```diff
@@ -17,6 +17,6 @@
     "no-trailing-spaces": 1,
     "no-underscore-dangle": 0,
     "quotes": [1, "single"],
-    "space-before-function-parentheses": [1, "always"]
```
phillipj: `space-before-function-parentheses` has been deprecated

fhemberger:
Good catch, thank you!

PR #39, opened 2015-08-18
moved downloads-table into reusable table class and small design fixes

AustinWinstanley:
Instead of a 'downloads-table' class, I moved it into a 'table' class so it can be reused in other tables if it needs to be. Also fixed some design issues and created a helpers.styl for common usages. And moved the colors for table into _variables


layouts/css/layout/_helpers.styl
```diff
@@ -0,0 +1,8 @@
+.text-center
+    text-align center
+	
+.text-left
```
fhemberger: As we will also have RTL language translation, better use `start` and `end` rather than `left` and `right` for `text-align`.

mikeal:
This probably broke with the new downloads work that got merged. Should this get rebased or is it no longer relevant?

AustinWinstanley:
Well, it reminded me to merge and fix the stying :) It looks good, but the table font looks a little strange on windows/chrome.

untitled

fhemberger:
How about merging this and deal with the overall font-tuning later? I guess there are still some other cases we need to check.

AustinWinstanley:
I rebased and merged with master. It should be up to date.

fhemberger:
Unfortunately, the changes broke the Download matrix. Can you please take a look at it?

AustinWinstanley:
Working on it. Will update soon.

bnb:
@AustinWinstanley You working on this? If not, we can close it.

fhemberger:
Closing this due to inactivity.

PR #43, opened 2015-08-24
Less versions coupling in .hbs templates

phillipj:
Simpler way of fetching current version of the project: `project.currentVersion` rather than `project.versions.0.version` meaning less coupling and in a sence logic in the .hbs templates.

Also replaced and removed the `version` property in `locale/en/index.md` as it should be read from the versions configuration rather than a locale property.

### DIFF
--- a/build.js
+++ b/build.js
@@ -15,8 +15,9 @@ const path = require('path');
 const fs = require('fs');
 const ncp = require('ncp');
 
-const filterStylusPartials = require('./plugins/filter-stylus-partials.js');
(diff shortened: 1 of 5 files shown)


fhemberger:
We still need the entire list of versions for the download page.

phillipj:
It's still there as before (`project.versions`), but the current / latest version is alot easier to get a hold of.

fhemberger:
Ah, sorry. I overlooked the line in the diff.

phillipj:
Np! I'm playing with the download matrix, this cleans up **alot** of verbosity in that markup too.

PR #48, opened 2015-08-26
Refactored some global styling into .class

phillipj:
This removes some global styling on `ul` and `small` was added along with the download matrix.

Also refactored out some specific styling on small-elements inside the download matrix, extracting it into reusable .class'es instead.

I prefer adding reusable classes to elements rather than handcrafting specific styling all the time. Would appreciate your thoughts about adding more reusable .class'es going forward rather than very specific styling as most of the site has ATM.

### DIFF
--- a/layouts/css/base.styl
+++ b/layouts/css/base.styl
@@ -19,9 +19,6 @@ h4,
 h5
     font-weight normal
 
-ul
-    padding 0
-
 a,
 a:link,
 a:active
@@ -58,9 +55,6 @@ pre
     code
         color #f0f0f0
 
-small
-    font-size 10px
-
 // Import specific page sections and layout parts
 @import 'layout/_sticky-footer'
 @import 'layout/_grid'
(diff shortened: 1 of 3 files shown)


fhemberger:
> I prefer adding reusable classes to elements rather than handcrafting specific styling all the time. Would appreciate your thoughts about adding more reusable .class'es going forward rather than very specific styling as most of the site has ATM.

It is better to have module specific classes (e.g. ".download-table") defining a certain behavior than coupling classes to visual appearance ("no-padding", "color-green"), etc.

fhemberger:
I merged the changes, I think we should address the "CSS styleguide" at a later point, as we don't have much time left until the new site needs to go online for the v4.0.0 release.

phillipj:
> I think we should address the "CSS styleguide" at a later point

Totally agree! Much easier to contribute with good changes if there's some kind of guidelines.

Issue #20, opened 2015-08-05
Next Steps


mikeal:
Let's try to write out all the next steps to get the website up and running and assign them to people so that we can avoid overlap.
## Content

For more details please refer to this earlier thread about the taxonomy. https://github.com/nodejs/new.nodejs.org/issues/19#issuecomment-125007795
- [x] Home
- [x] About.md
  - [x] governance.md
  - [x] working-groups.md
- [x] Security.md
- [x] Downloads.md
- [x] Foundation.md
  - [x] members.md
  - [x] TSC.md
  - [x] TSC meeting minutes
- [x] Contribute.md
- [x] Docs.md
  - [x] FAQ.md
  - [x] ES6.md
- [x] API (link to https://iojs.org/api/ for now)
- [x] Blog.md
- [x] 404.md
## Templates
- [x] About.hbs
- [x] Security.hbs
- [x] Downloads.hbs
- [x] Foundation.hbs
- [ ] Contribute.hbs
- [x] Docs.hbs
- [ ] Blog.hbs
## Style Fixes
- [ ] Home
- [ ] Blog
- [ ] General style for all the basic markdown pages like About and Foundation.
## Blog
- [x] Merge in old io.js posts from the evangelism repo. (Mikeal is working on this)
## RSS feeds
- [x] Blog posts
- [x] TSC meeting minutes
## Other
- [x] Inventory and map out any links breaking on the new site
- [x] Properly permanent redirect relevant old content to new urls

mikeal:
@fhemberger can you edit the checklist add any additional work.

diagramatics:
By the additional .md files, are they content writing? I'm able to help in any way I can if the quality of the weekly updates I churned out are satisfactory.

fhemberger:
@diagramatics Uhm, at least I used the checkboxes for content migration from nodejs.org/iojs.org, not necessarily for content (re)write or editing, I think this will be the next step.

mikeal:
@diagramatics @fhemberger Corrent, with the exception of the Resources page which I think should be something new and Docs WG should decide/write that.

leo:
So this whole things isn't about making a real **new** site (including a different design), it's just about merging the contents of the io- and node-site, right?

Does this mean that a re-design would be completely inappropriate? Or at least suggestions for a new layout? Because if I take a look at the current state, nothing has changed regarding styles.

fhemberger:
[...]

PR #50, opened 2015-08-26
WIP: support multiple projects

phillipj:
Working on getting _both_ io.js and node supported, mainly related to downloads and versions. So far its been brute force just to get it up-n-running.

As we dont have a unified `versions.json` yet, as @mikeal mentioned in nodejs/build#163, I've just created a simple `unified_versions.json` for mocking purposes in the meantime.

Just duplicating _everything_ doesnt make sense and obviously doesnt look that great either. The frontpage didnt get too bad though, but the downloads page with its matrix and previous released versions is a whoooole different story.

image

Any thoughts about how this?


fhemberger:
There's no need to support multiple projects. From version 4 on, there will be only Node.js.
Older versions of node/io.js will only be shown in the overall release list, which will by then become a separate page (with a looooong list).

phillipj:
Ahh, so when you said

> Both. There still needs to be some work done to integrate the builds automatically into the site.

in issue #20 you only meant the previous release list?

fhemberger:
Yes. The list should not only the converged release (starting with 4.0.0), but all io.js and node.js releases in a single spot. Sorry, if this caused a misunderstanding.

phillipj:
Np, created a WIP to find out what needed to be done - great success in other words :tada:

Sounds alot simpler than what I had in mind. Closing this PR as it sounds trivial enough to fix quickly after build has a unified versions in `versions.json` ready.

mikeal:
This came up in the TSC meeting a bit as well, we want to have a merged/complete Changelog as well and I want to try and create automated release posts from the Changelog+versions.js

PR #54, opened 2015-08-27
Scroll to top button added

bogas04:
screen shot 2015-08-27 at 11 07 38 pm
I use onscroll event to show scroll to top button only when it is required.
In my initial tests, the button seems to be placed alright for all resolutions.
CSS scroll-behaviour is used to give smooth scroll effect. I feel it is the cleanest way with best case scenario for backward compatibility.

### DIFF
--- a/layouts/partials/footer.hbs
+++ b/layouts/partials/footer.hbs
@@ -1,13 +1,13 @@
-<button id="scrollToTop">&uarr;</button>
+<a href="#" id="scrollToTop">&uarr; <span>Scroll To Top</span></a>
 
 <script type="text/javascript">
     var $scrollToTop = document.getElementById('scrollToTop');
-    $scrollToTop.innerHTML = window.innerWidth > 400 ? 'Scroll To Top' : '&uarr;';
-    window.onscroll = function() {
+    (window.onscroll = function() {
         $scrollToTop.style.display = (window.pageYOffset > window.innerHeight) ? 'block' : 'none';
-    };
+    })();
     $scrollToTop.onclick = function() {
         window.scrollTo(0, 0);
+        return false;
     };
 </script>
 <footer role="contentinfo">
--- a/layouts/css/page-modules/_scrollToTop.styl
+++ b/layouts/css/page-modules/_scrollToTop.styl
@@ -3,10 +3,19 @@ html
     scroll-behavior smooth
 
 #scrollToTop
-    display none
+    font inherit
+    font-size 1rem
+    color $node-gray
+    background-color #eee
+    border 1px solid #ddd
+    border-radius 2px
+    cursor pointer
     position fixed
     bottom 10%
-    right 5%
-    color $node-gray
-    background-color #D9EBB3
-
+    right 0
+    min-width 20px
+    text-align center
+    padding 0 5px
+    @media (max-width: 600px)
+        span
+            display none


fhemberger:
Changed a few things:
- `window.pageYOffset` instead of `e.pageY` for cross browser compatibility
- Use `<button>` instead of `<a href="#">` (when there's no href, always use a button for semantics).
- Update button styling
- Indentation

```diff
diff --git a/layouts/css/page-modules/_scrollToTop.styl b/layouts/css/page-modules/_scrollToTop.styl
index 495f1c1..58186d1 100644
--- a/layouts/css/page-modules/_scrollToTop.styl
+++ b/layouts/css/page-modules/_scrollToTop.styl
@@ -7,6 +7,12 @@ html
     position fixed
     bottom 10%
     right 5%
+    cursor pointer
     color $node-gray
-    background-color #D9EBB3
+    background-color #eee
+    border 1px solid #ddd
+    border-radius 2px
+    font inherit
+    font-size 1rem
+

diff --git a/layouts/partials/footer.hbs b/layouts/partials/footer.hbs
index b7a6ba6..f5a41d9 100644
--- a/layouts/partials/footer.hbs
+++ b/layouts/partials/footer.hbs
@@ -14,14 +14,13 @@
     </div>
 </footer>

-<a href="#" id="scrollToTop">Scroll To Top</a>
+<button id="scrollToTop">Scroll To Top</button>
 <script type="text/javascript">
-  $scrollToTop = document.getElementById('scrollToTop');
-  window.onscroll = function(e) {
-    $scrollToTop.style.display = (e.pageY > window.innerHeight) ? 'block' : 'none';
-  };
-  $scrollToTop.onclick = function(e) {
-    window.scrollTo(0, 0);
-    return false;
-  };
+    $scrollToTop = document.getElementById('scrollToTop');
+    window.onscroll = function () {
+        $scrollToTop.style.display = (window.pageYOffset > window.innerHeight) ? 'block' : 'none';
+    };
+    $scrollToTop.onclick = function () {
+        window.scrollTo(0, 0);
+    };
 </script>
```



However, I'm not completely happy with this solution at the moment, because on smaller/mobile screens the button will be placed over the content.




So maybe someone has an idea how to improve this. (Just show an up arrow, e.g. ▲?)
/cc @nodejs/website

fhemberger:
Oh, and `position: fixed` might still be a problem on mobile. Haven't tested that yet.

bogas04:
I chose anchor tag for cases when javascript is disabled (href="#" would still lead to top).

I think we can use a media query to change the text to the HTML entity for smaller resolutions. I am ready to take it up.

phillipj:
Have you considered throttling the `onscroll` handler? As it fires *alot* when scrolling. Or another trick to prevent touching the DOM when not necessary.

bogas04:
@phillipj  should I use a [debounce function](http://davidwalsh.name/javascript-debounce-function) ?  Or is there an alternative to it ? Or we should show the button what-so-ever ?

fhemberger:
By the way: How about having a regular "Back to top" button at the end of the page (above the footer) and enhance it with JS to the current behavior? Would be the best solution IMHO.

bogas04:
How does it look now ? 

* window.innerWidth > 400px
screen shot 2015-08-28 at 6 47 55 pm
* window.innerWidth <= 400px
screen shot 2015-08-28 at 6 48 13 pm


Note: I haven't added an event listener for resize of browser window for it seems to be overkill for such a small thing. On a default view, mobile user will get the arrow and desktop user will get "Scroll To Top". Mobile user will be okay with the arrow even for landscape view and desktop user will be okay with Scroll To Top even on reducing width, at least that is what I believe.

fhemberger:
Way better. :+1:

For accessibility (and also for not maintaining strings in the JS part), I'd use something like this:

```html
<a href="#" id="scrollToTop"><span>&uarr;</span> Scroll To Top</a>
```

(restore the default behavior to jump to the top w/o JavaScript) and 

```stylus
#scrollToTop
	...
	overflow hidden

	span
		display none

	@media (max-width: 400px)
		text-indent -100rem

		span
			display inline
```

Untested, but you get the idea.

bogas04:
I think the arrow with text for desktop and only arrow for mobile users (600px or less) would be neat. CSS is handling all that now. 
Also I'm calling the onclick handler on page load so as to not display it if not required. 

Screenies : 

Desktop screen shot 2015-08-28 at 7 40 00 pm Mobile screen shot 2015-08-28 at 7 40 10 pm

FYI : [scroll-behavior](https://developer.mozilla.org/en-US/docs/Web/CSS/scroll-behavior) is only available on FF36 & Chrome(with flag enabled)

Fishrock123:
Why add a scroll to top button? Most people's mice wheels, trackpads, touchscreens, etc work quite well, no?

bogas04:
@Fishrock123 I felt the need while browsing through /contribute/development & /blog. I feel they're long enough to make you realize the time spent scrolling through it.

phillipj:
> Now that I think of it, not having javascript will not show the button at all.

You will have to add `display:none` inline or in a .styl for that to be true, as JS is the only thing toggling the display style atm.

layouts/partials/footer.hbs
```diff
@@ -1,3 +1,15 @@
+<a href="#" id="scrollToTop">&uarr; <span>Scroll To Top</span></a>
+
+<script type="text/javascript">
+    var $scrollToTop = document.getElementById('scrollToTop');
+    (window.onscroll = function() {
+        $scrollToTop.style.display = (window.pageYOffset > window.innerHeight) ? 'block' : 'none';
```
phillipj: In regards to my previous throttling question, the simplest fix for as little style changes as possible I could think of:
```js
var newVal = (window.pageYOffset > window.innerHeight) ? 'block' : 'none';
if (this.currentVal !== newVal) {
    $scrollToTop.style.display = newVal;
    this.currentVal = newVal;
}
```

fhemberger:
@nodejs/website Do we want to merge this feature for the go-live?

bogas04:
Let's merge it, it'll be my first code contribution to the project :grinning:

fhemberger:
@therebelrobot Fixed the flashing of the button on load
@bogas04 Thank you, merged it.

PR #61, opened 2015-08-30
style: make site more readable on mobile

evanlucas:
The side nav breaks and is centered above article now.

pre blocks also have scrolling for overflow-x on mobile.

Before:

screen shot 2015-08-30 at 7 21 10 am

After:

screen shot 2015-08-30 at 7 19 13 am


fhemberger:
Wee, awesome. Thanks.

PR #63, opened 2015-08-30
Make footer look better on mobile

evanlucas:
Before:

screen shot 2015-08-30 at 8 33 46 am

After:

screen shot 2015-08-30 at 8 33 24 am

### DIFF
--- a/layouts/css/page-modules/_footer.styl
+++ b/layouts/css/page-modules/_footer.styl
@@ -2,6 +2,7 @@ footer
     // display flex
     margin-top 2em
     background-color $node-gray
+    padding: 0 0px
 
     a,
     a:link,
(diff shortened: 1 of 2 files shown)


fhemberger:
@evanlucas Have you changed the SVG file as well? If so, please add the file to the PR as well (put it in /static/images) and use [svgo](https://github.com/svg/svgo) to optimize it before including it as base64. Thanks!

evanlucas:
I did not change it at all. It is the exact same as the non-mobile one

fhemberger:
Okay, then there's no need to include the file again, please just change the background-position in the media-query.

evanlucas:
ah ok

evanlucas:
Done

fhemberger:
Cool, thanks!

PR #67, opened 2015-08-30
Improve error message and exit code in load-versions.js

phillipj:
When something fails, exit with code 1 too signal failure to whomever started the script.

Also included which URL failed when writing informative error to stderr as the errors from `https` tend to be too lowlevel to see which URL was requested.


fhemberger:
Thanks! This script was only meant as a quick hack unless both files are merged officially. ;)

PR #68, opened 2015-08-31
styl,hbs: add header dropdown menus

Fishrock123:
Proof of concept to see if the @nodejs/website folks like it.

I think I can do an even more clever and nicer effects by changing some stuff around.


Fishrock123:
cc @fhemberger Thoughts?

fhemberger:
@Fishrock123 Sorry for the late reply, totally forgot about this one.
Rebased against current master. Here are some screenshots for @nodejs/website:

fhemberger:
I must say I'm not entirely happy yet with this solution. At the moment, the sub-navigation and its items don't align well, making it look a bit odd. Maybe this can be fixed with a bit of CSS tweaking.

Fishrock123:
@nodejs/website If this is something you feel would fit well into the design, I'll put a considerable amount of effort in trying to make it work correctly.

mikeal:
I actually don't think this is a great idea. From prior experience i know that the rate at which people click on the "main" page from that top nav drops to near zero if there is a nav below it. I'd like to see those landing pages get a lot of views and to be a good entry point for newcomers. We can't control how people enter some of these more advanced topics if we go this route.

fhemberger:
I feel similar to what @mikeal said. It also requires quite a bit of CSS fine tuning, which might become an issue as well as soon as we add localizations (different text lengths, RTL languages), so I think the work you have to put into it doesn't match the outcome.

bnb:
Has this come to a decision?

Fishrock123:
> It also requires quite a bit of CSS fine tuning, which might become an issue as well as soon as we add localizations (different text lengths, RTL languages), so I think the work you have to put into it doesn't match the outcome.

I feel quite confident in the ability to make it work quite resizable. :)

That being said I'll agree with @mikeal. Closing.

Issue #57, opened 2015-08-30
Website testing on mobile and small screens


fhemberger:
We need to test the website on small screens as well. This means browser tests (e.g. with Chrome's device mode), but also real browser tests on devices (Safari iOS, Chrome for Android, Opera Mobile).

- [ ] Test overall website
- [x] Test site performance (Wifi, EDGE, 3G)
- [ ] Check font size, margins, table- and blog layout
- [ ] Adjust layout of main navigation
- [x] Adjust layout of sub-navigation
- [ ] …

rvagg:
you have less than 2 days btw

fhemberger:
Aaaahhhhh! *running around panicking*

Okay, well. I guess this means we'll do this after the initial release.

mikeal:
@rvagg this is the biggest blocker we have right now https://github.com/nodejs/build/issues/163#issuecomment-134747962 we need all the old node.js builds to show up in versions.

evanlucas:
Do we want/need to adjust the layout of the main navigation? If so, anything I can help with?

I am going through and checking each page to see if any alignments are off and things like that now.

fhemberger:
@evanlucas Adjusting the margin/padding should be fine, ne need for a hamburger menu, et. al.

ddprrt:
I did some performance tests on the current https://new.nodejs.org website.

## Webpagetest

Using Cable connection, from Dulles. Click [here](http://www.webpagetest.org/result/150830_X6_PWT/) for the results. Fully loaded under 1 second, Speed Index < 1000. Repeated view somewhere in the 400s. Excellent stats

Using Fast 3G: Click [here](http://www.webpagetest.org/result/150830_TN_PVW/) for the results. Fully loaded under 3 seconds, speed Index ~2500. This is okay. The repeated view is again excellent. Your cache works wonders.

Using Edge: Click [here](http://www.webpagetest.org/result/150830_FW_NFM/) for the results. Fully loaded in 16 seconds, Start render at 14 seconds. This is a bummer. 

Looking at the waterfall diagram (especially of the Edge render), we can easily define the culprits: Blocking CSS from three files. Solutions:

- We could go full critical and inline the critical CSS and asynchronously load the remaining styles. I made some tests if this would be feasible for the new page, but due to the unique design of the contents, we would need to run the critical CSS task for each and every page, which would lead to huge build times and possible memory leaks.

[...]

PR #69, opened 2015-08-31
asynchronously loading the webfont + caching it

ddprrt:
This is related to #57: 
- `webfont.css` contains "Source Sans Pro" in weights 300, 400, 600 and 700. I removed the italic styles, mainly because they aren't used anywhere on the site.
- `footer.hbs` contains a JavaScript, ajax-ing `webfont.css` and putting the contents into localStorage. 
- `html-head.hbs` loads CSS from localStorage and puts it as the first styles, practically inlining the webfonts.
- Removing `prism-tomorrow.css` and putting it at the end of `layouts/about.hbs` -- this might be not valid, but browsers behave friendly: They load it as the last resource, non-blocking.


layouts/partials/footer.hbs
```diff
@@ -13,3 +13,23 @@
...
+    try {
+      var xhr = new XMLHttpRequest();
+      xhr.addEventListener('load', function() {
+        if(this.responseText) {
+          addStyle(this.responseText);
+          window.localStorage.fonts_css_cache = this.responseText;
+        }
+      });
+      xhr.open('get', '/static/fonts/webfont.css');
+      xhr.send();
```
ddprrt: btw.: yes, we load the webfonts.css every time, and also update the cache every time. The user still gets a fast experience, the webfonts.css will be cached, but in case of new and different fonts, we still get an update.

silverwind:
Guess this is kind of obsolete now with https://github.com/nodejs/nodejs.org/commit/aa4b373526fd3feb2c30bd6271e2872001e56c05.

PR #70, opened 2015-08-31
fixed canonical URLs

ddprrt:
Turns out the `{{canonical}}` field does not exist. Glued it together from two different fields.


fhemberger:
Thanks! Had to modify it a bit for the trailing slash.

Issue #58, opened 2015-08-30
First release checklist


mikeal:
This is the list of blockers for getting the new site live. There's plenty of additional things to do to improve the website, this list is **only** things that **MUST** be done in order to go live.

# Content

* [x] Docs.md (main docs page content)
* [x] versions.json w/ all io.js and node.js releases
* [x] https://iojs.org/api/ -> https://new.nodejs.org/api/
* [x] Google Analytics

What else?

fhemberger:
Move/link API docs from iojs.org to (new.)nodejs.org

rvagg:
https://new.nodejs.org/download/release/index.json
https://new.nodejs.org/download/release/index.tab
or
https://new.nodejs.org/dist/index.json
https://new.nodejs.org/dist/index.tab

Note that this server is now populated with all of the old node dist directory

I wasn't planning on merging the two but I'm getting the impression that that's being suggested.

fhemberger:
@rvagg Can you sort the list by date? If merging both lists is a problem at the moment, we could do this at build time as well.

rvagg:
"date" is kind of hard to determine, which file do you test and what is that date indicating anyway?

Instead, I've pulled in node-semver and used it to compare for a sort, hope that helps

rvagg:
I've turned on CloudFlare for new.nodejs.org and am going to bed, let me know if anything odd turns up but we're getting very close to this being ready. I did want to switch iojs.org over to this new server but it's too late and I don't want to be switching it just before I head off to sleep so I'll do it tomorrow.

fhemberger:
@rvagg Wrote a little script `load-versions.js` which merges the two files for our release list.

fhemberger:
Okay, I'm calling it a day as well. ;)

rvagg:
You can do https://iojs.org/api/ -> https://new.nodejs.org/api/ now, it's active

Also, I don't know if it helps at all but I'm putting `latest-v0.10.x`, `latest-v0.12.x`, `latest-v4.x` etc. in all of the download directories (/dist/ a.k.a. /downoads/release/, /download/nightly/, /download/rc/, etc.), these'll get updated on each release. But you have the index.json data you can do that from anyway.

fhemberger:
@rvagg I still end up on the Node.js v0.12.7 docs instead of the iojs docs.

rvagg:
[...]

PR #71, opened 2015-09-01
Node versions before iojs

phillipj:
Addresses @rvagg [comment](https://github.com/nodejs/new.nodejs.org/issues/58#issuecomment-136496901) in #58 about having v0.12.7 as current version instead of v3.2.0.

Wait until node v4 is out before sorting combined node and iojs versions by semver.


fhemberger:
We can keep the script unchanged, just replace `source/versions.json` with `https://new.nodejs.org/dist/index.json`.

Or we just change https://github.com/nodejs/new.nodejs.org/blob/master/build.js#L55 to `v0.12.7` for now.

phillipj:
> We can keep the script unchanged, just replace source/versions.json with https://new.nodejs.org/dist/index.json.

Ahh, so we're okey with just having node version at first?

phillipj:
> Or we just change https://github.com/nodejs/new.nodejs.org/blob/master/build.js#L55 to v0.12.7 for now.

What I did initially but it felt **really** confusing displaying table of previous releases.. Having node versions at the top wasnt that disturbing.

fhemberger:
> Ahh, so we're okey with just having node version at first?

To be honest: I don't know. ;) /cc @mikeal

rvagg:
yes please, v0.12 first please, v4 will be out later this week or at the start of next week so it won't matter for long anyway

phillipj:
Alright, closing this as @fhemberger just pushed a fix to master instead

PR #72, opened 2015-09-01
copy robots.txt from old site

rvagg:


fhemberger:
Ah, haven't noticed that there was already one in place …

PR #78, opened 2015-09-02
download windows x64 .msi files from x64/ for 0.x.x

rvagg:
I haven't actually tested this works at all but windows 64-bit files for 0.x.x need to be prefixed with x64/, because reasons, so the big download button on the front page is giving a dead link for those visitors

### DIFF
--- a/static/js/download.js
+++ b/static/js/download.js
@@ -14,7 +14,7 @@
       db.innerText = dlLocal + ' OS X (x64)';
       break;
     case 'Win':
-      db.href += 'node-' + version + '-' + x + '.msi';
+      db.href += (version[0] == '0' && x == 'x64' ? x + '/' : '') + 'node-' + version + '-' + x + '.msi';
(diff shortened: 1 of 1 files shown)


static/js/download.js
```diff
@@ -14,7 +14,7 @@
       db.innerText = dlLocal + ' OS X (x64)';
       break;
     case 'Win':
-      db.href += 'node-' + version + '-' + x + '.msi';
+      db.href += (version[0] == '0' && x == 'x64' ? x + '/' : '') + 'node-' + version + '-' + x + '.msi';
```
fhemberger: Version is prefixed with 'v', so it must be `version[1]`. LGTM otherwise.

fhemberger:
Merged with 3cdb5229326e4e8094cffad3325251f8ac2ae54d and 045fae21bde6fb90900e1e0c04f17601c0fc9fcd.

PR #81, opened 2015-09-02
css: Fix default background-color

Delapouite:
Without specifying this explicit color, the site is currently unreadable for people who have configured their browser to display white on black texts.


fhemberger:
Thanks!

PR #82, opened 2015-09-02
fixed typo in home

a0viedo:


fhemberger:
It's not a typo. Words already ending on an `s` use `s'` instead of `s's`. ;)

PR #83, opened 2015-09-02
rephrasing in get involved view

a0viedo:
as it seems better fit


fhemberger:
Thanks for pointing this out, changed what -> why. Now the sentence makes sense. ;)

PR #87, opened 2015-09-03
Added SunOS and ARM binaries to download matrix

phillipj:
SunOS and 32-bit Mac binaries was not added to the download matrix in the first place as iojs doesnt have these binaries and this site consisted only of iojs as of last week.

Also prepended http://nodejs.org to all downloads URLs, frustrating to get 404 requests when clicking these downloads link when running locally..

Will node v4 have SunOS and 32-bit mac binaries @rvagg?

#### BEFORE

image

#### AFTER

image

### DIFF
--- a/layouts/download.hbs
+++ b/layouts/download.hbs
@@ -16,21 +16,21 @@
                     <div class="download-hero full-width">
                         <ul class="no-padding">
                             <li>
-                                <a href="/dist/{{project.currentVersion}}/node-{{project.currentVersion}}-x86.msi">
+                                <a href="https://nodejs.org/dist/{{project.currentVersion}}/node-{{project.currentVersion}}-x86.msi">
                                     <img src="/static/images/platform-icon-win.png" height="50" width="45" alt="">
                                     Windows Installer
                                     <p class="small color-lightgray pad-bottom-small">node-{{project.currentVersion}}-x86.msi</p>
                                 </a>
                             </li>
                             <li>
-                                <a href="/dist/{{project.currentVersion}}/node-{{project.currentVersion}}.pkg">
+                                <a href="https://nodejs.org/dist/{{project.currentVersion}}/node-{{project.currentVersion}}.pkg">
                                     <img src="/static/images/platform-icon-osx.png" height="50" width="45" alt="">
                                     Macintosh Installer
                                     <p class="small color-lightgray pad-bottom-small">node-{{project.currentVersion}}.pkg</p>
                                 </a>
                             </li>
                             <li>
-                                <a href="/dist/{{project.currentVersion}}/node-{{project.currentVersion}}.tar.gz">
+                                <a href="https://nodejs.org/dist/{{project.currentVersion}}/node-{{project.currentVersion}}.tar.gz">
                                     <img src="/static/images/platform-icon-generic.png" height="50" width="45" alt="">
                                     Source Code
                                     <p class="small color-lightgray pad-bottom-small">node-{{project.currentVersion}}.tar.gz</p>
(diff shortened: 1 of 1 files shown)


Fishrock123:
cc @nodejs/build

I don't think we'll have 32-bit mac ones? Not sure about smartos.

@phillipj 
This page, however is greatly missing the ARM binaries, which are probably more important. :)

Fishrock123:
Also I suggest listing the `.xz` versions of any *nix (not mac) binaries. But that may not be correct. I'll let others chime in.

phillipj:
@Fishrock123 ARM indeed! Maybe I should make a separate PR ready to be merged when v4 arrives, which has ARM and `.xz`? As this was primarly meant to fix what's lacking in production at the moment for v0.12...

Fishrock123:
See: https://iojs.org/dist/v3.3.0/

I'd target 4.0.0. it should drop next monday/tuesday.

rvagg:
eeek, tbh I hadn't factored in `sunos` binaries for v4 but I should have a quick look at adding that back in, I just don't know much about producing those so no promises

phillipj:
Ok @Fishrock123, I'll revise to target v4 instead.

@rvagg np, just say the word if I should remove SunOS again.

mhdawson:
The other thing I think that is different from the previous site is that for at least for 0.12.X there were links to downloads where the pLinux, AIX and zLinux builds were available.  @jasnell

phillipj:
Pushed some changes targeting node v4 and new paths:
* ARM binaries
* Changed paths to 32-bit windows binaries 
* Mac binaries are not universal but 64-bit (as commented in #88 [comment](https://github.com/nodejs/new.nodejs.org/pull/88#issuecomment-137722521))

**Dont merge until v4 has been released!**

### UPDATED

image

rvagg:
Just to complicate matters we also have ARMv8 in the mix now, also reffered to as AARCH64 although "ARMv8" is probably good enough, maybe "ARMv8 64-bit".

Also, I'm not sure how Solaris folks feel about naming and whether "SunOS" makes sense these days. Honestly I have no idea but /cc @nodejs/build for help on naming here. Strictly speaking they are SmartOS.

phillipj:
Would appreciate some suggestions about exposing `.xz` files aswell if we want those in there aswell. Little hesitant to almost doubling the amount of rows. What I've played with locally has ended up looking like a mess IMO, but I'm not the most creative UX person around :)

rvagg:
You could always have a link down the bottom that says "All download options" that just links to the dist directory and they can figure it out, then you could avoid the more obscure platforms in the main table

fhemberger:
I think that's the best solution, only show the most popular platforms.

jbergstroem:
@rvagg can't really back any claims up, but I reckon the smartos/sunos binaries can be run on Illumos, OmniOS et al as well (SmartOS is actually a derivative of Illumos). Just saying that `smartos` might not be the most correct choice hence staying with as-is is at least neutral.

rvagg:
@jbergstroem right, but is "SunOS" right here? Should it be "Solaris"?

rvagg:
https://nodejs.org/download/nightly/v4.0.0-nightly201509079cae65c510/ has all of the download asset types that we're going to be going live with when we do v4.0.0, including "sunos-x64", "sunos-x86" and "linux-arm64" which are missing from RC 1. See https://nodejs.org/download/nightly/ for an index.tab and index.json with the full data if you want that for testing.

phillipj:
Included three ARM types and added link for all downloads at the bottom.

The SunOS / Solaris / SmartOS is the last remaining puzzle.

### UPDATED

image

rvagg:
Wikipedia:

> SunOS is a version of the Unix operating system developed by Sun Microsystems for their workstation and server computer systems. The SunOS name is usually only used to refer to versions 1.0 to 4.1.4 of SunOS. These versions were based on BSD, while SunOS version 5.0 and later are based on UNIX System V Release 4, and are marketed under the brand name Solaris.

Just go with SunOS I guess, wait till someone complains, which I'm sure they will if that's wrong

jbergstroem:
@rvagg I wonder if it even works on solaris.. reckon less change is better for now.

fhemberger:
I merged this already into a separate [node-4.0-release](https://github.com/nodejs/new.nodejs.org/tree/node-4.0-release) branch, so all required changes can go live at the same time.

phillipj:
Sounds like a good plan @fhemberger :+1: I'm closing this PR then.

PR #89, opened 2015-09-04
Fix lint script after moving files

phillipj:
The contents of the previous `./plugins` directory got moved into `./scripts` in 7448a271 which resulted in breaking the `$ npm run lint` script.


fhemberger:
Whoops, sorry. Haven't noticed this one … thanks!

PR #92, opened 2015-09-05
Fix typo

grawlinson:
Corrected a small typo in the get involved title.


fhemberger:
I think we should all mot get in*love*d. ;) Thanks!

PR #93, opened 2015-09-05
css: show vertical scrollbar on desktop browsers

ChALkeR:
Without this, everything gets shifted left and right again when switching between pages with vertical scrollbar and no vertical scrollbar.
This shouldn't affect browsers that have zero-width scrollbars by default (mobile and safari).

### DIFF
--- a/layouts/css/base.styl
+++ b/layouts/css/base.styl
@@ -7,6 +7,7 @@ body
     color #333
     margin 0
     background-color #fff
+    overflow-y scroll
 
 header,
 #main,


fhemberger:
-1 on that. Don't show a dummy scroll bar when none is needed.

ChALkeR:
@fhemberger For a large enough screen, scrollbars are not needed on some pages, but are still needed on some of them (the most part, in fact).
If you want to observe the problematic behaviour that adding a scrollbar to all pages (when using a desktop browser) fixes, do this:
 1. Open some browser that has no-zero-width scrollbars (not Safari).
 2. Navigate to https://nodejs.org/en/ (first item in the menu).
 3. Zoom it out a bit, so that the scrollbar gets hidden.
 4. Press «About» (second item in the menu), _the page, including the menu under your cursor, gets shifted left_.
 5. Press «Downloads» (third item in the menu)¸  _the page, including the menu under your cursor, gets shifted right_.
 6. Press «Docs» in the menu (fourth item in the menu)¸  _the page, including the menu under your cursor, gets shifted left_.

That's just annoying.

fhemberger:
The behavior you describe exists since the dawn of web browsers on all websites that have some pages longer than the current viewport. So personally, I don't really consider this an issue, but I'm happy to hear @nodejs/website's thoughts on that.

Fishrock123:
I think I got around something like this on my site (https://searchbeam.jit.su/about), at least in webkit, by screwing with the scrollbar css.

ChALkeR:
@Fishrock123  No, it doesn't work, at least for me.

ChALkeR:
@fhemberger Yes. ANd this was the standard way of fixing it.

Fishrock123:
Ah yeah, I turned it back on. Here's the CSS for removing them completely in webkit (eeehhh probably don't do this.) https://github.com/Fishrock123/Searchbeam/blob/master/src/less/override.less#L14-L26

Ideally the scrollbar would be styleable and float ontop of content.

Maybe we can make a fake one if the page height  < screen height though. Idk.

silverwind:
First rule of web development: don't mess with the scrolling.

No, seriously: This is more of an browser bug than anything. The only acceptable solution I see it some JS that obtains the width of the scrollbar and if non-zero, offset the page to the right by width/2. If that's even possible cross-browser.

PR #96, opened 2015-09-05
Adds back in flex box and fixes footer issue.

### LINKED ISSUE #95: Fix sticky footer to work with IE11
We had an issue with a combination of using flexbox, the <main> element and the sticky footer, which rendered the page completely useless in IE11, as the whole content was covered:

See: https://twitter.com/jaffathecake/status/639776927839072257

I did a quick fix yesterday, but also ended up disabling the sticky footer for now. It would be cool if someone could investigate this issue and fix the footer for IE11.
(Maybe this helps: http://thatemil.com/blog/2013/11/03/sticky-footers-flexbox-and-ie10/)

paulgrock:
Should close #95. Uses the technique referenced on that issue to solve the flex box problem. I checked this in FF, Chrome, Safari, IE 10 and 11 and things look fine.

### DIFF
--- a/layouts/css/base.styl
+++ b/layouts/css/base.styl
@@ -64,7 +64,7 @@ pre
         color #f0f0f0
 
 // Import specific page sections and layout parts
-// @import 'layout/_sticky-footer'
+@import 'layout/_sticky-footer'
 @import 'layout/_grid'
 @import 'layout/_main'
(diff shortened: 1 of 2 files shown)


fhemberger:
Merged 804013bec64343cfd9a490b48c84584220be76ab. Thanks!

PR #97, opened 2015-09-05
Removes border around node.js logo in IE10

paulgrock:
While I was testing #96 I noticed a border around the logo in IE10.
This removes that border.


fhemberger:
Thanks!

PR #100, opened 2015-09-07
CSS: Fix sticky footer on IE10/11

fhemberger:
The sticky footer PR in #96 was not complete, it didn't stick to the bottom on IE 10/11.
Adding height for IE as in be2460293dbeaac0f9388632ccb03c26dccf9793 caused problems in Safari.
So instead, I'm loading a small custom Modernizr build only for IE as described here: http://thatemil.com/blog/2013/11/03/sticky-footers-flexbox-and-ie10/


fhemberger:
cc @nodejs/website

mikeal:
looks good

PR #102, opened 2015-09-07
German Node.js community added

nodecode:
I've added http://nodecode.de, largest German project about Node.js.


fhemberger:
Well, not sure about the "largest" part … ;)

PR #104, opened 2015-09-07
Updated readme new.nodejs.org -> nodejs.org and contributors

phillipj:
Removed mentions of new.nodejs.org, contribution tip and added myself to the contributors list. 

Any suggestions about the last bit of the repo description? Is the old nodejs.org still deployed somewhere or should we just remove the reference to it all together?

> `nodejs.org` represents an effort by the newly formed [Node.js Foundation](https://nodejs.org/foundation/) to build on the merged community's past website projects to form a self-publishing, community-managed version **of the old [nodejs.org](https://nodejs.org)**.


fhemberger:
At the moment, you can access the former website through old versions of the node api docs up to v0.12

PR #107, opened 2015-09-08
rm 0.12 stuff

Trott:
This removes the explication of 0.12.x behavior which was probably there just to contrast with io.js behavior or something. But now that we're all one happy family and 4.0.0 is Node, let's keep this document simpler and shorter by just talking about the present.


fhemberger:
Thanks, looks much better now!

PR #109, opened 2015-09-08
Prefer developers.google.com/v8/ for V8 link

s3ththompson:
The old code.google.com link is no longer the best link for more V8 information.  The developers.google.com link is the most up-to-date home for info.


fhemberger:
Great, thank you. Looks way more informative (as I guess not everybody wants to start diving into V8's code right away).

PR #112, opened 2015-09-08
blog: Add author and date to blog post

### LINKED ISSUE #111: Add dates and optional authors to blog post pages
Dates are missing, I don't think this is intentional, they are on the index, not sure where they should go though.

It's also be good to be able to read an author from the metadata and put that somewhere if there is one, then we could have posts by named people in the foundation and the broader community, not just faceless text. This'd have to be an optional field I suppose.

fhemberger:
Fixes #111

### DIFF
--- a/layouts/blog-post.hbs
+++ b/layouts/blog-post.hbs
@@ -9,7 +9,10 @@
         <div class="container">
 
             <article>
-                <h1>{{title}}</h1>
+                <div class="blogpost-header">
+                    <h1>{{title}}</h1>
+                    <span class="blogpost-meta">{{#if author}}{{i18n site.by}} {{author}}, {{/if}}<time datetime="{{ date }}">{{ strftime date }}</time></span>
+                </div>
 
                 {{{ contents }}}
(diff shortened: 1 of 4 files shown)


layouts/blog-post.hbs
```diff
@@ -9,7 +9,10 @@
         <div class="container">
 
             <article>
-                <h1>{{title}}</h1>
+                <div class="blogpost-header">
+                    <h1>{{title}}</h1>
+                    <span class="blogpost-meta">{{#if author}}{{i18n site.by}} {{author}}, {{/if}}<time datetime="{{ date }}">{{ strftime date }}</time></span>
```
evanlucas: Is `date` always present?

rvagg:
sgtm

layouts/blog-post.hbs
fhemberger: Yes. It's mandatory to have the blog posts sorted on the website. `strftime` also handles invalid dates (returning the original value).

PR #113, opened 2015-09-08
Mobile tweaks

rnsloan:
Part of #57 

Move in-the-news styles into own stylesheet and remove bad `:` character so the padding is correctly applied
before:
image

after:
image

-----

increase spacing between foundation member logos
before:
image

after:
image


fhemberger:
Thanks!

PR #116, opened 2015-09-08
css: banner text drop shadow (hard to read otherwise)

csabapalfi:
I noticed that the white text on the Node.js Interactive banner is hard to read because of the background so added some shadows to the text. Hope you also find it useful.


fhemberger:
Thanks!

mikeal:
This looks great!

PR #120, opened 2015-09-09
change ES6 in progress example

targos:
People are trying to use the --harmony_modules flag and see that it does not work at all.
Suggest `--harmony_destructuring` instead. It is less buggy.
Ref: https://github.com/nodejs/node/issues/2760


fhemberger:
Cool, thanks.

PR #126, opened 2015-09-10
Footer: Add "Report website issue" link

fhemberger:
As people keep posting issues to nodejs.org, I added a link to the footer:

Desktop view:

Mobile view:

/cc @nodejs/website

### DIFF
--- a/layouts/css/base.styl
+++ b/layouts/css/base.styl
@@ -162,19 +162,20 @@ pre
     clear both
     overflow hidden
 
-    @media screen and (max-width: 640px)
-        a.linuxfoundation-logo,
-        a.issue-link
-            float none
-            width: 100%
 
     .linuxfoundation-logo
+        width 60%
         float left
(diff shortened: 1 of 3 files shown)


Fishrock123:
Maybe highlight the color a bit?

Also are we using any sort of glyphs? We could add a bug glyph from like, fontawesome or something maybe.

LGTM if it works.
Perhaps we should add a "report node.js issue" above or below it?

fhemberger:
a) The link color is the same as for the others in the footer.
b) No, we don't use any sort of glyph font (better use SVG for that).

Good idea about the node.js issue.

fhemberger:
Updated desktop view:


Updated mobile view:

Fishrock123:
LGTM (did not look deeply at the css or anything)

fhemberger:
Not really much CSS to look at. ;)

PR #129, opened 2015-09-12
get-involved:Add official Korean Node.js community

JungMinu:
Added official Node.js(io.js) community


JungMinu:
This is the official korean node.js community
https://github.com/nodejs/iojs-ko

PR #130, opened 2015-09-12
added an npm start script

simonmcmanus:
Makes npm start command available. Just runs 'npm run serve' but means people can start the website using npm start without having to look at the package.json file.


Fishrock123:
LGTM

PR #132, opened 2015-09-12
remove mailing list from discussion section in get-involved

MassimilianoMura:


mikeal:
We all know that I'm :100:

PR #137, opened 2015-09-12
Reserve some space for the logo

fabiosantoscode:
I observed that on slow networks, the logo takes a bit of time to load, then pushes the whole page down, so I reserved some space for it so that it wouldn't do that.

### DIFF
--- a/layouts/partials/header.hbs
+++ b/layouts/partials/header.hbs
@@ -2,7 +2,7 @@
     <div class="container">
 
         <a href="/{{site.locale}}" id="logo">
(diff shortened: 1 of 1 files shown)


mikeal:
+1

fhemberger:
It would be nice if the logo keeps adjusting to the site's width (as it is supposed to do now via CSS)

fabiosantoscode:
Then if it does, we can apply a padding-bottom hack in CSS.

Since it doesn't resize right now maybe this changeset is okay?

fhemberger:
Thanks, merged.

PR #140, opened 2015-09-12
Add: Code + Learn page in Get Involved section

andygout:


mikeal:
Can we generalize the language because there will be multiple Code + Learns around the world. Then we should just make mention specifically of the London branch.

Issue #52, opened 2015-08-27
index_md anchor links (# appearing onhover on headings in nodejs.org) are missing. 


bogas04:
These anchor links are neat way of referring to a section of a page on social networks/forums. They are missing on new.nodejs.org 

Screenshot from nodejs.org:

![screen shot 2015-08-27 at 9 28 14 pm](https://cloud.githubusercontent.com/assets/6177621/9525617/9080e24e-4d02-11e5-8230-f19f67fbb3c4.png)

fabiosantoscode:
Expect a solution soon, gonna work on this.

phillipj:
Fixed with 294af4f

PR #143, opened 2015-09-12
Anchor links

fabiosantoscode:
Okay, this implements anchor links in headings from markdown text!

I love anchor links, so I really had to do this one.

### DIFF
--- a/build.js
+++ b/build.js
@@ -11,6 +11,7 @@ const markdown = require('metalsmith-markdown');
 const prism = require('metalsmith-prism');
 const stylus = require('metalsmith-stylus');
 const permalinks = require('metalsmith-permalinks');
+const marked = require('marked');
 const path = require('path');
 const fs = require('fs');
 const ncp = require('ncp');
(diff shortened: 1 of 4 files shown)


fabiosantoscode:
(Fixes #52)

fhemberger:
Great idea, still needs some fine tuning (see the link on the bottom right):

https://nodejs.org/en/about/working-groups/#website
bildschirmfoto 2015-09-14 um 18 40 23

Also a little bit of padding would be nice, so the hash doesn't touch the headline. Maybe also a more subtile hover style for those links without the green background?

fabiosantoscode:
Yes, good idea. The font might be too bold too.

phillipj:
Nice, looking forward to this landing 👍

mrkiffie:
I've forked @fabiosantoscode's branch [here](https://github.com/mrkiffie/new.nodejs.org/tree/feature/anchor-links), resolved conflicts and extended the renderer function to support markdown headings that contain links. Here is the [comparison with master](https://github.com/nodejs/new.nodejs.org/compare/master...mrkiffie:feature/anchor-links). Any thoughts?

phillipj:
Nice refactoring w/tests @mrkiffie :+1: Maybe you could address the styling issues @fhemberger pointed out aswell while you're at it?

> Also a little bit of padding would be nice, so the hash doesn't touch the headline. Maybe also a more subtile hover style for those links without the green background?

mrkiffie:
Ok, I'll have a look at those style tweaks

mrkiffie:
@phillipj, I've updated the styles to be similar to those of the attachment in #52.
I also simplified the markup that was being generated as it doesn't appear to serve a purpose.

phillipj:
Very nice @mrkiffie! I would say your branch looks mergeable after squashing your commits into one and opening a PR from your fork.

phillipj:
Fixed with 294af4f, thanks to both of you :+1:

PR #148, opened 2015-09-13
fix broken link to TC info

wonderdogone:
fix broken link to TC info. Simple type on path.


fhemberger:
Good catch, thank you!

PR #161, opened 2015-09-15
Add API Documentation links to previous versions table.

brycebaril:
I have had to url hack this so many times, was stoked that the io.js website had it, let's get it on nodejs.org as well.

selection_033

### DIFF
--- a/layouts/download-releases.hbs
+++ b/layouts/download-releases.hbs
@@ -42,6 +42,9 @@
                                         <a href="{{changeloglink version}}">
                                             {{i18n 'links.pages.changelog'}}
                                         </a>
+                                        <a href="{{apidocslink version}}">
+                                            {{i18n 'docs.text'}}
+                                        </a>
                                     </td>
                                 </tr>
                             {{/project.versions}}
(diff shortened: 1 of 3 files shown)


layouts/download-releases.hbs
```diff
@@ -42,6 +42,9 @@
                                         <a href="{{changeloglink version}}">
                                             {{i18n 'links.pages.changelog'}}
                                         </a>
+                                        <a href="{{apidocslink version}}">
+                                            {{i18n 'releases.documentation'}}
```
Fishrock123: I suggest using `docs.text` instead, for consistency. (defined here: https://github.com/nodejs/new.nodejs.org/pull/161/files#diff-7c141f233ad3e5f2b1dd471e292aec5dR59)

brycebaril:
Updated with @Fishrock123's suggestion:

selection_034

rvagg:
`+    return semver.satisfies(version, '>=1.0.0 <4.0.0') ?` not quite, 3.3.0+ does `docs` .. I'll manually add a symlink for 3.3.1 but a better solution might be to make a redirect for everything on iojs.org so this isn't necessary.

beside that, lgtm

Fishrock123:
LGTM with Rod's nits.

brycebaril:
Updated to use `docs` for all io.js releases as per @rvagg's new rewrite rules on the io.js website. Should be good to go

fhemberger:
Cool, thanks!

Issue #152, opened 2015-09-14
Download page needs redirect


johnkpaul:
In Google, and potentially bookmarks, the node download page still points to https://nodejs.org/download/. That URL is now an Apache index page, I'm pretty sure. The new URL has the 'en' locale in it - https://nodejs.org/en/download/. A redirect is needed between the new and the old page. 

I'm completely guessing that this is Apache and I think that I could whip this up with an .htaccess file, but let me know if that is the preferred solution.

rvagg:
it's nginx, and it's intentionally exposed as such, it can be used to manually navigate to the different builds available for download with /download/release being to what /dist actually points but there is also /download/rc, /download/nightly, etc.

johnkpaul:
Ah I see. Then maybe it should be removed from Google via robots.txt? I don't know if this is intentional or not, but right now Google searches for "download node" go there. 

<img width="839" alt="screenshot 2015-09-15 09 58 11" src="https://cloud.githubusercontent.com/assets/475621/9878405/7c3cbfdc-5b90-11e5-9569-5211583246ab.png">

PR #162, opened 2015-09-16
disallow most of /download/

rvagg:
/cc @johnkpaul - thanks for pointing this out, disallowing /dist/, which it already did, is effectively the same as disallowing /download/ that has been newly introduced since v4.


phillipj:
LGTM

Linking with #152

PR #165, opened 2015-09-16
Add Chinese website

h7lin:


fhemberger:
We were planning to add localization to the nodejs.org website as we had on iojs.org - how about adding your translation under `content/cn`, so releases, template and css fixes always stay up to date for the Chinese version as well? nodejs.cn could redirect to nodejs.org/cn/ as well.

fhemberger:
Pinging @nodejs/iojs-cn as well.

XadillaX:
@fhemberger +1

fengmk2:
Yep, follow the iojs.org way.

laosb:
Good point.

PR #173, opened 2015-09-18
Increase body font weight for readability

silverwind:
This was brought up on [HN](https://news.ycombinator.com/item?id=10234265) (Also more feedback in there regarding scrolling), and I think I have to agree that the body text has reduced readability on non-retina displays at least. Here's the status quo:

screen shot 2015-09-18 at 1 18 50 pm

And here's with this change:

screen shot 2015-09-18 at 1 19 20 pm

We import weights 300,400,600 from Google Fonts right now, and if we're sure 300 is used nowhere else, I'll update the PR to remove this weight for performance.

### DIFF
--- a/layouts/css/base.styl
+++ b/layouts/css/base.styl
@@ -3,7 +3,7 @@
 // Base styles
 body
     box-sizing border-box
-    font 300 20px/1.5 'Source Sans Pro', Open Sans, Roboto, Helvetica, Arial, sans-serif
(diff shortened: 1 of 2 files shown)


phillipj:
LGTM

Cant find `300` anywhere else in the .styl's which should make that weight removable.

silverwind:
Alright, removed the `300` weight.

Issue #123, opened 2015-09-09
contributing to localization in locale directory


JungMinu:
I would like to contribute to localization in `/locale/ko`
Can I ask you please explain how to contribute to `/locale/{}` for localization?
Is there any guidelines related with this issue?
Thanks in advance!

fhemberger:
We're working on that, once everything has settled a bit.

There is also a Korean community around iojs: https://github.com/nodejs/iojs-ko, which has been taken care of by the @nodejs/iojs-ko team. I think it's a good idea to coordinate the translation efforts with them.

JungMinu:
@fhemberger Thanks!

mikeal:
You can translate any markdown file in the english locale to the same file on the other locale and send a pull request for it. We haven't written the docs yet though, we should open an issue for that :)

PR #175, opened 2015-09-18
Download: Switch Windows link between x64/x86

fhemberger:
The homepage (/) points to the x64 installer, but the link on /downloads
(the one above the table of platforms/architectures) points to the x86 installer.

See https://github.com/nodejs/nodejs.org/issues/123

### DIFF
--- a/static/js/download.js
+++ b/static/js/download.js
@@ -4,24 +4,35 @@
   var os = n.platform.match(/(Win|Mac|Linux)/);
   var x = n.userAgent.match(/x86_64|Win64|WOW64/) ||
     n.cpuClass === 'x64' ? 'x64' : 'x86';
+  var text = 'textContent' in d ? 'textContent' : 'innerText';
   var db = d.getElementById('home-downloadbutton');
-  if (!db) { return; }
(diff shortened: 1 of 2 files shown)


static/js/download.js
```diff
@@ -4,24 +4,35 @@
...
-      db[text] = dlLocal + ' Windows (' + x +')';
-      break;
-    case 'Linux':
-      db.href += 'node-' + version + '-linux-' + x + '.tar.gz';
-      db[text] = dlLocal + ' Linux (' + x + ')';
-      break;
+  var version;
+  if (db) {
+    version = db.dataset.version;
+    var dlLocal = db.dataset.dlLocal;
```
phillipj: Still alot of users coming from < IE 11. Just verified locally that this doesnt work for IE 9. Maybe you could fix this issue in this PR aswell?

Using `db.getAttribute('data-version')` and `db.getAttribute('data-dl-local')` respectively does the trick.

fhemberger:
@phillipj Just fixed that. Can you please have a look if it works now?

phillipj:
Works like a charm, good one! :+1:

PR #177, opened 2015-09-19
Fix markdown for a link

yous:
See https://nodejs.org/en/blog/announcements/interactive-2015/.


fhemberger:
Good catch, but better make it `[@nodebotanist](https://twitter.com/nodebotanist)`

yous:
Agree, updated.

fhemberger:
Cool, thank you!

PR #182, opened 2015-09-21
Make the preposition consistent

thefourtheye:
As it is, the title of the page says `ES6 in Node.js`, but the main heading says `ES6 on Node.js`. This patch changes the main heading to `ES6 in Node.js`


fhemberger:
:+1:

PR #186, opened 2015-09-21
Verify binaries while creating release blog post

phillipj:
At the moment of creating the release blog post, there's a chance not all binaries has been uploaded to nodejs.org/dist/ yet.

In these circumstances those specific binaries should have "Coming soon" rather than the actual download URL as those URLs would fail with a 404 page anyway.

This change enables that behaviour by verifying all the binaries about to be included in the blog post, by doing a HEAD request against each of the URLs where they're supposed to be located.

Refs https://github.com/nodejs/new.nodejs.org/issues/156

cc @nodejs/release

### DIFF
--- a/scripts/helpers/downloads.js
+++ b/scripts/helpers/downloads.js
@@ -1,0 +1,69 @@
+const extend = require('util')._extend
+
+const downloads = [
+  {
+    'title': 'Windows 32-bit Installer',
+    'templateUrl': 'https://nodejs.org/dist/v%version%/node-v%version%-x86.msi'
+  },
+  {
+    'title': 'Windows 64-bit Installer',
+    'templateUrl': 'https://nodejs.org/dist/v%version%/node-v%version%-x64.msi'
+  },
+  {
+    'title': 'Windows 32-bit Binary',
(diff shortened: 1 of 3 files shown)


rvagg:
Merged and I used this for v4.1.1. I had to fix the ARMv7 link to make it `armv7l` (missing `l`) but it worked aside from that.

I did a couple of extra things for this release that are not covered by the script:
- removed the `##` header of the release, I don't think we need it given that all of that information is contained in the metadata of the post
- added `author: Rod Vagg`, it would be nice to automate this, we could either use the GitHub API or just web scrape to look up the release tag (`v4.1.1`) and find the commit for that tag and look at the author behind the tag (you can get the `git.name` which is probably the best one to use here). Another alternative is to scrape the README @ https://github.com/nodejs/node/blob/v4.1.1/README.md#current-project-team-members and look at the username / name mapping and match that against the username included in the CHANGELOG.md `##` heading. Or .. just hardwire a mapping in the script here might be easiest!
- added a bit of summary text at the top, but this doesn't need to be automated and often won't be there

great work on this though!

phillipj:
Thanks! I'll add two tasks for the improvements you mentioned.

PR #193, opened 2015-09-24
Tests for ./scripts/release-post.js

phillipj:
Created a test suite for the ever growing release blog post script, with a mix of:
- tape for running tests
- nock for overriding network responses
- proxyquire for mocking `require()`-d modules
- sinon for spying on calls made on fake callbacks

This doesnt make the script 100% covered with tests, although it's a start and ensures most of the current behavior still works when doing changes in the future.

Any thoughts about chosen frameworks or tests?

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,7 +8,7 @@
     "serve": "node build.js serve",
     "load-versions": "node scripts/load-versions.js",
     "start": "npm run serve",
-    "test": "standard build.js scripts/*.js scripts/**/*.js"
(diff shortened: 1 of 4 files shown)


fhemberger:
Haven't worked with tape before (usually using hapi's [lab](https://github.com/hapijs/lab), which includes code coverage as well), but I'm fine with whatever works. And as you are the first one who wrote tests at all, I'm grateful for having them. ;)

PR #194, opened 2015-09-25
variables weren't passed on properly

PatrickHeneise:
fixed:

```
TypeError: Path must be a string. Received undefined
    at assertPath (path.js:8:11)
    at Object.posix.join (path.js:477:5)
    at buildlocale (/Sites/new.nodejs.org/build.js:48:25)
```

and added Barcelona to the NodeConf list.


fhemberger:
Thanks for the update!

PR #204, opened 2015-10-02
Add missing weekly updates.

mikeal:
Ported weekly updates from evangelism repo.


fhemberger:
I think it's fine to push those straight to master in the future without creating a PR first. Not much to review here anyway (and they have already been published).

mikeal:
In the future these will likely be posted as "drafts" and edited by the evangelism WG. These can get posted right away because they already went through that in the old repo but I've sent a PR to change that process so that they get posted directly here.

PR #209, opened 2015-10-06
fix typo

Trott:
It looks like when io.js was replaced with Node.js, it got replaced
with nothing instead in at least one place.


fhemberger:
Thanks for spotting this!

PR #210, opened 2015-10-06
Add "strict mode only" to block scoping in es6 page.

bentinata:
While I was testing node v4.x block scoping, I get error message `uncaughtException SyntaxError: Block-scoped declarations (let, const, function, class) not yet supported outside strict mode`. Perhaps documentation could be a little clearer about this.

### DIFF
--- a/locale/en/docs/es6.md
+++ b/locale/en/docs/es6.md
@@ -15,7 +15,7 @@ All ES6 features are split into three groups for **shipping**, **staged**, and *
 ## Which ES6 features ship with Node.js by default (no runtime flag required)?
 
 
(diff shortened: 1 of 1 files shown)


fhemberger:
You're right, thanks for adding this.

bentinata:
@fhemberger I wonder, why [Travis build is failing](https://travis-ci.org/nodejs/new.nodejs.org/builds/83863996)?

fhemberger:
@bentinata Don't worry, it's not related to your PR. Restarting the build fixed it.
"Go home, Travis, you're drunk!"

PR #218, opened 2015-10-07
Add guide: "Anatomy of an HTTP Request"

Qard:
Note: This is part of the migration of content files from https://github.com/nodejs/docs, see: https://github.com/nodejs/docs/issues/44

### DIFF
--- a/locale/en/docs/guides/anatomy-of-an-http-transaction.md
+++ b/locale/en/docs/guides/anatomy-of-an-http-transaction.md
@@ -397,8 +397,8 @@ http.createServer(function(request, response) {
 }).listen(8080);
 ```
 
-We've now covered most of the basics of handling HTTP requests. At
-this point, you should be able to:
+We've now covered most of the basics of handling HTTP requests. At this point,
+you should be able to:
 
 * Instantiate an HTTP server with a request handler function, and have it listen
 on a port.
--- a/locale/en/docs/guides/anatomy-of-an-http-transaction.md
+++ b/locale/en/docs/guides/anatomy-of-an-http-transaction.md
@@ -397,7 +397,7 @@ http.createServer(function(request, response) {
 }).listen(8080);
 ```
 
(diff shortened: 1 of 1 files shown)


Qard:
@nodejs/documentation @nodejs/website Can I get a LGTM, just for good measure? :)

drewfish:
+1

bengl:
The only thing I would add here is that the links to API docs are all linking to iojs.org, which was because at the time, nodejs.org did not have modern API docs. This should probably be changed, and now's a good time.

Qard:
Ah, yep. I'll fix that. This was just a quick port of the commits. I didn't really look too closely at the content, since it had already gone through review on https://github.com/nodejs/docs.

phillipj:
Nice! Any plans for listing these guides? A completely new section or under docs maybe?

silverwind:
@phillipj yes, I think a 'Guides' section would be great for these kind of things. I had basically the same idea over at https://github.com/nodejs/docs/issues/42.

Qard:
The original plan for the docs wg, discussed back at nodeconf, was to sort docs into topics, guides and reference. We kind of dropped the ball on coordinating with website to figure out how this stuff would all appear though, which is what I'm working on fixing now. :smile_cat:

phillipj:
Looking forward to having these resources on the website :+1:

@Qard so lets merge this content now, and figure out the best way to group/present them later in a separate issue?

Qard:
Ok, it should be ready now. I just had to fix some links.

I'm good with merging and starting another PR to sort out linking to this content.

fhemberger:
Now we have PRs for adding pages as docs, meta docs, guides, topics and knowledge base. This is confusing for the users (at least it is for me now). I think having less categories (and/or a clearer structure) would be a better approach.

Otherwise LGTM.

fhemberger:
Merged, although it's not linked on the site yet.

Issue #153, opened 2015-09-14
How can we reduce 404's


iancrowther:
Moving to "new" code is great but is it leaving a trail of dead bookmarks and blog articles?

Example from http://apmblog.dynatrace.com/2015/04/09/node-js-is-hitting-the-big-time-in-enterprise-markets/

https://nodejs.org/static/video/

This is a 2015 article!

How can we best analyse and add redirects?

rvagg:
You can see the complete current list @ https://raw.githubusercontent.com/nodejs/build/master/setup/www/resources/config/nodejs.org, a lot of it was based on analysing the logs and narrowing down _most_ of the active URLs being hit and deciding what to do with them with @fhemberger. There's no doubt that we ignored a bunch of much less used URLs but we assumed that we'd got most of them.

My plan was to review after a time to see what redirects were actually being used and try and trim the list over time. I guess we should also do a review of 404's to see what's coming up. I could do some log grepping and get back to you on that if you like, I still have the old site's contents that we could cross-reference with and figure out what's real and what's not.

fhemberger:
The video wasn't part of the repository, some files/directories were placed directly on the old server, so for some rare parts, we simply don't know what was linked where. Maybe we still can find the video and put it online again.

As @rvagg already said, we went through all the urls of the old website, mapping them to the new content were possible (except those bits of content which were removed, which now rightfully return a 404), then going through a list of over 4000 URLs returning non-200 responses collected over a month on the old server. After weeding out automatic scanning attempts for things like Wordpress configs or admin interfaces, we tried to find a way to map those to the current website as well.

We will continue to check the logs and add further redirects.

rvagg:
https://gist.github.com/rvagg/50a283b0a35657cdc0f7

top 404s for the site since the server move sorted by frequency, I've removed some of the obvious bogus ones and left off all the 1's and 2's which are pretty much all guaranteed to be bogus. If you take out `ivy.xml` and `.jar$` then you reduce the list by ~3k down to ~5k.

Some creative grepping leaves with this list of first-round candidates:
- 6813 /static/images/icons-interior.png
- 6810 /static/images/platform-icons.png same as above
- 5802 /static/images/twitter-bird.png
- 3395 /static/images/footer-logo-alt.png

These are all mostly external referrers but also http://nodejs.org/dist/v0.10.32/docs/api/assets/style.css and other 0.10.x docs. I'm not inclined to "fix" them tbh but perhaps have other opinions?
[...]

PR #227, opened 2015-10-09
Add legacy website images

fhemberger:
See #153


bnb:
+1. Looks fine to me. I read the other issue, looks like covers what can we can do at the moment.

mikeal:
Should we start using large file storage for this stuff?

fhemberger:
I just added around 15 small files and optimized the images' compression before checking them in. That's all.

Issue #156, opened 2015-09-15
Future release blog post improvements


phillipj:
Just before node v4 got released, a script ([scripts/release-post.js](https://github.com/nodejs/new.nodejs.org/blob/master/scripts/release-post.js)) for automatically creating release blog post ([locale/en/blog/release/v4.0.0.md](https://github.com/nodejs/new.nodejs.org/blob/master/locale/en/blog/release/v4.0.0.md)) was created in #88, atleast a first draft of it.

I would like to create a list of tasks to further improve that script, but it requires feedback about the current state of it.
- [x] Fetch changelog from tag instead of master
- [x] Add CLI argument to specify version (w/fallback to https://nodejs.org/dist/index.json)
- [x] Leave pleaceholder for shasums if they're not found in /dist/
- [x] Check to see if all of the binaries are available before putting links to them
- [x] Remove `##` header from changelog
- [x] Add `Author: <full name>` in blog header, see [#186 comment](https://github.com/nodejs/new.nodejs.org/pull/186#issuecomment-142472144) for hints

Most importantly: was it useful at all while releasing v4 @rvagg? If not, what has to change?
### Current flow

There were some assumptions made about how the latest version number should be resolved, what changelog too use etc.

```
a) Fetch latest version from https://nodejs.org/dist/index.json
    |
    v
b) Fetch changelog from node's master branch, find changelog for latest version
    |
    v
c) Fetch shasums for latest version from https://nodejs.org/dist/
    |
    v
d) Render blog post with Handlebars template ./scripts/release.hbs
    |
    v
c) Write content to file ./locale/en/blog/release/v<VERSION>.md
```
### Changelog

One thing I got curious about while observing the release, was which changelog should be used. It didnt look the v4's changelog got merged into master at the time of creating the release blog post, therefore the changelog retrieval would probably fail due to version mismatch.

What would be best approach? (1) from the [release proposal v4.0.0](https://github.com/nodejs/node/pull/2715) or (2) [v4.x tagged changelog](https://github.com/nodejs/node/blob/v4.x/CHANGELOG.md)?

Any other thoughts?

Fishrock123:
cc @nodejs/release

rvagg:
For v4.0.0 I did the reverse and generated the post before the release, perhaps it's a special case because I did a writeup in there as well and maybe it'll be useful to generate this after the release is cut.

[...]

PR #231, opened 2015-10-11
Resolved full name of release author in release blog

phillipj:
In our release blog posts, we want to display the full name of the dev who cut the release, instead of the github login such as `rvagg` or `Fishrock123`. This was proposed by @rvagg in [#186 comment](https://github.com/nodejs/new.nodejs.org/pull/186#issuecomment-142472144).

These changes fixes that by resolving the full name via github's API, based on their github login specified in the changelog.

Also did some refactoring of some hairy changelog regex matching, to something a bit less complex and readable.

Refs #156.

/cc @nodejs/release

### DIFF
--- a/scripts/release-post.js
+++ b/scripts/release-post.js
@@ -31,18 +31,25 @@ const downloads = require('./helpers/downloads')
 
 function request (uri, method, cb) {
   return new Promise(function (resolve, reject) {
-    const opts = extend({ method }, url.parse(uri))
+    // user-agent is required when by api.github.com
+    const opts = extend({
+      method,
+      headers: {
+        'user-agent': 'nodejs.org release blog post script'
+      }
+    }, url.parse(uri))
+
     let data = ''
 
     https.request(opts, function (res) {
       if (res.statusCode !== 200) {
-        return reject(new Error('Invalid status code (!= 200) while retrieving ' + url + ': ' + res.statusCode))
+        return reject(new Error(`Invalid status code (!= 200) while retrieving ${uri}: ${res.statusCode}`))
       }
 
       res.on('data', function (chunk) { data += chunk })
       res.on('end', function () { resolve(data) })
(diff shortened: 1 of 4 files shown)


mikeal:
Looks like there's a failure in the slug test.

phillipj:
> Looks like there's a failure in the slug test.

Great! Didnt touch that code at all, and it passed after a couple of Travis restarts. Dont see any apparent reason for it to be unstable, we'll have keep an eye on that slug test.

mikeal:
weird, should probably log another issue for that test being flaky.

scripts/release-post.js
```diff
@@ -71,41 +72,56 @@ function findLatestVersion (cb) {
...
+    .then((author) => download(`https://api.github.com/users/${author}`))
+    .then(JSON.parse)
+    .then((githubRes) => githubRes.name)
+}
+
 function fetchChangelog (version) {
+  // matches a complete release section,
+  // support release sections with headers like:
+  // ## 2015-09-22, Version 4.1.1 (Stable), @rvagg
+  const rxSection = new RegExp(`## \\d{4}-\\d{2}-\\d{2}, Version ${version} \\(Stable\\)[\\s\\S]*?(?=## \\d{4})`)
```
rvagg: We're going to need to extend this, it'll include "LTS" and "Maintenance" too. Perhaps just `\\([^\\)]+\\)` would do.

phillipj: Sure thing! Just pushed an update.

phillipj:
@nodejs/release I just pushed an update to this PR, supporting version policy and codename aswell, as a response to #239.

Also removed version policy from the blog post slug, as I dont see the big value of it, other adding more complexity into this script. Agreed @nodejs/website?

phillipj:
Merging this now so author will be part of the next release post.

PR #242, opened 2015-10-14
Add localization resources

ghost:
I'll just start adding localization stuff to make this happen. The file I added is an altered version of the one found in the nodejs/iojs.org repo, which is changed to reflect the new localization repos and accommodate to the new file structure.

If I missed anything, let me know, since I still need to actually look at how the website itself works :smile:

### DIFF
--- a/TRANSLATION.md
+++ b/TRANSLATION.md
@@ -40,7 +40,7 @@ would like to contribute to the translation of nodejs.org, please refer to the f
 * Ensure that any site translations are done as pull requests into this repo. This will ensure the build process, layout, and styling, remain consistent across the different translations of the site.
 * You can find the appropriate language folder within `locale/`
 * There needs to be the following files in your language folder:
-    * `site.json` (this fills in the buttons and title bar with the appropriate translation)
+    * `site.json` (this contains the basic settings for the website version and the button and title bar translations)
     * `index.md` (this contains the markdown translation for the home page. The paragraph order is important here, so please maintain it)
     * `styles.styl` (this imports the necessary Stylus files)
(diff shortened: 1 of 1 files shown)


bnb:
Are you still working on this? If so, I think we should restart the list on the base translation file - I don't think we want/need all the io.js ones (correct me if I'm wrong!).

bnb:
Do you want me to modify this once it is accepted, or do you want me to comment on it as-is? The former would probably be better in terms of modularity of issues.

mikeal:
I say we merge it, although we're going to have to update most of those links soon since the locale groups are going through name changes.

ghost:
@mikeal Fair enough, I wasn't too sure on how to include those there.

outsideris:
You missed some repos for localization.

```
[iojs-ar](https://github.com/nodejs/iojs-ar)
[iojs-bg](https://github.com/nodejs/iojs-bg)
[nodejs-ko](https://github.com/nodejs/nodejs-ko)
[iojs-ms](https://github.com/nodejs/iojs-ms)
[nodejs-ta](https://github.com/nodejs/nodejs-ta)
```

fhemberger:
LGTM, please add the groups @outsideris mentioned and we're good to go.

ghost:
Real quick, what country exactly is `nodejs-ta`? I couldn't find anything.

fhemberger:
@sup: Not a country, it's the language Tamil: https://en.wikipedia.org/wiki/Tamil_language

ghost:
Alright, it should be good.

fhemberger:
Great, thanks!

PR #251, opened 2015-10-16
Update TRANSLATION.md

fhemberger:
- Define naming structure for language directories
- Update meaning of `site.json`and `index.md`
- Add some spacing (makes reading the raw file easier)


phillipj:
LGTM. Should we link this from the readme, just to be more explicit about its existence?

fhemberger:
Linking from the README and CONTRIBUTING should be fine for now.

Issue #234, opened 2015-10-12
Log of unstable tests


phillipj:
We're having some issues with unstable and flaky test runs on Travis. To list incidents and not forget about them, please add a comment with the name of the test of you experience it when creating PRs / merging to master.

phillipj:
`anchorMarkdownHeadings/makes pretty slugs` when opening PR with unrelated changes #231. Fixed after two Travis restarts.

mrkiffie:
Do we have any more details of the `anchorMarkdownHeadings/makes pretty slugs` test failure or steps to replicate?

phillipj:
The short answer is no. Wasnt able to re-produce it locally, nor see anything unusual in the Travis logs.

fhemberger:
Can we increase the timeout for those tests (`t.timeoutAfter(ms)`) and see if the situation improves? Default seems to be 500ms, maybe Travis takes sometimes just a tad longer?

phillipj:
TBH I would like to see timeouts occur, before trying to increase the test timeouts. The regex code related to making these slugs should _really_ not use more than 500ms.

Very sorry I didnt copy the Travis output when it was failing :disappointed:

@fhemberger I've seen you commenting some PRs with "ignore the errors, Travis is drunk". Do you remember which tests were failing in those cases?

stevemao:
I have some experience with unknown travis fails with some bigger project (https://github.com/ajoslin/conventional-changelog/commit/40ab546af99b230d1ec0a43e86605ee7786e9dcf and https://github.com/ajoslin/conventional-changelog/commit/104b3aae5d65e2fc73b7cbee48ff5f56050a50e8). I believe it might be travis own problem. Timeout might not be the issue (I've increased twice). The worse part is the error message doesn't tell much and it's hard to replicate :(

phillipj:
Closing this for now, lets open it again if unstable tests starts showing again.

PR #257, opened 2015-10-17
Changing tests to prevent tests from intermittently timing out

mrkiffie:
#234 mentions the intermittent failure of the test `anchorMarkdownHeadings/makes pretty slugs` on Travis-CI.

### DIFF
--- a/tests/scripts/anchor-mardown-headings.test.js
+++ b/tests/scripts/anchor-mardown-headings.test.js
@@ -5,6 +5,8 @@ const test = require('tape')
 test('anchorMarkdownHeadings', (t) => {
   const anchorMarkdownHeadings = require('../../scripts/plugins/anchor-markdown-headings')
 
+  t.timeoutAfter(500)
+  t.plan(4)
   t.test('correctly pharses markdown heading without links', (t) => {
     const text = 'Simple title'
     const level = 1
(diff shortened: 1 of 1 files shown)


tests/scripts/anchor-mardown-headings.test.js
```diff
@@ -5,6 +5,8 @@ const test = require('tape')
 test('anchorMarkdownHeadings', (t) => {
   const anchorMarkdownHeadings = require('../../scripts/plugins/anchor-markdown-headings')
 
+  t.timeoutAfter(500)
```
phillipj, line 8: Will this have any effect as 500ms is the default already? https://github.com/substack/tape#testname-opts-cb

phillipj:
Have you been able to reproduce these failures by any chance?

tests/scripts/anchor-mardown-headings.test.js
mrkiffie, line 8: Base on what I was able to gather from an [issue logged for tape](https://github.com/substack/tape/issues/178), this forces the timeout to be measured from the time of execution instead of load time. I could be completely wrong.

mrkiffie:
I've been able to replicate the `operator: fail` failure by reducing the timeout to single digits.
I saw the following log that fails: https://travis-ci.org/nodejs/new.nodejs.org/builds/85753833

Locally I get the following:

```
  not ok 1 test timed out after 5ms
    ---
      operator: fail
    ...
```

The error in the log is slightly different - `test exited without ending` instead of `test timed out ...`.

I'm hoping that specifying the number of plans will resolve the unexpected ending.

phillipj:
Okey, lets give it a try and keep our eyes open for these failures..

stevemao:
This can probably be reverted as https://github.com/nodejs/new.nodejs.org/pull/265 is merged

Issue #200, opened 2015-09-29
Prepare for v4 LTS and v5


rvagg:
From https://github.com/nodejs/node/issues/3000 where this question is something to be resolved by the @nodejs/website WG:
- [x] How do publicise and promote LTS lines? Does @nodejs/website have ideas on how to make this happen when we have something called "LTS" and then start to get new "Stable" releases with a `5` at the front? Ideally, users will be self-selecting and would end up with the kind of release suited to them, perhaps we need to outline on the front page of the website and/or on the download page why you would choose LTS over Stable?

Until the first item in the list @ https://github.com/nodejs/node/issues/3000 is resolved I think you need to be prepared for the following scenarios:
1. a v5 and a v4 LTS: this will be the eventual situation no matter what but this is a possible scenario from **next week** onward
2. a v4 LTS _only_: do you do anything special for this or just treat it as yet another v4? It's possible we have this scenario **next week**

Also, we probably should have wording for what LTS actually means, to put _somewhere_, feel free to kick that over to @nodejs/lts and make them do it if you want, either open a new issue and describe exactly what you need or add it in https://github.com/nodejs/LTS/issues/43.

phillipj:
Will there be changes to https://nodejs.org/dist/index.json?

Fishrock123:
cc @nodejs/evangelism 

@mikeal can we bring anyone from foundation-level marketing in to discuss this?

rvagg:
@phillipj perhaps a `lts` column/field in index.json and index.tab would be good, it could include the codename for lts releases

phillipj:
@rvagg sounds like a good plan. I'll could do some tests with that in mind for scenario 1, probably in the weekend / early next week.

For scenario 2, as long as file structures on /dist/ stays the same and we find it in index.json, no changes would be needed for us. Although adding some hard coded LTS strings around could be helpful to emphasize it, e.g. `Current version: v4.x.x LTS` on the frontpage

bnb:
From an Evangelism perspective, there shouldn't be much of a problem. Weekly or twice-weekly tweets, as well as listing of updates in blog posts, are easy. However, will we be attempting to pull people in to either using or contributing to LTS? Will it have independent contributors, outside of core?

phillipj:
Alright, to get the ball rolling I'll post some screenshots with some minimal changes necessary to host _two_ release lines at the same time. All kind of feedback is much appreciated!

[...]

PR #260, opened 2015-10-19
Display LTS and Stable on the frontpage

phillipj:
Until v5.0.0 exists in https://nodejs.org/dist/index.json, only v4.x LTS will be displayed, in other words just as before besides some design changes:

lts screenshot

Will display two buttons side by side automagically, as soon as `index.json` contains a v5.0.0 release which is not marked as `lts`:

skjermbilde 2015-10-19 kl 10 52 47

Any thoughts?

Refs https://github.com/nodejs/new.nodejs.org/issues/200.

### DIFF
--- a/layouts/index.hbs
+++ b/layouts/index.hbs
@@ -17,14 +17,32 @@
                         {{ labels.current-version }}: {{ project.currentVersion }}<br>
                         {{{ project.banner.content }}}
                     </p>
-                {{else}}
-                      <p class="home-version">{{ labels.current-version }}: {{ project.currentVersion }}</p>
                 {{/if}}
 
-                <a href="https://nodejs.org/dist/{{ project.currentVersion }}/" id="home-downloadbutton" data-version="{{ project.currentVersion }}" data-dl-local="{{ labels.download-for }}">{{ labels.download }}</a>
+                <h2 id="home-downloadhead" data-dl-local="{{ labels.download-for }}">{{ labels.download }}</h2>
+
+                <a href="https://nodejs.org/dist/{{ project.currentVersions.lts }}/" class="home-downloadbutton" title="{{ labels.download }} {{ project.currentVersions.lts }} {{ labels.lts }}" data-version="{{ project.currentVersions.lts }}">
+                    {{ project.currentVersions.lts }} {{ labels.lts }}
+                    <small>{{ labels.tagline-lts }}</small>
+                </a>
+
+                {{#if project.currentVersions.stable}}
+                    <a href="https://nodejs.org/dist/{{ project.currentVersions.stable }}/" class="home-downloadbutton"  title="{{ labels.download }} {{ project.currentVersions.stable }} {{ labels.stable }}"  data-version="{{ project.currentVersions.stable }}">
+                        {{ project.currentVersions.stable }} {{ labels.stable }}
+                        <small>{{ labels.tagline-stable }}</small>
+                    </a>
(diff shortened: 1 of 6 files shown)


layouts/index.hbs
```diff
@@ -17,14 +17,30 @@
...
                         {{{ project.banner.content }}}
                     </p>
-                {{else}}
-                      <p class="home-version">{{ labels.current-version }}: {{ project.currentVersion }}</p>
                 {{/if}}
 
-                <a href="https://nodejs.org/dist/{{ project.currentVersion }}/" id="home-downloadbutton" data-version="{{ project.currentVersion }}" data-dl-local="{{ labels.download-for }}">{{ labels.download }}</a>
+                <h2 id="home-downloadhead" data-dl-local="{{ labels.download-for }}">{{ labels.download }}</h2>
+
+                <a href="https://nodejs.org/dist/{{ project.currentVersions.lts }}/" class="home-downloadbutton" title="Download {{ project.currentVersions.lts }} LTS" data-version="{{ project.currentVersions.lts }}">
```
fhemberger: Texts need to be fully translatable, even "LTS" might need to be translated as well.

locale/en/index.md
```diff
@@ -4,7 +4,7 @@ labels:
   current-version: Current Version
   download: Download
   download-for: Download for
-  other-downloads: Other Downloads
+  other-downloads: Other LTS Downloads
```
fhemberger: Should probably be two fields: "Other Downloads", "Other Stable/LTS Downloads"

layouts/index.hbs
phillipj: :+1:

locale/en/index.md
phillipj: Just pushed an update fixing these issues.

phillipj:
Just too clarify, this can be merged now rather than syncing with the release of v5.0.0. It was made that way intentionally as I noticed @rvagg commented (somewhere?) that whenever a new release got promoted to http://nodejs.org/dist, it automatically kicked of a rebuild of the website.

bnb:
+1, a solid job and indicates that LTS is important, too.

phillipj:
I think it would be valueable for us to track clicks on these links. Any chance I could access to the GA stats for nodejs.org @mikeal?

jbergstroem:
I'd like to see a simple explanation of LTS if we're to introduce both buttons. "What should I download"

phillipj:
@jbergstroem would a couple of keywords like shown in [#200 comment](https://github.com/nodejs/new.nodejs.org/issues/200#issuecomment-149282667) suffice?

phillipj:
> I'd like to see a simple explanation of LTS if we're to introduce both buttons. "What should I download"

In response to @jbergstroem and other similar comments in #200, I just pushed an update with @rvagg suggested taglines.

#### v4.2.1 only (before v5 exists)

skjermbilde 2015-10-28 kl 11 53 47

#### v4.2.1 and v5

skjermbilde 2015-10-28 kl 11 53 06

rvagg:
+1 let's queue this up, aiming for a v5 release tomorrow, if that slips then next Monday or Tuesday at the latest

fhemberger:
Merged: e53ddb3.

I changed the wording from "Other LTS Downloads/Stable" to "Other Downloads: LTS/Stable".

phillipj:
Nice @fhemberger :+1:

PR #261, opened 2015-10-19
Display LTS and Stable on the download page

phillipj:
A pragmatic approach to toggling between LTS and Stable downloads. Duplicated the download page to have two separate pages: `/download` for LTS and `/download/stable` for Stable.

Two links in the upper right corner will appear as soon as the https://nodejs.org/dist/index.json contains _both_ an LTS and a Stable v5.x release. Until then, no extra links will be visible.
#### Default is LTS

screen shot 2015-10-19 at 20 15 14
#### When toggled to Stable

screen shot 2015-10-19 at 20 15 21

Any thoughts? Are we okey with two different pages/URLs or would doing some JS trickery by changing the link URLs on the page be preferred?

Refs #200.

### DIFF
--- a/layouts/css/page-modules/_download.styl
+++ b/layouts/css/page-modules/_download.styl
@@ -1,3 +1,20 @@
+.download-header
+
+    h1
+        width 40%
+        float left
+
+    ul
+        margin-top 2rem
+        width 40%
+        float right
+        text-align right
+
+.download-header::after
+    display block
+    clear both
+    content ""
+
 .download-hero
     display table
     margin-bottom 1em
--- a/layouts/download-stable.hbs
+++ b/layouts/download-stable.hbs
@@ -10,20 +10,17 @@
             <article>
                 <div class="download-header">
                     <h1>{{downloads.headline}}</h1>
-
-                    {{#if project.currentVersions.stable}}
                     <ul class="list-divider-pipe">
                         <li>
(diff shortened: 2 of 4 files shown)


bnb:
+1, LGTM.

phillipj:
@fhemberger @mikeal in regards to translation efforts, are you okay with having duplicate translation (and layout) files for the LTS and Stable download pages? The biggest advantages of doing this is possibly having different content on the these two pages and less JS trickeries.

fhemberger:
@phillipj I would prefer having a partial where you inject either `project.currentVersion.stable` or `project.currentVersion.lts`, e.g:

```
{{> download-matrix version=project.currentVersion.stable}}
```

This would be much easier to keep in sync.

mikeal:
@phillipj also, you can already add localization specific styles if what you nee to accomplish is doable with just CSS.

phillipj:
@fhemberger good idea about using a partial :+1: Just pushed an update adding `download-matrix`

phillipj:
@fhemberger I just fixed the merge conflict after you merged #260

fhemberger:
@phillipj Dang, just merged it (fixing the conflict). Can you please have a look if I merged the correct solution in 6007cb7461b31ab07df8c6348d4f383bd503b7cd?

phillipj:
@fhemberger no worries, both versions of the conflict worked :)

PR #266, opened 2015-10-21
Test: Use default timeout

stevemao:
Since #265 is merged and the timeout bug is fixed in tape.

This partially reverts #257.


phillipj:
Just did some digging, and couldnt find any default timeout in tape at all. Made a simple test which waited for 5s before ending, which works as expected, no errors.

``` js
const test = require('tape')

test('no default timeout', (t) => {
  setTimeout(() => {
    t.end()
  }, 5000)
})
```

In other words this change removes the timeout settings completely. That might be just fine, as the initial reason `timeoutAfter(500)` got there in the first place, was because we didnt want long test runs to fail the tests.

mrkiffie:
:+1:  LGTM

PR #268, opened 2015-10-21
Update ES6 docs for v5.0.0

evanlucas:
With node v5.0.0 will come an updated v8 that includes additional ES6
features. This adds some of those additional features to the ES6 doc

**This should not be merged until v5.0.0 is released**

### DIFF
--- a/locale/en/docs/es6.md
+++ b/locale/en/docs/es6.md
@@ -55,11 +55,23 @@ All ES6 features are split into three groups for **shipping**, **staged**, and *
 
 * [Arrow Functions](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Arrow_functions)
 
+* [new.target](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/new.target)
+
+* [Object.assign](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/assign)
+
+* [Spread operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Spread_operator)
+
 You can view a more detailed list, including a comparison with other engines, on the [compat-table](https://kangax.github.io/compat-table/es6/) project page.
 
 ## Which ES6 features are behind the --es_staging flag?
 
 * [`Symbol.toStringTag`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol) (user-definable results for `Object.prototype.toString`, behind flag `--harmony_tostring`)
+
+* [`Array.prototype.includes`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/includes) (determine whether
+an array includes a certain element, behind flag `--harmony_array_includes`)
+
+* [Rest Parameters](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/rest_parameters) (represent an indefinite number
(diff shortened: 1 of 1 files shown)


locale/en/docs/es6.md
```diff
@@ -55,12 +55,26 @@ All ES6 features are split into three groups for **shipping**, **staged**, and *
 
 * [Arrow Functions](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Arrow_functions)
 
+* [new.target](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/new.target)
+
+* [Object.observe](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/observe)
```
mgol: Object.observe is not in ES6 (there is no standard for it so far) and Node 4 contains it as well. I don't think it belongs on this list.

stevemao: Agree @mzgol 

@evanlucas this should go to ES2016 (Is there a list of ES2015/ES2016 that v8 supports? This could be in sync with that list)

mgol: Perhaps ES.Next or sth to that effect would be better; it's not final yet whether `Object.observe` will make it into 2016 AFAIK.

evanlucas: Ah good catch. While I agree it shouldn't be in the ES6 section, I do think it should be mentioned (somewhere). I'll remove it from here though

evanlucas:
Ok, updated. PTAL

mgol:
Looks good now. :)

phillipj:
Ooh, looking forward to getting these features natively :)

locale/en/docs/es6.md
```diff
@@ -55,12 +55,24 @@ All ES6 features are split into three groups for **shipping**, **staged**, and *
 
 * [Arrow Functions](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Arrow_functions)
 
+* [new.target](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/new.target)
+
+* [Object.assign](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/assign)
```
mgol, line 60: This is already in Node 4 BTW.

evanlucas, line 60: Yea, I don't think this document was updated then.

evanlucas:
Should we change this title to `ES 2015` since that is what the official spec is now?

fhemberger:
@evanlucas You are right … everybody still calls it ES6 for short. ;) What wording does v8 use? Just had a quick look but couldn't find anything. How about "ECMAScript 2015 (ES6)" (taken from MDN)?

targos:
@fhemberger V8 team uses exactly what you are suggesting: http://v8project.blogspot.ch/

fhemberger:
Ok, I'm adding it to the commit and merge it into the v5.0.0 release branch

fhemberger:
Merged to `v5.0.0`: 8564c4f, waiting for release.

phillipj:
How could these v5 features be marked exclusively for v5? As the site now hosts _both_ v4 and v5, this ES2015 features page also has to take that into consideration.

fhemberger:
@phillipj Just add a second footnote "Only available in Node.js >= 5.x.x"

locale/en/docs/es6.md
mgol: `Object.observe` is dead: https://mail.mozilla.org/pipermail/es-discuss/2015-November/044684.html

PR #283, opened 2015-10-25
replace custom logic in loadVersions with node-version-data package

rvagg:
I'm encouraging the npm team to use this same package to fetch the data they need for mapping node to npm versions

### DIFF
--- a/package.json
+++ b/package.json
@@ -41,6 +41,7 @@
     "metalsmith-serve": "0.0.3",
     "metalsmith-stylus": "1.0.0",
     "ncp": "2.0.0",
+    "node-version-data": "1.0.0",
     "octonode": "0.7.4",
     "require-dir": "0.3.0",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -41,6 +41,7 @@
     "metalsmith-serve": "0.0.3",
     "metalsmith-stylus": "1.0.0",
     "ncp": "2.0.0",
+    "node-version-data": "~1.0.0",
```
evanlucas: do we need to pin to an exact version here? It seems like that is the current way of doing it.

evanlucas:
LGTM

package.json
stevemao: Yes, we are using greenkeeperio here. example: https://github.com/nodejs/new.nodejs.org/pull/279

rvagg: whoops, didn't realise! fixed

PR #284, opened 2015-10-26
Update optional build tools instruction.

mattnwa:
Need to have sudo in the apt-get install portion.

### DIFF
--- a/locale/en/download/package-manager.md
+++ b/locale/en/download/package-manager.md
@@ -57,7 +57,7 @@ apt-get install --yes nodejs
 To compile and install native addons from npm you may also need to install build tools:
 
 ```
-apt-get install --yes build-essential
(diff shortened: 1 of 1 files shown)


stevemao:
I think this is a permission issue on your computer. Have you tried doing research on this?

mattnwa:
I believe following the instructions this would happen to any new linux user, until they were to send the sudo command. I know the password is usually cached for 5 minutes but sudo command is still needed to install the optional build tools if user is not su... If i am totally wrong I apologize!

fhemberger:
I updated the instructions based on the latest nodesource documentation (which is the basis for this document):

``` sh
# use `sudo` on Ubuntu or run this as root on debian
apt-get install -y build-essential
```

Thanks for pointing this out!

PR #286, opened 2015-10-27
move index page style block into stylus files

rnsloan:
- move interactive banner styles into new stylus file
- add parameter to footer partial on homepage so it has a specific html class attribute and has specific css applied only on that page

### DIFF
--- a/layouts/partials/footer.hbs
+++ b/layouts/partials/footer.hbs
@@ -1,6 +1,10 @@
 <a href="#" id="scrollToTop">&uarr; <span>{{i18n.scrollToTop}}</span></a>
 
-<footer role="contentinfo">
+{{#if className}}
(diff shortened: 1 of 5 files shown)


layouts/partials/footer.hbs
```diff
@@ -1,6 +1,10 @@
 <a href="#" id="scrollToTop">&uarr; <span>{{i18n.scrollToTop}}</span></a>
 
-<footer role="contentinfo">
+{{#if className}}
```
fhemberger: `<footer{{#if className}} class="{{className}}"{{/if}} role="contentinfo">`

rnsloan: @fhemberger updated conditional

fhemberger:
Thanks!

PR #287, opened 2015-10-27
metalsmith-layouts: Handles partials > removed map-handlebars-partials

woodyrew:
Metalsmith-layouts automatically registers all partials from within a
directory if a string is passed to `partials` with a directory.

_Note:_ **No version bump has been applied.**


build.js
```diff
@@ -143,7 +142,7 @@ function buildlocale (source, locale) {
     .use(layouts({
       engine: 'handlebars',
       pattern: '**/*.html',
-      partials: mapHandlebarsPartials(metalsmith, 'layouts', 'partials'),
+      partials: 'layouts/partials',
```
phillipj, line 145: Does `metalsmith-layouts` handle this in a windows compat manner or should you use `path.join()`?

phillipj:
Nice, deleting code == great success :+1:

fhemberger:
Nice, so this was finally fixed in `metalsmith-layouts`. :+1:

build.js
fhemberger, line 145: Rest of the script uses `path.join()` so we should stick to that.

@woodyrew Can you add this small change, please? Then it's good to merge.

woodyrew, line 145: `metalsmith-layouts` handles the paths. See https://github.com/superwolff/metalsmith-layouts/blob/master/lib/helpers/read-partials.js#L22

`partials: path.join(__dirname, 'layouts', 'partials'),`
Does **not** work.

PR #292, opened 2015-10-29
footer: change "here" link

Trott:
Context: Avoiding link text consisting of `here` (or `click here` or `more` etc.) is beneficial to users of screen readers and others who may tab from link to link.


fhemberger:
Good catch, thanks!

Trott:
Related: It would be great if that link could be to a web page rather than a PDF doc. But I totally understand that may be something not within our control.

PR #299, opened 2015-10-29
LTS and Stable changelog links on the frontpage

phillipj:
We're currently only link to the stable changelog on the frontpage. This adds a link to LTS changelog aswell.

skjermbilde 2015-10-29 kl 23 17 32

### DIFF
--- a/layouts/index.hbs
+++ b/layouts/index.hbs
@@ -45,7 +45,11 @@
                             <a href="https://nodejs.org/dist/{{ project.currentVersions.lts }}/">{{ labels.other-lts-downloads }}</a>
                         {{/if}}
                     </li>
-                    <li><a href="https://github.com/nodejs/node/blob/{{ project.currentVersion }}/CHANGELOG.md">{{ labels.changelog }}</a></li>
+                    <li>
+                        {{ labels.changelog }}:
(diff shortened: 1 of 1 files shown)


layouts/index.hbs
```diff
@@ -45,7 +45,10 @@
                             <a href="https://nodejs.org/dist/{{ project.currentVersions.lts }}/">{{ labels.other-lts-downloads }}</a>
                         {{/if}}
                     </li>
-                    <li><a href="https://github.com/nodejs/node/blob/{{ project.currentVersion }}/CHANGELOG.md">{{ labels.changelog }}</a></li>
+                    <li>
+                        {{ labels.changelog }}:
+                        <a href="https://github.com/nodejs/node/blob/{{ project.currentVersions.lts }}/CHANGELOG.md">{{ labels.lts }}</a> &#47; <a href="https://github.com/nodejs/node/blob/{{ project.currentVersions.stable }}/CHANGELOG.md">{{ labels.stable }}</a>
```
stevemao: Can we break this line the same as "downloads"?

stevemao:
LGTM :)

layouts/index.hbs
phillipj: > Can we break this line the same as "downloads"?

Sure, just updated :)

PR #300, opened 2015-10-30
add LTS to releases table

rvagg:


scripts/load-versions.js
```diff
@@ -5,7 +5,16 @@
...
 const nodeVersionData = require('node-version-data')
 
-module.exports = nodeVersionData
+module.exports = function (callback) {
+  nodeVersionData((err, versions) => {
+    if (err) {
+      return callback(err)
+    }
+
+    versions.forEach((v) => v.lts === false && delete v.lts)
```
stevemao: I believe Handlebars.js can ignore `false` values so no need to `delete` `false` property

layouts/download-releases.hbs
```diff
@@ -33,6 +34,7 @@
                             {{#project.versions}}
                                 <tr>
                                     <td data-label="Version">{{name}} {{version}}</td>
+                                    <td data-label="LTS">{{lts}}</td>
```
stevemao: So instead of deleting the property, this could be `{{#if lts}}{{lts}}{{/if}}`

scripts/load-versions.js
rvagg: no, I added this exactly because it didn't ignore it and I ended up with a bunch of `false`s in the column

stevemao: @rvagg how do you like the `{{#if lts}}{{lts}}{{/if}}` solution I suggested below?

rvagg: sorry, missed it, have updated to use that instead, tbh I didn't know that was possible although I did a quick google for such syntax!

stevemao: oh ok :) not saying this is better though.

stevemao:
`lts` fields are deleted. @rvagg can you rerun `npm run load-versions` :)

PR #304, opened 2015-10-30
Document node v5.x exclusive ES2015 features

phillipj:
As the site now hosts both v4 and v5, ES2015 features page should document which features are exclusive for v5.

Refs #268.


fhemberger:
LGTM

PR #306, opened 2015-10-30
Show both stable and LTS docs links on homepage

evanlucas:
Fixes: https://github.com/nodejs/new.nodejs.org/issues/305


fhemberger:
Thanks!

phillipj:
Wow, fast as lightning there @evanlucas! :+1:

shelleyp:
Please re-open this. You need to show both sets of docs for both active distributions on this page at a minimum https://nodejs.org/api/ - or you need to eliminate this page in favor of links in the front page to both sets of docs. And I would suggest a link to the overall description of the release cycle, too.

Note: the links need to show at the top-level menu, not in small print under the releases.

PR #315, opened 2015-11-01
Removes reference to nodebug.me in FAQ

alesandroortiz:
Nodebug.me appears to currently be down (on 10/31/2015). There's no mentions or activity on the Web about it this year (2015). Repo (https://github.com/nodebugme) also appears dead since late last year, only with some issues made earlier this year before the Node.js foundation was formed.

Considering it seemed like a great tool, would be great to revive it, but in the interim, this removes the dead link.


fhemberger:
Thank you!

alesandroortiz:
You're welcome!

PR #317, opened 2015-11-02
Use * instead of _ as em tags in markdown

xcatliu:
In [this page](https://nodejs.org/en/blog/release/v5.0.0/),

`_node_modules_` compiled to a wrong result: `_node<em>modules</em>`

 Problem solved when using `*` instead of `_` as `<em>` tags in markdown.


phillipj:
Thanks! :+1:

Issue #305, opened 2015-10-30
Support multiple documentation version sets


shelleyp:
You're supporting two Node versions at a time. 

But you're only providing one set of documents--for the version of Node that's going away in April. 

If you're going to support two versions of Node, you have to provide access to the API documentation for both versions.

fhemberger:
If you look at https://nodejs.org/en/docs/, we are providing API docs for [0.10.x](https://nodejs.org/docs/latest-v0.12.x/api/), [0.12.x](https://nodejs.org/docs/latest-v0.10.x/api/), the [current version](https://nodejs.org/api/) and in fact to all Node.js versions back to 0.1: https://nodejs.org/docs/.

fhemberger:
@rvagg Can we have a 'latest-lts' symlink for API docs as well?

shelleyp:
If you're going to support two versions of Node at the same time, you need to provide access to both APIs at the top most level of the web site, at the same time. If you're going to have two downloads for Node on your download page, you really need to have two APIs at the top most API page.

phillipj:
Maybe we should have two API links in the aside on the docs page aswell? E.g:
![skjermbilde 2015-10-30 kl 14 17 13](https://cloud.githubusercontent.com/assets/1231635/10846610/3ebb85cc-7f11-11e5-95a5-9d277b47bedf.png)

shelleyp:
No, not closed. 

This page needs to list both. https://nodejs.org/api/

shelleyp:
Agree with @phillipj suggestion.

mikeal:
Updated the title to be more productive.

mikeal:
I'm not convinced that https://nodejs.org/api/ needs to have both. That link has always been to the latest current docs, since literally forever :)

Totally agree that the docs page needs to link to both.

Have we considered also linking to the API docs in the download page for each release and in the automated release post of each release?

mikeal:
Reopening as I don't see the comment that closed this with links to new issues :(

dominykas:
I very much disagree that "this has been this way forever" is an argument to keep things as is. In fact, it's an argument for the opposite - this was a _problem_ forever, esp. during the months when I couldn't upgrade, say, 0.8 to 0.10 (which reminds me I need to update my bookmark which currently points to 0.12...)

[...]

PR #324, opened 2015-11-04
Display LTS and Stable API links on docs page

phillipj:
Part of the multiple version docs discussion in #305.

skjermbilde 2015-11-04 kl 08 54 44

Using the same URLs as we've got on the frontpage, e.g:
- **v4.2.2**: https://nodejs.org/dist/4.2.2/docs/api/
- **v5.0.0**: https://nodejs.org/api/

Are those URLs appropriate or would we want to use any of these aliases?
- latest -> v5.0.0
- latest-argon -> v4.2.2
- latest-v0.10.x -> v0.10.40
- latest-v0.12.x -> v0.12.7
- latest-v4.x -> v4.2.2
- latest-v5.x -> v5.0.0

### DIFF
--- a/layouts/index.hbs
+++ b/layouts/index.hbs
@@ -36,7 +36,7 @@
                           <a href="https://github.com/nodejs/node/blob/{{ project.currentVersions.lts }}/CHANGELOG.md">{{ labels.changelog }}</a>
                       </li>
                       <li>
-                        <a href="https://nodejs.org/dist/{{ project.currentVersions.lts }}/docs/api/">{{ labels.api }}</a>
+                        <a href="{{majorapidocslink project.currentVersions.lts}}">{{ labels.api }}</a>
                       </li>
                   </ul>
 
@@ -58,7 +58,7 @@
                             <a href="https://github.com/nodejs/node/blob/{{ project.currentVersions.stable }}/CHANGELOG.md">{{ labels.changelog }}</a>
                         </li>
                         <li>
-                          <a href="{{ site.docs.api.link }}">{{ labels.api }}</a>
+                          <a href="{{majorapidocslink project.currentVersions.stable}}">{{ labels.api }}</a>
                         </li>
                     </ul>
 
--- a/build.js
+++ b/build.js
@@ -158,7 +158,8 @@ function buildlocale (source, locale) {
         i18n: require('./scripts/helpers/i18n.js'),
         changeloglink: require('./scripts/helpers/changeloglink.js'),
         strftime: require('./scripts/helpers/strftime.js'),
-        apidocslink: require('./scripts/helpers/apidocslink.js')
+        apidocslink: require('./scripts/helpers/apidocslink.js'),
+        majorapidocslink: require('./scripts/helpers/majorapidocslink.js')
       }
     }))
     .destination(path.join(__dirname, 'build', locale))
@@ -188,7 +189,6 @@ function fullbuild () {
     const source = {
       project: {
         versions,
-        currentVersion: versions[0].version,
         currentVersions: {
           stable: latestVersion.stable(versions),
           lts: latestVersion.lts(versions)
--- a/layouts/css/base.styl
+++ b/layouts/css/base.styl
@@ -153,6 +153,9 @@ pre
 .color-lightgray
     color $light-gray
 
+a:hover .color-lightgray
+    color $white
+
 .no-padding
     padding 0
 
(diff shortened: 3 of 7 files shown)


dominykas:
May I suggest adding "LTS" to every LTS version label? i.e. display:
- v4.2.2 LTS API
- v5.0.0 API

rvagg:
perhaps use https://nodejs.org/docs/latest-v4.x/api/ for the link instead, you could also do the same with the v5.x link .. maybe. Also note that https://nodejs.org/docs/latest-argon/api/ is a thing if you want to reinforce the LTS name which is not a bad idea.

phillipj:
> May I suggest adding "LTS" to every LTS version label?

If so, why shouldn't we add "Stable" to v5 also?

thefourtheye:
> If so, why shouldn't we add "Stable" to v5 also?

Yes, that would be better.

dominykas:
> > If so, why shouldn't we add "Stable" to v5 also?
> 
> Yes, that would be better.

No it would not.

The idea is to _highlight_ the LTS release - if each release is labeled, then the information density increases and it becomes hard to see the trees in the forest. If the general recommendation is "when unsure - choose LTS", then LTS should be highlighted. Or, you can think about it this way - both LTS and Stable are "stable", but "LTS" is "special Stable", therefore it deserves an honorable mention.

If "unstable" / "alpha" / "beta" releases existed - they should also be marked as such (although better yet, they should only be displayed to power users, i.e. they should be absent from the lists on primary pages).

I am falling back on the laws of simplicity here - [Simplicity is about subtracting the obvious, and adding the meaningful.](http://lawsofsimplicity.com/los/law-10-the-one.html) - but similar concepts can be found in plenty of other generic usability advice.

layouts/index.hbs
```diff
@@ -51,12 +51,8 @@
                         &#47; <a href="https://github.com/nodejs/node/blob/{{ project.currentVersions.stable }}/CHANGELOG.md">{{ labels.stable }}</a>
                     </li>
                     <li>
-                      {{#if project.currentVersions.stable}}
```
phillipj: This if statement was put into place _before_ we had Stable and LTS on the website - as that's old news, it could be removed.

phillipj:
Just pushed a couple of updates:
- Use https://nodejs.org/docs/latest-v4.x/api/ and latest-v5.x URLs
- To reduce duplication and ensure consistency, I extracted docs URL generation into a Handlebars helper, which is also used on the frontpage
- Added "LTS" behind v4.x label

TBH I dont have a strong opinion about putting LTS into the v4.x link or not, but I agree with @dominykas reasoning.

skjermbilde 2015-11-11 kl 21 51 59

phillipj:
An updated screenshot now that we've got the `Guides` link aswell.

Any thoughts on this anyone? It's ready to go as soon as it gets a collaborator thumbs up.

skjermbilde 2015-11-18 kl 22 38 31

rnsloan:
:+1:

bnb:
-1 on the last proposed solution. 

LTS and Stable should both be marked - giving context to one and withholding context from other is a _very_ bad idea. 

Additionally, from a design view, there is **Way** too much going on in the green links, making them extremely hard for humans to parse. Having a `v, number, full stop, number, full stop, number, full stop, text element one, text element two` is far too much. We need to simplify this.

So here are my suggestions:
- Remove `v` - giving the number implies version
- Remove `LTS` from `4.2.2`
- Append `LTS` and `Stable` in smaller gray lettering or a gray tag

rnsloan:
When you have an LTS version of software, personally, I have never found the need to state everywhere that the later release is **stable**. Ubuntu do not do it: http://www.ubuntu.com/server , https://help.ubuntu.com/

dominykas:
@bnb you're assuming the reader knows the difference between "LTS" and "Stable" - the label needs to highlight it - both are "Stable", but only one is "LTS".

bnb:
@rnsloan @dominykas Fair play. However, I still maintain that those changes (minus adding `Stable`) should be made.

phillipj:
Something like this @bnb?

screen shot 2015-12-07 at 07 50 13

stevemao:
Looks good :)

phillipj:
Thanks for the feedback everyone! :+1:

@bnb I like your thoughts about removing the `v` from `v4.2.3` and `v5.1.1`! I thought it better fits a new PR/issue, as it should probably be done throughout the site, not only the docs page - up for grabs anyone :)

PR #326, opened 2015-11-05
Use sudo on Debian as well

oliversalzburg:
I don't understand the distinction that is being made here. Both platforms behave identical in regards to the installation process.

### DIFF
--- a/locale/en/download/package-manager.md
+++ b/locale/en/download/package-manager.md
@@ -30,13 +30,8 @@ Node.js is available from the [NodeSource](https://nodesource.com) Debian and Ub
 **NOTE:** If you are using Ubuntu Precise or Debian Wheezy, you might want to read about [running Node.js >= 4.x on older distros](https://github.com/nodesource/distributions/blob/master/OLDER_DISTROS.md).
 
 ```bash
-# Using Ubuntu
 curl -sL https://deb.nodesource.com/setup_4.x | sudo -E bash -
 sudo apt-get install -y nodejs
-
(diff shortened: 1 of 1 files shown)


thefourtheye:
LGTM. For future reference, quoting the [debian's official documentation](https://wiki.debian.org/sudo),

> **Why sudo?**
> Using `sudo` is better (safer) than opening a session as root for a number of reasons, including:
> - Nobody needs to know the root password (`sudo` prompts for the current user's password). Extra privileges can be granted to individual users temporarily, and then taken away without the need for a password change.
> - It's easy to run only the commands that require special privileges via `sudo`; the rest of the time, you work as an unprivileged user, which reduces the damage that mistakes can cause.
> - Auditing/logging: when a `sudo` command is executed, the original username and the command are logged.
>   For the reasons above, switching to root using `sudo -i` (or `sudo su`) is usually deprecated because it cancels the above features.

rvagg:
-1, we've avoided this in the source docs @ https://github.com/nodesource/distributions because some variants of Debian don't come with `sudo` so the "use root" is a way of getting around that.

thefourtheye:
@rvagg Oh, shall I revert the commit then?

rvagg:
either that or make a comment that `sudo` may not be available

oliversalzburg:
If you add a note that `sudo` isn't available by default, you should also add that `curl` is neither.

stevemao:
"use root" :+1:

PR #337, opened 2015-11-09
Node Interactive blog post: Correct CJ's info

isaacs:
Correct @ceejbot's last name and position at npm, Inc.


evanlucas:
Thanks!

PR #338, opened 2015-11-09
Remove redundant information

thefourtheye:


mikeal:
Actually, please leave this :) To date we've gotten much better emails than we have form submits so it's best to leave them both.

thefourtheye:
Oh okay. No problem then :-)

PR #341, opened 2015-11-09
Improve write to file

thefourtheye:
`writeToFile` function uses `exists` which is deprecated. This patch replaces that with `access` function.
Apart from that, the `writeToFile` function is fully asynchronous now. This patch removes `sinon` dependency as well. `writeToFile` tests have been written in such a way that the spying is not necessary.

### DIFF
--- a/scripts/release-post.js
+++ b/scripts/release-post.js
@@ -181,12 +181,21 @@ function renderPost (results) {
 function writeToFile (results) {
   const filepath = path.resolve(__dirname, '..', 'locale', 'en', 'blog', 'release', `v${results.version}.md`)
 
-  if (fs.existsSync(filepath)) {
-    return Promise.reject(new Error(`Release post for ${results.version} already exists!`))
-  }
+  return new Promise(function (resolve, reject) {
+    fs.access(filepath, fs.F_OK, function (err) {
+      if (!err) {
+        return reject(new Error(`Release post for ${results.version} already exists!`))
+      }
 
-  fs.writeFileSync(filepath, results.content)
-  return Promise.resolve(filepath)
+      fs.writeFile(filepath, results.content, function (err1) {
+        if (err1) {
+          return reject(new Error(`Failed to write Release post: Reason: ${err1.message}`))
+        }
+
+        resolve(filepath)
+      })
+    })
+  })
 }
 
 function slugify (str) {
--- a/tests/scripts/release-post.test.js
+++ b/tests/scripts/release-post.test.js
@@ -3,7 +3,6 @@
 const test = require('tape')
 const nock = require('nock')
 const proxyquire = require('proxyquire').noCallThru()
-const sinon = require('sinon')
 const path = require('path')
 
 test('explicitVersion(<version>)', (t) => {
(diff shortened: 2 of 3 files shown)


scripts/release-post.js
```diff
@@ -181,12 +181,21 @@ function renderPost (results) {
 function writeToFile (results) {
   const filepath = path.resolve(__dirname, '..', 'locale', 'en', 'blog', 'release', `v${results.version}.md`)
 
-  if (fs.existsSync(filepath)) {
-    return Promise.reject(new Error(`Release post for ${results.version} already exists!`))
-  }
+  return new Promise(function (resolve, reject) {
+    fs.access(filepath, fs.R_OK | fs.W_OK, function (err) {
+      if (!err) {
```
lpinca: Shouldn't this be `if (err) {`?

thefourtheye: @lpinca If there is no error, it means that the file exists. [`fs.access`](https://nodejs.org/api/fs.html#fs_fs_access_path_mode_callback)

lpinca: Correct, my bad.

lpinca: Can't we just use `fs.F_OK`?

``` js
fs.access(filepath, function (err) {
```

tests/scripts/release-post.test.js
```diff
@@ -258,10 +257,11 @@ test('findLatestVersion<version>', (t) => {
 })
 
 test('writeToFile<object>', (t) => {
-  let fileExists
+  let accessResult
+
   const fs = {
-    existsSync: () => fileExists,
-    writeFileSync: sinon.spy()
+    access: (filename, flags, cb) => accessResult ? cb(accessResult) : cb(),
```
lpinca: I think that keeping `fileExists` makes it easier to grok.

``` js
access: (filename, cb) => fileExists ? cb() : cb(New Error('ENOENT')),
```

scripts/release-post.js
thefourtheye: @lpinca Sure, I'll change it.

tests/scripts/release-post.test.js
thefourtheye: @lpinca You are right. Replaced it now. PTAL.

thefourtheye:
@lpinca Addressed your comments. PTAL.

lpinca:
LGTM.

scripts/release-post.js
```diff
@@ -181,12 +181,21 @@ function renderPost (results) {
...
-  if (fs.existsSync(filepath)) {
-    return Promise.reject(new Error(`Release post for ${results.version} already exists!`))
-  }
+  return new Promise(function (resolve, reject) {
+    fs.access(filepath, fs.F_OK, function (err) {
+      if (!err) {
+        return reject(new Error(`Release post for ${results.version} already exists!`))
+      }
+
+      fs.writeFile(filepath, results.content, function (err1) {
```
lpinca, line 190: @thefourtheye any reason for using `err1` here?
I think we can use `err` again as it will shadow the other `err`.

thefourtheye, line 190: @lpinca Exactly the point. I thought people will not be confused when they read the code with `err1`.

lpinca, line 190: Ok, I'm fine with that :)

phillipj:
Nice work @thefourtheye, LGTM!

PR #345, opened 2015-11-10
Update st to version 1.0.0

lpinca:
This updates [`st`](https://github.com/isaacs/st) to version `1.0.0`.

The breaking change is described here https://github.com/isaacs/st/issues/67. I see no problems, but correct me if I'm wrong.

### DIFF
--- a/package.json
+++ b/package.json
@@ -45,7 +45,7 @@
     "octonode": "0.7.4",
     "require-dir": "0.3.0",
     "semver": "5.0.3",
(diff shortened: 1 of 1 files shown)


thefourtheye:
Lets have @rvagg make the call.

rvagg:
trivial update that doesn't impact on production since `st` is only used for dev (we serve static files via nginx on nodejs.org)

PR #346, opened 2015-11-10
Update metalsmith-feed to version 0.2.0

lpinca:
This updates [`metalsmith-feed`](https://github.com/hurrymaplelad/metalsmith-feed) to version `0.2.0`.

There are 2 new options, `postDescription` and `postCustomElements`, but no breaking changes:
- `postDescription` defaults to `(file) -> file.less or file.excerpt or file.contents` like in version `0.0.6`.
- `postCustomElements` is disabled by default.

### DIFF
--- a/package.json
+++ b/package.json
@@ -32,7 +32,7 @@
     "marked": "0.3.5",
     "metalsmith": "2.1.0",
     "metalsmith-collections": "0.7.0",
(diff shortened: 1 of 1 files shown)


stevemao:
How come greenkeeperio does not auto PR?

lpinca:
I guess it's because the bot has been added after the update.

PR #349, opened 2015-11-10
Add link to Code of Conduct from contributing section

junosuarez:
I was trying to find the Node community code of conduct to refer to, and it was more difficult to find on the site than I think it should be. This PR adds it to the left-hand sidebar in the "contributing" section.


bnb:
LGTM.

MylesBorins:
Not too far off from https://github.com/nodejs/node/pull/3722

Discoverability of the COC right now is a problem imho

mikeal:
provisionally +1 but i would like to note that we have some structural improvements we need to make as well. we don't currently have a foundation-wide CoC so the one we are linking to is specific to core, nor do we probably document the working groups that you should be getting involved in (which all have their own CoCs as well).

but ya, this is better than nothing which is what we have now :)

ghost:
LGTM, gonna merge now since there are no complaints

PR #351, opened 2015-11-11
Update package-manager.md

jperkin:
Add installation instructions for NetBSD, SmartOS, and other illumos distributions.  Add pkgsrc as an alternative for OSX.

### DIFF
--- a/locale/en/download/package-manager.md
+++ b/locale/en/download/package-manager.md
@@ -16,7 +16,9 @@ title: Installing Node.js via package manager
 * [Arch Linux](#arch-linux)
 * [Void Linux](#void-linux)
 * [FreeBSD and OpenBSD](#freebsd-and-openbsd)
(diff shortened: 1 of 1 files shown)


lpinca:
:ok_hand:

bnb:
+1 on content, but can anyone run through these commands on the OSes and make sure they work? Testing them to be 100% sure is necessary to prevent problems down the line. Also, are there version numbers of the package managers that are required? And what about LTS/Stable updates? Will they be pushed automatically?

lpinca:
@bnb The document does not differentiate between LTS and stable atm. If we want that we should update the whole document not only this pr. We should probably update it anyway as its content is outdated.

If I'm not wrong there is `nodejs4` (LTS) and `nodejs` (Stable) in `pkgsrc`.

I merge this.

PR #353, opened 2015-11-12
Update the description of the Node Knockout event

lpinca:
The hackathon is over, we are in the judging period now.


lpinca:
I didn't touch anything sensitive and I don't know what's up with Travis, but here is the build https://travis-ci.org/nodejs/new.nodejs.org/builds/90737551.

fhemberger:
LGTM, merged.

PR #362, opened 2015-11-16
update TC with TSC

thefourtheye:


bnb:
Other than the above, LGTM.

thefourtheye:
@bnb I fixed it now. PTAL. Also, I feel that it would be better if we got an LGTM from @nodejs/tsc on this.

bnoordhuis:
LGTM

ghost:
yeah, LGTM

thefourtheye:
Thanks for the review people :-) Landed the patch in 0dd684cf21d278ba8aa178db0a20ebc6d587c58e

PR #365, opened 2015-11-17
Remove old nodejs.org link from GOVERNANCE.md

ghost:
since the old repo has now been 'archived', we can remove the reference to it being published on nodejs.org in the GOVERNANCE.md. the reason why i opened a PR on this was because i'm not sure what to put in place of it. should i just remove the text like i did in my initial commit?

### DIFF
--- a/GOVERNANCE.md
+++ b/GOVERNANCE.md
@@ -8,7 +8,7 @@ related projects which include:
 
 - [nodejs/iojs.org](https://github.com/nodejs/iojs.org) published to https://iojs.org
 - [nodejs/nodejs.org](https://github.com/nodejs/iojs.org) published to https://nodejs.org
-- [joyent/node-website](https://github.com/joyent/node-website) published to https://nodejs.org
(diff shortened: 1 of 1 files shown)


GOVERNANCE.md
```diff
@@ -8,7 +8,7 @@ related projects which include:
 
 - [nodejs/iojs.org](https://github.com/nodejs/iojs.org) published to https://iojs.org
 - [nodejs/nodejs.org](https://github.com/nodejs/iojs.org) published to https://nodejs.org
-- [joyent/node-website](https://github.com/joyent/node-website) published to https://nodejs.org
+- [joyent/node-website](https://github.com/joyent/node-website)
```
Fishrock123: Seems fine to just link. Mind updating it to point at https://github.com/nodejs/nodejs.org-archive?

ghost: ah, totally forgot that

bnb:
LGTM.

Issue #355, opened 2015-11-13
Travis PR integration broken


phillipj:
Seems there's an issue with the Travis integration in the newest PRs. Travis still picks up the [commits and run builds](https://travis-ci.org/nodejs/new.nodejs.org/builds), but it's not visible inline in the PRs :(

Examples: https://github.com/nodejs/new.nodejs.org/pull/349, https://github.com/nodejs/new.nodejs.org/pull/350, https://github.com/nodejs/new.nodejs.org/pull/351, https://github.com/nodejs/new.nodejs.org/pull/353
Compared to: https://github.com/nodejs/new.nodejs.org/pull/313, https://github.com/nodejs/new.nodejs.org/pull/334 which has Travis checks inline.

I can contact Travis about the issue, but I'll like to know the Travis webhook status from our perspective first, but I dont have sufficient repo access. Could some of the repo owners please check `Settings -> Webhooks & Services`?

/cc @mikeal @rvagg

lpinca:
Unrelated but I also wonder if greenkeeper bot is working. `node-version-data` has been updated to version `1.0.1` but we didn't receive a pr yet.

mikeal:
I see one webhook to a random IP address and another to greenkeeper, so maybe it got turned off?

phillipj:
Hmmm.. Not part of the `Settings -> Webhooks & Services -> Services` list either @mikeal?

Pretty sure @rvagg has said we've got a webhook setup too autodeploy the site, that might be the random IP.

mikeal:
oh, that is there :) it say travis is working and is wired up to @rvagg's account.

phillipj:
Thanks! I'm in contact with Travis support about this now.

BanzaiMan:
Hello there! The integration seems to be off because we lost some crucial data about this repository while syncing. Could an admin of this repository (is it you, @mikeal?) log out and back in to https://travis-ci.org, and resync data to see if that fixes the issue?

phillipj:
@BanzaiMan I just removed the old Travis service link, hooked up with @rvagg's account, then created a new link with my account. Unfortunately didn't work. Travis doesn't know about this repo https://travis-ci.org/nodejs/nodejs.org, even though we have had 3 commits since the Travis service links was altered. 

And it still looks like the old service link is active somehow, as new commits are getting built: https://travis-ci.org/nodejs/new.nodejs.org.

[...]

PR #372, opened 2015-11-21
Use Shields.IO for all the badges

lpinca:
This updates the README to use http://shields.io/ for all the badges.

Right now the Travis badge is not working because the repo need to be resynced (the name has changed). This will also hopefully fix the pr integration. See https://github.com/nodejs/nodejs.org/issues/355#issuecomment-157890226.


ghost:
LGTM

stevemao:
I don't know if this is necessary. Do these services internally cache badges from shields?

lpinca:
It's more for consistency than anything else.

PR #381, opened 2015-11-26
Update code sample on about page with ES6 features

Chris911:
Small update on the about page to use ES6 features in code sample supported on on Node latest.

Looks like:
screen shot 2015-11-26 at 12 55 05 am

### DIFF
--- a/locale/en/about/index.md
+++ b/locale/en/about/index.md
@@ -11,14 +11,17 @@ connections can be handled concurrently. Upon each connection the callback is
 fired, but if there is no work to be done Node is sleeping.
 
 ```javascript
-var http = require('http');
+const http = require('http');
 
-http.createServer(function (req, res) {
-  res.writeHead(200, {'Content-Type': 'text/plain'});
+const hostname = '127.0.0.1';
+const post = 1337;
+
+http.createServer((req, res) => {
+  res.writeHead(200, { 'Content-Type': 'text/plain' });
   res.end('Hello World\n');
-}).listen(1337, '127.0.0.1');
-
-console.log('Server running at http://127.0.0.1:1337/');
+}).listen(post, hostname, () => {
+  console.log(`Server running at http://${hostname}:${post}/`);
+});
 ```
 
 This is in contrast to today's more common concurrency model where OS threads


fhemberger:
Nice, LGTM

rvagg:
maybe it's time to do a proper job and introduce template strings at the same time?

``` js
const http = require('http');

  res.writeHead(200, { 'Content-Type': 'text/plain' });
  res.end('Hello World\n');
}).listen(1337, '127.0.0.1', function () {
  console.log(`Server running at http://${this.address().address}:${this.address().port}/`);
});
```

or is that too complicated?

fhemberger:
@rvagg Good point! Maybe add:

``` js
const connection = this.address();
console.log(`Server running at http://${connection.address}:${connection.port}/`);
```

ghost:
LGTM, regarding the complexity, maybe link to the ES6 page right below the code block? something like `This code example uses EcmaScript 2015 (ES6) features. Read more _here_`

Chris911:
@rvagg, @fhemberger, I can make the change the add the template string but personally I think it adds complexity by using `this.address().address`. Since this is a small introduction snippet maybe something like: 

``` js
const http = require('http');

const HOSTNAME = '127.0.0.1';
const PORT = 1337;

  res.writeHead(200, { 'Content-Type': 'text/plain' });
  res.end('Hello World\n');
}).listen(PORT, HOSTNAME, function () {
  console.log(`Server running at http://${HOSTNAME}:${PORT}/`);
});
```

This also matches the terminology used in the [`server.listen` documentation](https://nodejs.org/api/http.html#http_server_listen_port_hostname_backlog_callback).

fhemberger:
@Chris911 Good idea, let's settle for this.

Chris911:
@fhemberger updated.

stevemao:
:+1:

locale/en/about/index.md
```diff
@@ -11,14 +11,17 @@ connections can be handled concurrently. Upon each connection the callback is
...
-  res.end('Hello World\n');
-}).listen(1337, '127.0.0.1');
+const HOSTNAME = '127.0.0.1';
+const PORT = 1337;
 
-console.log('Server running at http://127.0.0.1:1337/');
+http.createServer((req, res) => {
+  res.writeHead(200, { 'Content-Type': 'text/plain' });
+  res.end('Hello World\n');
+}).listen(PORT, HOSTNAME, function () {
```
phillipj: Last nitpick, use arrow function on this callback aswell?

rvagg:
I'd vote for lower-case variable names, there's inconsistency in `http` vs `PORT` and `HOSTNAME` and upper-casing constants is not necessarily idiomatic Node or even JavaScript even though it's common in other languages.

locale/en/about/index.md
Chris911: Good catch.

Chris911:
Fine with me @rvagg.

Chris911:
Updated.

phillipj:
Nice :+1:

PR #383, opened 2015-11-26
Error in build.js/copystatic()  `ENOENT: no such file or directory, open '.../nodejs.org/build/static/event-geo.json'`

justsml:
After I switched to nodejs v5.0, running a clean build no longer worked.
It doesn't seem to be creating the directories before the `fs.mkdir` callback(s)??

It's a build script, so hopefully a synchronous fix isn't the worst...

### DIFF
--- a/build.js
+++ b/build.js
@@ -172,13 +172,11 @@ function buildlocale (source, locale) {
 
 function copystatic () {
   console.time('[metalsmith] build/static finished')
-  fs.mkdir(path.join(__dirname, 'build'), function () {
-    fs.mkdir(path.join(__dirname, 'build', 'static'), function () {
-      ncp(path.join(__dirname, 'static'), path.join(__dirname, 'build', 'static'), function (err) {
-        if (err) { return console.error(err) }
-        console.timeEnd('[metalsmith] build/static finished')
-      })
-    })
+  fs.mkdirSync(path.join(__dirname, 'build'))
+  fs.mkdirSync(path.join(__dirname, 'build', 'static'))
+  ncp(path.join(__dirname, 'static'), path.join(__dirname, 'build', 'static'), function (err) {
+    if (err) { return console.error(err) }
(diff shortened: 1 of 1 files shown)


phillipj:
Strange, not able to reproduce this issue on my machine. What's your OS/version?

stevemao:
If this was a problem then there might be some issue in the core.

justsml:
I'm on linux mint debian edition v2. 
It works on my ubuntu 15. 
On my fresh fedora install it works only the 2nd attempt. 
I'm migrating to a new HDD tonight, I'll see if that fixes it...

lpinca:
@phillipj @stevemao I'm able to reproduce this and please note that we are on the `events` branch.

phillipj:
> If this was a problem then there might be some issue in the core.

Agreed, which probably should be reported.

I wouldn't mind replacing two levels of callbacks with sync here, as it's simpler to grasp and doesn't really matter for performance reasons as @justsml mentioned.

Trying to build with these changes results in an error for me though...

``` bash
$ npm start

> nodejs.org@1.0.0 start /nodejs.org
> npm run serve


> nodejs.org@1.0.0 serve /nodejs.org
> node build.js serve

fs.js:842
  return binding.mkdir(pathModule._makeLong(path),
                 ^

Error: EEXIST: file already exists, mkdir '/nodejs.org/build'
    at Error (native)
    at Object.fs.mkdirSync (fs.js:842:18)
    at copystatic (/nodejs.org/build.js:175:6)
...
```

stevemao:
The problem here is not sync or performance. It's why it fails.

bnb:
I've started using metalsmith, and got this same error. No idea why it worked, but deleting the existing `/build` directory solved it.

mikeal:
could you re-target to master now that the branch is merged

mikeal:
I pushed a hotfix for this just pushing the geo create to before the timer end.

PR #382, opened 2015-11-26
Map distance+position fix for mapbox/leaflet

justsml:
@mikeal

PR #389, opened 2015-11-28
Events page auto-centers based on top 10 events

justsml:
Hey @mikeal Here's a clean PR RE #382

### DIFF
--- a/layouts/events.hbs
+++ b/layouts/events.hbs
@@ -41,7 +41,6 @@
               </div>
               
               <div class="main-content">{{{contents}}}</div>
-
                 {{#each regions}}
                 <div class="region">
                   <h1>{{region}}</h1>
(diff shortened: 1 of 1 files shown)


layouts/events.hbs
```diff
@@ -149,26 +149,19 @@
...
+                , _lon = feature.geometry.coordinates[0]
+                , dist = haversine(lat, lon, _lat, _lon)
+                ;
+              return {lon: _lon, lat: _lat, dist: dist}
+            })
+            .sort(function (a, b) {
+              return a.dist > b.dist ? 1 : 
+                     a.dist < b.dist ? -1 : 0;
+            })
+            var nearest = Array.isArray(points) && (points.length > 10 ? points.slice(0, 9) : points.slice(0, points.length-1));
```
mikeal: I don't think this safety is necessary because we're in the feature layer's onready, so points will always be an array and it'll always be huge (much more than 10) because we load the array from the giant geojson url.

justsml: Ah yes, thx for spotting that. One sec...

justsml:
Hopefully that looks better

justsml:
Hey @mikeal Hope you had a good thanksgiving.

Did you notice my latest update?
... Presumably after you've woken up out of the food coma. :sleeping:

mikeal:
This looks great! I've merged locally and am modifying just a little bit. I realized that we also need a minimum threshold of distance so that when you're in a very dense area you don't get a hyper localized map.

mikeal:
This was merged as part of https://github.com/nodejs/nodejs.org/commit/b3efd13a7e56262c350cf73a71da40755f60b290

PR #391, opened 2015-11-29
Fix events map link color

### LINKED ISSUE #390: node-green for a:hover leads to hard to read blue-on-green mapbox leaflet titles
How it is:
screen shot 2015-11-28 at 1 08 12 pm

Expected: 
screen shot 2015-11-28 at 1 08 32 pm

Fix is easy but where should the code go? base.styl has all anchor styles but adding code that depends on mapbox in a core file seems "hacky". Should one worry too much about stuff like this ? 

``` styl
.leaflet-container
    a:hover
        color $white !important
```

rnsloan:
fixes #390
- fix map link color
- make images not overflow map marker box

Before:

After:

### DIFF
--- a/layouts/css/page-modules/_events.styl
+++ b/layouts/css/page-modules/_events.styl
@@ -4,8 +4,17 @@
     margin-top 20px
 
     .marker-image img
+        width 100%
         max-width 320px
         max-height 100px
+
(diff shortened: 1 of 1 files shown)


ghost:
LGTM

PR #398, opened 2015-12-01
Changed url for PereiraJS (Colombia) meetup

pin3da:
We don't use the meetup.com platform at this moment. It would be cool if you use http://pereirajs.org as referral page.

Let me know if we need to write the data in another format.

Thanks (:


cronopio:
:+1:

lpinca:
The changes look good to me but they will be overwritten when the  `pull-meetup.js` script is run.

indexzero:
Agreed @lpinca. I think the script @mikeal wrote will overwrite this.

cronopio:
woot! Are we the only ones in the world NOT using meetup? :scream: 
BTW, I think is not good looking create an if stament in `pull-meetup.js` just to avoid Pereira been included there. Any other ideas?

stevemao:
From https://nodejs.org/en/get-involved/events/

> If you aren't listed on those sites, or you are the organizer of a conference that is not listed, you can add it manually by editing events.md and sending a pull request to the website repo.

If it will be overwritten this is a problem?

lpinca:
I was thinking about something like this:

``` diff
diff --git a/events/custom.json b/events/custom.json
new file mode 100644
index 0000000..5048232
--- /dev/null
+++ b/events/custom.json
@@ -0,0 +1,19 @@
+[
+  {
+    "id": 18205029,
+    "created": 1393477200000,
+    "link": "http://pereirajs.org/",
+    "description": "Comunidad de aprendizaje del lenguaje Javascript",
+    "organizer": {
+      "member_id": "",
+      "name": "",
+      "photo": {
+        "highres_link": "",
+        "photo_id": "",
+        "photo_link": "",
+        "thumb_link": ""
+      }
+    },
+    "members": 30
+  }
+]
diff --git a/events/pull-meetup.js b/events/pull-meetup.js
index 05749c1..d1a3391 100644
--- a/events/pull-meetup.js
+++ b/events/pull-meetup.js
@@ -1,7 +1,9 @@
 var request = require('request').defaults({json: true, headers: {'user-agent': 'pull-meeting-0.1'}}),
   url = 'https://api.meetup.com/2/groups',
   auth = process.env.MEETUP_TOKEN,
+  merge = require('lodash.merge'),
   qs = require('querystring'),
+  custom = require('./custom'),
   yml = require('./yaml-sync'),
   opts =
   { topic: 'nodejs',     category: 34,     upcoming_events: true,     key: auth
@@ -29,6 +31,12 @@ function finish (events) {
     }
     var region = yml.getRegion(countryMap[event.country])
     if (!region.meetups) region.meetups = []
+    custom.some(elem => {
+      if (event.id === elem.id) {
+        event = merge(event, elem)
+        return true
+      }
+    })
     clean(event)
     yml.replace(region.meetups, 'name', event.name, event)
   })
index fb04c03..9e9f611 100644
--- a/package.json
+++ b/package.json
@@ -30,6 +30,7 @@
     "html-to-text": "^1.5.0",
     "js-yaml": "^3.4.5",
     "junk": "1.0.2",
+    "lodash.merge": "^3.3.2",
     "map-async": "0.1.1",
     "marked": "0.3.5",
     "metalsmith": "2.1.0",
```

so you can simply add your customized event in `custom.json` and in this way changes are not overwritten. The problem is that the event id is not shown and has to be extracted from meetup api.
Truthfully, I think this is cumbersome so if you have better ideas please share.

mikeal:
When we pull data in from another source we consider that resource the source of truth for that information. If we want the local copy to be the source of truth we should add a flag in the markdown and make sure our scripts honor it and don't overwrite it. I _don't_ think we should try and merge information we have locally with information we pull from another source, that's going to get messy :)

stevemao:
Can we just simply include two md files: one is generated from the script and the other is not?

pin3da:
I tried to do what @mikeal said : db53e2909d11a375356eb431f6b5a7009ae0b10d

Tell me if I need to change something.

cc @cronopio

events/yaml-sync.js
```diff
@@ -16,6 +17,22 @@ exports.getRegion = function (region) {
   return reg
 }
 
+exports.isSoT = function(region, city) {
+  for (var reg in store.regions) {
```
lpinca: Can't we reuse `exports.getRegion`?

events/yaml-sync.js
```diff
@@ -16,6 +17,22 @@ exports.getRegion = function (region) {
...
+    if (store.regions[reg].region == region) {
+      var meetups = store.regions[reg].meetups
+      for (var i in meetups) {
+        if (meetups[i].city == city) {
```
lpinca: I think we should find a better way to match this. What happens if there are 2+ events from the same region and city?

fhemberger:
@nodejs/website @pin3da What's the status of this PR?

pin3da:
@fhemberger I just fixed the conflicts. I also added the @lpinca 's suggestions.

Let me know if there is something missing/wrong.

Thank you (:

events/pull-meetup.js
```diff
@@ -29,7 +29,9 @@ function finish (events) {
     const region = yml.getRegion(countryMap[event.country])
     if (!region.meetups) region.meetups = []
     clean(event)
-    yml.replace(region.meetups, 'name', event.name, event)
+    if (!yml.isSoT(region, event.city, event.name)) {
```
lpinca: Can we use `region.meetups` here instead of `region`?

events/yaml-sync.js
```diff
@@ -20,6 +20,19 @@ function getRegion (region) {
   return reg
 }
 
+function isSoT (region, city, name) {
+  let meetups = region.meetups
```
lpinca: I think we can remove this line if the first argument of the function is changed to `meetups`.

events/yaml-sync.js
```diff
@@ -20,6 +20,19 @@ function getRegion (region) {
...
+  for (const i in meetups) {
```
lpinca: I wouldn't use `for...in` to iterate over an array.

lpinca:
Left some comments but LGTM.

lpinca:
All good here, :ship: ?

ghost:
aye

lpinca:
Thank you @pin3da!

pin3da:
Thank you all :smiley_cat:

PR #402, opened 2015-12-04
0.10.41, 0.12.9, 4.2.3, 5.1.1 releases and summary post

rvagg:
the big 'un


Trott:
The markdown files refer to nodejs/node-private but links to that repo are not visible to users. Not sure if this is really an issue but not sure it's _not_ an issue so... `¯\_(ツ)_/¯`

rvagg:
yeah, I've been manually removing them from the changelogs but obviously missed a few! will rectify, thanks

PR #403, opened 2015-12-04
Display security announcement for december security releases

phillipj:
Should we display an announcement about the recent security releases? It links to the [December Security Release Summary](https://nodejs.org/en/blog/vulnerability/december-2015-security-releases/) post.

/cc @rvagg 

screen shot 2015-12-04 at 08 41 06

### DIFF
--- a/build.js
+++ b/build.js
@@ -194,8 +194,8 @@ function fullbuild () {
           lts: latestVersion.lts(versions)
         },
         banner: {
-          visible: false,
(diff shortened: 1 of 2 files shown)


rvagg:
good call @phillipj, I forgot all about it .. I did have this on my list a few weeks ago when we first started planning but the whole thing got blown way up into a major exercise. I need to write up a formal policy and document the procedures to include little things like this.

Issue #281, opened 2015-10-24
i18n: Create Handlebars helper for dynamic navigation


fhemberger:
At the moment, the main and sidebar navigation are all hardwired to the templates. So if a translation team decides to have a different structure or to leave out pages (e.g. the blog section), there will be broken links.

We need to create a Handlebars helper to build (partial) navigation based on the page structure in `site.json` instead. This way templating would be easier to handle, e.g.

`{{navigation}}` – renders the main navigation
`{{navigation 'about'}}` – renders the sub navigation matching the given key

mrkiffie:
I like this idea. It would make adding additional items to the navigation much simpler. 

I've had a look at the structure of the `site.json` and it currently doesn't contain anything that we could use to determine the order of the menu items.

Maybe adding an `order` property to the items could be used to achieve the desired ordering of the links in the secondary nav.

```
      "docs": {
        "link": "docs",
        "text": "Docs",
        "es6": {
            "link": "docs/es6",
            "text": "ES6 in Node.js",
            "order": 1
        },
        "faq": {
            "link": "docs/faq",
            "text": "FAQ",
            "order": 2
        },
        "api": {
            "link": "/api/",
            "text": "API",
            "order": 3…
```

Or perhaps we could restructure the data for the secondary navigation into an array - something like this for example:

```
    "docs": {
        "link": "docs",
        "text": "Docs",
        "nav": [
            {"link": "docs/es6", "text": "ES6 in Node.js"},
            {"link": "docs/faq", "text": "FAQ"},
            {"link": "/api/", "text": "API"},
            {"link": "knowledge", "text": "Knowledge Base", "nav": […]}
        ]
    }
```

Either of these could provide what the data the Handlebars helper might need.

fhemberger:
The order of the menu items equals the order in the json. Please don't add more complexity than necessary. ;)

mrkiffie:
I've always considered the ordering of objects to be somewhat volatile. I might be lacking in my understanding of the way node handles JSON objects. 

My understanding is more theoretical based and I haven't gone into much testing.

Does node reliable maintain the order of properties within JSON objects?

> An object is an unordered collection of zero or more name/value pairs 
> <cite> - [JSON RFC](http://www.ietf.org/rfc/rfc4627.txt)</cite>

mrkiffie:
The sidebar's menu's have started to grow in complexity and have deviated from a simple list managed by the `site.json`. 

[...]

PR #409, opened 2015-12-05
Adding dynamic navigation based on `site.json`

mrkiffie:
Implements a solution for the menus, partially as described in #281.
Changes to `site.json` are to replicate the current state of the various sidebar menus

**Menu Links**
- Are objects that have both a `link` and a `text` property
- The `link` can be an internal path relative to the local. e.g `docs/faq` 
- The `link` can be relative to the root. e.g `/api/`
- The `link` can be an external url. e.g `https://github.com/nodejs/nodejs.org`

**Main menu**
- Uses an inline loop instead of a template as it is only used in once. 
- Consists of menu links found on the first level.
- Displays menu links in the order in which they appear in `site.json`. 
- _I removed `trademark` from the object to prevent it showing up in the main name_

**Sub menus**
- All menu link that appear in the main menu are also traversed to extract sub menus.
- Sub menus include a reference to themselves as the first link (to replicate the current sidebar menu behaviour)
- Example usage   `{{> navigation key='about'}}`

**Navigation Helper**
The navigation helper adds the menus to the metadata under the property `nav`


mrkiffie:
@sup I've updated the PR - added a short description as you suggested. I also made a couple of formatting changes in the navigation helper that were failing on travis but not picked up pre-commit.

mikeal:
can we get a screenshot :)

mrkiffie:
Hey @mikeal, this doesn't change the appearance at all. It only changes way the navigation is managed. 

Instead of having to maintain different menu templates for different sidebar menus, they share a generic template and the menus are defined in each locale's `site.json`. 

The benefit will be more noticeable for additional translations of the site as they won't require the menu structure to be identical to the english version. (I'm not sure if this behaviour is desirable or not?)

mikeal:
ahhhh, ok :) :+1:

mrkiffie:
The merge conflict above is a result of deleting `layouts/partials/docs-menu.hbs` which has since had additions (#324). The merge conflict can be easily resolved, but real conflict in this PR is in the direction it takes the sidebar menu.

mrkiffie:
I have resolved the merge conflict. 

As far as my testing goes, I think that I've been able to replicate the current navigation.

I have added additional logic in the template to support the API doc's "LTS" label and the text is an optional `subtext` property in the locale's `site.json`.

Let me know if there are any additional considerations I have missed.

fhemberger:
@nodejs/website Can you please take a look and test it? This would be a **huge** improvement.

locale/en/site.json
```diff
@@ -72,16 +84,18 @@
             "link": "docs/faq",
             "text": "FAQ"
         },
-        "api": {
-            "text": "API"
+        "api-lts": {
+            "link": "dist/latest-v4.x/docs/api",
```
hiroppy: Is this ok?
I think that  `dist/latest-v4.x/docs/api` becomes `/en/dist/latest-v4.x/docs/api/` after generation.
However true url is `https://nodejs.org/dist/latest-v4.x/docs/api/`.

`dist/latest-v5.x/docs/api` is also the same.

Fishrock123:
Seems to work fine. LGTM

locale/en/site.json
Fishrock123: oh hmm, no that isn't ok. The docs are not translated to begin with. (due to various other issues)

mrkiffie: Good catch @abouthiroppy. I accidentally introduced this while trying to figure out how to handle the "LTS" label on the version 4 docs. I'll push a patch shortly.

mrkiffie:
I've update the PR with the following changes:
- Fix urls to LTS and Stable API docs - made relative to the domain's root instead of the locale
- Fix malformed html by adding missing closing tags (whoops 😳)
- Moving subtext conditional - because API docs start with a `/`

Please let me know if you spot anything else out of the ordinary.

hiroppy:
LGTM :+1:

fhemberger:
Merged, thank you all!

PR #414, opened 2015-12-09
add momentum scrolling

moorebrett0:
Add momentum scrolling for IOS makes reading the docs easier on mobile.

### DIFF
--- a/layouts/css/layout/_sticky-footer.styl
+++ b/layouts/css/layout/_sticky-footer.styl
@@ -9,6 +9,8 @@ body
     flex-direction column
     min-height 100vh
     margin 0
+    overflow-y scroll; /* has to be scroll, not auto */
+    -webkit-overflow-scrolling touch
 
 #main
     flex 1 0 auto
--- a/layouts/css/layout/_sticky-footer.styl
+++ b/layouts/css/layout/_sticky-footer.styl
@@ -9,7 +9,7 @@ body
     flex-direction column
     min-height 100vh
     margin 0
-    overflow-y scroll; /* has to be scroll, not auto */
+    overflow-y scroll /* has to be scroll, not auto */
     -webkit-overflow-scrolling touch
 
 #main


layouts/css/layout/_sticky-footer.styl
```diff
@@ -9,6 +9,8 @@ body
     flex-direction column
     min-height 100vh
     margin 0
+    overflow-y scroll; /* has to be scroll, not auto */
+    -webkit-overflow-scrolling touch
```
stevemao, line 13: are we using autoprefixer for this?

moorebrett0, line 13: Looks like the stylus-auto prefixer plugin was installed?    Is there anything special I need to do to configure it?

stevemao, line 13: I think its already configured.

`overflow-scrolling touch` should just work but its not really standard.

moorebrett0, line 13:
```
What's the standard way to turning on the momentum scrolling?    I really do enjoy browsing through the docs on my phone and it's such a pain on my thumb when I have to use it so much to find my spot again (this is the only way I know how :/).    
```

stevemao, line 13: I'm not saying your doing wrong. I'm saying its not standard feature. https://developer.mozilla.org/en-US/docs/Web/CSS/-webkit-overflow-scrolling

And I'm just saying autoprefixer should handle the prefixing.

moorebrett0, line 13: Ah, I see.    Sorry about that!

silverwind:
https://github.com/nodejs/node/pull/5878 adds momentum scrolling to the docs. If we're going with it, we should probably do it for the main site too.

cc: @lpinca

layouts/css/layout/_sticky-footer.styl
```diff
@@ -9,6 +9,8 @@ body
     flex-direction column
     min-height 100vh
     margin 0
+    overflow-y scroll; /* has to be scroll, not auto */
```
silverwind: Unnecessary semicolon?

lpinca:
Looks fine, but I'm gonna try it first.

lpinca:
Hmm I'm not sure why but I can't see the difference here with it enabled.
Anyway the property is not auto prefixed.

fhemberger:
@moorebrett0 Hi, can you please have a look at the comments above and update your PR? I'd like to get it merged.

lpinca:
I tried pulling again this change and can't see the difference on my iPhone. Scroll works the same with or without this change.
I vote for closing this if someone can confirm that there is no difference.

moorebrett0:
Hey there,   sorry for the delayed response - removed the semi colon.      This seems to work for me on ios-9 can anyone confirm?

lpinca:
@nodejs/website collaborators with iOS devices, can you please test this pull request?
Do you notice any difference when scrolling?

preco21:
There is no difference on my iPhone too.

Used device: iPhone 6+
Browser: Chrome for iOS, Safari

lpinca:
I close this as there is no advantage in getting this merged. If someone can prove otherwise, please reopen.

PR #416, opened 2015-12-11
80-wrap and reflink in simple-profiling

bengl:
As per https://github.com/nodejs/docs/blob/master/GETTING-STARTED.md#how-we-write


ghost:
LGTM!

PR #378, opened 2015-11-25
Comprehensive Events Page

mikeal:
Ok, here we go :)

This is code to:
- Pull data from meetup.com
- Pull data from nodeschool.io
- Display data from meetup, nodeschool, and any manually added data in the events.md yaml

The display is _terrible_, would love some help there before we merge this.

PR #418, opened 2015-12-13
Take advantage of the request dependency

lpinca:
With #378 we added the `request` dependency. Now that we have it I think that it makes sense to use it to make the code easier to read.


mikeal:
LGTM but I've never spent time in these files so I'd like to see someone who has weigh in.

phillipj:
Big :+1: for cleaning this up.

PR #426, opened 2015-12-21
Change index.js to server.js

Spaxe:
`npm start` defaults to run `node server.js`, which causes the previous example to fail because `index.js` was ignored.

This minor name change allows the example to be run properly.

See https://docs.npmjs.com/misc/scripts#default-values.


fhemberger:
LGTM

ghost:
LGTM!

PR #437, opened 2016-01-02
Extracted the server code out of build.js

phillipj:
Some refactoring for the code to be easier to grasp, and keep concerns separate.

To keep the build process found in build.js as focused as possible, this extracts the server portion into it's own file.

### DIFF
--- a/build.js
+++ b/build.js
@@ -2,8 +2,7 @@
 
 'use strict'
 
-// BUILD.JS: This file is responsible for building static HTML pages and a
-// server for local development.
+// BUILD.JS: This file is responsible for building static HTML pages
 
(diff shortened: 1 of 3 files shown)


stevemao:
Should `build.js` be called `client.js`?

ghost:
@stevemao in tradition with metalsmith, i'd leave it at `build.js`

phillipj:
With these changes, `build.js` strictly consist of build related code. Why would we call it `client.js` @stevemao?

stevemao:
I just thought the code in `build.js` is the client side opposite to `server.js`

fhemberger:
LGTM

PR #439, opened 2016-01-04
Point to Node.js v4 as a first choice for Red Hat installs

mcollina:
As titled.

Possibly a different solution might be found, but at least this is a quick fix. We want everybody on Node 4 asap :D.


lpinca:
:+1:

PR #445, opened 2016-01-05
Make position of "Edit on GitHub" link consistent

strawbrary:
Currently if the headline is long, it will push the "Edit on GitHub" link down as can be seen on the "Get Involved" page:
screen shot 2016-01-04 at 20 02 49

This commit moves the link before the headline so that it is always in the top right:
screen shot 2016-01-04 at 20 03 33


fhemberger:
Good catch, but we should try to move the edit link out of the `<h1>` completely for search engine indexing/a11y reasons.

strawbrary:
@fhemberger Thanks, I hadn't considered that. Probably would have been bad to have all the search engine results start with "Edit on GitHub" heh. Moved it outside and increased the `font-size` for `.edit-link` since it doesn't inherit the `h1` `font-size` anymore.

fhemberger:
Great, thank you! Merged.

PR #448, opened 2016-01-06
Fix missing link on some foundation pages

strawbrary:
The "Board" link is not present on the "Announcements" or "In the News" pages because they have separate templates whose menus don't include it. I moved the foundation menu into a partial and included it in the three foundation templates to fix it.


fhemberger:
This should be cross-checked with #409.

strawbrary:
Confirmed #409 fixes it.

PR #463, opened 2016-01-13
default to .xz files for all Linux downloads

rvagg:
I reckon it's time for this. Any objections from @nodejs/build?

Leaving source tarball alone for now for portability, OSX doesn't have xz installed by default.

### DIFF
--- a/layouts/partials/download-matrix.hbs
+++ b/layouts/partials/download-matrix.hbs
@@ -64,22 +64,22 @@
             </tr>
 
             <tr>
-                <th>Linux Binaries (.tar.gz)</th>
-                <td colspan="3"><a href="https://nodejs.org/dist/{{version}}/node-{{version}}-linux-x86.tar.gz">32-bit</a></td>
-                <td colspan="3"><a href="https://nodejs.org/dist/{{version}}/node-{{version}}-linux-x64.tar.gz">64-bit</a></td>
+                <th>Linux Binaries (.tar.xz)</th>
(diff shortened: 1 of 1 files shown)


Starefossen:
LGTM

jbergstroem:
LGTM

mhdawson:
This may be a dumb question but how long have linux distributions supported xz ?  Just want to be sure that on platforms like RHEL 6 and CENTOS 5.x it is already installed or easily available

jbergstroem:
@mhdawson can't speak for rhel6 but centos5 [doesn't have xz support](https://github.com/nodejs/build/issues/258) out of the box.

mhdawson:
Do we need some info on the download page that says if you can't use xz then the gz are under the "All download options"  ?  Might be too obvious.

rvagg:
.xz is getting pretty common these days, ever since the linux kernel project switched to it exclusively there's been a steady stream of adaption, it's even in `tar` these days (`-J`), even for distros where it's not available by default I bet people are getting used to dealing with it. I'd prefer to leave it alone but here's an option:

Current

screen shot 2016-01-14 at 5 01 23 pm

.xz note:

screen shot 2016-01-14 at 5 03 26 pm

jbergstroem:
How about treating the source tarball with a similar treatment?

rvagg:
cause xz on OS X is a pain in the backside, that's why I left it off, we could mention it in the note however

PR #465, opened 2016-01-14
#464 Change the footer year to 2016

arturoromeroslc:
Update the footer from ©2015 to ©2016.

Issue #464, opened 2016-01-14
Update the year on the footer


arturoromeroslc:
This one is a simple update to te website's footer. 

The edit will be changing ©2015 to ©2016.

Knighton910:
You didn't want to do a pull request & go about it that way? If not I'll handle it.

arturoromeroslc:
I will work on the pull request now, thanks.

Knighton910:
No worries :)

Knighton910:
:+1: @stevemao

arturoromeroslc:
Pull request is open.  https://github.com/nodejs/nodejs.org/pull/465

arturoromeroslc:
Pull request merged. Thank you @kelthenoble

Knighton910:
Anytime man :100:

PR #466, opened 2016-01-14
Footer: Set year for copyright automatically

fhemberger:
Little update to #464 and #465, so we don't have to update the footer each year. ;)

### DIFF
--- a/scripts/helpers/copyright-year.js
+++ b/scripts/helpers/copyright-year.js
@@ -1,0 +1,6 @@
+'use strict';
+
+module.exports = function () {
+
+    return (new Date()).getUTCFullYear();
+};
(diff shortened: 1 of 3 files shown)


Knighton910:
That makes sense, nice work @fhemberger dynamically :+1:

scripts/helpers/copyright-year.js
```diff
@@ -0,0 +1,6 @@
+'use strict';
+
+module.exports = function () {
+
+    return (new Date()).getFullYear();
```
stevemao: This should probably be `getUTCFullYear()`

fhemberger: @stevemao Oh, you're right. Thanks.

phillipj:
Just pushed some style issues introduced with this PR in 87db4eea4d83597f8dfdc1f5037df6db1e5a94da, too bad we can't get Travis up-n-running again due to org restrictions :(

Issue #431, opened 2015-12-30
against es6 in documentation


forgotPassword:
i don't know if it was decided/voted or not, but i am against es6 features in documentation or code examples.
the docs and examples should be as simple as possible, and now even the classic createServer example ('about' page) has const, arrow functions, and template strings.

before you could just say 'copy this example, see how easy it is to create a server in node?', now you also have to explain what all those special signs and characters are.

stevemao:
I believe these examples are in the docs of nodejs that support them. If you switch to an older version they shouldn't be there.

ghost:
thing is, the latest releases all support most ES6 features out of the box. even the LTS does that, so it's officially encouraged.
but what we might be able to do is add a link to an ES6 introduction under the code example on the about page in case someone is confused

but what @stevemao says does apply, the pre-4.x docs are still in ES5

wonderdogone:
I think it's good to push ES6 unless it's a feature set that is outside of Node internally. For example some ES6 that needs to be transpiled by external tool like babel should not be in docs because then you have to explain that workflow.

stevemao:
> but what we might be able to do is add a link to an ES6 introduction under the code example on the about page in case someone is confused

I don't think this is necessary. We already have docs about features in es2015. If we add a link to all the pages that have es2015 code it would be too noisy. Not to mention that there are many people already using es2015 and a there are a lot of material about es2015 online.

Closing this now. @forgotPassword if you find anything that does not work against the corresponding node version out of the box please fire a bug :) thanks.

rvagg:
Thanks for the link @stevemao, I knew there was an issue in here about this recently that I couldn't find. 

Folks in here, please see #467 for _possible_ way forward. Only discussion at this stage.

PR #467, opened 2016-01-14
Add ES5/ESnext switcher on about page

rvagg:
This is a proof of concept, my frontend JS is super-rusty so this will likely need a more skilled hand if we go with it. My styling skills also leave a lot to be desired. The purpose of this is for discussion and a possible way to address the growing concern and confusion about the increasing use of ES6+ in docs. Specifically being discussed here: https://github.com/nodejs/docs/issues/55 (and perhaps elsewhere?).

Here's what my PR adds:

eswut?

The markdown only needs to have `Modern JavaScript:` added before a code block, then `ES5:` followed by a code block with ES5. The work of putting it together is done in the browser, if for some reason that can't be done the fall back just displays the two blocks with titles above them so it gracefully degrades. This approach could be used across all docs, API docs included.

### DIFF
--- a/locale/en/about/index.md
+++ b/locale/en/about/index.md
@@ -10,7 +10,9 @@ scalable network applications. In the following "hello world" example, many
 connections can be handled concurrently. Upon each connection the callback is
 fired, but if there is no work to be done Node is sleeping.
 
-```javascript
+Modern JavaScript:
+
+```js
 const http = require('http');
 
 const hostname = '127.0.0.1';
@@ -21,6 +23,22 @@ http.createServer((req, res) => {
   res.end('Hello World\n');
 }).listen(port, hostname, () => {
   console.log(`Server running at http://${hostname}:${port}/`);
+});
+```
+
+ES5:
+
+```js
+var http = require('http');
+
+var hostname = '127.0.0.1';
+var port = 1337;
+
+http.createServer(function(req, res) {
+  res.writeHead(200, { 'Content-Type': 'text/plain' });
+  res.end('Hello World\n');
+}).listen(port, hostname, function() {
+  console.log('Server running at http://%s:%d/', hostname, port);
 });
 ```
 
--- a/layouts/about.hbs
+++ b/layouts/about.hbs
@@ -44,6 +44,84 @@
         </div>
     </div>
 
+    <style>
+        .es-switcher > p.active, .es-switcher > p:hover {
+            color: #FFF;
+            background-color: rgb(51, 51, 51);
+            border-color: rgb(51, 51, 51);
+        }
+        .es-switcher > p {
+            display: inline-block;
+            margin: 0px;
+            cursor: pointer;
+            padding: 0em 21px;
+            border-style: solid;
+            border-color: rgb(255, 255, 255);
+            border-width: 3px;
+            background-color: rgb(238, 238, 238);
+            border-top-right-radius: 3px;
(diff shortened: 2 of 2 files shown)


stevemao:
Ref: #431

phillipj:
Looks like a good approach 👍

benjamingr:
Related https://github.com/nodejs/docs/issues/55

benjamingr:
Looks like a great idea ! Good work!

I gladly volunteer to port all other existing doc pages to have the two versions if the help is needed.

eljefedelrodeodeljefe:
@rvagg skimming through all the code blocks on #4733 that looks really useful to me. Bearing in mind future changes to the language. Also, through adding dimension to the docs I can imagine lots of other use cases for tutorials and examples etc.... I'll grab this up, I guess.

locale/en/about/index.md
```diff
@@ -24,6 +26,22 @@ http.createServer((req, res) => {
 });
 ```
 
+ES5:
+
+```javascript
```
stevemao: this should be ```js (https://github.com/nodejs/docs/blob/master/STYLE-GUIDE.md bottom).

rvagg:
@benjamingr since frontend is not my forte, it's been too long since I kept track of what browsers support what, and I was putting this up as a POC initially, would you mind taking this over? The branch is in the repo so you can take it and do what you like with it until you're happy with it.

eljefedelrodeodeljefe:
@benjamingr, @rvagg also I had a quick look a while ago, and I think the API docs are using features like `display: flex`, which liekly exclude IE8 support anyhow. Since Microsoft is dropping support for it, I wouldn't put too much work into compat.

xcatliu:
@benjamingr @rvagg @eljefedelrodeodeljefe 

According to [Web Browser Market Share Trends](http://www.w3counter.com/trends), IE8 has only 1.1% users in total (2015-12).
And I tried https://nodejs.org/en/about/ in IE8, it looks like this:

image

So I think it's time to say goodbye to IE8.

benjamingr:
@eljefedelrodeodeljefe `display: flex` excludes IE9 too IIRC,

I vote to disable support for any browsers under latest (since it's for developers). JavaScript developers don't use old browsers anyway. The only case I can think of that makes me sad here is developers stuck debugging things on computers with old IE (like IIS servers) and want to open the docs - that's not really that big of a deal though. 

@rvagg assuming this idea (switching between modern JS and ES5) gets accepted - I can gladly revamp the code base for it and as I said before add such examples for all code examples on the site docs. I just don't want to do a lot of work while we're not sure yet it'll be used (and the PoC does a good job at illustrating the usefulness).

stevemao:
> So I think it's time to say goodbye to IE8.

Thanks for testing this @xcatliu , lets say good bye to old ies :+1:

silverwind:
@benjamingr Flexbox support starts with IE10 (buggy and prefixed), IE11 should be fine.
@eljefedelrodeodeljefe API docs don't need Flexbox and work for me in IE11's IE7 mode with a few minor issues.

chrisdickinson:
Noting this here for future folks: the [Docs WG resolution](https://github.com/nodejs/docs/issues/55#issuecomment-173324666) was not to apply this approach to all of the docs due to the maintenance burden of keeping up two examples for every existing example, and instead look for a better way to highlight confusing syntax (like backticks.)

Issue #468, opened 2016-01-14
Unable to access workflow page


sumitsharma0999:
URL: http://nodejs.org/documentation/workflow/

![image](https://cloud.githubusercontent.com/assets/1701967/12324710/9da4ab08-baea-11e5-801b-75a9211e338c.png)

ghost:
for me, it redirects to https://nodejs.org/en/docs/. paging @phillipj, do you know where this is supposed to lead?

stevemao:
It redirects for me too.

ghost:
@sumitsharma0999 from where did you get to that URL?

phillipj:
> for me, it redirects to https://nodejs.org/en/docs/

Yupp, that's deliberate by the following [nginx rules](https://github.com/nodejs/build/blob/master/setup/www/resources/config/nodejs.org).

The workflow URL in question is mentioned in TJ's next-chapter blog post: https://nodejs.org/en/blog/community/next-chapter/ which is available in the old nodejs.org-archived repo: https://github.com/nodejs/nodejs.org-archive/blob/master/doc/contribute/code_contributions/workflow.md

As good as that workflow documentation is, it doesn't apply anymore AFAIK. Today's equivalent would probably be https://github.com/nodejs/node/blob/master/CONTRIBUTING.md

sumitsharma0999:
For me also, it redirects now to https://nodejs.org/en/docs/ - this looks like a temporary issue.

Regarding the source of URL, I was browisng through https://nodejs.org and I am unable to the recall where exactly the link was. But I am pretty sure it was either on the home page or atmost 1 level down. I tried to find the link now, but was unable to find it.

rnsloan:
Should the nginx redirect be changed to the document that has superseded the old workflow document https://github.com/nodejs/dev-policy

phillipj:
> Should the nginx redirect be changed to the document that has superseded the old workflow document https://github.com/nodejs/dev-policy

Maybe... I was hesitant to reference it since it hasn't been updated in 8 months and it still says draft in the repo title. Any thoughts from @nodejs/documentation?

fhemberger:
@phillipj New Development Policy can be found here: https://nodejs.org/en/get-involved/development/

phillipj:
[...]

PR #473, opened 2016-01-19
Change dev policy URL on page about core contribution

phillipj:
This changes the three dev policy links displayed in the [Contribute](https://nodejs.org/en/get-involved/contribute/) page, to what's the updated and future dev policy / workflow documentation: https://nodejs.org/en/get-involved/development/ as mentioned in https://github.com/nodejs/nodejs.org/issues/468#issuecomment-172612818

Refs #468.

/cc @nodejs/documentation

### DIFF
--- a/locale/en/get-involved/contribute.md
+++ b/locale/en/get-involved/contribute.md
@@ -24,10 +24,10 @@ The Node.js project is currently managed across a number of separate GitHub repo
 
 ## Code contributions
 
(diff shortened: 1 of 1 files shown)


phillipj:
I'm merging this as I've got the impression that the links we're currently display (https://github.com/nodejs/dev-policy) are outdated and not going to be maintained.

/cc @nodejs/documentation

PR #480, opened 2016-01-22
Added missing , in docker example npm script

phillipj:
..this also serves as a test if Travis CI picks this PR up and displays a inline build status.


phillipj:
Too document for later; Travis integration still not visible in PRs, even though Travis had 3d party access for a while as said in https://github.com/nodejs/nodejs.org/issues/355#issuecomment-173989061.

PR #483, opened 2016-01-26
Instruction how to get the ip-address of the current machine

stefanwalther:


silverwind:
Also have to agree that `docker-machine` and the likes have no place in this guide. That information is just to specific to a certain setup.

fhemberger:
Agreed, this is a bit too specific for this guide. How to work with Docker in general on your environment is part of Docker's own getting started guides. But thanks for bringing this up anyway.

PR #490, opened 2016-01-28
Falling back to english when page hasn't been translated locally

phillipj:
Currently translation working groups will have to translate _all_ pages to their respective
language, if not most pages would end in displaying a (english) 404 page. That's a massive
requirement on translation groups, and keeping up with future changes are probably near impossible.

Lowering the barrier and letting translation groups start on the frontpage and work their way
down a couple of page depths, should be alot simpler and more rewarding.

NB! This only affects local development as nodejs.org runs nginx serving strictly static content.
Getting the same behaviour in production requires changes to the nginx config.

Example of the current 404 issue; navigate any of the headers links in the current Korean translation: https://nodejs.org/ko/. I haven't done my dues when it comes to translating (yet), this is one of the main reasons I've decided not to.

Any thoughts? Should we do this, and roll it out in production aswell?

### DIFF
--- a/server.js
+++ b/server.js
@@ -10,7 +10,8 @@ const chokidar = require('chokidar')
 const mount = st({
   path: path.join(__dirname, 'build'),
   cache: false,
-  index: 'index.html'
+  index: 'index.html',
+  passthrough: true
 })
 
 const build = require('./build')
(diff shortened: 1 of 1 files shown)


stevemao:
roll it out in production +1

ghost:
yup, agree with @stevemao here

Knighton910:
LGTM

phillipj:
Thanks for reviewing :+1: I've opened https://github.com/nodejs/build/pull/316 to fix this in production aswell, not much point in merging this before that PR gets some positive feedback and lands.

Did a minor refactor to get more inline with what nginx will end up doing. Rather than serving the fallback file from disk, it now responds with a 302 redirect to the equivalent english URL.

FYI I'm going on one week vacation in a couple of days 🏊🌴Will follow this up when I'm back on the keyboard again.

phillipj:
If anyone's up for some nginx exercise, feel free to join the discussion in https://github.com/nodejs/build/pull/316 :)

fhemberger:
Waiting for https://github.com/nodejs/build/pull/316 to land.

ghost:
this would be really nice to have now that we have a mostly unfinished chinese translation

phillipj:
@tentakel thought we had nginx solved, but obviously not; https://github.com/nodejs/build/pull/316#issuecomment-219318546. Hopefully I'll get an ansible setup of nodejs.org soon, too debug what's wrong with the suggested changes.

Fishrock123:
https://github.com/nodejs/build/pull/316 has been deployed live now and seems to be going fine, I think this can be merged/reconsidered

Fishrock123:
cc @fhemberger

fhemberger:
@Fishrock123 Great, thanks for letting me know. Merging now!

Issue #493, opened 2016-01-29
LTS Roadmap on website


mikeal:
We've been getting questions about LTS a lot and it would be good to have it one the website rather than just in the README on the LTS repo.

@nodejs/documentation 

For reference:

![screen shot 2016-01-29 at 12 44 22 pm](https://cloud.githubusercontent.com/assets/579/12683554/ba04b088-c66e-11e5-928a-87c913960833.png)

eljefedelrodeodeljefe:
The easiest I can come up with immediately would be... (Link to: https://raw.githubusercontent.com/nodejs/LTS/master/schedule.png)

![node_js](https://cloud.githubusercontent.com/assets/3899684/12684312/8a350a56-c6be-11e5-900a-3a20f4954894.png)

gtewallace:
+1 thanks @eljefedelrodeodeljefe

mikeal:
I think we need more that just the graphic, the graphic is great but a lot of the context for what LTS is we need text for.

eljefedelrodeodeljefe:
Should we make it incremental? ...have the above not and create some whole subpage? This would probably take way bigger effort, since I wouldn't just append `LTS` to the navbar.

mikeal:
Ya, you're right, getting _something_ up soon is best, I'm just wondering if we should link to the whole README or just to the graphic?

eljefedelrodeodeljefe:
If we could have an anchor on the readme above the image that should be fine. I'll open a pull request for the above and on the readme.

MylesBorins:
Long term I think having some sort of interactive info graphic that is accurate as of the current time would be very useful

eljefedelrodeodeljefe:
The interim change would be ready.

Trott:
There is now a link to the LTS schedule on the front page. I think this can be closed as stale/inactive/done-ish. Feel free to re-open if you think I'm wrong.

PR #494, opened 2016-01-29
website: hotfix -> have link to lts schedule

eljefedelrodeodeljefe:
As per discussion on #493

### DIFF
--- a/locale/en/index.md
+++ b/locale/en/index.md
@@ -13,6 +13,8 @@ labels:
   tagline-lts: Mature and Dependable
   changelog: Changelog
   api: API Docs
+  version-schedule-prompt: Or have a look at the
+  version-schedule-prompt-link-text: LTS schedule
 ---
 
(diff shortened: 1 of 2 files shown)


locale/en/index.md
```diff
@@ -13,6 +13,9 @@ labels:
   tagline-lts: Mature and Dependable
   changelog: Changelog
   api: API Docs
+  version-schedule-prompt: Or have a look at the
+  version-schedule-prompt-link: https://raw.githubusercontent.com/nodejs/LTS/master/schedule.png
```
phillipj: This URL doesn't need to be different in other languages, right? If so it could just be placed directly inside the href-attribute in `layouts/index.hbs`

eljefedelrodeodeljefe: That's right I'll wait for nodejs/LTS#73 and change it then.

phillipj: :+1:

phillipj:
Thanks!

PR #498, opened 2016-01-31
not to mention deprecated domains module

### LINKED ISSUE #497: Guide 'Anatomy of an http transaction' refers to deprecated module 'Domain'.
https://nodejs.org/en/docs/guides/anatomy-of-an-http-transaction/

> There are other ways of handling these errors such as using domains or other ...

stevemao:
Fixes #497

### DIFF
--- a/locale/en/docs/guides/anatomy-of-an-http-transaction.md
+++ b/locale/en/docs/guides/anatomy-of-an-http-transaction.md
@@ -126,7 +126,7 @@ request.on('error', function(err) {
 });
 ```
 
-There are other ways of [handling these errors][] such as using [`domains`][] or
+There are other ways of [handling these errors][] such as
 other abstractions and tools, but always be aware that errors can and do happen,
 and you're going to have to deal with them.
 


locale/en/docs/guides/anatomy-of-an-http-transaction.md
```diff
@@ -126,7 +126,7 @@ request.on('error', function(err) {
 });
 ```
 
-There are other ways of [handling these errors][] such as using [`domains`][] or
+There are other ways of [handling these errors][] such as
```
Trott, line 129: With the specific example of `domains` removed, the phrase `such as other abstractions and tools` seems a little awkward and uninformative. I'd remove those words too. I'd even consider removing the entire sentence.

ghost, line 129: @Trott my mistake. i've removed the entire sentence for now, do you think we can replace it in some way?

ghost, line 129: actually, that was another mistake - i committed directly, argh. should i revert the commit manually?

Trott, line 129: I actually don't know. Things are a bit looser in the nodejs/nodejs.org repo than they are in the nodejs/node repo. My best guess is that if all you did was modify this guide and it's a small modification, don't worry about it too much. Make a PR to note the change and what further changes you propose.

ghost, line 129: @Trott how would i go about making a PR for changes that have already landed though?

Trott, line 129: I just mean that if you have additional changes, mention the changes that already happened in the PR for the additional changes. Although now that I'm thinking about this, it's easy enough to just send a revert commit and then submit a proper PR. So, you know, if you want to set a good example and be a good citizen, might as well do that.

PR #516, opened 2016-02-12
link to .xz file from front page for Linux

### LINKED ISSUE #329: Default downloads for linux still tar.gz on main node page
@sxa555 mentioned that when you hit the main page (https://nodejs.org/en/) with a linux machine the downloads are still tar.gz even though the default on the download page has been switched to tar.xz

rvagg:
completes: #463
fixes: https://github.com/nodejs/build/issues/329


MylesBorins:
LGTM

rvagg:
confirmed working, thanks @stevemao

PR #528, opened 2016-02-23
release posts: add ppc64le & link to tar.xz for Linux & SunOS

rvagg:
/cc @nodejs/build 

matches what I did @ http://nodejs.org/en/blog/release/v5.7.0/

@mhdawson I called it "Linux PPC LE 64-bit Binary" there, and in this PR, is that an idiomatic way to refer to it?

### DIFF
--- a/scripts/helpers/downloads.js
+++ b/scripts/helpers/downloads.js
@@ -30,31 +30,35 @@ const postMergeDownloads = [
   },
   {
     'title': 'Linux 32-bit Binary',
-    'templateUrl': 'https://nodejs.org/dist/v%version%/node-v%version%-linux-x86.tar.gz'
+    'templateUrl': 'https://nodejs.org/dist/v%version%/node-v%version%-linux-x86.tar.xz'
   },
   {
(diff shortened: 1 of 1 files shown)


MylesBorins:
LGTM. Curious as to why the switch from `.gz.` to `.xz`

rvagg:
`.xz` is much tiny, many wow, compare sizes @ https://nodejs.org/download/release/v5.7.0/, xz compression is available on every modern Linux distro now and you can even use `tar -J` to uncompress them. Still a hassle to do it on OSX hence leaving the darwin binaries and source tarball .gz by default.

ghostbar:
Why the source code is on `.tar.gz` yet? For OSX people as well? (That would make sense, just confirming)

mhdawson:
"Linux PPC LE 64-bit Binary" sounds ok to me but I'll double check with our power team

mhdawson:
Feeback from power team is that ""Linux PPC LE 64-bit Binary" is fine

rvagg:
great, OTY @nodejs/website

PR #530, opened 2016-02-23
Express Annoucement

mikeal:
This should have gone up a while ago.


Fishrock123:
Looks like the date formatting is incorrect? https://github.com/nodejs/nodejs.org/issues/531

rvagg:
americans and their weird dates

perhaps this would be a good addition to the precommit hook? check date is valid and not too far in the future

stevemao:
@rvagg oh well `02-10-2016` would be Americans. I'd like to see how to write a hook for that :)

rvagg:
`2016-10-02` > `2016-02-24`, that's a pretty easy hook

stevemao:
Oh I thought you'd find all dates in the page. If just the header one its easy.

Issue #531, opened 2016-02-23
Wrong date formatting on Express Announcement


julianduque:
Current date shows post as of Oct 02

![](https://cldup.com/Hqy7eD0v_x.png)

PR #532, opened 2016-02-23
blog: fix date formatting for express announcement

Fishrock123:
refs: https://github.com/nodejs/nodejs.org/issues/531 & https://github.com/nodejs/nodejs.org/pull/530

r= @mikeal


ghost:
good catch on the spaces there.

Fishrock123:
@hg oh uh, that wasn't really intentional, my editor does it automatically (Atom). If we don't want that for markdown we should add an `.editorconfig` [like core](https://github.com/nodejs/node/blob/master/.editorconfig).

stevemao:
@Fishrock123 I really hate using trailing spaces for newlines. We should avoid it.

PR #534, opened 2016-02-23
Update icons to new logo

silverwind:
Iconized the new logo as a favicon:

favicon

The new PNG uses only 1.6kB as opposed to the previous ICO at 15kB and the new ICO at 6kB, reducing download size for modern browsers. IE10 and below [can't handle png favicons](http://caniuse.com/#feat=link-icon-png), but should fallback to the ICO, I think.


mikeal:
Oh that is really cool!

Need to get this on the live.nodejs.org site as well :)

PR #536, opened 2016-02-25
update node-green to pass WCAG AA

### LINKED ISSUE #501: $node-green #80bd01 on white background does not pass accessible contrast ratio
Currently standard links on the website appear in `#80bd01` most often on a white background. Chrome accessibility tools indicate the contrast ratio is 2.28 and passing for AA level accessibility at the ~15px font-size on the API docs, like this page https://nodejs.org/api/fs.html would be a contrast ratio of ~3 which equates to a suggested `#6ea400`.

I personally think it is important that the website is properly accessible to everyone, but realize this creates some conflict between a Node brand color and what color combinations are accessible.

Please let me know if I can be of any assistance in resolving this, I felt it was premature to just open a PR suggesting a new color where it is a brand issue.

jrit:
Based on the new colors defined with the updated logo,
this is the the closest green that passes on a white
background for small text where a contrast ratio
of 4.5 is required.

Resolves #501 

Here are some previews of how this will look:

nodejs-docs
nodejs-home


mikeal:
The big buttons seem a little dark. I'm worried about them looking actionable enough, maybe a slightly brighter green and the green in this screenshot as the rollover color?

jrit:
Thanks @mikeal - updated the buttons, I agree they looked dull. I shifted the rollover color a bit based on the default state. Also had to update the download page a bit since the gray background during hover made the new node-green nearly unreadable.

button revised

downloads

ghost:
i actually prefer the first set of button colors :X the second set looks a bit too much like `color: green` or something

jrit:
I'll let this stand until at least tomorrow to maybe get some broader consensus. Thanks.

ghost:
@jrit thank you for doing this :smile:

stevemao:
I generally like the new colours. Thanks @jrit :)

phillipj:
I prefer the brighter ones in your updated version

jrit:
Sounds like 2 votes for bright buttons and one for dull. I'm leaning toward the brighter one as well. I'm on standby if anyone has any further requirements before this can be merged.

ghost:
we should wait for @mikeal to respond since he was the one that originally asked for the color change :P 
will merge after that though!

Fishrock123:
Woot, the website's colors have felt just not right for like a week now since the new logo haha. :P

layouts/css/_variables.styl
```diff
@@ -2,7 +2,8 @@ $body-max-width = 980px
 
 $white = #fff
 
-$node-green = #80bd01
+$node-green = #43853d
+$bright-green = #028a00
 $light-green = #f1fbda
 $hover-green = #d9ebb3
```
stevemao: Should these two be updated with the new color guide?

layouts/css/page-modules/_home.styl
```diff
@@ -51,7 +51,7 @@
     transition .2s background-color ease-in-out
 
     &:hover
-        background-color saturation($node-green, 80%)
+        background-color darken($bright-green, 20%)
```
stevemao: Then this could use `$hover-green`?

layouts/css/_variables.styl
jrit: I'll take a look at where/how they are being used

jrit:
Thanks all for your feedback. The suggestion to consolidate a few styles led me to consolidate some of the visuals as well so the active/hover/on states feel more consistent. I also fixed the contrast of green text on the light-green background. One other very minor change is adjusting the line-height and padding on the left navigation so when an item wraps, like "Technical Governance" it doesn't look like it is two separate links. Arrows on the below indicate hover states.

nodejs-foundation
downloads
nodejs-home

@stevemao @mikeal

silverwind:
Damn, I totally missed this PR. I almost filed my own PR with these changes:

``` patch
-$node-green = #80bd01
-$light-green = #f1fbda
-$hover-green = #d9ebb3
+$node-green = #44883e
+$light-green = #e4f2e3
+$hover-green = #c0e0bd
```

And the download button hover:

``` patch
     &:hover
-        background-color saturation($node-green, 80%)
+        background-color saturation($node-green, 60%)
```

What you might want to adapt is my `$hover-green` and `$light-green` which are different lightness shades of `#44883e`.

silverwind:
Oh, and disregard me, I should've read the changes first. Overall LGTM, except those platform logos which are still the old color. I'd convert these to SVG, but that can likely happen at a later point in time.

layouts/css/_variables.styl
```diff
@@ -2,9 +2,10 @@ $body-max-width = 980px
 
 $white = #fff
 
-$node-green = #80bd01
-$light-green = #f1fbda
-$hover-green = #d9ebb3
+$node-green = #43853d
```
silverwind, line 5: Why not `#44883e` (http://rgb.to/pantone/7741-c)? I'd say it's good enough contrast.

jrit:
@silverwind I agree about the platform icons on both color and file format. If you want to open a PR for that I bet it could just follow on this one being merged.

layouts/css/_variables.styl
jrit, line 5: I think that was the original spec color (correct me if I'm wrong) and it turned out to just barely not pass WCAG, which was actually the impetus for this PR originally. It is my hope in the future we can include WCAG validation as part of the site build process, so even the minor difference would cause a failure. Just taking care of the most important contrast differences for now, baby steps.

silverwind, line 5: Well, It's barely noticeable, so I don't have much of an opinion, but I'd lean on more the side of using the original color than to make it match the guideline.

silverwind:
Filed https://github.com/nodejs/nodejs.org/pull/548 for SVG icons, they use `$node-green`.

silverwind:
And those SVGs are in master. I don't think you need to change anything in your PR, but best to double-check.

jrit:
@silverwind made a small update to fill SVGs with white on hover as they were the same color as the background otherwise. Should be good now.

silverwind:
Okay, merged. Should have squashed to commits in hindsight, hope it's fine.

jrit:
Thank you!

jrit:
I thought maybe https://nodejs.org/dist/latest-v5.x/docs/api/ would get updated, but that CSS lives somewhere else I guess. Do you know where that needs to be changed @silverwind ?

silverwind:
API is handled here: https://github.com/nodejs/node/blob/master/doc/api_assets/style.css

I'm preparing a pull request for it right now, but it will only get live on next release of each branch, unfortunately.

jrit:
:+1:

ricordisamoa:
FWIW, I think the old buttons looked more friendly, playful, tempting, less serious.

stevemao:
The API docs are still using the old theme.

jrit:
@stevemao see a couple comments above yours

PR #538, opened 2016-02-25
comment gitignore and add commonly ignored stuff

ghost:
was missing the infamous `.DS_Store`. also, are there any windows/linux files to be commonly ignored? i'm running OSX, so i don't really know.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,11 @@
+# Generated HTML and other static files
 build/
+# Commonly ignored Node.js files
 node_modules/
 npm-debug.log
+.npm/
+
+# OSX system files, the bane of our existence
+.DS_Store
+.AppleDouble
+.LSOverride


phillipj:
[Travis build passed](https://travis-ci.org/nodejs/nodejs.org/builds/111858467) :+1:

phillipj:
😎

.gitignore
```diff
@@ -1,3 +1,11 @@
+# Generated HTML and other static files
 build/
+# Commonly ignored Node.js files
 node_modules/
 npm-debug.log
+.npm/
+
+# OSX system files, the bane of our existence
```
stevemao, line 8: These should be in your global I think.

ghost, line 8: i always put them in there for safety :S

phillipj, line 8: I'm also used to including these in every project. I never remember to add these globally, so I like to not "require" others to do it either.

PR #548, opened 2016-03-02
Use SVG logos on the download page

silverwind:
Replaced platform logos with SVG and use `$node-green` on them. Also updated the Windows logo in process:

cc: @jrit


ghost:
[Travis build passed](https://travis-ci.org/nodejs/nodejs.org/builds/113218589) :+1:

ghost:
LGTM

silverwind:
Thanks!

Trott:
Now if we could just get an SVG version of the new Node.js logo instead of an 1843x1129 PNG resized in CSS on the home page, WE'D BE GOLD!!!!!!!

(Not meant to disparage anyone or anything. I mean, it's not like I'm capable of creating an SVG myself without messing it up horribly. Nothing but respect. But uh, if someone wants to SVG-ize the Node logo, that would be great so long as the gradients don't cause the file size to be ridiculous. Or is that basically the problem with that approach?)

Trott:
(Also, if we have a reasonable SVG version, maybe we can put it in the nodejs/node repo README for EXTRA PIZZAZZ!)

silverwind:
Yes +100, that downscaled scaled png logo is pure image abuse.

@mikeal do you have the original illustrator resource for the logo? If you could export that as svg, I can work on cutting it down to a reasonable size.

PR #551, opened 2016-03-03
v0.12.11 release post

rvagg:
@phillipj FYI this is my first use of the release post script for <v4 and it worked like a charm, thankyou! I manually edited the ChangeLog text to be more markdowny but it was mostly fine as is. One tweak that could be made is that `Notable changes:` doesn't come with a `###` on <v4, if we had that automatically added then the rest would be fine to go as is (although I'd personally still want to make the changelog more markdowny but that's not strictly necessary).


phillipj:
Good to hear it work 👍 I'll take a look at fixing the header you mentioned

PR #553, opened 2016-03-04
v0.10.43

rvagg:
@nodejs/website `./scripts/release-post.js 0.10.43` is trying to fetch the ChangeLog from nodejs/node-v0.x-archive rather than nodejs/node, it looks like the changelog-url library needs to be updated to know when to start looking at nodejs/node for v0.10 releases.


phillipj:
^^ https://github.com/phillipj/changelog-url

PR #554, opened 2016-03-04
Typo fix

ahmetanilgur:
Installed void linux distro and tried the package manager and it fails to download the given file. The actual package name is nodejs.


ghost:
[Travis build passed](https://travis-ci.org/nodejs/nodejs.org/builds/113660565) :+1:

fhemberger:
Good catch, thanks!

PR #564, opened 2016-03-09
add ppc binaries to download-matrix

MylesBorins:
Both LTS and Stable now have PPC 64le versions. Let's publish that on
the downloads page of nodejs.org

edit: simplified this pr

/cc @mhdawson @jasnell

### DIFF
--- a/layouts/css/page-modules/_download.styl
+++ b/layouts/css/page-modules/_download.styl
@@ -1,8 +1,8 @@
 .download-header
-
     h1
         width 40%
         float left
+        margin-bottom: -1rem
 
     ul
         margin-top 2rem
@@ -102,6 +102,7 @@
             height 100%
             width 100%
             text-align center
+            font-size: 0.8em;
 
         a:hover, a:active
             background $node-green
--- a/layouts/download-stable.hbs
+++ b/layouts/download-stable.hbs
@@ -12,38 +12,8 @@
                     <h1>{{downloads.headline}}</h1>
                 </div>
 
-                {{> download-matrix version=project.currentVersions.stable versionTypeStable="true"}}
-
-                <section>
-                    <h2>{{additional.headline}}</h2>
-                    <p>{{additional.intro}}</p>
-
-                    <table class="download-matrix full-width">
-                        <tbody>
-                            <tr>
-                                <th>{{additional.platform}}</th>
-                                <th>{{additional.provider}}</th>
-                                <th></th>
-                            </tr>
-                            <tr>
-                                <th>AIX on Power Systems</th>
-                                <td><span>IBM</span></td>
-                                <td><a href="http://www.ibm.com/developerworks/web/nodesdk/">{{download}}</a></td>
-                            </tr>
-                            <tr>
-                                <th>Linux on Power Systems</th>
-                                <td><span>IBM</span></td>
-                                <td><a href="http://www.ibm.com/developerworks/web/nodesdk/">{{download}}</a></td>
-                            </tr>
-                            <tr>
-                                <th>Linux on System z</th>
-                                <td><span>IBM</span></td>
-                                <td><a href="http://www.ibm.com/developerworks/web/nodesdk/">{{download}}</a></td>
-                            </tr>
-                        </tbody>
-                    </table>
-
-                </section>
+                {{> primary-download-matrix version=project.currentVersions.stable versionTypeStable="true"}}
+                {{> secondary-download-matrix version=project.currentVersions.stable versionTypeStable="true"}}
 
             </article>
 
--- a/layouts/download.hbs
+++ b/layouts/download.hbs
@@ -12,38 +12,8 @@
                     <h1>{{downloads.headline}}</h1>
                 </div>
 
-                {{> download-matrix version=project.currentVersions.lts versionTypeLts="true"}}
-
-                <section>
-                    <h2>{{additional.headline}}</h2>
-                    <p>{{additional.intro}}</p>
-
-                    <table class="download-matrix full-width">
-                        <tbody>
-                            <tr>
-                                <th>{{additional.platform}}</th>
-                                <th>{{additional.provider}}</th>
-                                <th></th>
-                            </tr>
-                            <tr>
-                                <th>AIX on Power Systems</th>
-                                <td><span>IBM</span></td>
-                                <td><a href="http://www.ibm.com/developerworks/web/nodesdk/">{{download}}</a></td>
-                            </tr>
-                            <tr>
-                                <th>Linux on Power Systems</th>
-                                <td><span>IBM</span></td>
-                                <td><a href="http://www.ibm.com/developerworks/web/nodesdk/">{{download}}</a></td>
-                            </tr>
-                            <tr>
-                                <th>Linux on System z</th>
-                                <td><span>IBM</span></td>
-                                <td><a href="http://www.ibm.com/developerworks/web/nodesdk/">{{download}}</a></td>
-                            </tr>
-                        </tbody>
-                    </table>
-
-                </section>
+                {{> primary-download-matrix version=project.currentVersions.lts versionTypeLts="true"}}
+                {{> secondary-download-matrix version=project.currentVersions.lts versionTypeLts="true"}}
 
             </article>
 
--- a/layouts/partials/download-list.hbs
+++ b/layouts/partials/download-list.hbs
@@ -1,0 +1,9 @@
+<section>
+  <ul>
+      <li><a href="https://nodejs.org/dist/{{version}}">{{site.all-downloads}}</a></li>
+      <li><a href="/{{site.locale}}/{{site.download.package-manager.link}}">{{site.download.package-manager.text}}</a></li>
+      <li><a href="/{{site.locale}}/{{site.download.releases.link}}/">{{site.download.releases.text}}</a></li>
+      <li><a href="https://nodejs.org/download/nightly/">{{site.nightly}}</a></li>
+  </ul>
+  <p>{{downloads.buildDisclaimer}}</p>
+</section>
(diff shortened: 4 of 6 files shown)


layouts/partials/download-matrix.hbs
```diff
@@ -68,6 +68,11 @@
                 <td colspan="3"><a href="https://nodejs.org/dist/{{version}}/node-{{version}}-linux-x86.tar.xz">32-bit</a></td>
                 <td colspan="3"><a href="https://nodejs.org/dist/{{version}}/node-{{version}}-linux-x64.tar.xz">64-bit</a></td>
             </tr>
+            
+            <tr>
+                <th>Linux PPC Binaries (.tar.xz)</th>
+                <td colspan="6"><a href="https://nodejs.org/dist/{{version}}/node-{{version}}-linux-ppc64le.tar.xz">64le</a></td>
```
MylesBorins: would this be the way you would like it displayed @mhdawson? `64le`

ghost:
[Travis build passed](https://travis-ci.org/nodejs/nodejs.org/builds/114750539) :+1:

rvagg:
pls don't merge this yet, discussion going on here needs to come to a conclusion: https://github.com/nodejs/build/issues/205#issuecomment-187464137

MylesBorins:
thanks for the context @rvagg

Just noticed it was missing after being added and didn't connect the dots

How it looks now for example

screen shot 2016-03-09 at 3 01 15 am

I don't think it looks terribly cluttered, but I have no problem closing this in lieu of a better fix

rvagg:
My concerns are both about clutter and usability, which are obviously connected concepts. I believe it's becoming increasingly difficult for users to select the downloads that they need and the expanding list is adding to the confusion.

We have targets on this list that have very low usage, PPC is clearly going to be one of those as it's not a mass-market platform, it also carries the confusing history of being an Apple architecture up until the switch to Intel. "SunOS" is problematic because it's really just SmartOS (these binaries won't work on Solaris) and its download numbers are tiny (SmartOS users are pkgin to get their Node most of the time I think). The "Docker Image" isn't a download link per se and arguably doesn't belong on this particular list. ARM Binaries don't come near the Intel-based platforms in terms of download numbers and users appear to have problems self-selecting which one they need—based on issue regularly filed against nodejs/node, so we could probably move it and improve how we present them.

And to blunt, inclusion of SmartOS and PPC binaries on this page are more about marketing for their respective platform owners than about giving actual users a place to download binaries. It's much more likely that both of these platforms will have more convenient ways of consuming binaries than downloading from here. I have no problem with flag waving here, we all benefit from being able to show the diversity of support Node has, I'd just rather it be done in a way that increases difficulty for the vast bulk of visitors to this page who have no interest in uncommon platforms.

/cc @nodejs/build

mhdawson:
Based on earlier discussion I think we'd discussed 2 sections.  One with the downloads for Linux, Windows and Mac and then a second section with the rest.  I was going to try to look at that today.

mhdawson:
This is a first cut at what I had in mind:

downloadpage-updated

mhdawson:
@rvagg is what I have any better in respect to your concerns ?

rvagg:
Yeah, although I think we could do some better wording.

Some suggestions, mainly for someone implementing this on @nodejs/website:
- Remove "Primary Downloads", it doesn't really add much
- Change heading of second section to "Advanced download options" (can't think of anything better atm)
- Move the Python note far down to the bottom (using a footnote symbol?), or even remove it, it's a distraction
- The "Current version" bit could be improved, maybe at the top somewhere.

fhemberger:
Why not use the term "Additional Platforms" for ARM, SunOS, PPC, Docker and the downloads provided by IBM?

rvagg:
I could live with that

silverwind:
Maybe replace that Python notice with a "See below for additional platforms."?

fhemberger:
We could also put the unordered list below the "additional platforms" section. I think it even makes more sense that way.

How about this for the "Current version" issue?


We could get rid of the "download-header" class as well, it's no longer used.

silverwind:
@fhemberger +1 to that.

silverwind:
On a unrelated note, could someone update @nodejs-github-bot's avatar to use the same one as the org? The current one doesn't work too well on dark backgrounds:

fhemberger:
Quick mockup of combined changes

silverwind:
1) Get rid of the Python notice, build requirements are mentioned better in the README.md contained in the source package.
2) These are ranked by popularity, right? Doesn't ARM get more downloads than SunOS nowadays which could justify moving it above SunOS.

fhemberger:
@silverwind :+1:

silverwind:
Oh, and "Install Node.js via package manager" could be shortened to "Install via package manager".

mhdawson:
The point of the original change is that the ppc le binaries are now official and provided on the community download site.  So I'd want something like:
additionalplatofrms

phillipj:
> On a unrelated note, could someone update @nodejs-github-bot's avatar to use the same one as the org?

@silverwind it's updated now

stevemao:
Why do we need to list 3 IBM links that all link to the same page?

silverwind:
@phillipj thanks, looking good now!

silverwind:
@fhemberger any reason why you changed the LTS label in these screenshots? Imho, it's not ideal that the text now wraps to another line and the old text was fine imho.

jbergstroem:
What differentiates the ppc builds from for instance stuff we build on joyent or arm? I mean, what does 'official' imply here? For instance, we already know that calling the joyent builds `sunos` is incorrect since it won't really run on sunos. What we need is a sunos host for that to actually start working properly again. Would that denote its 'offical' standing?

We're deploying the servers, setting up the ci and building it through our release with shasum's generated by us. Might make sense to clarify this.

fhemberger:
@silverwind The screenshot was taken from the live website, text was changed in #570.

silverwind:
Ah, right. Well then let's make the button slightly wider to stop the wrapping.

mhdawson:
As I think @jbergstroem indicates, in my mind the "Unofficial provided by IBM" is for the cases were we don't have the builds in the community ci yet.   The original purpose of this PR was to reflect that we now build PPC le binaries as part of the regular releases and to have the page provide links to the binaries which are already available in "All downloads" directory.  For example: https://nodejs.org/dist/v4.4.0/

silverwind:
@fhemberger I fixed the download button width in https://github.com/nodejs/nodejs.org/commit/22bf803874aec29200e959d34b13403f8f1e2d01.

mhdawson:
@fhemberger  any comments on my proposed modification ?

fhemberger:
@mhdawson Sure, this makes sense. So go ahead. ;)

mhdawson:
@rvagg  you good with my suggestion as well ? If so we'll either update this PR to reflect that or submit a new one.

MylesBorins:
I have updated the PR to now split up the downloads as has been suggested. 

@nodejs/website thoughts?

edit: it is worth mentioning that I expanded the copy on the PPC release to keep the table from getting wonky

Each of these examples are with span=3 for the property on the table element.

Open to whatever change people may suggest

screen shot 2016-03-15 at 3 12 00 pm

screen shot 2016-03-15 at 3 21 02 pm

MylesBorins:
@fhemberger do you have any suggestions about the styles? Not getting a good lay out in the table due to one column having too much text

fhemberger:
@TheAlphaNerd Just a quick mock, how about this:



``` css
.download-matrix td span {
    font-size: 0.8em;
}
```

``` html
<tr>
    <th>Linux on Power Systems</th>
    <td colspan="3">
        <a href="#">64-bit le</a>
        <span>Official Node.js release</span>
    </td>
    <td colspan="3">
        <a href="#">64-bit be</a>
        <span>Unofficial, provided by IBM</span>
    </td>
</tr>
```

---

As you're already working on it, can you please also move the "current version" below the headline? (see https://github.com/nodejs/nodejs.org/pull/564#issuecomment-194825611)

``` css
/* style for .download-header h1 can be removed */
.download-header span {
    display: block;
    color: #999; /* "light-gray" in stylus */
    font-size: 0.8em;
}
```

``` html
<div class="download-header">
    <h1>Downloads <span>Current version v4.4.0</span></h1>
</div>
```

Then I think we're good to go.

MylesBorins:
@fhemberger Thanks for the detailed notes, that made things so much simpler!!!

Moving the version to the top not only makes sense, but it solves the problem with the fold!

I believe I've gotten everything right now except for remove the python notice at the bottom which was mentioned by @silverwind. I feel like that should perhaps have it's own discussion. 

Here is the latest version, please take a look at the style changes I made to make sure they are above board

screen shot 2016-03-16 at 4 54 23 pm

rvagg:
Could we put ARM above SmartOS? It gets more downloads and this is a primary place for some of those folks to go, particularly ARMv6 which isn't served by any other source afaik. SmartOS have their pkg ecosystem as the primary place to get builds already.

MylesBorins:
@rvagg not a problem updating . How would you feel about arm being included primary list?

screen shot 2016-03-16 at 6 44 50 pm

rvagg:
ARM just isn't in the same league as x (yet), we can reevaluate if that changes significantly

arm no competition

fhemberger:
@TheAlphaNerd There's still a wide margin between the "Downloads" headline and the version. Otherwise LGTM. Thank you for taking care of all those details … :+1: 

As soon as you push your changes, I think we're good to go.

mhdawson:
lgtm from me

MylesBorins:
@fhemberger What do you think of this?

screen shot 2016-03-17 at 9 17 53 am

fhemberger:
@TheAlphaNerd Perfect, go for it!

PR #565, opened 2016-03-09
Update versions

MylesBorins:
- path resolving now platform agnostic
- update versions
  - done using `scripts/load-versions`

### DIFF
--- a/scripts/load-versions.js
+++ b/scripts/load-versions.js
@@ -4,6 +4,7 @@
 
 const fs = require('fs')
 const nodeVersionData = require('node-version-data')
+const path = require('path')
 
 module.exports = nodeVersionData
 
@@ -16,7 +17,7 @@ if (require.main === module) {
     }
 
     fs.writeFileSync(
-      __dirname + '/../source/versions.json'
+      path.join(__dirname, '..', 'source', 'versions.json')
       , JSON.stringify(versions, null, 2)
     )
   })
--- a/source/versions.json
+++ b/source/versions.json
@@ -1,5 +1,287 @@
 [
   {
+    "version": "v5.7.1",
+    "date": "2016-03-02",
+    "files": [
+      "headers",
+      "linux-arm64",
+      "linux-armv6l",
+      "linux-armv7l",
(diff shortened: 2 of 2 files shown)


ghost:
[Travis build passed](https://travis-ci.org/nodejs/nodejs.org/builds/114750853) :+1:

williamkapke:
I was curious about this code- I can't tell how it is used now. There were uses in the code of `require('./versions.json')` but they have been removed. Now there just seems to be a link to the `makefile`... but I don't understand how **that** is used. If you're in this file- do you mind adding a comment indicating why/where `version.json` is used? (assuming you know) I would be super helpful for the next person.

MylesBorins:
@williamkapke I have no idea... I'll do a code audit later today and try and find out what's going on

phillipj:
In nodejs.org's early days, `versions.json` was used to read which versions of io.js and Node.js was available. Today we're using the [node-version-data](https://github.com/rvagg/node-version-data) module to retrieve this data from `https://{nodejs|iojs}.org/download/release/index.json`. In other words, we don't use `versions.json` anymore IIRC.

williamkapke:
@phillipj yes, I churned through the commits to discover that- however when @rvagg [switched it to his module](https://github.com/nodejs/nodejs.org/commit/372fd2e7cca993c17a78e57ee91ddc73c567cdca), there was an `if` statement put around the part that writes it to disk [when it is run as the target of the node process](https://github.com/nodejs/nodejs.org/blob/f9e1c8daef5346734c57057b3200948092f6bb45/scripts/load-versions.js#L10) only. Making me think there was a real use for when it is called from `make`. Also, his commit message was `load versions on build`

phillipj:
I've been wanting to make loading of versions offline compatible, meaning we could decide to actually use `versions.json` at startup for convenience sake with a special CLI parameter. When commuting I've noticed the pain of being **required** to be online.

That's the only real use of `versions.json` I know of today.

phillipj:
Merging this as whether or not too keep `versions.json` should be discussed in a separate issue.

PR #570, opened 2016-03-09
Rephrase mature and dependable to recommended for most users

max-mapper:
Based on real world feedback from many nodeschool attendees I believe the choice between 'LTS' and 'Stable' is too nuanced and difficult to reason about for many first-time node users.

I like the way Ubuntu does it:

screen shot 2016-03-09 at 2 22 07 pm

This PR changes the wording on LTS to say 'Recommended for most users':

screen shot 2016-03-09 at 2 23 20 pm

I think this will help people not aware of what 'LTS' stands for have a better experience getting started with node.

### DIFF
--- a/locale/en/download/index.md
+++ b/locale/en/download/index.md
@@ -7,7 +7,7 @@ downloads:
     lts: LTS
     stable: Stable
     tagline-stable: Latest Features
-    tagline-lts: Recommended for most users
+    tagline-lts: Recommended for Most Users
     display-hint: Display downloads for
     intro: >
         Download the Node.js source code or a pre-built installer for your platform, and start developing today.
--- a/locale/en/download/index.md
+++ b/locale/en/download/index.md
@@ -7,7 +7,7 @@ downloads:
     lts: LTS
     stable: Stable
     tagline-stable: Latest Features
-    tagline-lts: Mature and Dependable
+    tagline-lts: Recommended for most users
     display-hint: Display downloads for
     intro: >
         Download the Node.js source code or a pre-built installer for your platform, and start developing today.
--- a/locale/en/download/index.md
+++ b/locale/en/download/index.md
@@ -7,7 +7,7 @@ downloads:
     lts: LTS
     stable: Stable
     tagline-stable: Latest Features
-    tagline-lts: Recommended for Most Users
+    tagline-lts: Recommended For Most Users
     display-hint: Display downloads for
     intro: >
         Download the Node.js source code or a pre-built installer for your platform, and start developing today.
--- a/locale/en/download/stable.md
+++ b/locale/en/download/stable.md
@@ -7,7 +7,7 @@ downloads:
     lts: LTS
     stable: Stable
     tagline-stable: Latest Features
-    tagline-lts: Recommended for most users
+    tagline-lts: Recommended for Most Users
     display-hint: Display downloads for
     intro: >
         Download the Node.js source code or a pre-built installer for your platform, and start developing today.
--- a/locale/en/download/stable.md
+++ b/locale/en/download/stable.md
@@ -7,7 +7,7 @@ downloads:
     lts: LTS
     stable: Stable
     tagline-stable: Latest Features
-    tagline-lts: Mature and Dependable    
+    tagline-lts: Recommended for most users
     display-hint: Display downloads for
     intro: >
         Download the Node.js source code or a pre-built installer for your platform, and start developing today.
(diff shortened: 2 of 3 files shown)


ashleygwilliams:
i think this is an excellent compromise for now, given that we are still deciding is we want to change the name "stable". it definitely helps clarify a serious problems users have, which is understanding what Node they should download. LTS is definitely in-group slang, and most of our users (beginners!) don't know what it means. thanks @maxogden !

ghost:
[Travis build passed](https://travis-ci.org/nodejs/nodejs.org/builds/114918744) :+1:

MylesBorins:
Lgtm!

Trott:
One contributing factor (for me, anyway): It has always bothered me that if you want stability, you should _not_ choose the "Stable" version but should instead choose the LTS version.  If I had a time machine, I would go back and argue for "Latest" or something like that instead of "Stable". (That was probably considered and rejected, I imagine.)

Trott:
Oh, and: I like this change. LGTM

Trott:
Nit: Capitalization does not match between the two buttons. Should either change `Latest Features` to `Latest features` or else change `Recommended for most users` to `Recommended For Most Users`.

max-mapper:
@trott changed to Recommended for Most Users

max-mapper:
@Trott ah dang, I forgot to capitalize the F, just fixed

rvagg:
lgtm!

A job on my list is to improve the page you land on when you click the LTS schedule link, it consistently confuses people.

Trott:
@ashleygwilliams wrote:

> i think this is an excellent compromise for now, given that we are still deciding is we want to change the name "stable".

Oh, I didn't know that conversation was happening! That would also be a most welcome change (by me, anyway).  Although that's probably a semver major name change... :-P

mikeal:
should we bold "Recommended for most users" the way that Ubuntu does?

Trott:
@mikeal My opinion only, but bolding text in a call-to-action button may be funky, especially when there are two of those buttons right next to each other and only one has the bold text. I'd say let's start with this and see how it works and revisit typography changes only if this change is insufficient.

max-mapper:
@mikeal I don't think bold is necessary here since its the only thing on the button

julianduque:
This is a great and simple suggestion! LGTM

Fishrock123:
@maxogden So I'd like to address a question you posed in one of you're tweets, but the response is a bit too long for twitter, so I'll do it here:

https://twitter.com/denormalize/status/707689699251359745

> also for a thing you want feedback on, maybe 'Experimental' is a better name? Since LTS should convey stability

So, I think there is an underlying misunderstanding here that I'm not sure how to concisely convey for the website. 

The entire release setup, including all of our testing and pr / commit processes, are designed to let the "Stable" branch have a certain amount of stability. The goal is that it is usable in everyday usage. It is very much not experimental; features added are guaranteed supported for at least the life of the branch. Everything is thorough and intentional to ensure things do not break as best as possible.

The only place we'd ever add and remove exposed or flagged features without any significant support thoughts would be in and only in the master branch alone, and in nightlies / "canary" releases. Those would actually be / are "experimental".

Now, there's still an unavoidable a chance things will break in stable because computers and software and humans etc, but really, the thing we call "Stable" really is designed to be stable. All in all, I think it's pretty impressive what we have managed considering things virtually go from landing in master directly to release if they are not `semver-major`.

That being said, YMMV depending on your definition of stable; stable doesn't equate to lengthy support here in our internal context, at least.

Hopefully that helps a bit. :)

stevemao:
> i think this is an excellent compromise for now, given that we are still deciding is we want to change the name "stable".

+1 on that. It should really say "Latest". Can you link where the discussion is happening please? Thanks!

max-mapper:
@Fishrock123 got it, makes sense. My main concern is that if newbies use node v5 they are much more likely to hit errors in the ecosystem, things like "Module version mismatch" and thousands of other bugs in modules due to API changes. Even if the code in node v5 is stable, the ecosystem isn't. Obviously it's a good thing to encourage people to use the "next" version of node, but I don't think it's a good default. I think "Next" would be a lot more descriptive than "Stable" for example.

Also, this is kind of off topic for this PR, if you can link me and @stevemao I'd be happy to move the conversation there.

Fishrock123:
Right, the ecosystem thing is something I think we should consider more carefully and is essentially why I'm not objecting to this. :)

PR #575, opened 2016-03-14
Update guide: "Easy profiling for Node.js"

### LINKED ISSUE #447: Profiling Guide Requires Updates
In light of https://github.com/nodejs/node/pull/3032, there is some stuff in the [simple profiling guide](https://github.com/nodejs/nodejs.org/blob/030e1dd985654c1daa3d619d68b1b637116c7fb0/locale/en/docs/guides/simple-profiling.md) that can be simplified. Maybe @matthewloring can indicate what needs to change? It might just be the part about making sure to use the same version of tick-processor as came with the source and how to fire up the tick processor? Not sure anything else...

matthewloring:
Instructions on using the tick processor now use the flag on the Node.js
binary --prof-process for all versions and remove mention of processing
scripts (which are now removed).

Fixes #447

### DIFF
--- a/locale/en/docs/guides/simple-profiling.md
+++ b/locale/en/docs/guides/simple-profiling.md
@@ -21,7 +21,7 @@ code-creation,Stub,2,0x2d5000a33e00,507,"NumberToStringStub"
 ```
 
 In the past you need the V8 source code to be able to interpret the ticks.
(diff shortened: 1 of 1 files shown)


ghost:
LGTM!

Issue #578, opened 2016-03-15
Pardot integration (email subscription) needs help


williamkapke:
Pardot is used for the "subscribe to news about the Node.js Foundation" in these locations:
https://github.com/nodejs/nodejs.org/search?q=pardot

After filling out the form, it just disappears... and stays missing even after refresh. Not even an email confirmation.

After some digging I found:

> By default, a prospect who fills out a form and then returns to the form page within a 10 minute window will not see the form again and will instead see the Thank You Content (or be redirected based on your other settings). This option disables that 10 minute window.

... I'm guessing that the _Thank You Content_ is blank...?

See: [Control What A Prospect Sees After the Form Submits](http://help.pardot.com/customer/portal/articles/2132705-forms-overview#control-what-a-prospect-sees-after-the-form-submits)

It doesn't look like this can be fixed with a PR- I think it requires someone with access to the Pardot account? :(

stevemao:
Yup, and I don't think i have received any news.

fhemberger:
ping @mikeal

williamkapke:
To date, I still have not gotten any emails. 2 board meetings have been hosted online now- and I didn't know either of them were happening. I would think I an email would be sent about that :/

mikeal:
We haven't announced the meetings anywhere but twitter, not have we really used that email list much.

We're working on a new newsletter (kicking off in July when the new budget rolls around) that will replace that.

@hackygolucky is working on a PR to add a page about the board meetings to the website.

williamkapke:
Should we just remove this feature then?

mikeal:
This does sign you up for something, I'm just not sure that we've ever used it :/

gtewallace:
Hi all,

I'll look into the thank you field.

It adds you to our master email list, which we haven't used too much but we do use it.

Lately we've been sending emails to encourage people to submit talks for Node.js Interactive and also to announce early bird registration prices.

Earlier in the year we used it to ask people to take the Node.js User Survey.

Definitely more we can be doing with this channel, and would love suggestions.

Greg

williamkapke:
@gtewallace great thanks! FWIW, I searched my inbox and found just 1 email on May 4th about Node.js Interactive. Nothing about early bird prices or a survey though.

gtewallace:
survey email would have been in January, so maybe before you signed up?

some of our emails only go to certain geos - for instance the early bird pricing email is going only to EMEA contacts.

[...]

PR #583, opened 2016-03-16
css: newsletter-signup

LoganArnett:
Found this in relation to #57, the iframe for the newsletter signup has a large amount of empty whitespace underneath it with a default `height="500"` on both desktop and mobile. Just changed to `max-height` to avoid this.

### DIFF
--- a/locale/en/foundation/newsletter.md
+++ b/locale/en/foundation/newsletter.md
@@ -5,4 +5,4 @@ layout: foundation.hbs
 
 To subscribe to news about the Node.js Foundation please signup below.
 
-<iframe src="https://go.pardot.com/l/6342/2015-09-15/2sgqpp" width="100%" max-height="500" type="text/html" frameborder="0" allowTransparency="true" style="border: 0"></iframe>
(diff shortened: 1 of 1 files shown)


ghost:
[Travis build passed](https://travis-ci.org/nodejs/nodejs.org/builds/116511131) :+1:

williamkapke:
Is this related to #578?

LoganArnett:
@williamkapke I dont believe so, this is for when the form is present not after it has been filled out and disappears

lpinca:
@LoganArnett what you are doing here is changing the HTML `height` attribute to `max-height` which is not a valid HTML attribute.

With this change the iframe will display with its default height which is 150px.
This is not sufficient to display the full iframe contents on screens which are between 480px and 580px wide unless we are ok to have a vertical scroll bar.

Setting `max-height` in the `style` attribute also does not help.

That said, reducing the current value from 500 to 300 should be safe and partially fix the issue.

LoganArnett:
Yea I do not know why I changed it that way, yesterday was a long day haha I can update it to the 300px

lpinca:
LGTM.

ghost:
yup, LGTM from me as well

PR #596, opened 2016-03-23
let codes follow ES2015

hiroppy:
Hello :)
## change of the syntax
- `Arrow functions`
- `Template strings`
- `variable declaration`
## Edit range
- `build.js`
- `scripts/`
- `events/`
## Check nodejs versions
- [x] 4.2.2
- [x] 5.9.1

Cheers

### DIFF
--- a/build.js
+++ b/build.js
@@ -44,28 +44,17 @@ const markedOptions = {
 // in DEFAULT_LANG as a fallback to prevent any strings that aren't filled out
 // from appearing as blank.
 function i18nJSON (lang) {
-  var defaultJSON = require(`./locale/${DEFAULT_LANG}/site.json`)
-  var templateJSON = require(`./locale/${lang}/site.json`)
-  var finalJSON = JSON.parse(JSON.stringify(defaultJSON))
-  var merge = function (targetJSON, customJSON) {
-    Object.keys(customJSON).forEach(function (key) {
-      let value = customJSON[key]
-      if (typeof value === 'object') {
-        merge(targetJSON[key], value)
-      } else {
-        targetJSON[key] = value
-      }
-    })
-  }
-  merge(finalJSON, templateJSON)
-  return finalJSON
+  const defaultJSON = require(`./locale/${DEFAULT_LANG}/site.json`)
+  const templateJSON = require(`./locale/${lang}/site.json`)
+
+  return Object.assign({}, defaultJSON, templateJSON)
 }
 
 // This is the function where the actual magic happens. This contains the main
 // Metalsmith build cycle used for building a locale subsite, such as the
 // english one.
 function buildLocale (source, locale) {
-  console.time('[metalsmith] build/' + locale + ' finished')
+  console.time(`[metalsmith] build/${locale} finished`)
   const siteJSON = path.join(__dirname, 'locale', locale, 'site.json')
   const metalsmith = Metalsmith(__dirname)
   metalsmith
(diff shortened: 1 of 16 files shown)


ghost:
[Travis build passed](https://travis-ci.org/nodejs/nodejs.org/builds/118107855) :+1:

MylesBorins:
@abouthiroppy did you verify that all of the instances of const are for objects not being mutated? If you just did a regex or a find and replace there might be some bugs in there

hiroppy:
@TheAlphaNerd  I checked `npm run build`, `npm load-version`, `npm test` and `node *.js`. 
So, there might be a danger. Is `let` all right?

MylesBorins:
Yes and no. We recently found a bunch of problems in core with let being up to 3x slower in a for loop. That is generally only a problem in hot code.

We tend to be stricter on core than in this project though.  @fhemberger do you have any thoughts?

stevemao:
CC @nodejs/documentation

fhemberger:
Personally, I'm already using ES6 all over the place (including `let` and `const`). I'm not really working with time critical/hot code most of the time, so even if `let` is still slower at the moment, that's no deal breaker for me. Upcoming versions of V8 (and other engines as well) will improve on that particular case as well. Also regarding the build process of the website, I don't think there will be a measurable impact at all.

@phillipj may want to take a look as well, as many of the scripts were written by him.

ghost:
i do like this a lot. the only thing i'd like to see would be an information under the `// BUILD.JS: etc` comments that points out that this is, in fact, ES6 code.

fhemberger:
I don't see a need for that. ES6 is the current official standard of the JavaScript language.

ghost:
@fhemberger true

build.js
```diff
@@ -44,12 +44,12 @@ const markedOptions = {
...
-  var defaultJSON = require(`./locale/${DEFAULT_LANG}/site.json`)
-  var templateJSON = require(`./locale/${lang}/site.json`)
-  var finalJSON = JSON.parse(JSON.stringify(defaultJSON))
-  var merge = function (targetJSON, customJSON) {
-    Object.keys(customJSON).forEach(function (key) {
-      let value = customJSON[key]
+  const defaultJSON = require(`./locale/${DEFAULT_LANG}/site.json`)
+  const templateJSON = require(`./locale/${lang}/site.json`)
+  const finalJSON = JSON.parse(JSON.stringify(defaultJSON))
+  const merge = (targetJSON, customJSON) => {
```
lpinca: Is deep clone required here? If not we can use `Object.assign()`.

Edit: It doesn't seem to be required, the cloned object is only passed to `metalsmith.metadata` so I guess it's ok to use `Object.assign()`.

build.js
```diff
@@ -208,20 +208,20 @@ function buildLocale (source, locale) {
...
-    Object.keys(files).forEach(function (path) {
+    Object.keys(files).forEach((path) => {
       if (!isEditable.test(path)) {
         return
       }
 
-      var file = files[path]
-      var url = 'https://github.com/nodejs/nodejs.org/edit/master/locale/' + options.locale + '/' + path.replace('.html', '.md')
+      const file = files[path]
+      const url = 'https://github.com/nodejs/nodejs.org/edit/master/locale/' + options.locale + '/' + path.replace('.html', '.md')
```
lpinca: I would use template string here for consistency.

events/pull-nodeschool.js
```diff
@@ -28,11 +31,11 @@ request('http://nodeschool.io/chapters/list.json', {json: true}, function (e, re
   })
 
   function _geo () {
-    if (chapters.length === 0) return yml.save()
-    var chapter = chapters.shift()
-    geocoder.geocode(chapter.location, function (err, res) {
+    if (chapters.length === 0) { return yml.save() }
```
lpinca, line 34: Does linter complain about curly braces?

lpinca:
Regarding arrow functions, are round brackets required in our style when there is a single argument?

eljefedelrodeodeljefe:
The discssion popped up every now and then and I think concsensus was that they are required always, even with no argument. See [process](https://nodejs.org/api/process.html) for reference.

lpinca:
@eljefedelrodeodeljefe yeah I noticed that core always uses round brackets, there is also a linter rule for it. I guess we should do the same then.
AFAIK Standard (which is what we are using here) does not enforce it.

events/pull-nodeschool.js
hiroppy, line 34: No, but curly brace is written in other files.

lpinca, line 34: You are right, there isn't a consistent style for this, not even in this diff.
Forget the above comment.

lpinca:
LGTM.

hiroppy:
I also think we should attach a round bracket.

build.js
phillipj: I'd also like to see `Object.assign()` here. I'm okey with that change being a separate PR though.

phillipj:
LGTM

@fhemberger I completely agree with you about performance not being an issue for us.

build.js
hiroppy: Oh, I changed it here. Thanks:)

build.js
```diff
@@ -44,28 +44,18 @@ const markedOptions = {
...
-      }
-    })
-  }
-  merge(finalJSON, templateJSON)
-  return finalJSON
+  const defaultJSON = require(`./locale/${DEFAULT_LANG}/site.json`)
+  const templateJSON = require(`./locale/${lang}/site.json`)
+  const finalJSON = JSON.parse(JSON.stringify(defaultJSON))
+
+  return Object.assign(finalJSON, templateJSON)
```
lpinca: What about:

``` js
return Object.assign({}, defaultJSON, templateJSON)
```

so we can get rid of clone via stringify/parse.

lpinca:
:shipit: ?

hiroppy:
Thank you for many reviews and opinions😊

ghost:
[Travis build passed](https://travis-ci.org/nodejs/nodejs.org/builds/118288644) :+1:

build.js
phillipj: @abouthiroppy ^^ notice that @lpinca removed `finalJSON` completely

build.js
```diff
@@ -44,28 +44,18 @@ const markedOptions = {
...
-      } else {
-        targetJSON[key] = value
-      }
-    })
-  }
-  merge(finalJSON, templateJSON)
-  return finalJSON
+  const defaultJSON = require(`./locale/${DEFAULT_LANG}/site.json`)
+  const templateJSON = require(`./locale/${lang}/site.json`)
+  const finalJSON = JSON.parse(JSON.stringify(defaultJSON))
```
lpinca: As noted by @phillipj there is still one nit. Remove this line, and use `defaultJSON` in `Object.assign`:

``` js
return Object.assign({}, defaultJSON, templateJSON)
```

Thanks!

hiroppy: Yeah, definitely. We didn't need `JSON.parse(JSON.stringify(obj))` for deep copy.
I will modify it:)

Issue #593, opened 2016-03-22
Website WG Meetings


williamkapke:
From: https://github.com/nodejs/nodejs.org/blob/master/GOVERNANCE.md#wg-meetings

> The WG meets weekly on a Google Hangout On Air. The meeting is run by a designated moderator approved by the WG. Each meeting should be published to YouTube.

Which YouTube account has the videos?
Where can I view the past agendas/minutes?
Where do future meetings get announced?

MylesBorins:
@williamkapke afaik there has not been a WG meeting for the website team since I joined the org in october...

I bet that copy is old and should likely be updated

phillipj:
No WG meetings since I got involved in august either. Could be there were some meetings in the io.js periode, possibly something @fhemberger knows more about.

Both `README.md` and `GOVERANCE.md` really needs updating.

fhemberger:
I only participated once on an io.js call last year when I started working on the website. ;) So yes, this is old and should be updated. The GitHub flow worked so well for the website team so far, there simply wasn't the need to get together in a call. Also we have far more contributors working on the site compared to last year, so everything is just running smoothly. :+1:

justsml:
Good to know, I thought I was missing out.

mikeal:
This WG works really well without meetings, adding them would only end up excluding some people who have timezone difficulties.

fhemberger:
Closed via a1e54a8cc3f6b81fea8b3c45813a491696f7e384.

williamkapke:
This is a FAST group! :D 

...however I missed my chance to comment before the merge:
https://github.com/nodejs/nodejs.org/pull/598#issuecomment-200964250

I had just said:

> I think it would be better to say something that indicates that there are no regularly scheduled meetings, but the group may choose to schedule one at its discretion.
> 
> Having nothing might make someone unsure if they're just not looking the the right place.

Thoughts?

phillipj:
Just answered your comment in #598.

PR #598, opened 2016-03-24
Removed WG Meetings section from GOVERNANCE.md

phillipj:
The Website WG has done extremely well without the need for _any_ meetings so far. Having a big section about such meetings in our `GOVERNANCE.md` seems unnecessary. Should we go ahead and remove that section all together?

Refs #593

### DIFF
--- a/GOVERNANCE.md
+++ b/GOVERNANCE.md
@@ -91,36 +91,6 @@ the WG membership shares an employer, then the situation must be
 immediately remedied by the resignation or removal of one or more WG
 members affiliated with the over-represented employer(s).
 
-## WG Meetings
-
-The WG meets weekly on a Google Hangout On Air. The meeting is run by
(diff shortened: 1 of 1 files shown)


ghost:
[Travis build passed](https://travis-ci.org/nodejs/nodejs.org/builds/118279385) :+1:

fhemberger:
LGTM.

williamkapke:
I think it would be better to say something that indicates that there are no regularly scheduled meetings, but the group may choose to schedule one at its discretion.

Having nothing might make someone unsure if they're just not looking the the right place.

phillipj:
> Having nothing might make someone unsure if they're just not looking the the right place.

That argument might also be made for someone getting unsure when reading about WG meetings, but  can't seem to find any WG meeting minutes anywhere in the repo.

PR #603, opened 2016-03-25
Delete `background-repeat`

hiroppy:
Because background is disappear if the screen is large(more than 1850px)
## check browsers
- chrome
- chromium
- firefox
- safari
### safari(v8.0.4)

CSS of other parts are broken :cry: 

ps. Sorry, my Safari version is too low...:(
http://caniuse.com/#search=flex

### DIFF
--- a/layouts/index.hbs
+++ b/layouts/index.hbs
@@ -76,7 +76,6 @@
         width: 100%;
         height: 220px;
         background-image: url("/static/images/interactive/background.png");
-        background-repeat: repeat-y;
         text-align: center;
         display: flex;
         flex-wrap: nowrap;


ghost:
[Travis build passed](https://travis-ci.org/nodejs/nodejs.org/builds/118429780) :+1:

ghost:
cc @mikeal

lpinca:
I can confirm the issue. The fix works.
I've also noticed that on Firefox the images are not resized, see below:

screen shot 2016-03-25 at 15 41 19

mikeal:
the only reason it is there is for mobile, can you move the rule down to the mobile section rather than removing entirely.

hiroppy:
@mikeal  `background-repeat`'s default is `repeat-x` and `repeat-y`.
What is the problem?

hiroppy:
@lpinca I wasn't able to confirm it in My firefox(v47.0a2).

lpinca:
Maybe mine is too old (45.0.1 on OS X), don't know.

hiroppy:
Let’s put that on hold:)

mikeal:
I can reproduce the Firefox issue, it seems like flexbox isn't working how it's suppose to :(

PR #611, opened 2016-03-28
Fix a Flexbox issue in NodeJS Interactive images in Firefox. Closes #610.

### LINKED ISSUE #610: Mozilla Firefox 45.0.1 home page bug
screen shot 2016-03-28 at 16 09 00
screen shot 2016-03-28 at 16 08 56
I think there are come problems with the width for Firefox..

diagramatics:
As documented in #610, this fixes it. It's a flexbox issue documented in Philip Walton's flexbugs [here](https://github.com/philipwalton/flexbugs#1-minimum-content-sizing-of-flex-items-not-honored). Firefox got omitted from the list of   browsers with the issue in philipwalton/flexbugs#103.

I have to admit, I don't like how the CSS is written or injected to the document as an inline style, but this is what I can do for a quick fix.

### DIFF
--- a/layouts/index.hbs
+++ b/layouts/index.hbs
@@ -70,7 +70,7 @@
 
         </div>
     </div>
-    
+
     <style>
       div.interactive {
         width: 100%;
@@ -85,7 +85,15 @@
       div.interactive a, div.interactive span {
          border: 20px solid transparent;
       }
-      
+
+      #interactive-left,
+      #interactive-center,
+      #interactive-right {
+        /* Fix Flexbox inconsistency of items not shrinking to parent size.
+           https://github.com/philipwalton/flexbugs#1-minimum-content-sizing-of-flex-items-not-honored */
+        min-width: 0;
+      }
+
       @media (max-width: 480px) {
         #interactive-center {
           display:none;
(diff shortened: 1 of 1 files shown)


ghost:
[Travis build passed](https://travis-ci.org/nodejs/nodejs.org/builds/119016467) :+1:

lpinca:
Works fine, :+1:.

layouts/index.hbs
```diff
@@ -140,18 +148,18 @@
...
       }
     </style>
-    
+
     <div class="interactive">
       <span id="interactive-first">
         <img src="/static/images/interactive/nodejs-interactive-logo-center.png" />
       </span>
-      <a href="http://events.linuxfoundation.org/events/node-interactive-europe">
+      <a id="interactive-left" href="http://events.linuxfoundation.org/events/node-interactive-europe">
```
lpinca, line 156: Can't we simply use `style="min-width:0;"` or add the same CSS class for both anchor elements?

MylesBorins:
I landed this in dfec618

I was experiencing the same bug and testing the fix on my machine.

Moved forward on this quickly because we will potentially be getting a lot of traffic with a release of each version happening today

layouts/index.hbs
MylesBorins, line 156: Hey @lpinca I ended up merging this as is due to the release tonight. Do you think it would make sense to submit that optimization in another pr?

lpinca, line 156: Yes it's totally fine.

PR #622, opened 2016-04-01
Rename TC to CTC

williamkapke:
The TC was renamed to CTC in the [node](https://github.com/nodejs/node/) repo here:
https://github.com/nodejs/node/commit/18abb3ccc2e0917e1371ab9afae4852bdeaa298c#diff-9e8744ec73212c1966bb19fa22fc3e0a

..but the website still has a few refs to TC. This is a simple rename from TC to CTC and 1 fixed # link.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -125,7 +125,7 @@ CoC](https://github.com/rust-lang/rust/wiki/Note-development-policy#conduct).
 * Private harassment is also unacceptable. No matter who you are, if
   you feel you have been or are being harassed or made uncomfortable
   by a community member, please contact one of the channel ops or any
-  of the TC members immediately with a capture (log, photo, email) of
+  of the CTC members immediately with a capture (log, photo, email) of
   the harassment if possible.  Whether you're a regular contributor or
   a newcomer, we care about making this community a safe place for you
   and we've got your back.
--- a/locale/en/about/working-groups.md
+++ b/locale/en/about/working-groups.md
@@ -5,11 +5,11 @@ title: Working Groups
 # Working Groups
 
  Working Groups are autonomous projects created by the
-[Technical Committee (TC)](https://github.com/nodejs/node/blob/master/GOVERNANCE.md#technical-committee).
+[Core Technical Committee (CTC)](https://github.com/nodejs/node/blob/master/GOVERNANCE.md#core-technical-committee).
 
-Working Groups can be formed at any time but must be ratified by the TC.
+Working Groups can be formed at any time but must be ratified by the CTC.
 Once formed the work defined in the Working Group charter is the
-responsibility of the WG rather than the TC.
+responsibility of the WG rather than the CTC.
 
 It is important that Working Groups are not formed pre-maturely. Working
 Groups are not formed to *begin* a set of tasks but instead are formed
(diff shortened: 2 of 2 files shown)


ghost:
[Travis build passed](https://travis-ci.org/nodejs/nodejs.org/builds/119989531) :+1:

stevemao:
cc @nodejs/ctc @nodejs/documentation

CONTRIBUTING.md
```diff
@@ -125,7 +125,7 @@ CoC](https://github.com/rust-lang/rust/wiki/Note-development-policy#conduct).
 * Private harassment is also unacceptable. No matter who you are, if
   you feel you have been or are being harassed or made uncomfortable
   by a community member, please contact one of the channel ops or any
-  of the TC members immediately with a capture (log, photo, email) of
+  of the CTC members immediately with a capture (log, photo, email) of
```
Trott, line 128: I'm pretty sure this one should be TSC and not CTC.

Trott, line 128: It looks like the text was copied from the nodejs Code of Conduct, which has this as the TSC, not the CTC, so yeah, this one should almost certainly be TSC and not CTC. See third-from-last bullet point at https://github.com/nodejs/node/blob/5fc6938cff48f978642d7d83e257de36660977d3/CODE_OF_CONDUCT.md#conduct

Trott:
Other than my one CTC -> TSC comment, LGTM.

stevemao:
I don't know the reason of the name change and https://github.com/nodejs/TSC is not changed.

Trott:
@stevemao I'm going to get details wrong, I'm sure, but broadly speaking: In the beginning, there was the `TC`. At some point, it was decided that there needed to be a group that handled administrative/legal/policy stuff and a separate group to handle day-to-day technical decision-making for core. So the `TC` morphed into two groups: the `TSC` and the `CTC`. Because the `T` in both names stands for `Technical`, we are forever doomed to confusion about which one is which.

stevemao:
@Trott There are more places that's wrong about the name. I can see some link in  https://github.com/nodejs/TSC is broken. If you `@nodejs/ctc` it says "(previously the TSC)" and it sounds like "TSC" doesn't exist any more to me.
https://github.com/nodejs/node/blob/master/WORKING_GROUPS.md doesn't mention TSC.

williamkapke:
Actually, I now noticed:
https://github.com/nodejs/TSC/blob/master/WORKING_GROUPS.md

...which has them as TSC even though [they are all under the CTC](https://github.com/nodejs/TSC/issues/76#issuecomment-200573281)
except for Inclusivity.

Confused now.

williamkapke:
Ah, it would seem that the confusion is in the difference between a Top-Level Working Group (created by the TSC) and a Node.js Core Working Groups (created by the CTC).

https://github.com/nodejs/node/blob/master/WORKING_GROUPS.md
vs.
https://github.com/nodejs/TSC/blob/master/WORKING_GROUPS.md

@mikeal @rvagg I know the org is considering some shifts, but it will be a while before that solidifies and this page could be reworked for that. With this PR, I was just hoping to remove the old `TC` usages from the main website... but now I'm wondering if this bottom section (starting from [Starting a WG](https://nodejs.org/en/about/working-groups/#starting-a-wg) down) could/should be removed altogether? 

Besides, IMO, someone considering creating a WG will probably reference the github repos rather than the stuff at the end of this page!

williamkapke:
This was ultimately addressed by #623. Closing.

PR #623, opened 2016-04-01
Update Working Groups page.

williamkapke:
- Includes Top-Level Working Groups
- Pulls in many update/changes that were made to the CTC's [WORKING_GROUPS.md](https://github.com/nodejs/node/blob/master/WORKING_GROUPS.md)
- HTML comments referencing the mirror of the information outside of the website to help the changes of updates getting propagated to both.

The information was just a copy+paste from:
https://github.com/nodejs/TSC/blob/master/WORKING_GROUPS.md and 
https://github.com/nodejs/node/blob/master/WORKING_GROUPS.md

...and then some more detailed sections were deleted (i.e.: how to create a Working Group) since this is going in the `about` section of the main website and those levels of details are better left in GitHub.

The focus is to update this info only. If you see errors or want to make suggestions that exist in both copies, I would rather do that in different PRs on both repos to keep this PR simple.

This supersedes #622

### DIFF
--- a/locale/en/about/working-groups.md
+++ b/locale/en/about/working-groups.md
@@ -3,13 +3,51 @@ layout: about.hbs
 title: Working Groups
 ---
 # Working Groups
+There are 2 types of Working Groups:
+* [Top-Level Working Groups](#top-level-working-groups)
+* [Core Working Groups](#core-working-groups)
 
- Working Groups are autonomous projects created by the
-[Technical Committee (TC)](https://github.com/nodejs/node/blob/master/GOVERNANCE.md#technical-committee).
+## Top-Level Working Groups
+Top-Level Working Groups are created by the 
+[Technical Steering Committee (TSC)](https://github.com/nodejs/TSC#top-level-wgs-and-tlps).
 
-Working Groups can be formed at any time but must be ratified by the TC.
+Core Working Groups can be formed at any time but must be ratified by the TSC. 
+Once formed the work defined in the Working Group charter is the responsibility 
+of the WG rather than the TSC.
+
+### Current Top-Level Working Groups
+* [Inclusivity](#inclusivity)
+
+#### [Inclusivity](https://github.com/nodejs/inclusivity)
+The Inclusivity Working Group seeks to increase inclusivity and diversity for 
+the Node.js project:
+
+* Increasing inclusivity means making the Node.js project a safe and friendly 
+place for people from diverse backgrounds.
+* Increasing diversity means actively onboarding people from diverse backgrounds 
+to the Node.js project and maintaining their participation.
+
+Its responsibilites are:
+* Foster a welcoming environment that ensures participants are valued and can 
+feel confident contributing or joining discussions, regardless of any aspect of 
+their identity.
+* Proactively seek and propose concrete steps the project can take to increase 
+inclusivity.
+* Serve as a resource for the development and enforcement of workflows that 
+protect community members and projects from harassment and abuse.
+* Acknowledge and celebrate existing diversity accomplishments within the project 
+while seeking to build upon them.
+* Identify ways to measure diversity and inclusivity within the project and report 
+them at regular intervals.
+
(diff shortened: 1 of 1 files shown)


ghost:
[Travis build passed](https://travis-ci.org/nodejs/nodejs.org/builds/120006939) :+1:

locale/en/about/working-groups.md
```diff
@@ -157,27 +201,41 @@ Their responsibilities are:
...
-* Create Pull Requests for relevant changes to [Roadmap.md](https://github.com/nodejs/node/blob/master/ROADMAP.md)
+* Create Pull Requests for relevant changes to [Roadmap.md](./ROADMAP.md)
 
 
-### [Docker](https://github.com/nodejs/docker-node)
+### [Docker](https://github.com/nodejs/docker-iojs)
 
 The Docker working group's purpose is to build, maintain, and improve official
-Docker images.
+Docker images for the `Node.js` project.
```
ghost, line 235: why is Node.js formatted this way here, but not in, say, line 137?

ghost:
overall good job! :+1: should we link to the page where the process of creating a working group is explained now, or later when it gets actually created?

locale/en/about/working-groups.md
williamkapke, line 235: Good question...? That'll be something to take up with the CTC's version of the info.

ghost, line 235: alright :)

williamkapke:
My opinion is that the detail in the `about` section of the website should remain somewhat high level. Creating a WG is also somewhat confusing... are you creating a Top-Level WG or a Core WG? Chances are, if you're really thinking about creating one- you're probably beyond the info on the about page.

nebrius:
I think the section on the Inclusivity WG reads well, but I notice that there is quite a bit of information about that WG, while the others only have their name listed. Was this intentional?

If we are going to keep the other WG listed with just there name, I would recommend at least removing the "Responsibilities" section of the Inclusivity WG to make the section shorter, if not remove all of the info all together and just list the name like the others.

However, an argument could be made that we should add similar info for the other working groups too and normalize info that way :)

locale/en/about/working-groups.md
```diff
@@ -3,22 +3,47 @@ layout: about.hbs
...
+
+* Increasing inclusivity means making the Node.js project a safe and friendly 
+place for people from diverse backgrounds.
+* Increasing diversity means actively onboarding people from diverse backgrounds 
+to the Node.js project and maintaining their participation.
+
+Its responsibilites are:
+* Foster a welcoming environment that ensures participants are valued and can 
+feel confident contributing or joining discussions, regardless of any aspect of 
+their identity.
```
nebrius: I noticed the "including but not limited to" section was left off. The list is an important part of this statement. If we are going to list the responsibilities section then I think it should be listed in full, not summarized.

williamkapke:
@nebrius The info for Inclusivity was just taken from: https://github.com/nodejs/TSC/blob/master/WORKING_GROUPS.md

> I noticed the "including but not limited to" section was left off.

I don't see that in that document...? Help me find what you're looking at. I don't want to leave anything out!

> while the others only have their name listed
> ... I would recommend at least removing the "Responsibilities" section

All of the other groups have a Responsibilities section.

nebrius:
> I don't see that in that document...? Help me find what you're looking at. I don't want to leave anything out!

Iiiiinteresting. https://github.com/nodejs/inclusivity/#list-of-responsibilities <= that is the definitive document. I submitted https://github.com/nodejs/TSC/pull/86 to update the TSC copy.

> All of the other groups have a Responsibilities section.

Ugh, reading comprehension fail on my part, you're right.

williamkapke:
@nebrius I copied in your Inclusivity changes from https://github.com/nodejs/TSC/pull/86

nebrius:
LGTM, thanks for the updates :)

PR #614, opened 2016-03-30
Fix Spain world region

lpinca:
I was browsing the list of events in Africa when I noticed something strange:

screen shot 2016-03-30 at 17 31 23

Madrid, Barcelona, Sevilla, Valencia are in Africa? This is new...
I guess we should rebuild `events.md` but I don't have the required API keys.

PR #625, opened 2016-04-01
Update the list of global events

lpinca:
Follow up of https://github.com/nodejs/nodejs.org/pull/614.
This updates `events.md` to fix the Spain region issue.

I've used my own API key for meetup.com and no API key for node-geocoder (it seems to work fine).


ghost:
[Travis build passed](https://travis-ci.org/nodejs/nodejs.org/builds/120021492) :+1:

fhemberger:
1. lat/lng coordinates for the following events were deleted. Was this intentional?
   - Chennai NodeSchool
   - Jakarta NodeSchool
   - NodeSchool Kanpur
   - Tabriz NodeSchool
   - Tehran NodeSchool
   - NodeSchool Tokyo
   - Chernivtsi NodeSchool
   - Dublin NodeSchool
   - Faro NodeSchool
   - Nodeschool Lodz
   - NodeSchool Madrid
   - Lyon NodeSchool
   - NodeSchool Strasbourg
   - NodeSchool Stuttgart
   - NodeSchool Vienna
   - NodeSchool Cleveland
   - NodeSchool: LA Edition
   - NodeSchool Manchester
   - NodeSchool Philadelphia
   - NodeSchool Raleigh-Durham
   - NodeSchool DC
   - NodeSchool Vancouver
   - Western Massachusetts NodeSchool
2. We should adjust the event script:
   - Collapse new lines in description texts. One entry added **390 empty new lines**.
   - We don't show things like member count (which is also changing constantly), I vote for dropping at least this field to keep future diffs easier to handle.

lpinca:
It wasn't intentional and I noticed the empty lines issue.
The scripts need indeed some love.

lpinca:
I'll try to fix this later or tomorrow.

lpinca:
The new line issue seems to be a bug in `js-yaml`:

``` js
const data = yaml.safeLoad(contents);
const dump = yaml.safeDump(data); // => unwanted new lines.
```

See https://github.com/nodeca/js-yaml/issues/221, https://github.com/nodeca/js-yaml/issues/222. Not sure what to do about it. Those issues seem to be fixed in master and it looks like that a new version of `js-yaml` will be released soon but the new dumper uses a sligtly different syntax so we will have a lof of diff noise.

Maybe it makes sense to rebuild `events.md` from scratch.

fhemberger:
Hmm, in the meantime, how about something like `description.replace(/\n{3,}/g, '\\n\\n')`?

lpinca:
@fhemberger I tried that, not so useful, the problem is that the major part of the new lines are added when data is dumped with [`yaml.dump(store)`](https://github.com/nodejs/nodejs.org/blob/948cd36624172683a4be5302bd0d430979f40831/events/yaml-sync.js#L53).

We can probably work around that issue using the [`lineWidth`](https://github.com/nodeca/js-yaml#safedump-object---options-) option (e.g. setting it to something like 2000) but we will generate so much noise that it will be very hard to review the changes.

That's why I think it's probably better to regenerate `events.md` from scratch.

lpinca:
I've updated the pr to work around the `js-yaml` issue with spurious new lines.
Now a line width is not enforced so `yaml.dump()` works as expected.

There is a lot of diff noise but I tried to make the review process a little "easier" by splitting the changes over multiple commits.

If we wait for the next version of `js-yaml`, 52f4895 can be removed.

events/pull-meetup.js
```diff
@@ -47,9 +36,12 @@ function pull (opts) {
     body.results.forEach((result) => {
       const title = result.name.toLowerCase()
 
-      if (title.includes('nodeschool')) return
-      if (title.includes('mongodb') && title.includes('node')) return
-      if (title.includes('find a tech job') && title.includes('node')) return
+      if (
+        title.includes('find a tech job') ||
+        title.includes('nodeschool') ||
+        / mongodb (?:user group|meet ?up)$/.test(title) ||
```
fhemberger, line 42: Why not `title.includes('mongodb')`?

events/yaml-sync.js
```diff
@@ -5,31 +5,25 @@ const fs = require('fs')
...
-const lines = fs.readFileSync(p).toString().split('\n')
-const begin = lines.indexOf('---') + 1
-const end = lines.indexOf('---', begin)
-const store = yaml.safeLoad(lines.slice(begin, end).join('\n'))
-
-function getRegion (region) {
-  let reg
-  for (reg in store.regions) {
-    if (store.regions[reg].region === region) return store.regions[reg]
+const contents = fs.readFileSync(p, { encoding: 'utf8' }).trim().slice(4, -4)
```
fhemberger, line 12: Can you please add a comment about slicing? (YAML head, I know, but it's not that obvious.)

events/yaml-sync.js
```diff
@@ -5,31 +5,25 @@ const fs = require('fs')
...
+    store.regions.push(region)
   }
-  reg = { region: region }
-  store.regions.push(reg)
-  return reg
+
+  return region
 }
 
 function isSoT (meetups, city, name) {
```
fhemberger, line 34: Hmm, what's this `source_of_truth`? Can we find a better (function) name for that? Or at least a short comment?

events/pull-meetup.js
lpinca, line 42: Because there are some MongoDB+Node.js events.

events/yaml-sync.js
lpinca, line 12: Will do.

events/pull-meetup.js
fhemberger, line 42: Ah, ok. :+1:

events/yaml-sync.js
lpinca, line 34: I'll add a reference to #398.

events/pull-meetup.js
lpinca, line 42: There are also a lot of irrelevant events like Angular Ember etc but it's kinda hard to have ONLY relevant Node.js events. This a flaw in Meetup categories and topics.

fhemberger:
I haven't run the scripts myself, but the overall changes look good to me.

@lpinca f9746165, 650ddc77, 29ab04f9 and d4931de1 look like manual changes. Will they be overwritten the next time someone runs the script?

lpinca:
They will be updated yes, but not deleted. New events will be added.
The purpose of https://github.com/nodejs/nodejs.org/commit/29ab04f9ea348c334ab29a8c8558e1136f1b5fd2 is exactly to prevent it from being overwritten.

fhemberger:
Great, thank you!

lpinca:
@fhemberger I've added the requested comments, let me know if they should be changed/improved.

fhemberger:
@lpinca LGTM! :+1:

PR #627, opened 2016-04-01
Simplify CSS for Node.js interactive banner

lpinca:
This patch brings the following changes:
- Banner height no longer have a fixed value. This value is computed automatically.
- Removed `background-repeat: no-repeat;` (background image is 1852px wide).
- Fixed a bug that could make the Node.js logo disappear (when the screen was exactly 480px wide).
- Cleaned up fix (#611) for Flexbox bug.

Screenshots:

**320**

**375**

**425**

**768**

**1024**

**1440**

Tested on:
- Firefox
- Chrome
- Safari
- iPhone 5 (Safari)

### DIFF
--- a/layouts/index.hbs
+++ b/layouts/index.hbs
@@ -72,96 +72,53 @@
     </div>
 
     <style>
-      div.interactive {
-        width: 100%;
-        height: 220px;
+      #interactive-wrapper {
+        flex: none;
+      }
+
+      #interactive {
         background-image: url("/static/images/interactive/background.png");
-        text-align: center;
         display: flex;
-        flex-wrap: nowrap;
         align-items: center;
         justify-content: space-around;
       }
-      div.interactive a, div.interactive span {
-         border: 20px solid transparent;
-      }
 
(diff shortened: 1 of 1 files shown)


ghost:
[Travis build passed](https://travis-ci.org/nodejs/nodejs.org/builds/120074774) :+1:

hiroppy:
I checked Edge and IE11 on win10.

Edge is ok, but IE11 fails(the master branch is similar, too) :cry:

lpinca:
@abouthiroppy I've updated the pr to handle the issue with IE11. This is how it looks on my machine:

screen shot 2016-04-03 at 15 13 08

Can you confirm that is fixed?

P.S. There is also another issue in IE11 (in the footer), but it's not related to this patch. I'll fix it in another pr.

hiroppy:
@lpinca nice:+1: 
I confirmed this PR in IE11 and this problem was solved:)

fhemberger:
Thank you both! Merged.

PR #635, opened 2016-04-06
README: fixed git clone link

Knighton910:
The old one wasn't working.

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,7 +17,7 @@ This repo's issues section has become the primary home for the Website WG's coor
 Please contribute! There's plenty of [good first contributions](https://github.com/nodejs/nodejs.org/labels/good%20first%20contribution) to do.
 
 ```bash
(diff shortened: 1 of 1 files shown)


stevemao:
@lordKnighton Why is it not working? Did you setup your ssh properly?

fhemberger:
@stevemao Old link was also working fine.

@lordKnighton Please don't PR+merge in one step, give others the time to comment and give feedback as well.

stevemao:
Hm... I prefer to use ssh as I have two factor enabled. But it could just simply say "clone the repo" here.

ghost:
@stevemao i think it's better as setting up ssh is harder than just cloning it from https. it lowers the entry barrier by a bit

lpinca:
It's really subjective and I also prefer to use ssh but @hg has a point here.

Issue #636, opened 2016-04-06
Discuss about the design at `foundation/in-the-news/`


hiroppy:
I think that  it is better for this page to use a list.
This is because I think that this page is hard to watch on Smartphone if it is not the list.
### PC

<img width="1066" alt="2016-04-07 0 22 53" src="https://cloud.githubusercontent.com/assets/3367801/14322692/bc743404-fc58-11e5-8ff6-03faab85302d.png">
### Smartphone

![screenshot_20160407-001812](https://cloud.githubusercontent.com/assets/3367801/14322707/c5b7f91a-fc58-11e5-8e65-5ca47165b09a.png)
### PC

<img width="1088" alt="2016-04-06 22 37 16" src="https://cloud.githubusercontent.com/assets/3367801/14322725/db905db8-fc58-11e5-9439-bd538bd8035a.png">
### Smartphone

![screenshot_20160407-001834](https://cloud.githubusercontent.com/assets/3367801/14322750/e9a536bc-fc58-11e5-8e03-a9d58d499482.png)

How does everybody think?

lpinca:
I don't have a preference I'm fine with both.
Maybe it would be better to use a date as prefix instead of a bullet.
I mean something like the "Announcements" page.

fhemberger:
It would be great if both "Announcements" and "In the News" would use the same style (regardless of list bullets or not).

fhemberger:
Added a PR. @abouthiroppy Would you still favor the bullet point list variant?

hiroppy:
@fhemberger  Thank you so much! I think that's ok.

PR #638, opened 2016-04-07
Use same style for "In the News" and "Announcements"

fhemberger:
As mentioned in #636, "In the News" and "Announcements" should use the same listing style. I also added date parsing for the spreadsheet data.

Tested on Chrome, Firefox and Safari on OS X.

**Small width:**
small width

**Medium width:**
medium width

**Desktop width:**
desktop width

/cc @nodejs/website @abouthiroppy @lpinca

### DIFF
--- a/layouts/in-the-news.hbs
+++ b/layouts/in-the-news.hbs
@@ -45,13 +45,29 @@
                                  callback: showInfo,
                                  simpleSheet: true } )
 
+                function pad(str) { return ("0"+str).substr(-2) }
+
+                function parseDate(date) {
+                  try {
+                    var parsedDate = date.split('.')
+                    return new Date('20' + parsedDate[2], parsedDate[0] - 1, parsedDate[1])
+                  } catch(_){
+                    return date
+                  }
+                }
+
+                function formatDate(date) {
+                  return date == "Invalid Date" ? '' : date.getFullYear()+'-'+pad(date.getMonth()+1)+'-'+pad(date.getDate())
+                }
+
                 function showInfo(data, tabletop) {
                   data.reverse()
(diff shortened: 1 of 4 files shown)


hiroppy:
awesome!!! 

I checked IE11 and Edge. And they were ok :+1: 

I have one question. When there is no date , is this ok?

fhemberger:
@abouthiroppy It's just one entry and I hope that @mikeal knows the right person to poke to add the missing date there as well ;)

Also, I will add `time { width: 5.5em; }` to fix this.

lpinca:
LGTM.

layouts/in-the-news.hbs
```diff
@@ -43,14 +45,31 @@
                                  callback: showInfo,
                                  simpleSheet: true } )
 
+                function pad(str) { return ("0"+str).substr(-2) }
```
lpinca, line 48: No one can unpublish this right? :)

Edit: ok sorry left-pad jokes are no longer funny, useless spam.

fhemberger, line 48: Indeed, I was waiting for one to happen. ;)

lpinca:
Should we wait for that missing date to be fixed or can we ship this?

abouthiroppy:
@harshadsabne I felt that I should add `width`:)

fhemberger:
@lpinca I think this can go online even with the one missing date.

Merged.

layouts/in-the-news.hbs
stevemao, line 48: I own left-pad now I'll make sure it doesn't disappear again.

PR #649, opened 2016-04-12
Compress survey infographic

lpinca:
I've compressed the infographic with https://compressor.io/ (lossless mode). Size has been reduced by 61%.


lpinca:
Closing as the logo has been damaged and is very visible.

fhemberger:
I tried two different approaches:
1. ImageOptim (loseless) = 328k (-17,4%)
   [2016-survey-infographic](https://cloud.githubusercontent.com/assets/153481/14468514/3eee3b06-00e0-11e6-897c-eb1b3c087257.png)
2. ImgAlpha (lossy, 256 colors) + ImageOptim = 164k
   [2016-survey-infographic-dithered](https://cloud.githubusercontent.com/assets/153481/14468536/60ba579c-00e0-11e6-9048-0f1bd5ccd92c.png)

Logo is dithered in the second version, but the rest looks fine.

lpinca:
The difference between the second one and the original is noticeable but the first is perfect.

fhemberger:
Ok, I'll use the first one to replace the current graphic: d370b7af69f08cf21e42615e3c4d4c5c81118a91

mikeal:
So, we aren't actually displaying the graphic on any pages, it's just there for people to grab and use on their own, which is why I added it in full resolution. It's fine if we use a lossless compression but for this image I would stay away from anything that might degrade other people's ability to cut up the graphic, compress it themselves, etc.

lpinca:
@mikeal yeah I've only noticed that when I had already filed this pr (my bad for not checking where and how the image was actually used).
Anyway the image compressed by @fhemberger has no visible degradation so I think we should be fine.

PR #641, opened 2016-04-10
About: update example to use better practices

Fishrock123:
also for testing the latest https://github.com/nodejs-github-bot/github-bot

PR #656, opened 2016-04-17
Add back semicolons to the HTTP server example in about page

lpinca:
Reverts nodejs/nodejs.org#641

Not sure what are the better practices there, I guess the purpose was to test the bot.
Anyway I'm filing this pr because semicolons have been removed and this makes the example inconsistent with all the others (guides, API docs, etc).

cc: @Fishrock123

### DIFF
--- a/locale/en/about/index.md
+++ b/locale/en/about/index.md
@@ -11,20 +11,20 @@ connections can be handled concurrently. Upon each connection the callback is
 fired, but if there is no work to be done Node is sleeping.
 
 ```javascript
-const http = require('http')
+const http = require('http');
 
-const hostname = '127.0.0.1'
-const port = 3000
+const hostname = '127.0.0.1';
+const port = 3000;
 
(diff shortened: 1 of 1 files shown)


Fishrock123:
Feel free to re-add the semicolons but the rest is actually better practice, I.e. Not using `writeHead()` directly

lpinca:
Forgive my ignorance but what's wrong with `writeHead()`? It's a public api afaik.

Fishrock123:
Nothing is wrong with it, but chances are you are not going to be calling it directly from user code. :)

lpinca:
I'm still not sure why I should prefer

``` js
res.statusCode = 200;
res.setHeader('Content-Type', 'text/plain');
```

over

``` js
res.writeHead(200, { 'Content-Type': 'text/plain' });
```

but it doesn't matter. I've updated the pr to only re-add the semicolons.

fhemberger:
Cool, thanks!

PR #657, opened 2016-04-17
Cleanup contributor and collaborator guides

phillipj:
Noticed we've got some issues with our current .md guides:
- CONTRIBUTING.md has the contents of a collaborator guide, therefor it's renamed to COLLABORATOR_GUIDE.md
- Copied the baseline CONTRIBUTING.md from [TSC/BasePolicies](https://github.com/nodejs/TSC/tree/master/BasePolicies), changing TC -> WG in the last section, which seems more inline with what we've got here


lpinca:
Should we keep line length to 80 chars where possible? Apart from this LGTM.

fhemberger:
@lpinca Personally, I'm not entirely happy with this 80-chars-thingie in markdown files. It makes diffs harder to compare, when you have to hard-wrap a whole paragraph again, because you added/deleted only a couple of words. Editors can soft-wrap the content.

lpinca:
No problem, ~~I added that comment because that's how most of the markdown files are currently formatted.~~

Edit: Actually there isn't a consistent style, so ignore everything I wrote.

fhemberger:
I'm also not sure what "best practice" is in this case, but as rendered Markdown output is soft-wrapped as well and GitHub does a pretty good job with wrapping the content of diffs, I don't think we really need this 80 chars limit.

COLLABORATOR_GUIDE.md
```diff
@@ -0,0 +1,186 @@
+# Node.js Collaborator Guide
+
+**Contents**
+
+* [Issues and Pull Requests](#issues-and-pull-requests)
+* [Accepting Modifications](#accepting-modifications)
+ - [Involving the Website Admin Group](#involving-the-website-admin-group)
```
fhemberger, line 7: `*` for consistency, or it this a sub-item of "Accepting Modifications"?

CONTRIBUTING.md
```diff
@@ -1,186 +1,81 @@
...
-Collaborators may **close** any issue or pull request they believe is
-not relevant for the future of the Node.js project. Where this is
-unclear, the issue should be left open for several days to allow for
-additional discussion. Where this does not yield input from Node.js
-Collaborators or additional evidence that the issue has relevance, the
-issue may be closed. Remember that issues can always be re-opened if
-necessary.
+* A **Contributor** is any individual creating or commenting on an issue or pull request.
+* A **Committer** is a subset of contributors who have been given write access to the repository.
+* A **TC (Technical Committee)** is a group of committers representing the required technical
```
fhemberger: Should be changed to "Working Group" as well.

CONTRIBUTING.md
```diff
@@ -1,186 +1,81 @@
...
 
-Collaborators may **close** any issue or pull request they believe is
-not relevant for the future of the Node.js project. Where this is
-unclear, the issue should be left open for several days to allow for
-additional discussion. Where this does not yield input from Node.js
-Collaborators or additional evidence that the issue has relevance, the
-issue may be closed. Remember that issues can always be re-opened if
-necessary.
+* A **Contributor** is any individual creating or commenting on an issue or pull request.
+* A **Committer** is a subset of contributors who have been given write access to the repository.
```
fhemberger: Should be "Collaborator"

CONTRIBUTING.md
```diff
@@ -1,186 +1,81 @@
...
+or if it should land then it should be escalated to the TC. TC members should regularly
+discuss pending contributions in order to find a resolution. It is expected that only a 
+small minority of issues be brought to the TC for resolution and that discussion and 
+compromise among committers be the default resolution mechanism.
 
-- has a significant impact on the codebase,
-- is inherently controversial; or
-- has failed to reach consensus amongst the Collaborators who are
-  actively participating in the discussion.
+# Becoming a Committer
```
fhemberger: Collaborator (also in the following paragraphs)

CONTRIBUTING.md
phillipj: "Committer" was placed here deliberately when the base policy was created: https://github.com/nodejs/TSC/pull/40#discussion_r51919768

I'm currently somewhat confused about these terms though. I've raised an issue in the TSC-repo, where @Fishrock123 hinted we should try to use the term committer more: https://github.com/nodejs/TSC/issues/89#issuecomment-210475348

@fhemberger As mentioned I'm not really sure which of these to use. If you still feel "collaborator" is the better term for us, I'll happily change it.

COLLABORATOR_GUIDE.md
phillipj, line 7: > or it this a sub-item of "Accepting Modifications"?

yep

phillipj:
Thanks for reviewing! Replaced committer -> collaborator and some of TC -> WG.

PTAL

fhemberger:
LGTM.

PR #660, opened 2016-04-17
fix links to roadmap and docker-node

williamkapke:
- The links to the roadmap were broken.
- Pointed the Docker WG link to `docker-node` instead of `docker-iojs`.


fhemberger:
Good catch, thank you!

PR #661, opened 2016-04-18
Add a note about spread operator limitations

### LINKED ISSUE #653: Docs: Default es6 features should clarify type of Spread operator
Hi guys,

When I read through this list of features shipped by default:
https://nodejs.org/en/docs/es6/#which-features-ship-with-node-js-by-default-no-runtime-flag-required
... I thought it great that the Spread operator is now supported without any flags.

So I went and tried the spread operator on an object and was promptly met with an `Unexpected token ...` error, despite having node v5.10 installed.

I then had to be [educated by StackOverflow](http://stackoverflow.com/a/36666473/1688159) that it is only the _Array_ spread operator that is supported, but not the _Object_ spread operator.

I think this should be clarified on the page.

Many thanks.

lpinca:
This improves the existing note about the spread operator specifying that it only works with array literals and function calls on Node 5.x.

Closes #653.

### DIFF
--- a/locale/en/docs/es6.md
+++ b/locale/en/docs/es6.md
@@ -35,12 +35,13 @@ All ECMAScript 2015 (ES6) features are split into three groups for **shipping**,
 * [Arrow Functions](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Arrow_functions)
 * [new.target](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/new.target) [[2]](#ref-2)<span id="backref-2"></span>
 * [Object.assign](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/assign)
-* [Spread operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Spread_operator) [[2]](#ref-2)<span id="backref-2"></span>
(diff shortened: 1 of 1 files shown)


MylesBorins:
Should the language be explicit about node 6.x having greater support for the spear operator?

MylesBorins:
Actually... in thinking about it more the only way this change makes sense will be if we update the information regarding the spread operator at the end of the month when we bump to 6, as 5 will no longer be receiving support

lpinca:
Yeah, was thinking the same. I mean we can update it when node 6 will be released with something like:
"spread operator works on array literals, function calls and destructuring".

lpinca:
Or just wait for node 6 and use a sentence similar to the above one directly.

MylesBorins:
You could send this PR directly to 5.x and send the other to master.

lpinca:
Whoever reads site contents, reads master afaik.

MylesBorins:
the generated docs for v5 will be made from that branch's doc folder though

lpinca:
@TheAlphaNerd Are you suggesting to move this pr to node repo? Otherwise I don't get it.

MylesBorins:
@lpinca OH darn... this is what I get for responding via email when I am not fully awake yet.

I thought this was against the node repo. Sorry for the confusion

lpinca:
No problem :)

lpinca:
I've updated the pr, but it's probably better to close it.
The issue it tries to solve (#653) is probably invalid because the author ~~used the spread operator on a non-iterable object~~ tried to use object spread.
I don't see a real value in adding this note. This info should be (and actually is) in the operator documentation, not here.

fhemberger:
Ok, closing this now. Feel free to reopen this PR if required.

PR #666, opened 2016-04-19
Change "npm start" to "node index.js"

flowirtz:
I dont quite understand why you would use npm to call "node index.js", when you can do that directly.
Especially as it would keep the **number of processes low**, which is always interesting, even for a non-production environment (see [this document](https://github.com/nodejs/docker-node/blob/master/docs/BestPractices.md#cmd)).

If you see a major advantage in using `npm start` over `node server.js` then feel free to block this PR and discuss with me. I love to learn.

### DIFF
--- a/locale/en/docs/guides/nodejs-docker-webapp.md
+++ b/locale/en/docs/guides/nodejs-docker-webapp.md
@@ -31,10 +31,7 @@ create a `package.json` file that describes your app and its dependencies:
   "version": "1.0.0",
   "description": "Node.js on Docker",
   "author": "First Last <first.last@example.com>",
-  "main": "server.js",
-  "scripts": {
-    "start": "node server.js"
-  },
+  "main": "server.js"
   "dependencies": {
     "express": "^4.13.3"
   }
@@ -118,11 +115,10 @@ EXPOSE 8080
 ```
 
 Last but not least, define the command to run your app using `CMD` which defines
-your runtime. Here we will use the basic `npm start` which will run 
-`node server.js` to start your server:
+your runtime. Here we will use the basic `node server.js` to start your server:
 
 ```docker
-CMD [ "npm", "start" ]
+CMD [ "node", "server.js" ]
 ```
 
 Your `Dockerfile` should now look like this:
@@ -142,7 +138,7 @@ RUN npm install
 COPY . /usr/src/app
 
 EXPOSE 8080
-CMD [ "npm", "start" ]
+CMD [ "node", "server.js" ]
 ```
 
 ## Building your image


59:
i do like npm start for its standardized way to start a node.js program. when you clone a repo, you know you just have to run `npm start`, not some other, possibly more complicated program. heroku takes advantage of this, so that you don't even need a `Procfile` when you deploy a node app - it just runs `npm start`.

apart from that, i'm 👎  on this. let's keep it at what we have currently. if you want to, you can start it using `node index.js` - no problem with that :)

59:
also, even if it keeps the number of processes low, the website runs on a capable server afaik. i wouldn't imagine it to be a HUGE difference in memory usage.

flowirtz:
Also, the guide states that it is not intended for production, I cant quite relate though how this is ment: What would differ?

59:
@FWirtz nothing! i'm sure using `node index.js` would be better in some way. however, the current system as it is in production works pretty well, and we wouldn't want to break the build process by removing this line. what you _can_ do is open an issue in the [build](https://github.com/nodejs/build) repo asking if that command can be changed to start

59:
(actually i don't even know if it already uses `node index.js`, but i can't find my way through the build repo :P)

59:
anyways though, thanks for bringing this up! let's ask @nodejs/build how they do this

FWirtz:
@59 Well... Dont you want to remove the line **not for production** then from the guide?
I find that one particularly confusing in that context now...
Shall I open an issue and ask them what they do differently? You lost me there for a second, I am confused...

59:
@FWirtz i think we're talking about different things here, sorry about that! which guide are you referring to?

FWirtz:
@59 Not sure if we actually do, I probably just dont understand - this guide: https://nodejs.org/en/docs/guides/nodejs-docker-webapp/

ghost:
@FWirtz oh my god i think i understand now. i mistook your change to be in the `package.json` file of this repo and not in that guide, which is why i was sceptical about removing that. wow, what a gigantic mess this caused, hah

ghost:
let me reopen this then. can you delete your comments so we can make this look like this never happened _cough_

FWirtz:
Sure thing! :dancer:

flowirtz:
Well I am changing things in the `package.json` though! I am **removing the whole "script"** part, **would you rather leave that in there?**

ghost:
alright, in THIS case we should ask @nodejs/docker. they probably know best

flowirtz:
Clean up the comments before or after that? :smile:

ghost:
@FWirtz before this! also i was referring to https://github.com/nodejs/nodejs.org/blob/master/package.json - not the one in the guide

flowirtz:
Oh well... No, I am totally not messing with Node.js itself, just the package conf. in the guide! hahaha!
I am very sorry if my malformed description caused that trouble!
Thanks @59 for the nice chat!

ghost:
@FWirtz it's fine! thanks for bringing this up, again :P

flowirtz:
@59 Oh I thank **you**!

ghost:
@nodejs-github-bot reveal yourself!!!

phillipj:
Wooops, that last comment was mine. Made a mistake by answering from the bot email.

ghost:
but yeah, that came to me halfway in this discussion. another reason why i deleted my comments, to avoid spreading misinformation

phillipj:
Hah shoot, I tripped here as well. By just looking at the comments by mail, I thought this was related to the metalsmith setup we've got for nodejs.org. Ignore me and the bot 🙈

flowirtz:
Guys... You are so confusing! :relaxed:
@phillipj What if you cant build static files? Suppose having a REST api in Express.js, served via Node.js on Docker?
Oh, sorry, you said I was supposed to ignore this...

ghost:
@FWirtz we both made the exact same mistake 😢 we thought this was about the root-level `package.json` file in this repository

flowirtz:
**DISCLAIMER** THIS IS NOT ABOUT THE ROOT-LEVEL package.json!!!
@59 Thats probably why I dont understand very much of what you are writing here... :sweat_smile:

ghost:
i'm fairly sure this is because it's issue 666 :o

pesho:
`npm start` is **the** idiomatic way to run a Node app/package, and it's a good practice to teach in a guide. If you obtain some random Node app, you can be 99% certain that `npm start` will run it. Otherwise you'd have to research whether it needs `node index.js` or `node server.js` or `node --harmony main.js`, etc.

Replacing `npm start` with direct Node invocation is a valid micro-optimization, but probably not many apps really need that. So my preference would be to leave the guide as is.

pesho:
...also, the `-onbuild` Docker image relies on `npm start` being available.

evanlucas:
-1 from me. I think leaving it as `npm start` is better.

flowirtz:
@pesho @evanlucas  Oh okay, I see. I understand that.
Do you also have some information for me on how an actual production container/image would differ from the one described in the guide?

pesho:
@FWirtz a production image _might_ indeed differ by replacing `npm start` with direct Node invocation. This might make sense sometimes:
- if the image is running in a memory-constrained environment, e.g. in a Heroku dyno, or if we want to run many instances of the image.
- if we want to save some milliseconds from the start-up time

Apart from that, I don't think a production Dockerfile would differ much from the one in the guide. It's a good one.

flowirtz:
@pesho Oh ok, I understand. So even in a production environment, using `node xxx` directly only saves some ms over `npm start`, I see.
Well, but then thats really the part that got me **confused about the guide**: I also dont see much potential to improve the Dockerfile, but I am certainly not an expert.

I really think the line "_Not for production_" should then either
a) be **removed** or
b) linked to a new page that **describes a production environment** in greater detail.
Do you agree with me on that or would you see any difficulties?

fhemberger:
@FWirtz The "not for production" comment is there on purpose. The node app is still running as root in the Docker container, which can be a security issue. Details have been explained in the linked document 
"Node.js Docker Best Practices Guide", which are not in the scope of a basic "Getting started" guide.

flowirtz:
@fhemberger THANK YOU! Can you think of something else that would differ in _production_?

fhemberger:
@FWirtz I'm no Docker expert, both guides have been written by @nodejs/docker.

flowirtz:
@fhemberger Thanks anyways! I guess Ill open an issue on that one then and start a discussion.

Siilwyn:
@FWirtz one point missing in the above discussion is handling exit signals, normally you'd listen to signals and close your server, database, etc. If running through `npm start` this does not work. Just added this to the [best practices guide](https://github.com/nodejs/docker-node/blob/master/docs/BestPractices.md#cmd).

flowirtz:
Hah, I see! Thanks @Siilwyn for pointing that out - didn't know that!

PR #668, opened 2016-04-20
Remove unused variable in download script

lpinca:
The `version` variable (probably incorrectly added as global) seems to not be used on download page.


lpinca:
cc: @phillipj

phillipj:
LGTM, good catch :+1:

PR #674, opened 2016-04-21
Replace outdated ES6 page with node.green

### LINKED ISSUE #671: ES6 page should be ready to be updated when Node.js v6 drops.
Node.js v6.0.0 is in the rc phase. See https://github.com/nodejs/node/issues/5766 for status updates. This will include a jump to v8 5.0. That jump will greatly increase the amount of ES2015 features and we need to update the ES6 page (https://nodejs.org/en/docs/es6/).

fhemberger:
Fixes #671, requires https://github.com/nodejs/build/pull/394 to land as well.

### DIFF
--- a/locale/en/docs/es6.md
+++ b/locale/en/docs/es6.md
@@ -1,8 +1,8 @@
 ---
-title: ECMAScript 2015 (ES6)
+title: ECMAScript 2015 (ES6) and beyond
 layout: docs.hbs
 ---
-# ECMAScript 2015 (ES6) in Node.js
+# ECMAScript 2015 (ES6) and beyond
 
 Node.js is built against modern versions of [V8](https://developers.google.com/v8/). By keeping up-to-date with the latest releases of this engine, we ensure new features from the [JavaScript ECMA-262 specification](http://www.ecma-international.org/publications/standards/Ecma-262.htm) are brought to Node.js developers in a timely manner, as well as continued performance and stability improvements.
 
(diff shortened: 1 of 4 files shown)


locale/en/site.json
```diff
@@ -77,8 +77,8 @@
         "link": "docs",
         "text": "Docs",
         "es6": {
-            "link": "docs/es6",
-            "text": "ES6 in Node.js"
+            "link": "http://node.green",
+            "text": "ES6 and beyond"
```
evanlucas, line 81: should we go ahead and update this to ES2015 while we're at it?

locale/it/site.json
```diff
@@ -68,8 +68,8 @@
         "link": "docs",
         "text": "Documentazione",
         "es6": {
-            "link": "docs/es6",
-            "text": "ES6 in Node.js"
+            "link": "http://node.green",
+            "text": "ES6 e oltre"
```
lpinca, line 72: Good job :)

fhemberger, line 72: Google Translate FTW! 😄

locale/en/site.json
stevemao, line 81: ES6 is also a valid name. ES2015 is a bit long and the tab doesn't look as good.

benjamingr:
-1, the point of that page is also to be friendly.

williamkapke:
I have to agree with @benjamingr - there needs to be a page that explains the flags. _**Maybe**_ I could work something in to node.green for that- but I would want the node community's input on how that should look.

stevemao:
Hm... most people +1ed here as writing our own es6 would cause a lot of maintaining issues. I wanted to write something myself but I feel whatever I write, it wont solve the problem and I might end up writing a es6 tutorial. node.green is something I always wanted to see. I believe a little bit description on node.green would be helpful.

fhemberger:
Okay, I have an idea: Let's keep the current ES6 page to explain the flags, etc. and just replace the paragraph "Which features ship with Node.js by default (no runtime flag required)?" with a link to node.green – so @williamkapke doesn't have to come up with a way to introduce all the texts to his site.

stevemao:
Wouldn't it be easier if the text is on node.green? One less clicking is good :)

fhemberger:
Half the people want to get rid of the ES6 page, the other half doesn't … just figure something out, I'll adjust the PR then. 😉

phillipj:
> Okay, I have an idea: Let's keep the current ES6 page to explain the flags, etc. and just replace the paragraph "Which features ship with Node.js by default (no runtime flag required)?" with a link to node.green

+1

benjamingr:
I'm also in favor of what @fhemberger suggested: keep the page, replace the part with "which features" with node.green. I say we move ahead with that suggestion.

fhemberger:
Okay, I updated the PR: We keep the ES6 page and I added a reference to node.green.

Please take a look and let me know what you think.

locale/en/docs/es6.md
```diff
@@ -1,54 +1,21 @@
...
-
-## Which features are behind the --es_staging flag?
-
-* [`Symbol.toStringTag`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol) (user-definable results for `Object.prototype.toString`, behind flag `--harmony_tostring`)
-* [`Array.prototype.includes`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/includes) (determine whether
-an array includes a certain element, behind flag `--harmony_array_includes`) [[2]](#ref-2)<span id="backref-2"></span>
-* [Rest Parameters](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/rest_parameters) (represent an indefinite number
-of arguments as an array, behind flag `--harmony_rest_parameters`) [[2]](#ref-2)<span id="backref-2"></span>
+The website [node.green](http://node.green) provides an excellent overview over supported ECMAScript features in various versions of Node.js, based on kangax's compat-table:
+<br><a href="http://node.green" class="imagelink">![Screenshot of the node.green website](/static/images/node-green-screenshot.png)</a>
```
lpinca: Do we need a screenshot?

phillipj: @lpinca looks like it's already part og this PR

phillipj:
LGTM

locale/en/docs/es6.md
lpinca: Yeah I mean is it necessary?

phillipj: Ahh sorry. I assume you mean it's unnecessary and I agree. I don't think we need to "prepare" the end user for what he/she 's about to see. Another downside is we would have to keep the screenshot updated to whatever design changes node.green gets in the future.

Fishrock123:
Definitely -1 on a static screenshot alone.

MylesBorins:
I'm frame?

fhemberger:
Ok, removed the screenshot again, keeping only the link.

williamkapke:
I guess there could be more clarification on the difference of the `--es_staging` vs `--harmony` flags. From the existing page, I concluded they were the aliases of each other. I guess that isn't true:
See: https://github.com/williamkapke/node-compat-table/issues/1#issuecomment-214118438

fhemberger:
@williamkapke I opened up a separate issue for this in #694. With v6 released, we _need_ the updated ES6 page.

Issue #377, opened 2015-11-24
Investigate issues with greenkeeper bot


lpinca:
@phillipj has already created this issue https://github.com/greenkeeperio/greenkeeper/issues/68.

If anyone has a clue on why the bot is not working, please share.

This is the list of dependencies to update:

```
Package              Current  Wanted  Latest  Location
chokidar               1.2.0   1.2.0   1.3.0  nodejs.org
handlebars             4.0.4   4.0.4   4.0.5  nodejs.org
metalsmith-metadata    0.0.2   0.0.2   0.0.3  nodejs.org
nock                  2.18.2  2.18.2   3.3.0  nodejs.org
node-version-data      1.0.0   1.0.0   1.0.1  nodejs.org
semver                 5.0.3   5.0.3   5.1.0  nodejs.org
```

phillipj:
Just noticed there's an error on our side of the greenkeeper webhook. It says it got a 400 status last time it tried to contact greenkeeper. I'll try to sync and/or re-enable greenkeeper as soon as I've got the time.

christophwitzko:
Hi @lpinca,
Hi @phillipj,

renaming the repository from `new.nodejs.org` to `nodejs.org` might caused the problem.
You can fix that issue by running `greenkeeper sync`.

Have a great day,
Christoph

phillipj:
Thanks for answering @christophwitzko!

Just tried `$ greenkeeper sync` but it didn't have any effect as far as I can see, the webhook link in this repo's settings are still in a bad state. Also tried `$ greenkeeper enable` which ended in this message:

``` bash
→ greenkeeper enable
🌴  info enable The GitHub slug is: nodejs/nodejs.org
🌴  ERR! enable Couldn’t enable a project with this slug.
🌴  ERR! enable If you want to try your free private repo make sure to grant the necessary rights by running $ greenkeeper login --force --private
🌴  ERR! enable You have to have a plan for more than one private repo. To verify run $ greenkeeper whoami
🌴  ERR! enable If you have just recently created this repo try running $ greenkeeper sync
🌴  ERR! enable You need admin access to a repo to enable it.
🌴  ERR! enable If you think this error really shouldn’t appear let us look into it with $ greenkeeper support
```

Do you know if greenkeeper requires private access to the nodejs github organization? If so we're in bad luck, because we dont allow that kind of permission to any third-party app.

christophwitzko:
Hey @phillipj,

[...]

PR #675, opened 2016-04-21
Update all dependencies to their latest versions

paazmaya:
According to https://david-dm.org/nodejs/nodejs.org there is also security issue with `marked` which has not been fixed.
However all dependencies are updated to their latest versions via this PR


fhemberger:
Have you checked for any breaking changes, especially in `semver-major` updates?

package.json
```diff
@@ -31,42 +31,42 @@
...
     "changelog-url": "1.0.2",
-    "cheerio": "0.19.0",
-    "chokidar": "1.2.0",
-    "handlebars": "4.0.4",
-    "html-to-text": "^1.5.0",
-    "js-yaml": "^3.4.5",
+    "cheerio": "0.20.0",
+    "chokidar": "1.4.3",
+    "handlebars": "4.0.5",
+    "html-to-text": "^2.1.0",
```
lpinca, line 39: Is this safe? What is the breaking change here?

package.json
```diff
@@ -31,42 +31,42 @@
...
+    "metalsmith-layouts": "1.6.4",
     "metalsmith-markdown": "0.2.1",
-    "metalsmith-metadata": "0.0.2",
-    "metalsmith-permalinks": "0.4.0",
-    "metalsmith-prism": "2.1.1",
-    "metalsmith-stylus": "1.0.0",
+    "metalsmith-metadata": "0.0.4",
+    "metalsmith-permalinks": "0.5.0",
+    "metalsmith-prism": "2.2.0",
+    "metalsmith-stylus": "2.0.0",
```
lpinca, line 52: What are the breaking changes?

package.json
```diff
@@ -31,42 +31,42 @@
...
+    "st": "1.1.0",
     "strftime": "0.9.2"
   },
   "devDependencies": {
     "faucet": "^0.0.1",
-    "nock": "^2.17.0",
+    "nock": "^8.0.0",
     "pre-commit": "^1.1.2",
-    "proxyquire": "^1.7.3",
+    "proxyquire": "^1.7.4",
```
lpinca, line 68: Unnecessary change.

package.json
```diff
@@ -31,42 +31,42 @@
...
     "standard": "^6.0.8",
-    "tape": "^4.2.2"
+    "tape": "^4.5.1"
```
lpinca, line 70: Ditto.

package.json
```diff
@@ -31,42 +31,42 @@
...
     "metalsmith-yearly-pagination": "2.0.0",
     "ncp": "2.0.0",
-    "node-geocoder": "^3.4.1",
-    "node-version-data": "1.0.0",
-    "octonode": "0.7.4",
-    "request": "^2.67.0",
+    "node-geocoder": "^3.9.1",
+    "node-version-data": "1.0.1",
+    "octonode": "0.7.5",
+    "request": "^2.72.0",
```
lpinca, line 58: Not needed, caret versioning covers this.

package.json
```diff
@@ -31,42 +31,42 @@
...
+    "metalsmith-permalinks": "0.5.0",
+    "metalsmith-prism": "2.2.0",
+    "metalsmith-stylus": "2.0.0",
     "metalsmith-yearly-pagination": "2.0.0",
     "ncp": "2.0.0",
-    "node-geocoder": "^3.4.1",
-    "node-version-data": "1.0.0",
-    "octonode": "0.7.4",
-    "request": "^2.67.0",
+    "node-geocoder": "^3.9.1",
```
lpinca, line 55: Unnecessay change.

lpinca:
This is good, I also wanted to do this, but it's risky, we need to be sure that we are not breaking things.

package.json
```diff
@@ -31,42 +31,42 @@
...
-    "cheerio": "0.19.0",
-    "chokidar": "1.2.0",
-    "handlebars": "4.0.4",
-    "html-to-text": "^1.5.0",
-    "js-yaml": "^3.4.5",
+    "cheerio": "0.20.0",
+    "chokidar": "1.4.3",
+    "handlebars": "4.0.5",
+    "html-to-text": "^2.1.0",
+    "js-yaml": "^3.6.0",
```
lpinca, line 40: Same thing, not necessary.

package.json
stevemao, line 39: We should pin this too

package.json
stevemao, line 40: we should pin all deps and rely on future greenkeeper.io

lpinca, line 40: @stevemao https://github.com/nodejs/nodejs.org/issues/377

stevemao:
I suggest we only bump minor versions here and leave major version bump in other PRs.

package.json
```diff
@@ -31,42 +31,42 @@
...
   "dependencies": {
-    "autoprefixer-stylus": "0.8.1",
+    "autoprefixer-stylus": "0.9.2",
     "changelog-url": "1.0.2",
-    "cheerio": "0.19.0",
-    "chokidar": "1.2.0",
-    "handlebars": "4.0.4",
-    "html-to-text": "^1.5.0",
-    "js-yaml": "^3.4.5",
+    "cheerio": "0.20.0",
```
stevemao, line 36: Note for v0, a minor version bump contains breaking changes.

package.json
stevemao, line 40: I know, that's why I said "future" :)

stevemao, line 40: If we aren't going with greenkeeper.io, I suggest to use `~` for all deps.

lpinca:
I looked at the semver major updates and everything seems safe:
- `cheerio@0.20.0` → https://github.com/cheeriojs/cheerio/blob/master/History.md#0200--2016-02-01
- `html-to-text@2.1.0` → https://github.com/werk85/node-html-to-text/blob/master/CHANGELOG.md#version-200
- `metalsmith-stylus@2.0.0` → https://github.com/esundahl/metalsmith-stylus/pull/15

If there are no objections I would merge this and file a new pr where all deps (except dev dependencies) are pinned down or use tilde versioning.

ghost:
yeah, looks safe to me as well. no breaking changes that affect our usage

paazmaya:
I did not check anything other than running `npm test` and checking that site got served as it was before.

PR #677, opened 2016-04-21
[typo] correct the error on knowledge document

detailyang:
correct some errors on knowledge document:) :grin:

### DIFF
--- a/locale/en/knowledge/intermediate/how-to-log.md
+++ b/locale/en/knowledge/intermediate/how-to-log.md
@@ -65,11 +65,11 @@ Here is an example of setting up a `winston` logger.  This example includes most
           new winston.transports.File({ filename: 'path/to/all-logs.log' }),
           new winston.transports.Couchdb({ 'host': 'localhost', 'db': 'logs' }),
           new winston.transports.Riak({ bucket: 'logs' }),
-          new winston.transports.MongoDB({ db: 'db', level: 'info'}),
+          new winston.transports.MongoDB({ db: 'db', level: 'info'})
         ],
         exceptionHandlers: [
-          new winston.transports.File({ filename: 'path/to/exceptions.log' }),
-        ],
+          new winston.transports.File({ filename: 'path/to/exceptions.log' })
+        ]
       });
 
 Here, we have instantiated a new `winston` logger, and provided a number of logging transports.  Winston has built-in support for configurable logging levels, and provides alias methods for each configured logging level.  For example, `winston.warn(x)` is an alias for `winston.log('warn', x)`.  Thus, the following:
@@ -87,6 +87,6 @@ Because of the file transport we set up, winston also logged the warning to 'som
 
 Note that winston's file logger formats the logs differently for file logging (JSON in this case) than it does for the console transport.
 
-Winston also supports logging to Riak, CouchDB, MongoDB and even [Loggly](http://loggly.com).  The `logger.warn` call we used before also put the same message into each database, according to the options we gave to each transport.
+Winston also supports logging to Riak, CouchDB, MongoDB (need install winston-riak、winston-mongo and winston-couchdb to support Riak、CouchDB、MongoDB) and even [Loggly](http://loggly.com).  The `logger.warn` call we used before also put the same message into each database, according to the options we gave to each transport.
 
 For further information, please see the [thorough documentation for Winston.](https://github.com/indexzero/winston).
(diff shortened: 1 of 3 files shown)


locale/en/knowledge/intermediate/how-to-log.md
```diff
@@ -57,17 +57,19 @@ Here is an example of setting up a `winston` logger.  This example includes most
...
-          new winston.transports.File({ filename: 'path/to/all-logs.log' })
-          new winston.transports.Couchdb({ 'host': 'localhost', 'db': 'logs' })
-          new winston.transports.Riak({ bucket: 'logs' })
-          new winston.transports.MongoDB({ db: 'db', level: 'info'})
-        ]
+          new winston.transports.Console(),
+          new winston.transports.File({ filename: 'path/to/all-logs.log' }),
+          new winston.transports.Couchdb({ 'host': 'localhost', 'db': 'logs' }),
+          new winston.transports.Riak({ bucket: 'logs' }),
+          new winston.transports.MongoDB({ db: 'db', level: 'info'}),
```
lpinca: I would not add trailing commas here and below.

locale/en/knowledge/intermediate/how-to-log.md
```diff
@@ -57,17 +57,19 @@ Here is an example of setting up a `winston` logger.  This example includes most
 
       require('winston-riak').Riak;
       require('winston-mongo').Mongo;
+      require('winston-couchdb').Couchdb;
```
fhemberger, line 60: Not sure we need the CouchDB adapter in this example, I'd rather mention it in the paragraph at the end (line 90).

lpinca, line 60: @fhemberger I guess it's becuase it is used in the array?

locale/en/knowledge/intermediate/how-to-log.md
detailyang: why? here should be array.

locale/en/knowledge/intermediate/how-to-log.md
detailyang, line 60: As of winston@0.6.0 the CouchDB transport has been broken out into a new module: winston-couchdb. So winston dont have `winston.transports.Couchdb` default

detailyang, line 60: should both mongo and riak adapter be mentioned? @fhemberger

locale/en/knowledge/intermediate/how-to-log.md
lpinca: I don't like arrays written like this `[ 1, 2, 3, ]` but if other reviewers are ok with trailing commas I'll live with it.

Edit: The reason is that they caused issues in older versions of IE and now they trigger my OCD.

detailyang: @lpinca  haha, i know what you meaning:) it's up to us. :grin:

detailyang:
ping @fhemberger check this now, thx:)

locale/en/knowledge/intermediate/how-to-log.md
fhemberger, line 60: @lpinca No, I meant do we need to list every available transport module in a logging example? I think it's enough to show how to add multiple transports and mention the widely used ones at the end.

locale/en/knowledge/intermediate/how-to-log.md
```diff
@@ -85,6 +87,6 @@ Because of the file transport we set up, winston also logged the warning to 'som
 
 Note that winston's file logger formats the logs differently for file logging (JSON in this case) than it does for the console transport.
 
-Winston also supports logging to Riak, CouchDB, MongoDB and even [Loggly](http://loggly.com).  The `logger.warn` call we used before also put the same message into each database, according to the options we gave to each transport.
+Winston also supports logging to Riak, CouchDB, MongoDB (need install winston-riak、winston-mongo and winston-couchdb to support Riak、CouchDB、MongoDB) and even [Loggly](http://loggly.com).  The `logger.warn` call we used before also put the same message into each database, according to the options we gave to each transport.
```
fhemberger: I'd shorten it to something like: 

> Winston also supports logging to Riak, CouchDB, MongoDB and [many other transports](https://github.com/winstonjs/winston/blob/master/docs/transports.md). The `logger.warn` …

As there are a lot more than the ones already showed above, and with the link everybody can find out if their specific logging scenario is supported.

detailyang:
how about this :grin:,  ping @fhemberger

fhemberger:
LGTM, thank you! Let's get this merged. @lpinca what do you think?

lpinca:
LGTM.

PR #699, opened 2016-04-27
fix handlebars and stylus indentation

ghost:
corrects .hbs and .styl files to have 2 spaces of indentation, along with the rest of the site

### DIFF
--- a/.htmllintrc
+++ b/.htmllintrc
@@ -1,0 +1,12 @@
+{
+  "attr-bans": [],
+  "attr-name-style": false,
+  "attr-no-dup": false,
+  "attr-quote-style": false,
+  "attr-req-value": false,
+  "id-class-style": "dash",
+  "img-req-alt": false,
+  "indent-style": "spaces",
+  "indent-width": 2,
+  "tag-bans": []
+}
--- a/package.json
+++ b/package.json
@@ -9,7 +9,7 @@
     "load-versions": "node scripts/load-versions.js",
     "start": "npm run serve",
     "test": "npm run test:lint && npm run test:unit && npm run test:smoke",
-    "test:lint": "standard && htmllint **/*.hbs",
+    "test:lint": "standard && htmllint **/*.hbs && stylint layouts/css",
     "test:unit": "tape tests/**/*.test.js | faucet",
     "test:smoke": "tape tests/*.smoketest.js | faucet"
   },
@@ -68,6 +68,7 @@
     "pre-commit": "^1.1.2",
     "proxyquire": "^1.7.4",
     "standard": "^6.0.8",
+    "stylint": "^1.3.8",
     "tape": "^4.5.1"
   }
 }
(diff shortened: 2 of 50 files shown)


hiroppy:
LGTM :+1:

phillipj:
@tentakel could we enforce correct indentation going forward? Fixing all of this isn't all that valuable if changes could be merged w/wrong indentation, and I don't think we should spend time doing indentation nitpicking in PRs manually.

ghost:
@phillipj sure, i'd more than welcome that. should i add that to the `CONTRIBUTING.md` or make a new file for that?

fhemberger:
Well, personally, I'm not a big fan of 2 space indentation. Things use to get sloppy over time, which can lead to closing tags missing in HTML and so on.

But if everybody is okay with it, I'm fine.

@tentakel You rearranged some code in a few files. What's the reason for that? Would be easier to review the PR if it was just the indentation in one commit and the other changes in a separate one.

Please also change the `.editorconfig` to reflect the new indentation: https://github.com/nodejs/nodejs.org/blob/master/.editorconfig#L7

@phillipj `.editorconfig` is one place, not sure if we should/can add a linter checking all files for indentation.

ghost:
@fhemberger i did? 😨 all i remember is fixing one `<div>` that was supposed to be a closing tag. sorry, but could you point out an example?

phillipj:
@tentakel was thinking more in terms of _enforcing_ as part of the `npm test` script as we do with standard for .js files. Maybe give [htmllint-cli](https://www.npmjs.com/package/htmllint-cli) a shot?

fhemberger:
@tentakel Uh, never mind. The GitHub diff got a little confused because of all the indentation change.

ghost:
@phillipj that sounds like a good idea. i'll look into how to do that

ghost:
alright, htmllint is running together with standard. i've made sure to keep consistency with the standard patterns, too

package.json
```diff
@@ -63,6 +63,7 @@
   },
   "devDependencies": {
     "faucet": "^0.0.1",
+    "htmllint-cli": "0.0.4",
```
phillipj: Use `^` as the rest of our devDeps?

ghost: will do!

phillipj:
@tentakel nice! Care to try https://www.npmjs.com/package/stylint#indentpref--default-false-number-or-false- while you're at it to get .styl indentation enforced as well?

ghost:
@phillipj yep!

ghost:
stylint is added and runs smoothly

.htmllintrc
```diff
@@ -0,0 +1,12 @@
+{
```
stevemao, line 1: I wish there were similar projects as `standard` so that we don't need to create two extra metadata files.

phillipj:
LGTM 👍

Issue #694, opened 2016-04-27
ES6: Clarify difference between --es_staging vs --harmony flags


fhemberger:
Quoting @williamkapke https://github.com/nodejs/nodejs.org/pull/674#issuecomment-214133370

> I guess there could be more clarification on the difference of the `--es_staging` vs `--harmony` flags. From the existing page, I concluded they were the aliases of each other. I guess that isn't true:
> See: https://github.com/williamkapke/node-compat-table/issues/1#issuecomment-214118438

I've never used those flags myself, can someone explain the difference?
/cc @nodejs/collaborators

bnoordhuis:
In v4 and v5, they're synonyms for beta-level features: considered complete but not battle-tested yet; can still have bugs.

In v6, `--es_staging` is `--harmony` plus a few alpha-level features; currently `--harmony_regexp_lookbehind` and `--harmony_tailcalls` but that can change at any time.

ChALkeR:
Note that `--es-staging` is labeled as «for internal use only» and exposes more unstable features, while the change that got landed promotes `--es-staging` over `--harmony`.

```
  --es_staging (enable test-worthy harmony features (for internal use only))
  --harmony (enable all completed harmony features)
```

fhemberger:
@ChALkeR Good catch, the "for internal use" remark was not present in Node.js 4.3.1. So we should rather document the `--harmony` flag only and leave out `--es_staging`?

ChALkeR:
@fhemberger In Node.js 4.3.1 those two were equivalent.

> So we should rather document the `--harmony` flag only and leave out `--es_staging`?

/cc @nodejs/documentation

fhemberger:
@ChALkeR Yes, but in Node.js 6.0 they are not and `--es_staging` is flagged "for internal use". But in all versions `--harmony` does the same, so we should use this flag in our documentation and omit the internal flag `--es_staging`.

ChALkeR:
> But in all versions `--harmony` does the same,

… except for 0.10/0.12.

lpinca:
> so we should use this flag in our documentation and omit the internal flag `--es_staging`.

I agree.

fhemberger:
Fixed on 139393e.

PR #705, opened 2016-04-30
Document how harmony flags relate to v8

yoshuawuyts:
`es_` / `harmony` flags are deprecated by v8 without notice, which
potentially can cause errors to be thrown. Users should be informed
about this behavior.
- Original issue:
  https://github.com/nodejs/node/issues/6489

### DIFF
--- a/locale/en/docs/es6.md
+++ b/locale/en/docs/es6.md
@@ -10,7 +10,7 @@ All ECMAScript 2015 (ES6) features are split into three groups for **shipping**,
 
 * All **shipping** features, which V8 considers stable, are turned **on by default on Node.js** and do **NOT** require any kind of runtime flag.
 * **Staged** features, which are almost-completed features that are not considered stable by the V8 team, require a runtime flag: `--es_staging` (or its synonym, `--harmony`).
-* **In progress** features can be activated individually by their respective harmony flag, although this is highly discouraged unless for testing purposes.
+* **In progress** features can be activated individually by their respective harmony flag, although this is highly discouraged unless for testing purposes. Note: these flags are exposed by V8 and will potentially change without any deprecation notice.
 
 ## Which features ship with which Node.js version by default?
 


locale/en/docs/es6.md
```diff
@@ -9,7 +9,7 @@ Node.js is built against modern versions of [V8](https://developers.google.com/v
 All ECMAScript 2015 (ES6) features are split into three groups for **shipping**, **staged**, and **in progress** features:
 
 * All **shipping** features, which V8 considers stable, are turned **on by default on Node.js** and do **NOT** require any kind of runtime flag.
-* **Staged** features, which are almost-completed features that are not considered stable by the V8 team, require a runtime flag: `--es_staging` (or its synonym, `--harmony`).
+* **Staged** features, which are almost-completed features that are not considered stable by the V8 team, require a runtime flag: `--es_staging` (or its synonym, `--harmony`). Note: these flags are exposed by v8 and will potentially change without any deprecation notice.
```
MylesBorins: I think you want this down in **in progress** I believe this is for the generic `--harmony` flag

MylesBorins:
The copy looks good but I believe it is on the wrong entry

yoshuawuyts:
Whoops, updated!

MylesBorins:
LGTM

locale/en/docs/es6.md
```diff
@@ -10,7 +10,7 @@ All ECMAScript 2015 (ES6) features are split into three groups for **shipping**,
 
 * All **shipping** features, which V8 considers stable, are turned **on by default on Node.js** and do **NOT** require any kind of runtime flag.
 * **Staged** features, which are almost-completed features that are not considered stable by the V8 team, require a runtime flag: `--es_staging` (or its synonym, `--harmony`).
-* **In progress** features can be activated individually by their respective harmony flag, although this is highly discouraged unless for testing purposes.
+* **In progress** features can be activated individually by their respective harmony flag, although this is highly discouraged unless for testing purposes. Note: these flags are exposed by v8 and will potentially change without any deprecation notice.
```
thefourtheye: Shouldn't that be flags?

thefourtheye: V8 instead of v8

locale/en/docs/es6.md
```diff
@@ -10,7 +10,7 @@ All ECMAScript 2015 (ES6) features are split into three groups for **shipping**,
 
 * All **shipping** features, which V8 considers stable, are turned **on by default on Node.js** and do **NOT** require any kind of runtime flag.
 * **Staged** features, which are almost-completed features that are not considered stable by the V8 team, require a runtime flag: `--es_staging` (or its synonym, `--harmony`).
```
fhemberger, line 12: `--es_staging` and `--harmony` are not fully identical: https://github.com/nodejs/nodejs.org/issues/694

locale/en/docs/es6.md
yoshuawuyts: @thefourtheye what do you mean by "flags"? Updated to `V8`

locale/en/docs/es6.md
yoshuawuyts, line 12: Changing this wording seems past the scope of this PR. Or is the change I made not reflective of the reality? If so, what wording would you propose?

locale/en/docs/es6.md
thefourtheye: "by their respective harmony flag" or "by their respective harmony flags"?

locale/en/docs/es6.md
fhemberger, line 12: Well, `--es_staging` and `--harmony` are no synonyms, as is was the initial code name for ES6. And `--es_staging` already covers first ES7 features as well. So I think we should drop the `--harmony`-flag altogether.

locale/en/docs/es6.md
yoshuawuyts: I think singular is correct; a single feature is not toggled by multiple flags

Knighton910:
In general LGTM

locale/en/docs/es6.md
yoshuawuyts, line 12: I think there are valid reasons why features might be toggled individually - it's also part of the Node API, so I reckon we should document it unless it's actually removed from the API

yoshuawuyts:
Looks like it's got two rubber stamps for a one line change, can this be merged? Thanks!

stevemao:
I'm happy to merge this and leave the inline comments in a separate PR.

yoshuawuyts:
Thanks! :sparkles:

PR #707, opened 2016-05-01
Add GitHub issue template

fhemberger:
As general questions about Node.js and issues regarding the API docs often land here instead, I created a GitHub issue template based on the template in `nodejs/node`.

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -1,0 +1,22 @@
+<!--
+Thanks for wanting to report an issue you've found on the nodejs.org website.
+
+For issues regarding the Node.js API docs, please head over to:
+https://github.com/nodejs/node/issues/ (prefix your issue name with "doc")
+
+For general questions about Node.js:
+https://github.com/nodejs/help/issues/
+
+Please fill in the template below by replacing the HTML comments with an
+appropriate answer. If unsure about something, just do as best as you're able.
+If you are reporting a visual glitch, it will be much easier for us to fix it
+when you attach a screenshot as well.
+
+Thank you!
+-->
+
+* **URL**:
+* **Browser version**:
+* **Operating system**:
+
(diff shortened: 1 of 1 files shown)


hiroppy:
LGTM 👍

lpinca:
LGTM.

lpinca:
Maybe add something like "Steps to reproduce the issue"?

fhemberger:
@lpinca Added field "URL" instead. I guess there aren't many cases where there are multiple steps necessary to re-create an issue.

lpinca:
Yeah it makes sense.

mikeal:
I'm a little troubled by this template. Do we have a problem with bad issues that we're trying to solve? I can't recall a ton of erroneous issues we've had to deal with and I don't know if it's worth putting up a barrier to entry to solve a concern we don't have.

ghost:
@mikeal if it helps users to find the right repo to open an issue in, therefore reducing unnecessary noise, i'd say it's justified

fhemberger:
I don't see it as a barrier: You still can totally ignore the template content and post your stuff anyway. I took the general wording from node core.

We had some tickets in the past that were doc issues or general questions unrelated to the website. That's why I added links to both repositories, to help people find the right spot for their questions (which IMHO is better than closing tickets with "wrong repo" right away).

Affected URL and browser also helps us to reproduce issues – again, this is totally optional.

mikeal:
> You still can totally ignore the template content and post your stuff anyway.

_I_ can certainly do that (and do often) but new users, especially people new to contributing to Node.js i any way, will read all of this and try to tailor their comment to it and if they can't fit it into the box of the template they will probably just not log an issue at all.

At the very least I think that we can cut this way down.

fhemberger:
The main idea was to point people to the right place, if it's not about the website and they followed the link by accident (e.g instead of reporting an issue with Node instead).

If you have an idea for a better, more approachable template text, please go ahead! This was just a first draft.

PR #711, opened 2016-05-02
Point to Node.js v6 instead of v5 on package manager page

yodeyer:


detailyang:
LGTM, ping others :beers:

fhemberger:
LGTM as well, merging. Thank you!

PR #713, opened 2016-05-03
Add gzip task for production, fixes #712

### LINKED ISSUE #712: Store gzip representations of common files on deploy
Since we're using `gzip_static on` in the server we could shave off an considerable amount of cpu power by storing a gzip representation of each file we deploy (for instance `index.html` would be accompanied by `index.html.gz`). Additionally, since we default to `-6` in nginx, we'd likely make each file a bit smaller by using `-9` on deploy.

fhemberger:
Ref: #712
/cc @jbergstroem

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,6 +6,7 @@
   "scripts": {
     "build": "node build.js",
     "serve": "node server.js",
+    "gzip": "find -E build -type f -regex '.*(html|css|js|xml|json)$' -exec gzip -kf9 {} \\;",
     "load-versions": "node scripts/load-versions.js",
     "start": "npm run serve",
     "test": "npm run test:lint && npm run test:unit && npm run test:smoke",


jbergstroem:
That was quick! LGTM (haven't tested but flags and formats looks good to me); here are the content types we (would) gzip, so it should match well: `gzip_types text/plain text/css application/x-javascript text/xml application/xml application/xml+rss text/javascript;`

fhemberger:
@jbergstroem You don't gzip `text/html`?

jbergstroem:
@fhemberger from nginx manual: "Responses with the “text/html” type are always compressed."

fhemberger:
@jbergstroem Ah, ok. Merging this now.

jbergstroem:
Cool. Does this mean that it'll be shipped with next version of docs/website?

fhemberger:
I guess all that's left to do is replacing the line here (https://github.com/nodejs/build/blob/master/setup/www/resources/scripts/build-site.sh#L28):

``` diff
-  build_cmd="npm run build"
+  build_cmd="npm run build; npm run gzip"
```

jbergstroem:
Ok. Want to PR?

fhemberger:
@jbergstroem https://github.com/nodejs/build/pull/405

PR #714, opened 2016-05-03
ES6: Highlight difference between '--es_staging' and '--harmony' flag

fhemberger:
Both flags are not exact synonyms, as '--harmony' includes only staged ES6
features, while '--es_staging' also covers first features from ES7.

Refs:
- #705
- #694


lpinca:
LGTM, merging.

ChALkeR:
Note that `--es-staging` is labeled as «for internal use only» and exposes more unstable features, while this change promotes `--es-staging` over `--harmony`.

```
  --es_staging (enable test-worthy harmony features (for internal use only))
  --harmony (enable all completed harmony features)
```

PR #719, opened 2016-05-05
Remove unused CSS rules

lpinca:
It seems that these rules are no longer used. I think they have been replaced by the rules in `_linuxfoundation.styl`.

Please correct me if I'm wrong.


hiroppy:
LGTM

PR #735, opened 2016-05-12
Show shipped npm version on download page, fixes #731

### LINKED ISSUE #731: download pages need clear indicators about npm versions
As it stands, this page https://nodejs.org/en/download/current/ doesn't make it very clear at all which version of npm is included with a version of node. It's also pretty easy to be mistaken and think that you can only use the included npm with a given node release. Needs better messaging here.

fhemberger:
Exposes versions of node, npm, v8 and openssl for templates.


lpinca:
LGTM.

ghost:
LGTM

rvagg:
nice work

PR #736, opened 2016-05-12
Use Object.assign instead of util._extend

lpinca:
`util._extend` has been deprecated. This patch replaces it with `Object.assign`.


hiroppy:
LGTM

ghost:
LGTM!

PR #740, opened 2016-05-13
Fix  Korean for nodejs-ko

outsideris:
This was changed in https://github.com/nodejs/nodejs.org/commit/41fe898c4e6a24dd0620e12519504d4545e7dda9 . I don't know why it was changed in the commit. But we use it "한국어" not "조선말". So, I revert it.(related with https://github.com/nodejs/nodejs-ko/issues/344 ) @nodejs-ko


lpinca:
👍

stevemao:
@nodejs/nodejs-ko

taggon:
👍

PR #744, opened 2016-05-16
add attributes for accessibility

jrit:
This fixes a couple links and images that were missing any descriptive text or alt attributes.


fhemberger:
Good catch, thanks a lot!

PR #745, opened 2016-05-16
improve accessibility for header anchors

jrit:
Anchors shouldn't be without a label, so this uses aria-labelledby to associate the existing header text with the anchor. Also revised the text on the home page after further review of the resulting content calculation.

Is it correct that the API docs anchors would need to be updated separately on the /node repo?

### DIFF
--- a/layouts/index.hbs
+++ b/layouts/index.hbs
@@ -109,19 +109,17 @@
       <span id="logo-first" class="interactive-item">
         <img src="/static/images/interactive/nodejs-interactive-logo-center.png" />
       </span>
-      <a class="interactive-item" href="http://events.linuxfoundation.org/events/node-interactive-europe"
-        title="Node.js Interactive Europe">
+      <a class="interactive-item" href="http://events.linuxfoundation.org/events/node-interactive-europe">
         <img src="/static/images/interactive/nodejs-interactive-hero-banner-left-eu.png"
-          alt="Amsterdam, Netherlands, September 15-16, 2016"/>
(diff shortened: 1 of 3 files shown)


Fishrock123:
> Is it correct that the API docs anchors would need to be updated separately on the /node repo?

Yeah. let us know if you need any help with that. :)

layouts/index.hbs
```diff
@@ -109,19 +109,17 @@
       <span id="logo-first" class="interactive-item">
         <img src="/static/images/interactive/nodejs-interactive-logo-center.png" />
       </span>
-      <a class="interactive-item" href="http://events.linuxfoundation.org/events/node-interactive-europe"
-        title="Node.js Interactive Europe">
+      <a class="interactive-item" href="http://events.linuxfoundation.org/events/node-interactive-europe">
         <img src="/static/images/interactive/nodejs-interactive-hero-banner-left-eu.png"
-          alt="Amsterdam, Netherlands, September 15-16, 2016"/>
+          alt="Node.js Interactive Europe, Amsterdam, Netherlands, September 15-16, 2016"/>
```
Fishrock123, line 114: These aren't on anchors now though?

Fishrock123:
Seems fine to me.

layouts/index.hbs
jrit, line 114: I looked at it more closely in dev tools and having the alt tag on the image counted as the content of the anchor, so that overrode the title attribute completely, so the title never would have been read how it was.

ghost:
yeah, LGTM

PR #749, opened 2016-05-17
template: modify ppcbe to serve official tarball

MylesBorins:
We are now building ppcbe in production so we should be serving
that from the website

_note_: do not land this until a new version of v4.x is cut. The next
release will include ppcbe, and we can avoid a bunch of nasty custom
code in the meantime.

/cc @mhdawson


fhemberger:
LGTM.

mhdawson:
LGTM

PR #754, opened 2016-05-20
Add missing fields to the Italian site.json

lpinca:
This adds some missing fields to the Italian `site.json`.

cc: @nodejs/nodejs-it.


fhemberger:
Mostly rearranging fields, LGTM. Merging.

lpinca:
Yeah, "build notturne" may sound a little weird but the alternative is to use "nigtly builds".

PR #756, opened 2016-05-21
Update YLD logo

### LINKED ISSUE #755: swap current YLD logo with new one attached
- **URL**:
- **Browser version**:
- **Operating system**:

Hi all,

Node.js Foundation member YLD has gone through a rebrand and asks that we swap out their [old logo](https://nodejs.org/static/images/foundation/yldLogo2.png) on https://nodejs.org/en/foundation/members/ with the attached.

Many thanks!
[BLACK YLD logo (1).zip](https://github.com/nodejs/nodejs.org/files/274950/BLACK.YLD.logo.1.zip)

lpinca:
This updates the YLD logo as requested in #755.
The image has been compressed with ImageOptim.
Closes #755.


rvagg:
@nodejs/website I propose that we impose a policy on these logos that we only have one file per company, regardless of how it changes over time. Unfortunately now we now have https://nodejs.org/static/images/foundation/yldLogo.png, https://nodejs.org/static/images/foundation/yldLogo2.png and https://nodejs.org/static/images/foundation/yldLogo3.png and as we found when we migrated from the old Joyent site, we can't get rid of any of these static assets because of their deep linking all over the internet.

PR #761, opened 2016-05-27
Added Austin MEAN Stack to North American events list

cdnadmin:
Hey, y'all, simply wanted to add the [Austin MEAN Stack Dev meetup](http://www.meetup.com/Austin-MEAN-Stack-Development/) in the North American list of meetups.

PR #762, opened 2016-05-27
Added MK to map object; re-generated events.md

cdnadmin:
Added MK (Republic of Macedonia) to map object, per @lpinca 's  [nudge](https://github.com/nodejs/nodejs.org/pull/761#issuecomment-222180020) in   #761  :-) 

Went ahead an included re-generated events.md while I was at it.

### DIFF
--- a/events/country-map.js
+++ b/events/country-map.js
@@ -75,5 +75,6 @@ module.exports = {
   VE: 'Latin America',
   AU: 'South Pacific',
   NZ: 'South Pacific',
-  SG: 'South Pacific'
+  SG: 'South Pacific',
+  MK: 'Republic of Macedonia'
 }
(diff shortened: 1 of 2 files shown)


events/country-map.js
```diff
@@ -75,5 +75,6 @@ module.exports = {
   VE: 'Latin America',
   AU: 'South Pacific',
   NZ: 'South Pacific',
-  SG: 'South Pacific'
+  SG: 'South Pacific',
+  MK: 'Republic of Macedonia'
```
lpinca: Just add this pls. It should be Europe not Republic of Macedonia I guess. Also please add it in alphabetical order in the europe group.

lpinca:
Thanks!

lpinca:
Updating `events.md` makes sense but there is a lot of noise because:
1. The description of many events changed
2. The `js-yaml` dependency has been updated and it seems that it now uses a sligtly different syntax
3. New events are added

The result is a huge amount of changes hard to review.

PR #792, opened 2016-06-20
update stylint to 1.4.1

ghost:
should fix travis


lpinca:
It's not needed we are already using caret versioning but it doesn't hurt either.

ghost:
oh yeah the latest builds passed. i'm dumb

PR #793, opened 2016-06-21
Make the event-geo script work on Windows

lpinca:
We are now [`splitting`](https://github.com/nodejs/nodejs.org/blob/c32b2b37fc703f881d98a36e010609a896669a55/scripts/event-geo.js#L10) the `events.md` file contents using the `\n` end-of-line marker.
This doesn't work as expected on Windows because lines are terminated with `\r\n`. As result [`lines.indexOf('---')`](https://github.com/nodejs/nodejs.org/blob/c32b2b37fc703f881d98a36e010609a896669a55/scripts/event-geo.js#L11) returns `-1` making the `load()` function fail.

This patch fixes the issue.

### DIFF
--- a/scripts/event-geo.js
+++ b/scripts/event-geo.js
@@ -1,17 +1,16 @@
 'use strict'
+
+const htmlToText = require('html-to-text')
 const yaml = require('js-yaml')
 const fs = require('fs')
 const path = require('path')
-let p = path.join(__dirname, '..', 'locale', 'en', 'get-involved', 'events.md')
-let htmlToText = require('html-to-text')
+
+const p = path.join(__dirname, '..', 'locale', 'en', 'get-involved', 'events.md')
 
 function load () {
-  let buf = fs.readFileSync(p)
-  let lines = buf.toString().split('\n')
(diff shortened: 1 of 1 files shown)


scripts/event-geo.js
```diff
@@ -1,17 +1,16 @@
...
+const htmlToText = require('html-to-text')
 const yaml = require('js-yaml')
 const fs = require('fs')
 const path = require('path')
-let p = path.join(__dirname, '..', 'locale', 'en', 'get-involved', 'events.md')
-let htmlToText = require('html-to-text')
 
-function load () {
-  let buf = fs.readFileSync(p)
-  let lines = buf.toString().split('\n')
```
lpinca, line 10: We can use `os.EOL` here as an alternative fix, but I find the proposed solution easier to grok.

silverwind:
LGTM, another alternative could be `.split(/\r?\n/)`.

scripts/event-geo.js
```diff
@@ -1,17 +1,16 @@
 'use strict'
+
```
stevemao, line 2: nit: any reason to add a newline?

lpinca, line 2: All other files (or most of them) use a new line after `use strict`.

ghost:
LGTM

PR #794, opened 2016-06-21
Extending the noted time for Interactive to include collab summit.

mikeal:


MylesBorins:
LGTM

ghost:
LGTM

MylesBorins:
So I was just looking at the site, and it now makes it look like the events are twice as long. Should we perhaps be explicit about the dates for the summit? It would not be great if people end up booking longer trips and don't have a need for the summit

MylesBorins:
Also I just realized that the US dates should be to Dec 2 and the graphic says Dec 1

PR #815, opened 2016-07-06
Updating Amazon Linux Version

ryanmurakami:
I just verified that the package manager install steps work on the newest version of Amazon Linux (2016.3.3), so I'm updating the version on the package manager page.


lpinca:
Can you please remove the first 3 commits? They only create confusion, even if we squash when merging.

ryanmurakami:
@lpinca Sure thing! Sorry about that.

ryanmurakami:
@lpinca Alright, should be fixed now.

lpinca:
Great, thank you.

PR #817, opened 2016-07-06
Fix typo in the blog/release/v6.3.0.md

vsemozhetbyt:
compliment -> complement


fhemberger:
Good catch! Thank you!

PR #821, opened 2016-07-08
Bithound.metalsmith stylus.2.0.0

NubleX:
Back-up security.

### DIFF
--- a/package.json
+++ b/package.json
@@ -11,7 +11,9 @@
     "test": "npm run test:lint && npm run test:unit && npm run test:smoke",
     "test:lint": "standard",
     "test:unit": "tape tests/**/*.test.js | faucet",
-    "test:smoke": "tape tests/*.smoketest.js | faucet"
+    "test:smoke": "tape tests/*.smoketest.js | faucet",
+    "snyk-protect": "snyk protect",
+    "prepublish": "npm run snyk-protect"
   },
   "repository": {
     "type": "git",
@@ -35,7 +37,7 @@
     "changelog-url": "1.0.2",
     "cheerio": "0.19.0",
     "chokidar": "1.2.0",
-    "handlebars": "4.0.4",
+    "handlebars": "4.0.5",
     "html-to-text": "^1.5.0",
     "js-yaml": "^3.4.5",
     "junk": "1.0.2",
@@ -54,12 +56,13 @@
     "ncp": "2.0.0",
     "node-geocoder": "^3.4.1",
     "node-version-data": "1.0.0",
-    "octonode": "0.7.4",
+    "octonode": "0.7.6",
     "request": "^2.67.0",
     "require-dir": "0.3.0",
     "semver": "5.0.3",
-    "st": "1.0.0",
-    "strftime": "0.9.2"
+    "st": "1.1.0",
+    "strftime": "0.9.2",
+    "snyk": "^1.17.0"
   },
   "devDependencies": {
     "faucet": "^0.0.1",
@@ -68,5 +71,6 @@
     "proxyquire": "^1.7.3",
     "standard": "^6.0.8",
     "tape": "^4.2.2"
-  }
+  },
+  "snyk": true
 }
--- a/package.json
+++ b/package.json
@@ -51,7 +51,7 @@
     "metalsmith-metadata": "0.0.2",
     "metalsmith-permalinks": "0.4.0",
     "metalsmith-prism": "2.1.1",
-    "metalsmith-stylus": "1.0.0",
+    "metalsmith-stylus": "^2.0.0",
     "metalsmith-yearly-pagination": "2.0.0",
     "ncp": "2.0.0",
     "node-geocoder": "^3.4.1",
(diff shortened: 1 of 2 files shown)


ghost:
could you provide a little more information as to what exact security flaw is fixed here? also, it would be nice if you could fix the merge conflict 👍

NubleX:
I'm thinking long-term here since I'm building an OS to operate the central nerve system and artificial limbs.

ghost:
but... how does this relate to this website?

i do see the security flaw though. goes back to [this](https://snyk.io/vuln/npm:minimatch:20160620). fair enough, now we only need to fix this merge conflict

lpinca:
I have no strong opinions, -0.

fhemberger:
I think it would be a good idea, but I guess it won't be added to the org (just like greenkeeper was removed). I think @phillipj might know the reason.

package.json
```diff
@@ -49,16 +51,17 @@
     "metalsmith-metadata": "0.0.2",
     "metalsmith-permalinks": "0.4.0",
     "metalsmith-prism": "2.1.1",
-    "metalsmith-stylus": "1.0.0",
+    "metalsmith-stylus": "^2.0.0",
```
stevemao, line 54: I think these should be in another pr

package.json
```diff
@@ -11,7 +11,9 @@
     "test": "npm run test:lint && npm run test:unit && npm run test:smoke",
     "test:lint": "standard",
     "test:unit": "tape tests/**/*.test.js | faucet",
-    "test:smoke": "tape tests/*.smoketest.js | faucet"
+    "test:smoke": "tape tests/*.smoketest.js | faucet",
+    "snyk-protect": "snyk protect",
+    "prepublish": "npm run snyk-protect"
```
fhemberger, line 16: Website is not published to npm, so it should be

``` diff
-    "snyk-protect": "snyk protect",
-    "prepublish": "npm run snyk-protect"
+    "postinstall": "snyk protect"
```

package.json
fhemberger, line 54: +1 for "Update dependencies" PR

lpinca, line 54: `metalsmith-stylus` is already at version `2.0.0` on master.
Some other dependencies updated here are also already updated on master.

package.json
MylesBorins, line 16: isn't prepublish always run after install?

ghost, line 16: @TheAlphaNerd not always, but most times: 
screen shot 2016-07-09 at 23 32 30

fhemberger, line 16: I think the intention is better understandable with 'postinstall', even if 'prepublish' might work as well.

phillipj:
I wasn't aware of snyk, at first glance it seems like a good thing to have. Although I would prefer if it would be less obtrusive, especially thinking of the `.snyk` file.

> I think it would be a good idea, but I guess it won't be added to the org (just like greenkeeper was removed). I think @phillipj might know the reason.

You're correct about not being able to grant the snyk GH application access. We could get around that issue by running it via Travis. I did some tests locally and seems like running `$ snyk test` does the job.

image

How do we want to integrate this in our current flow? E.g. should we break the build or just view snyk issues as (spammy) warnings? Or maybe just add this badge to README.md [](https://snyk.io/test/github/nodejs/nodejs.org) (yes, that badge is for nodejs.org).

And seeing the [current snyk report](https://snyk.io/test/github/nodejs/nodejs.org), we're not able to fix all issues as some dependencies seem to have been abandoned. Might need to replace a couple of deps and possibly change some of our code.

fhemberger:
@phillipj `snyk protect` applies patches locally for those issues found, as long as no official fix/update is provided.

phillipj:
@fhemberger sounds like blackmagic. What does that mean in detail? Got any examples of that by any chance?

fhemberger:
@phillipj No black magic, all changes are openly available on GitHub:
- marked: https://github.com/Snyk/vulndb/tree/snapshots/master/patches/npm/marked
- minimatch: https://github.com/Snyk/vulndb/tree/snapshots/master/patches/npm/minimatch/20160620

phillipj:
Aah, so snyk installs its own version of e.g. minimatch in prepublish - that's the blackmagic I was looking for.

Again snyk seems too intrusive and possibly misleading for my taste, but I won't hold it back if most of you like it, so -0 from me.

NubleX:
i got 80% stability mark on my node [😣]

fhemberger:
@phillipj Not exactly. It installs the official package from npm, then patches the vulnerability (if patching is possibe).

E.g: Dependency `x@latest` still uses package `y@2.0.1`, which contains a vulnerability. It is reported to the author and gets patched, `y@3.0.0`. Still, the author of `x` takes their time to update the dependency (or isn't even aware of the update). So snyk tries to backport the patch to `y@2.0.1` (if possible) for the time being. It's the same "black magic" Linux package maintainers have been doing for years. 😃 

I've been trying myself to reach out to npm package maintainers to address dependency vulnerability issues, it's very hard work and PRs are being ignored some times for ages even if you make it clear, that a dependency is vulnerable and a security issue. Especially if the package is n-levels deep in the dependency chain. So snyk tries to mitigate this problem.

fhemberger:
I've added a clean version of this PR.

NubleX:
nice one :)

fhemberger:
Closing this for #841.

Issue #830, opened 2016-07-17
questions for "about" documents


outsideris:
We're working on translating the website as Korean, see https://github.com/nodejs/nodejs-ko/issues/381. We encounter a few issue with content.
1. In [locale/en/about](https://github.com/nodejs/nodejs.org/tree/master/locale/en/about/), there is an `advisory-board` folder, but we can't found where it is used on the website. we guess it is a missing link.
2. [locale/en/about/organization.md](https://github.com/nodejs/nodejs.org/blob/master/locale/en/about/organization.md) is same with above. 

So, we can't decide whether we put our effort to translate these documents into Korean. If these aren't used anymore, these should be removed. We can make a pull request for it, if so.

And we can't understand the meaning at [Addon API Working Group](https://nodejs.org/en/about/working-groups/#addon-api)

> The NAN project makes available an abstraction layer for native add-on authors for `both Node.js and Node.js`, assisting in the writing of code that is compatible with many actively used versions of `Node.js, Node.js`, V8 and libuv.

in `both Node.js and Node.js` and `Node.js, Node.js`, why is there two "Node.js". Does it means different respectively. It was introduced in [this commit](https://github.com/nodejs/nodejs.org/commit/41fe898c4e6a24dd0620e12519504d4545e7dda9#diff-2a9a3bc70d631e83d2e68bae34835c95).

stevemao:
> in both Node.js and Node.js and Node.js, Node.js, why is there two "Node.js". Does it means different respectively.

This must be a mistake.

Fishrock123:
> 1. In [locale/en/about](https://github.com/nodejs/nodejs.org/tree/master/locale/en/about/), there is an `advisory-board` folder, but we can't found where it is used on the website. we guess it is a missing link.
> 2. [locale/en/about/organization.md](https://github.com/nodejs/nodejs.org/blob/master/locale/en/about/organization.md) is same with above. 

You'd need to check the website nginx config: https://raw.githubusercontent.com/nodejs/build/master/setup/www/resources/config/nodejs.org

I've copied the relevant bits below:

<details>

``` nginx

    rewrite ^/about/advisory-board(.*)$                           https://$server_name/en/about/organization/ permanent;
    rewrite ^/advisory-board(.*)$                                 https://$server_name/en/about/organization/ permanent;
    rewrite ^/about/organization/tsc-meetings/?$                  https://$server_name/en/foundation/tsc/minutes/ permanent;
[...]

PR #831, opened 2016-07-19
Remove multiple "Node.js" in documents

outsideris:
It looks like all occurrences of "io.js" were replaced with "Node.js".
See, #830


lpinca:
LGTM.

PR #838, opened 2016-07-22
Make the release-post script more tolerant

lpinca:
Right now the release script does not work if a release section starts with a new line.
This patch fixes the issue by removing any leading or trailing new line.

### DIFF
--- a/scripts/release-post.js
+++ b/scripts/release-post.js
@@ -103,7 +103,7 @@ function fetchChangelog (version) {
     const rxSection = new RegExp(`<a id="${version}"></a>\\n([\\s\\S]+?)(?:\\n<a id="|$)`)
     const matches = rxSection.exec(data)
     return matches
-      ? matches[1]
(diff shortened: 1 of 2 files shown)


evanlucas:
Maybe add an additional test that fetches the 6.3.1 changelog?

lpinca:
@evanlucas yeah I thought about that but it requires a new fixture.

evanlucas:
IMO that is worth it, but up to you

lpinca:
I can do that but it should only test that the content is trimmed as the other parts (author / version policy extraction, shasum etc.) are already covered by the test suite.
Since there is already a test for the trim part here I think that adding another is redundant.
The test is also meaningful as a trailing new line is actually removed from 4.1.1 changelog.

evanlucas:
Fair enough. LGTM. Thanks for doing this btw!

hiroppy:
LGTM 👍

Issue #846, opened 2016-08-03
Download page's bottom link is missing.


wongzigii:
- **URL**: https://nodejs.org/zh-cn/download/
- **Browser version**: Version 51.0.2704.103 (64-bit)
- **Operating system**: Mac OSX 10.11

The bottom link is missing.

Knighton910:
Good job on the pull Req, once merged, come back & close issue https://github.com/nodejs/nodejs.org/pull/847

👍

PR #847, opened 2016-08-03
Add chinese missing link for download page

wongzigii:
Fix https://github.com/nodejs/nodejs.org/issues/846

### DIFF
--- a/locale/zh-cn/site.json
+++ b/locale/zh-cn/site.json
@@ -71,6 +71,10 @@
         "package-manager": {
             "link": "download/package-manager",
             "text": "使用包管理器安装 Node.js"
+        },
+        "shasums": {
+          "link": "SHASUMS256.txt.asc",
+          "text": "查看已签名的 SHASUM"
         }
     },
     "docs": {
--- a/locale/zh-cn/site.json
+++ b/locale/zh-cn/site.json
@@ -74,7 +74,7 @@
         },
         "shasums": {
           "link": "SHASUMS256.txt.asc",
-          "text": "查看已签名的 SHASUM"
+          "text": "查看发布文件的 SHASUM 签名"
         }
     },
     "docs": {


lpinca:
LGTM if the value of the `text` field is correct :)

phillipj:
Hm, this shows we've got a bug in the site.json fallback logic found in [build.js#L44](https://github.com/nodejs/nodejs.org/blob/master/build.js#L44), right?

..ofcourse not a blocker for this PR.

lpinca:
@phillipj correct, `Object.assign` is not suitable in this case. We need to use https://lodash.com/docs#defaultsDeep or something like that.

lpinca:
Looks like #848 is not enough to fix the fallback issue because the page uses [`site.download.shasums.text`](https://github.com/nodejs/nodejs.org/blob/024130aa6805dda4a44e23576be0fb318373d6f9/layouts/partials/download-list.hbs#L3) and [`site`](https://github.com/nodejs/nodejs.org/blob/024130aa6805dda4a44e23576be0fb318373d6f9/build.js#L65) is actually the parsed contents of the `site.json` not the result of the [`i18nJSON`](https://github.com/nodejs/nodejs.org/blob/024130aa6805dda4a44e23576be0fb318373d6f9/build.js#L48-L53) function.

This fallback thing needs a little work. This is what I have in mind:
- Remove `i18n` [here](https://github.com/nodejs/nodejs.org/blob/024130aa6805dda4a44e23576be0fb318373d6f9/build.js#L67) and the [i18n helper](https://github.com/nodejs/nodejs.org/blob/024130aa6805dda4a44e23576be0fb318373d6f9/scripts/helpers/i18n.js).
- Make `site` be the returned value of the `i18nJSON` function.
- Use `site` everywhere.

lpinca:
cc: @nodejs/nodejs-cn

locale/zh-cn/site.json
```diff
@@ -71,6 +71,10 @@
         "package-manager": {
             "link": "download/package-manager",
             "text": "使用包管理器安装 Node.js"
+        },
+        "shasums": {
+          "link": "SHASUMS256.txt.asc",
+          "text": "查看已签名的 SHASUM"
         }
```
fundon, line 78: “查看发布文件的 SHASUM 签名“ 不知道如何？

fengmk2:
+1

laosb:
LGTM except @fundon 's comment above.

locale/zh-cn/site.json
wongzigii: 查看已发布版本的 SHASUM 签名？

wongzigii:
Just update `site.json` for a better translation! Ready for merge. 😄

wongzigii:
Thanks for reviewing!

lpinca:
Thank you @wongzigii.

PR #848, opened 2016-08-03
Replace Object.assign with lodash.defaultsDeep

lpinca:
`Object.assign` does not work recursively so it's not suitable to merge the localization [data](https://github.com/nodejs/nodejs.org/blob/557f1e7f0bad5b26a66e0b2327ea270349397902/build.js#L48-L49).
If a nested item is missing from [`templateJSON`](https://github.com/nodejs/nodejs.org/blob/557f1e7f0bad5b26a66e0b2327ea270349397902/build.js#L49) that item will not be included in the final object, despite it being available in [`defaultJSON`](https://github.com/nodejs/nodejs.org/blob/557f1e7f0bad5b26a66e0b2327ea270349397902/build.js#L48).

This patch should fix the issue.

Ref: https://github.com/nodejs/nodejs.org/pull/847

### DIFF
--- a/package.json
+++ b/package.json
@@ -40,6 +40,7 @@
     "html-to-text": "2.1.3",
     "js-yaml": "3.6.1",
     "junk": "1.0.3",
+    "lodash.defaultsdeep": "4.5.1",
     "map-async": "0.1.1",
     "marked": "0.3.5",
     "metalsmith": "2.1.0",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -40,6 +40,7 @@
     "html-to-text": "2.1.3",
     "js-yaml": "3.6.1",
     "junk": "1.0.3",
+    "lodash.defaultsdeep": "4.5.1",
```
lpinca, line 43: I don't know if it makes sense to use full `lodash`, we may need more utilities in future.

ghost, line 43: let's keep it at this for now

ghost:
LGTM

PR #854, opened 2016-08-08
locale: Fix Ukrainian translation.

### LINKED ISSUE #88: Broken layout on the site
image

We have broken layout on the site.

@RReverser has already checked that `Документація` won't break layout and everything was going to be okay. But something happened. I suppose this problem was caused by font which probably doesn't support cyrillic characters.

What do we have to do with that? I think the easiest way is removing `Документація` word from this blocks.

cc @nodejs/nodejs-uk

denysdovhan:
This commit includes few important fixes:
- реком**_а_**ндовано → реком**_е_**ндовано — fix misspelled word.
- Fix https://github.com/nodejs/nodejs-uk/issues/88
- Chrome~~'s~~ V8 — there's no possessive case in Ukrainian.


denysdovhan:
cc/ @nodejs/nodejs-uk

ghaiklor:
LGTM

ghost:
LGTM!

PR #855, opened 2016-08-10
Uppercase Node.js in LICENSE

phillipj:
There has been an initiative to unify how we write "Node.js", this minor change fixes that in our LICENSE file.

..this also serves as a test to check if the @nodejs-github-bot webhook still works after being moved to a new server.


jasnell:
LGTM

phillipj:
Even the bot looks good 👍

Issue #851, opened 2016-08-04
Use standard in static JS?


zaynv:
I've noticed that you guys are using `standard` for linting, what do you think about also following the `standard` style for the client side as well for consistency? I understand you couldn't do it for the `modernizr.custom.js` but it could probably be easily done in the `download.js` file.

lpinca:
I'm fine with fixing the style of `download.js` but in the `static/legacy` folder there are other scripts which don't follow the `standard` style. All of them along with `modernizr.custom.js` should be ignored.

PR #856, opened 2016-08-11
Make the download script follow the standard style

lpinca:
This fixes the style of [`download.js`](https://github.com/nodejs/nodejs.org/blob/e0fe46531491965579da53e21e4c70c40cbdd716/static/js/download.js) as suggested in #851.
I've run `standard --fix static/js/download.js` to fix it.

### DIFF
--- a/static/js/download.js
+++ b/static/js/download.js
@@ -1,79 +1,82 @@
 ;
 (function (d, n) {
-  'use strict';
+  'use strict'
 
   // document.querySelectorAll polyfill for ancient IEs
   // https://gist.github.com/chrisjlee/8960575
   if (!document.querySelectorAll) {
     document.querySelectorAll = function (selectors) {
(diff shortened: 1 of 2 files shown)


static/js/download.js
```diff
@@ -1,79 +1,82 @@
 ;
 (function (d, n) {
-  'use strict';
+  'use strict'
 
   // document.querySelectorAll polyfill for ancient IEs
   // https://gist.github.com/chrisjlee/8960575
   if (!document.querySelectorAll) {
     document.querySelectorAll = function (selectors) {
-      var style = document.createElement('style'), elements = [], element;
```
lpinca, line 9: This change has been done manually as it wasn't fixed automatically.

saadq:
LGTM, passes `standard`.

ghost:
LGTM ✨

PR #863, opened 2016-08-17
Update case-studies.md, add NASA

gtewallace:
We have a new case study with NASA that we'd like to add to this page.

Thanks for reviewing it. 

If possible, we'd like it available by Thursday, Aug. 18 so we can point reporters and analysts to it.


lpinca:
LGTM.

lpinca:
For other reviewers, the pdf has been added in another commit (https://github.com/nodejs/nodejs.org/commit/4b7d8b788ff5a5617bb5ffa2212bf41072059558).

DavidTPate:
LGTM

PR #867, opened 2016-08-18
Use an optimized JPG instead of a huge PNG file for the background image of Interactive section

SamuelBolduc:
The site load feels very fast and snappy, except for this image which takes 1-2 seconds to fully load, from top to bottom (on a 30mbps connection, probably worse for many visitors).

The old PNG file was 377KB, the new JPG file is 68.8KB. The quality difference is barely noticeable, especially in this case where the image is only used as a background and is not meant to be the center of attention.


lpinca:
For anyone interested here are the two images:

**JPG**

background

**PNG**

background

The JPG version is very blurred but as @SamuelBolduc said it's not very noticeable. There is a logo over it and the event locations and dates.

+0 from me.

ghost:
+1 from me, as this has personally bugged me on slow connection speeds

DavidTPate:
+1 from me as well, I've noticed it a number of times on spotty connections.

lpinca:
Thank you.

PR #875, opened 2016-08-28
Adding Turkish user group

yasinaydinnet:
Thanks.

### DIFF
--- a/locale/en/get-involved/index.md
+++ b/locale/en/get-involved/index.md
@@ -62,3 +62,4 @@ right place. Explore our community resources to find out how you can help:
 - [Taiwan community](http://nodejs.tw)
 - [Nicaragua Node.js community](http://nodenica.com/)
 - [Korean Node.js community](http://nodejs.github.io/nodejs-ko/)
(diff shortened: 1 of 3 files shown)


fhemberger:
Hmm, looks more like a link list (books, jobs, template engines?!) than an actual community page.

yasinaydinnet:
@fhemberger  We are newly established and rapidly growing community so that's all we have. Plus, we are the only one so far in Turkey.
What else should we need?

yasinaydinnet:
@fhemberger I suggest you to look at the current community pages (If they still work).
Update: I moved them to where I think they belong: https://github.com/nodejs/nodejs.org/issues/876

fhemberger:
Yavaş yavaş, didn't say we won't add you. 😄 

The whole "community page" thing grew a bit organically, and we usually don't check the pages regularly once we added them. So thanks for pointing that out! Entries are supposed to be sorted by country name.

yasinaydinnet:
Ok, sorry :smile: Waiting for the merge, then. In the meantime, we will keep adding stuff!

yasinaydinnet:
Thanks :)

PR #880, opened 2016-08-30
Use a smaller image for the header logo

lpinca:
This reduces the size of the logo used in the header.
The new size is 245 x 150 which is 2x (for retina displays) the maximum size the image can be.
The image has been compressed with ImageOptim.


ghost:
LGTM

fhemberger:
Compressed the image … LGTM

lpinca:
I think it's possible to shave some more bytes with zopflipng 😄.

silverwind:
LGTM. It'd be awesome if it were a SVG, but I don't suppose a vector source image is available?

lpinca:
@silverwind there is a vectorial version (`nodejs-new-pantone-white.ai`) which could serve the purpose.

lpinca:
I tried to convert `nodejs-new-pantone-white.ai` to SVG with Inkscape but the colors look different. Here is the result https://dl.dropboxusercontent.com/u/58444696/nodejs-new-pantone-white.svg.

Can someone try with Illustrator?

silverwind:
I too only have inkscape available, but it seems the dull colors are there in the source .ai as well. Below is an attempt at conversion, I'd wager the gradient `<stop>`s could be removed completely, so it's a WIP at best.

silverwind:
https://github.com/nodejs/nodejs.org/pull/884 for my version of the logo.

PR #881, opened 2016-08-30
Update case-studies.md

gtewallace:
Hi,

We'd like to add to this page the embed code for the Slideshare of V1 of the Enterprise Case for Node.js presentation.

I _think_ I've done it correctly, and I greatly appreciate an expert set of eyes and hands here.

Thanks in advance!

Greg

PR #882, opened 2016-08-31
Update case-studies.md

gtewallace:
@lpinca thanks for the guidance :)

I edited the embed code to only include the iframe (I think)

thank you!


lpinca:
@gtewallace thanks but I think it's better to also clean up the `<iframe>` attributes :)
I close this, we can continue on the old pull request #881, sorry!

gtewallace:
#Githubnewbie thanks!!

PR #886, opened 2016-09-07
Add the Facebook Israeli group of node.js to the get involved page

gonenduk:
Add a link to the Facebook Israeli community of node.js on the get involved page.


fhemberger:
LGTM

PR #894, opened 2016-09-13
Updated URL to @fene's account in README

phillipj:
@fene would you agree, or are you about to change your username again? :smile:


ghost:
i would like to write a clever reply but yes, this can be changed :P

PR #903, opened 2016-09-18
Update ARM related wording on download page

### LINKED ISSUE #899: Linux Binaries should say Linux (x86) Binaries and ARM binaries should say Linux (ARM) Binaries?
https://github.com/nodejs/help/issues/278#issuecomment-247746229
This confuses me for a while and I'm a noob of systems so I was afraid to ask. But I thought ARM should be architectures for computer processors, and it's under "Additional Platforms", where as Linux is an OS. The page makes "ARM" look like an OS/Platform. Is there any reason to make it this way?

fhemberger:
- Renames "ARM binaries" to "Linux binaries (ARM)"
- Moves Linux ARM binaries to primary downloads
- Remove archive file extension where only one type is provided

Fixes #899

### DIFF
--- a/layouts/partials/primary-download-matrix.hbs
+++ b/layouts/partials/primary-download-matrix.hbs
@@ -64,9 +64,16 @@
       </tr>
 
       <tr>
-        <th>Linux Binaries (.tar.xz)</th>
+        <th>Linux Binaries (x86/x64)</th>
         <td colspan="3"><a href="https://nodejs.org/dist/{{version.node}}/node-{{version.node}}-linux-x86.tar.xz">32-bit</a></td>
         <td colspan="3"><a href="https://nodejs.org/dist/{{version.node}}/node-{{version.node}}-linux-x64.tar.xz">64-bit</a></td>
+      </tr>
+
+      <tr>
+        <th>Linux Binaries (ARM)</th>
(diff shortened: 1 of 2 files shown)


lpinca (approved):
LGTM

layouts/partials/primary-download-matrix.hbs
```diff
@@ -64,12 +64,19 @@
       </tr>
 
       <tr>
-        <th>Linux Binaries (.tar.xz)</th>
+        <th>Linux Binaries (x86)</th>
```
lpinca: Maybe it's better to use `x86/x64` here.

lpinca (changes requested):

layouts/partials/primary-download-matrix.hbs
fhemberger: Not entirely sure about this, IMHO `x86` is the architecture platform, 32/64bit are different versions of it. Any thoughts @nodejs/website, @nodejs/build?

lpinca: AFAIK `x86` is an instruction set and `x86-64` is an extension of `x86` so I'm also ok as is. I suggested the change because the links below are `linux-x86` and `linux-x64`.

silverwind: I think `x86/x64` would be clearer.

fhemberger: Ok, changed it.

lpinca (approved):

stevemao (approved):

PR #905, opened 2016-09-20
Remove double properties in `.editorconfig`

zlatanvasovic:
There's no reason to duplicate any property in editorconfig. :bulb:


fhemberger:
LGTM

PR #910, opened 2016-09-22
remove banner

Trott:
The security banner is 3 months old. While it's nice to leave it up for
a good while, we risk people not seeing it (because they're used to it
always being there) when we change it for a new security update banner.

(I hope I did this correctly!)


fhemberger:
Not sure if we need to keep it visible for the most recent OpenSSH security fixes …

lpinca:
I guess we can replace it with #911 once merged.

PR #915, opened 2016-09-23
update instructions for OpenBSD

qbit:
On OpenBSD we use a different port location (/usr/ports/lang/node), this updates the docs to reflect that.

Also show the difference between pkg-ng and OpenBSD's pkg_add stuff.

### DIFF
--- a/locale/en/download/package-manager.md
+++ b/locale/en/download/package-manager.md
@@ -224,11 +224,19 @@ Additionally, versions of **CentOS** and **Scientific Linux** corresponding to t
 
 Node.js is available through the ports system.
 
+**FreeBSD**:
+
 ```bash
 /usr/ports/www/node
 ```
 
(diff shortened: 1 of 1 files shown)


lpinca:
LGTM but my experience with FreeBSD/OpenBSD is zero.

qbit:
Here is the link to the port in cvs as well: http://cvsweb.openbsd.org/cgi-bin/cvsweb/ports/lang/node/

lpinca:
I'm inclined to merge this, it will not cause any harm and we can fix it later if it is not perfect.
cc: @nodejs/build.

jbergstroem (approved):
LGTM

lpinca:
@jbergstroem, thanks.

lpinca:
@qbit, thank you.

PR #921, opened 2016-09-27
Replacing mailing list signup.

mikeal:
We're going to use MailChimp for this so we need to replace the old Pardot form.


ghost:
LGTM i guess?

hackygolucky:
👍

stevemao:
@mikeal I have already signed on before. Should I sign on again with the new service?

PR #931, opened 2016-10-01
Revert "Blog: fix CVE ID for writeHead() bug / CVE-2016-5326"

rvagg:
Reverts nodejs/nodejs.org#930

@nodejs/security darn it, digging further I've realised that it was actually correct as is! CVE-2016-5326 is for a separate issue that we have decided not to handle as a security problem afterall and CVE-2016-5325 is the one we were going to fix in June but deferred and are now releasing, so it was right all along! Sorry for the confusion!

### DIFF
--- a/locale/en/blog/vulnerability/september-2016-security-releases.md
+++ b/locale/en/blog/vulnerability/september-2016-security-releases.md
@@ -35,9 +35,7 @@ Originally reported by Alexander Minozhenko and James Bunton (Atlassian).
 
 All versions of Node.js are **affected**.
 
-### CVE-2016-5326: `reason` argument in `ServerResponse#writeHead()` not properly validated
-
-***Update 1-Oct-2016: this was originally reported as CVE-2016-5325, it has been updated with the correct ID, CVE-2016-5326***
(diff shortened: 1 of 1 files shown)


indutny (changes requested):
Em... are you sure now? How can we check?

rvagg:
@indutny see https://github.com/nodejs/security/issues/63#issuecomment-224449526

rvagg:
"HTTP processing security defect (CVE-2016-5325)" is the one we reported and retracted because in the end we decided we weren't 100% on how we were handling it. Now we've committed to a solution and it's finally published. CVE-2016-5326 is something entirely different that probably won't ever be used (as it currently stands).

rvagg:
anyone willing to endorse me on retracting this embarrassing correction?

indutny (approved):

indutny:
LGTM

dougwilson (approved):
I will, since I 👍 the original :) They were both just reserved publically, so figured that at least the actual blog wording looks fine, assuming the change itself was correct :)

rvagg:
sorry, all my bad, didn't go deep enough when responding to the original request for clarification!

PR #940, opened 2016-10-06
Update links for AIX and Linux on Z downloads

mhdawson:
**NOTE:** This should not land until 6.X becomes LTS.  This is because binaries exist only for 6.X and later and I've not special cases LTS or Current.  Once 6.X becomes LTS then binaries will be available for both LTS and Current.

I'd like to get this reviewed in advance so that ideally it can land on the same day that 6.X becomes LTS.

Now that binaries are being built as part of release:
- Update the links for AIX and Linux on Z on the download
  page so that they point to the binaries available on
  the community download site instead of the
  external IBM download page.
- Add AIX, Linux s390x and Linux PPC BE to the helper
  used to generate release announcements.

### DIFF
--- a/layouts/partials/secondary-download-matrix.hbs
+++ b/layouts/partials/secondary-download-matrix.hbs
@@ -25,11 +25,11 @@
 
       <tr>
         <th>Linux on System z</th>
-        <td colspan="6"><a href="http://www.ibm.com/developerworks/web/nodesdk/">Download (Unofficial, provided by IBM)</a></td>
+        <td colspan="6"><a href="https://nodejs.org/dist/{{version.node}}/node-{{version.node}}-linux-s390x.tar.xz">64-bit</a></td>
       </tr>
       <tr>
         <th>AIX on Power Systems</th>
-        <td colspan="6"><a href="http://www.ibm.com/developerworks/web/nodesdk/">Download (Unofficial, provided by IBM)</a></td>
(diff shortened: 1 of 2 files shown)


mhdawson:
@rvagg can you review as you seem to have done the last updates to: scripts/helpers/downloads.js

jasnell (approved):
LGTM

rvagg:
OK, yeah, lgtm I guess this should work when we have LTS=6 and Current=7, I don't _think_ there's another way of getting that page for older builds. I'd like someone from @nodejs/website to sign-off though. Perhaps @phillipj?

thefourtheye:
Changes look okay. Was there any official communication stating that Node.js officially supports these builds?

mhdawson:
@thefourtheye we were just formalizing under https://github.com/nodejs/node/pull/8922 so no formal announce yet.  I'm hoping it will be part of the initial 6.X release/announcement as we have achieved consensus and its just down to getting the doc PR ready to land.

@fhemberger I know you commented before care to review as website rep?

zlatanvasovic (approved):
LGTM :shipit:

fhemberger:
@mhdawson LGTM as well.

fhemberger (approved):

mhdawson:
Ok, seems we are now  back to having LTS and Current and we should be good to land this.  Last chance to object, otherwise I'll land this later today.

MylesBorins:
LGTM, sorry we missed this on Tuesday

MylesBorins (approved):

mhdawson:
@fhemberger thanks.

Issue #944, opened 2016-10-08
A little confuse for "Current version" and "Current"


outsideris:
<img width="487" alt="2016-10-08 20 18 25" src="https://cloud.githubusercontent.com/assets/390146/19212811/c3c4de02-8d94-11e6-86b0-d66d0bb9ea68.png">

We call Node.js's latest version as "Current". So, I think there is a little confuse in [the download page](https://nodejs.org/en/download/) to beginners because there are two "current".

How do you think?

rauschma:
If you really want to promote LTS as “safe” and Current as “latest” then a separate link underneath would probably work better than a toggle than doesn’t look like a toggle (among other issues: its bottom half could be mistaken for a description of the top half).

rauschma:
Other UI options to express alternatives: tabs and radio buttons

mxstbr:
Yeah I can confirm, I did not get that's a toggle until @rauschma pointed it out.

rauschma:
As @outsideris  mentioned, the other confusing element is “Current version: v4.6.0”. Here, it may be better to write, e.g., “Version to download: …”.

mairead:
If I were to take a look at this who is the best person to contact re: browser support/testing list? Thanks, m

MylesBorins:
I've submitted a fix for this specific issue, thank you @outsideris for submitting the bug.

For other UI / UX related problems please open a separate issue to focus specifically on them.

outsideris:
@TheAlphaNerd Awesome!

mairead:
oh :/

MylesBorins:
@mairead I may have misunderstand your original comment. If you were volunteering to help I apologize for stepping in.

If you would like some help finding ways to collaborate on this project or on node.js core please feel free to reach out, you can find my contact info on my website https://thealphanerd.io

PR #945, opened 2016-10-09
change label for download to no longer say Current

MylesBorins:
This made sense prior to Stable being change to Current, now it is
just confusing.

Fixes: https://github.com/nodejs/nodejs.org/issues/944

Before: 

c3c4de02-8d94-11e6-86b0-d66d0bb9ea68

After:

screen shot 2016-10-09 at 1 01 15 pm

screen shot 2016-10-09 at 1 03 07 pm


lpinca (approved):
LGTM

PR #963, opened 2016-10-14
Simplify package.json

zlatanvasovic:
1. There's a shorter `owner/repository` syntax for GitHub repositories according to [npm docs](https://docs.npmjs.com/files/package.json#repository).
2. [normalize-package-data](https://github.com/npm/normalize-package-data#what-normalization-currently-entails) says bugs are needless if you've specified the repository.

### DIFF
--- a/package.json
+++ b/package.json
@@ -17,15 +17,9 @@
     "test:snyk": "snyk test",
     "postinstall": "snyk protect"
   },
-  "repository": {
-    "type": "git",
-    "url": "git+https://github.com/nodejs/nodejs.org.git"
-  },
+  "repository": "nodejs/nodejs.org",
   "author": "Node.js Website Working Group",
   "license": "MIT",
-  "bugs": {
-    "url": "https://github.com/nodejs/nodejs.org/issues"
-  },
   "engines": {
     "node": ">=4.0.0"
   },


lpinca:
+0, I see no real value in doing this.

zlatanvasovic:
Software development is based on constant improvement. If people reacted like you did now, we'd still be stuck on writing programs in machine language.

lpinca:
@zdroid I'm sorry if my comment looks like a judgment, that wasn't my intention.
I'm not opposing to this change (note the +0) but I can't see how it is useful.

pin3da:
@zdroid In my opinion is less readable because it hides information.  

btw, I don't think that '+0' is a bad opinion, is just appointing that there is no _improvement_ in doing that.

zlatanvasovic:
It doesn't hide any information. You don't need to be told again you're using a git repository hosted on https://github.com.

sotayamashita:
@zdroid 

> 1. There's a shorter owner/repository syntax for GitHub repositories according to npm docs.

If it is an optional choice, I think we do not have to change it.

> 2. normalize-package-data says bugs are needless if you've specified the repository.

I agree with this to follow the standard.

zlatanvasovic:
@sotayamashita Basically, both of those are already standard, especially the first. They're widely accepted by the npm community.

sotayamashita:
@zdroid I am sorry and I do not know it. If both of those are, I think it looks good

ryanmurakami:
+1
I like the reduction of boilerplate in a package.json. I think the assumptions of where the source code is stored is widely understood as being Github, so removing that just makes it cleaner and easier to read.

zlatanvasovic:
Merging this, as opinions were positive and the proposed changes don't affect general build at all.

PR #973, opened 2016-10-17
V6.x lts release

MylesBorins:
This is a band aid solution for now that changes the copy for v6.x to be LTS rather than Current. It replaces the references to Current with LTS, and modifies some of the copy. I have a separate fork that we should land that will revert this change and update the copy for v7.x


MylesBorins:
This has now been updated to monkey patch `scripts/helpers/latestversion.js` to guarantee we get the expected versions on the main page

This also removes the tests that were breaking after the change.

THIS SHOULD BE REVERTED WHEN v7 LANDS

lpinca (approved):
LGTM

SamuelBolduc (approved):

rvagg:
fixes the front page, which is probably most important, so lgtm, but it doesn't deal with https://nodejs.org/en/download/current/ which is still stuck on 6.8.1. I'm not sure how to deal with that one other than pretend that 6.9.0 is current for a week or so.

phillipj:
> I have a separate fork that we should land that will revert this change and update the copy for v7.x

Goodie 👍 Would be nice if you opened a PR on that one asap, that sounds like a more permanent change and may need more discussion

MylesBorins:
I'm merging this for now, will look into the other route a bit later tonight. if anyone else wants to take a stab in the mean time please dl

lpinca:
I was sure to have already put a LGTM yesterday here, but I didn't.

lpinca:
@rvagg https://nodejs.org/en/download/current/ seems to work fine.

phillipj:
> .. but it doesn't deal with https://nodejs.org/en/download/current/ which is still stuck on 6.8.1

I'm also seeing v6.9.0 now, I assume you saw the old cached version of that page.

jbergstroem:
Looks good to me!

PR #974, opened 2016-10-17
Update fedora

kasicka:
Update guidelines for installing from Fedora repositories

### DIFF
--- a/locale/en/download/package-manager.md
+++ b/locale/en/download/package-manager.md
@@ -214,9 +214,9 @@ Additionally, versions of **CentOS** and **Scientific Linux** corresponding to t
 
 **Supported Fedora versions:**
 
-* **Fedora Rawhide** (i686/x86_64/armv7hl)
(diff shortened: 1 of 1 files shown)


jbergstroem:
We have a testing gap for fedora25 in CI and still support fedora22 (which probs should be dropped) -- not sure how to handle this but otherwise LGTM.

kasicka:
I thought that was more about what is available from Fedora repositories, which is responsibility of Fedora packagers.

jbergstroem:
Upon reading the full page, the responsibility is pushed back to package maintainers but the word "supported" is ambiguous. Anyway, don't let me hold up progress.

fhemberger:
@kasicka @jbergstroem Sorry for the long delay: Can this be merged? If the range of supported versions/platforms is managed by the Fedora package maintainers, should we remove the list entirely?

fhemberger:
Merging this now … if the supported Fedora version range changes (or should be removed entirely), please open a new PR.

PR #976, opened 2016-10-18
blog: make shasums an h3, remove SHA512/SHA256 note

rvagg:
We should only have SHA256, both file and content hashes, to avoid confusion, so we don't need the note. Also making it a proper heading now. What think ye?


zlatanvasovic (approved):

phillipj:
👍 as there has been some confusion about this

lpinca (approved):
LGTM

Issue #985, opened 2016-10-18
Use one stylesheet for all locales


zlatanvasovic:
Basically, there are no changes in the `styles.styl` files of all locales. I am wondering why do we still have them.

In case we really need to add language-specific styles, there is [:lang pseudo-class](https://developer.mozilla.org/en-US/docs/Web/CSS/:lang) in CSS.

phillipj:
I'm all for removing complexity 👍

@fhemberger @mikeal maybe you remember what the actual need behind locale `.styl`s were back in the day?

mikeal:
I think we were worried that someone _might_ need to adjust something in a locale specific way so we wanted the option, but they should fall back to the english one when unused.

Being that nobody has actually used them it's probably best to get rid of them.

fhemberger:
+1 for removing them.

There should be no need for language specific styling. RTL languages for example could be handled in the global styles as well.

zlatanvasovic:
@nodejs/documentation Any other opinions? Let's decide on this to make the process of translating even easier.

Fishrock123:
I think we should use a single stylesheet if possible.

diagramatics:
I've created a preliminary PR #1085 if this is going to be the decision.

PR #988, opened 2016-10-20
fix(locale): add missing styles.styl file for de

junedev:
Currently the styling of https://nodejs.org/de/ is broken. I think the reason is that I did not copy the styles.styl into the translation folder "locale/de" when working on #971. Unfortunately the bug did not appear when testing everything locally. Fixing the issue in this PR. Sorry for the inconvenience.

### DIFF
--- a/locale/de/styles.styl
+++ b/locale/de/styles.styl
@@ -1,0 +1,1 @@
+@import 'base'


zlatanvasovic (approved):

zlatanvasovic:
Related to #985. If we used only one stylesheet for all locales, there'd be no such issues.

lpinca (approved):
LGTM

lpinca:
Thank you!

PR #984, opened 2016-10-18
Remove 'Security' and 'Foundation' from About nav

zlatanvasovic:
Fixes #914. It's really confusing to have those sections in two navbars.

PR #989, opened 2016-10-23
Remove 'Security' and 'Foundation' from About nav (German localization)

lpinca:
Refs: https://github.com/nodejs/nodejs.org/pull/984


fhemberger:
LGTM!

Issue #982, opened 2016-10-18
Plans for frontpage related to 2 x LTS releases + current


phillipj:
Any thoughts on how we could improve the current UX related to displaying 2 x LTS releases, plus the current release (v7) coming soon?

Tonight v6.9.0 transitioned into LTS 🎉 That means we currently have two LTS releases, and no current. With https://github.com/nodejs/nodejs.org/pull/973 merged we now display these two releases side by side, which I assume is quite confusing and not ideal in the long run.

<img width="580" alt="screen shot 2016-10-18 at 21 11 31" src="https://cloud.githubusercontent.com/assets/1231635/19492571/78d0b89a-9577-11e6-8c91-49ecabd39d63.png">

And what should we promote when a new current release is cut? Only display v6 LTS + v7 current? v4 is still LTS thought, so it's probably still valuable for those who don't want to jump the v6 train just yet.

![s1](https://github.com/nodejs/LTS/raw/master/schedule.png)

MylesBorins:
the simplest way to move forward is to revert the change I landed when 7
drops and simply have a single Lts v6 and current v7

I've thought about this quite a bit and think it is the right move. people
who want v4 will know how to find it. we can bring the "all versions " link
further up the page and make it more noticeable

phillipj:
Going back to promoting two releases in total would at least avoid "which LTS should I choose" questions which I see as the biggest confusion.

> people who want v4 will know how to find it

Sure enough, if not, we'll probably hear from the users struggling.

mairead:
I submitted a layout for a tabbed version in the previous issue #953, which Axel had originally suggested but I'm actually thinking if you want something scalable (so 1, or any future number of releases) and something mobile friendly you'd be much better putting it all in an accordion and having the preferred version expanded on page load. Horizontal layouts like tabs are lousy on mobile especially if the number of useful releases to promote is something which changes frequently. I am struggling to follow all the conversation here though. Is there actually a consensus on this yet?

gibfahn:
@mairead I think all the opinions on here so far are against adding more than 2 releases, so I think the current website version (one LTS, one Current) should be fine.

mhdawson:
[...]

PR #990, opened 2016-10-23
Disable current release line when not yet cut

phillipj:
These changes makes the website handle periodes where a current release line has not been cut yet, automatically. As discussed in https://github.com/nodejs/nodejs.org/issues/982, we could grey out the "current" option, instead of two LTS options, in periodes where we don't have a current release line.

Rationale behind handling this automatically is to prevent us having to land a big PR, which we probably haven't tested much, asap after v7 has been cut. This would _just work_ whenever v7 is out.

The main logic around this, is found in `scripts/helpers/latestversion.js` which picks which version should be displayed as LTS and current, plus "upcoming current" introduced with these changes. That logic is based around the list of releases in https://nodejs.org/download/release/index.json. In summary the logical rules are:
- **LTS**: latest version in with `.lts === true`
- **Current**: latest version with `.lts !== true` and version number greater than the latest **LTS**
- **Upcoming current**: whenever **current** doesn't exists, calculate upcoming current to be major of LTS + 1, e.g. `v6 + 1`

Download page should probably have something similar implemented. Would appreciate any thoughts about how that might look. Tried simply greying out the background for "current" in the download matrix, but that _really_ didn't look good, so I'm skipped that part for now.

Examples of how the frontpage looks depending on the versions found in https://nodejs.org/download/release/index.json.
#### Previously when LTS and current existed

screen shot 2016-10-23 at 20 06 19
#### When no current release is found (like now)

screen shot 2016-10-23 at 20 06 34
#### As soon as current exists again

screen shot 2016-10-23 at 20 07 28
## Tasks
- [x] Frontpage
- [x] Download page
- [ ] Download page for current when current don't exist

### DIFF
--- a/scripts/helpers/latestversion.js
+++ b/scripts/helpers/latestversion.js
@@ -9,9 +9,9 @@ const map = (release) => release && {
   openssl: release.openssl
 }
 
-exports.lts = (releases) => {
-  const match = releases.find((release) => release.lts && semver.lte(release.version, '5.0.0'))
+exports.current = (releases) => {
+  const match = releases.find((release) => !release.lts && semver.gte(release.version, '5.0.0'))
   return map(match)
 }
 
-exports.current = (releases) => map(releases.find((release) => release.lts))
+exports.lts = (releases) => map(releases.find((release) => release.lts))
--- a/scripts/helpers/latestversion.js
+++ b/scripts/helpers/latestversion.js
@@ -9,7 +9,7 @@ const map = (release) => release && {
   openssl: release.openssl
 }
 
-// might we undefined if the current release line has not been cut yet
+// might be undefined if the current release line has not been cut yet
 exports.current = (releases) => {
   const ltsVersion = exports.lts(releases).node
   const match = releases.find((release) => !release.lts && semver.gte(release.version, ltsVersion))
@@ -20,10 +20,14 @@ exports.current = (releases) => {
 // has not yet been cut
 exports.upcomingCurrent = (releases) => {
   const isCurrentReleased = exports.current(releases) !== undefined
+  if (isCurrentReleased) {
+    return undefined
+  }
+
   const ltsMajor = semver.major(exports.lts(releases).node)
   const nextCurrentVersion = `v${ltsMajor + 1}.0.0`
 
-  return !isCurrentReleased ? { node: nextCurrentVersion } : undefined
+  return { node: nextCurrentVersion }
 }
 
 exports.lts = (releases) => map(releases.find((release) => release.lts))
--- a/scripts/helpers/latestversion.js
+++ b/scripts/helpers/latestversion.js
@@ -9,9 +9,21 @@ const map = (release) => release && {
   openssl: release.openssl
 }
 
+// might we undefined if the current release line has not been cut yet
 exports.current = (releases) => {
-  const match = releases.find((release) => !release.lts && semver.gte(release.version, '5.0.0'))
+  const ltsVersion = exports.lts(releases).node
+  const match = releases.find((release) => !release.lts && semver.gte(release.version, ltsVersion))
   return map(match)
+}
+
+// calculates the next upcoming major release version when the current release line
+// has not yet been cut
+exports.upcomingCurrent = (releases) => {
+  const isCurrentReleased = exports.current(releases) !== undefined
+  const ltsMajor = semver.major(exports.lts(releases).node)
+  const nextCurrentVersion = `v${ltsMajor + 1}.0.0`
+
+  return !isCurrentReleased ? { node: nextCurrentVersion } : undefined
 }
 
 exports.lts = (releases) => map(releases.find((release) => release.lts))
--- a/scripts/helpers/latestversion.js
+++ b/scripts/helpers/latestversion.js
@@ -4,6 +4,7 @@ const semver = require('semver')
 
 const map = (release) => release && {
   node: release.version,
+  nodeMajor: `v${semver.major(release.version)}.x`,
   npm: release.npm,
   v8: release.v8,
   openssl: release.openssl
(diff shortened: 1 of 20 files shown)


scripts/helpers/latestversion.js
```diff
@@ -9,9 +9,21 @@ const map = (release) => release && {
   openssl: release.openssl
 }
 
+// might we undefined if the current release line has not been cut yet
```
lpinca: Typo: we -> be

scripts/helpers/latestversion.js
```diff
@@ -9,9 +9,21 @@ const map = (release) => release && {
...
+  const ltsVersion = exports.lts(releases).node
+  const match = releases.find((release) => !release.lts && semver.gte(release.version, ltsVersion))
   return map(match)
 }
 
+// calculates the next upcoming major release version when the current release line
+// has not yet been cut
+exports.upcomingCurrent = (releases) => {
+  const isCurrentReleased = exports.current(releases) !== undefined
+  const ltsMajor = semver.major(exports.lts(releases).node)
```
lpinca: It might make sense to exit early when `isCurrentReleased` is `true`.

lpinca:
I like the idea, 👍.

MylesBorins:
This LGTM

#992 can be closed if this lands before the v7 release.

scripts/helpers/latestversion.js
```diff
@@ -9,9 +9,21 @@ const map = (release) => release && {
   openssl: release.openssl
 }
 
-exports.lts = (releases) => {
-  const match = releases.find((release) => release.lts && semver.lte(release.version, '5.0.0'))
+// might we undefined if the current release line has not been cut yet
```
ryanmurakami: 'might be' instead of 'might we'

phillipj:
Been trying a couple of different approaches on the download page, it's quite obvious I'm no designer... What do you think of these ones? I'm more than open to other alternatives here :)

#### With upcoming current

screen shot 2016-10-24 at 22 19 04

#### Both release lines present

screen shot 2016-10-24 at 22 22 41

ryanmurakami:
@phillipj Would the "with upcoming current" grey text be disabled? I like that approach.

phillipj:
> @phillipj Would the "with upcoming current" grey text be disabled? I like that approach.

Yepp, when grey it's just text, not an `<a>` at all.

lpinca:
I think it could work. I have no better ideas (I'm also not a designer).

MylesBorins:
what would be the behavior if someone visited the 'Current' url?

phillipj:
Pushed a fix needed for the navigation menu found on `/docs/` to work without current present:

screen shot 2016-10-24 at 23 18 11

phillipj:
> what would be the behavior if someone visited the 'Current' url?

@TheAlphaNerd good catch, not particularly good I'm afraid :/ That page really doesn't handle **current** not being present -- see screenshot below.

I've gotta sleep now. Feel free to suggest how the `/download/current/` page should be handled. I'll be available tomorrow from 07.00 UTC.

screen shot 2016-10-24 at 23 20 07

phillipj:
^^ seems like a showstopper for merging this today IMO. Especially since `/download/current` has been valid for almost a week, pointing to LTS v6.

Thinking out loud, two options comes to mind for the `/download/current` when current doesn't exist:
1. Display it as if the latest current version actually exists (even tho it's end of life).
2. Display an almost empty page saying Node.js currently don't have an active current release.

Afraid I won't be able to get any of the above options implemented before v7 is hopefully released today, meaning PR https://github.com/nodejs/nodejs.org/pull/992 is probably the one to be merged today.

Fishrock123:
FWIW with how this past transition went I think it is pretty safe to say we'll probably do LTS a week before the next current again next year.

fhemberger:
@phillipj Hi, what's the current status of this PR?

phillipj:
@fhemberger thanks, needed the reminder on this one 👍 

So, this got stale cause I didn't have the time to do something with the `/download/current` URL. Thought I give option 2 from https://github.com/nodejs/nodejs.org/pull/990#issuecomment-255954922 a shot:

screen shot 2017-03-28 at 21 56 21

What do you think? Would displaying a more or less blank page like that be sufficient in the short period where *current* does not exist?

gibfahn:
Maybe add some info about why there isn't a release line. Otherwise it looks like a "whoops, couldn't find that link" page (like we misplaced it 😁 ).

Maybe something like:

>This is temporary for a week or two while the Current release line transitions into LTS.

Also if `the latest LTS release` was a link that would be good.

Otherwise seems reasonable to me.

phillipj:
More info about *why* there is no Current release line and linking to the LTS download page, sounds like a good idea 👍 Little unsure about the wording you suggested:

> This is temporary for a week or two while the Current release line transitions into LTS.

I read that as the Current that existed, e.g. v5, would transition into LTS after a week or two, but in reality v5 had already moved into end of life IIRC.

Maybe displaying the release plan graphics would be appropriate? Something like this:

gibfahn:
>I read that as the Current that existed, e.g. v5, would transition into LTS after a week or two, but in reality v5 had already moved into end of life IIRC.

When Node 7 came out, Node 5 had already been EoL for several months right? The previous Current was Node 6, and the reason there was a gap is that Node 6 was transitioning from Current -> LTS. Agreed that the wording isn't exactly obvious for people who don't already understand the release process.

phillipj:
Hm, you might be right, don't remember the exact details so I looked mostly at the screenshots I took initially when opening this PR:



The most important thing is that the visitors actually *understand* what we're trying to communicate. Not being a native english speaker, I'm not confident I'll find something better than what already has been suggested though.

fhemberger:
@phillipj Hmm, it's been a while … what's the current status of this PR? What do we need to get it merged?

phillipj:
I've got too much outside of GitHub these days, family and buying/building a house, so I don't have time to follow this up and push it through I'm afraid.

If anyone sees value in this and want to get it merged, please go ahead and hijack the branch.

Closing this for now.

fhemberger:
@phillipj Alright, take care!

phillipj:
@fhemberger thanks, appreciated 👍

PR #997, opened 2016-10-27
Adding Sorocaba Node.js User Group to the Meetups list

pablos4nt0s:

PR #998, opened 2016-10-27
Update events.md

pin3da:
Related to #997


lpinca:
@pin3da can you please try to downgrade `js-yaml` to version `3.5.5` and see if the diff is still so big?
This fix https://github.com/nodeca/js-yaml/pull/274 changed the dumper output syntax, see https://github.com/nodejs/nodejs.org/pull/625#issuecomment-207315411 for reference.

pin3da:
Hi @lpinca, here the results:
- js-yaml **3.6.1**: 9,462 additions, 5,887 deletions. 
- js-yaml **3.5.5**: 7,190 additions, 4,000 deletions. 

which should I use ?

lpinca:
Ok, it seems it doesn't help much.

I would go with the latest (3.6.1) and revert https://github.com/nodejs/nodejs.org/pull/625/commits/cc65d20287c3fe9462f78223102920addf212e20 as it is no longer needed.

A lot of irrelevant events (React, Angular, etc.) are pulled in but filtering them out is not easy with Meetup API.

pablos4nt0s:
Hi @pin3da, do you want further help on it? Thank you man!

pin3da:
Hi @rectius sorry for the delay, I was waiting for feedback.

@lpinca we need something else before merging ?...  or we must to wait for #1031 ?

lpinca:
@pin3da this shouldn't be blocked by #1031 imo. I would revert cc65d20, pull the new events and merge.

fhemberger:
@pin3da @lpinca What's the status on this PR? Can you please rebase it and resolve the conflict?

pin3da:
Sorry, I made a big mistake here. I will close this and then raise another clean PR with the updated list.

PR #1022, opened 2016-11-12
Docker: Add .dockerignore file in guide

tpreusse:
Without the `.dockerignore` file the npm modules installed inside the image possibly get overwritten by your local modules.

The overwriting happens here:
```docker
COPY . /usr/src/app
```
Given you have a `node_modules` folder locally.

### DIFF
--- a/locale/en/docs/guides/nodejs-docker-webapp.md
+++ b/locale/en/docs/guides/nodejs-docker-webapp.md
@@ -145,6 +145,15 @@ EXPOSE 8080
 CMD [ "npm", "start" ]
 ```
 
+## .dockerignore file
+
+Prevent your local modules and debug log from being copied into your Docker image:
+
+```
+node_modules
+npm-debug.log
+```
+
 ## Building your image
 
 Go to the directory that has your `Dockerfile` and run the following command to
--- a/locale/en/docs/guides/nodejs-docker-webapp.md
+++ b/locale/en/docs/guides/nodejs-docker-webapp.md
@@ -147,12 +147,16 @@ CMD [ "npm", "start" ]
 
 ## .dockerignore file
 
-Prevent your local modules and debug log from being copied into your Docker image:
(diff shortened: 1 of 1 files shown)


locale/en/docs/guides/nodejs-docker-webapp.md
```diff
@@ -145,6 +145,15 @@ EXPOSE 8080
 CMD [ "npm", "start" ]
 ```
 
+## .dockerignore file
+
+Prevent your local modules and debug log from being copied into your Docker image:
```
ryanmurakami: Maybe just make this statement a little clearer, something like:
To prevent your local modules and debug logs from being copied onto your Docker image, create a `.dockerignore` file with the below contents:

Also, you should mention where this file will go.

ryanmurakami (changes requested):

ryanmurakami (approved):

ryanmurakami:
LGTM. Thanks for making that change!

tpreusse:
Thank you for reviewing.

tpreusse:
How do you handle the update of translations?
The Ukrainian version seems to be just English:
https://github.com/nodejs/nodejs.org/blob/master/locale/uk/docs/guides/nodejs-docker-webapp.md
Should I change it there too?

There is also a Korean version:
https://github.com/nodejs/nodejs.org/blob/master/locale/ko/docs/guides/nodejs-docker-webapp.md

ryanmurakami:
@tpreusse You can ping the projects here. Like this: @nodejs/nodejs-uk @nodejs/nodejs-ko

outsideris:
I'm from nodejs-ko. We will update the Korean document after this PR merged.

ghaiklor:
@denysdovhan @nodejs/nodejs-uk I think we need to update it after PR is merged as well.

ryanmurakami:
@tpreusse Thanks for the contribution!

PR #1034, opened 2016-11-28
Use the appropiate article when talking about Node

nhsz:
- 'un' instead of 'una'


outsideris:
@nodejs/nodejs-es

vrunoa:
Nice catch

a0viedo (approved):
LGTM

PR #1035, opened 2016-11-28
Change 'library' translation

nhsz:
- 'librería' is a bad translation when we are talking about code in Spanish. We use the word 'biblioteca' instead, which refers to the place were we store, catalog and go to request/consult books.

### DIFF
--- a/locale/es/index.md
+++ b/locale/es/index.md
@@ -20,4 +20,4 @@ labels:
 Node.js® es una entorno de ejecución para JavaScript construido con el [motor de JavaScript V8 de Chrome](https://developers.google.com/v8/).
 Node.js usa un modelo de operaciones E/S sin bloqueo y orientado a eventos, que lo hace liviano y eficiente.
 El ecosistema de paquetes de Node.js, [npm](https://www.npmjs.com/), es el ecosistema mas grande de
-librerías de código abierto en el mundo.
+bibliotecas de código abierto en el mundo.


zlatanvasovic (approved):

zlatanvasovic:
Agreed. *biblioteca* is an usual Spanish word for library. I've never heard of *libreria* used in that context.

pin3da:
Hi @nhquiroz in Colombia we use "librería" when we talk about code. Other countries (like Mexico, also use that word), I don't think that is a bad translation. 

If you prefer we can put both (librerías/bibliotecas de código abierto).

¡ saludos :smile: !

lpinca:
@nhquiroz this is kinda funny, we have the same words in Italian and "biblioteca" is also used to refer to places where book are stored and can be rent like a university library but when we talk about code we use "libreria".

nhsz:
Ok, maybe this becomes `undefined` :stuck_out_tongue:

I think we tend to use _librería_ in Spanish because it sounds very similar to _library_, but _librería_ means _book shop_ and that's not the intended use in the programming metaphor.

edsadr:
this basically a regional problem, but I found that many official translations use *librería*, so that's what I used, a couple of examples here (warning... obviously is Spanish): 

* https://msdn.microsoft.com/es-es/library/hh875057.aspx
* https://developer.mozilla.org/es/docs/Gecko_SDK

we can vote anyway...

nhsz:
@edsadr I agree with that.

edsadr:
@nhquiroz so ... safe to close this one?

nhsz:
Yes.

PR #1037, opened 2016-12-01
Add information that ab is ApacheBench

anirudh24seven:
I had trouble figuring out what ab referred to, as the term is not SEO-friendly. It is useful to add information that ab is ApacheBench.


fhemberger:
You are right! Thank you for your contribution.

PR #1053, opened 2016-12-06
remove Node Interactive banner from home

edsadr:
Removing Node Interactive banner, this change revert the home page state to: https://web.archive.org/web/20160316165108/https://nodejs.org/en/


lpinca:
You might also want to remove https://github.com/nodejs/nodejs.org/blob/master/layouts/css/page-modules/_interactive-banner.styl and https://github.com/nodejs/nodejs.org/blob/master/layouts/css/base.styl#L83

edsadr:
@lpinca sorry forgot about those, just updated the PR

lpinca (approved):

lpinca:
👍

lpinca:
Maybe also the relevant images under /static/images can be removed, not sure.

edsadr:
I was thinking about that too, but images could be indexed and this will generate a 404... so I would say let's keep them...

lpinca:
Yup but we already removed the old ones, like the old Amsterdam background etc. ¯\_(ツ)_/¯

edsadr:
@lpinca images removed then, still looks like a policy about deleting assets should be set

lpinca:
@edsadr I agree.

outsideris:
LGTM

PR #885, opened 2016-09-06
Respect visitor's Do-Not-Track setting

phillipj:
There's been raised some concern about our use of external web tracking/analytics service. In that discussion, there has been agreement that we at least should respect the visitor's Do-Not-Track setting in their browser.

Also moves the Google Analytics script to the bottom of the page, as tracking is _not_ top priority IMO, even though GA-scripts are async.

**SHOULD BE MERGED ON A SATURDAY**

Refs: https://github.com/nodejs/node/pull/6601#issuecomment-243876849

PR #1068, opened 2016-12-10
Fix Google Analytics tracking across the site

### LINKED ISSUE #1028: Analytics drop for blog in Oct/Nov
Something is going on with the analytics on the blog.

In Oct/Nov we show something like 60 session total, where prior we had ~50K.

The analytics on the rest of the website seem unchanged, still showing over a million sessions per month.

Anyone ideas on what might have happened?

phillipj:
With https://github.com/nodejs/nodejs.org/pull/885 site wide Google Analytics tracking got moved from the `layouts/partials/footer.hbs` -> `layouts/index.hbs`.
That was a huge mistake (by me), which made the index / frontpage be the only page left with Google Analytics tracking, all other pages lost the tracking scripts.

This fixes that blooper by moving the tracking scripts back into `layouts/partials/footer.hbs`, but still ensuring we respect our visitor's Do-Not-Track setting.

Closes https://github.com/nodejs/nodejs.org/issues/1028

/cc @mikeal

### DIFF
--- a/layouts/index.hbs
+++ b/layouts/index.hbs
@@ -73,18 +73,6 @@
 
   {{> footer className="no-margin-top" }}
   <script src="/static/js/download.js" async defer></script>
-  <script src="/static/js/dnt_helper.js"></script>
-  <script>
-    if (!_dntEnabled()) {
-      !function(n,o,d,e,j,s){n.GoogleAnalyticsObject=d;n[d]||(n[d]=function(){
-      (n[d].q=n[d].q||[]).push(arguments)});n[d].l=+new Date;j=o.createElement(e);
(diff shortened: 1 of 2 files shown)


MylesBorins:
LGTM

would it maybe make sense to put the tracker in it's own partial so it could be included in pages without the footer? If all pages always have the footer than of course that is unnecessary

MylesBorins (approved):

phillipj:
> would it maybe make sense to put the tracker in it's own partial so it could be included in pages without the footer? If all pages always have the footer than of course that is unnecessary

Thought about he same, but all the .hbs-files found in `layouts/` includes the footer partial, so concluded it wouldn't be of much value.

lpinca (approved):

Issue #1051, opened 2016-12-04
By-Laws on website are outdated


williamkapke:
We need an updated version to PR.
[The current version](https://nodejs.org/static/documents/node-foundation-by-laws.pdf) is labeled "Node.js Foundation Bylaws 06-1-15 FINAL"

I also suspect the [Membership Agreement](http://f.cl.ly/items/0N1m3x0I3S2L203M1h1r/nodejs-foundation-membership-agreement-2015-march-04.pdf) is also since it is dated 2015-march-04.

/cc @rvagg @hackygolucky @mikeal @nodejs/tsc

williamkapke:
Closing in favor of: https://github.com/nodejs/board/issues/53

PR #1071, opened 2016-12-12
Updated bylaws

mikeal:
The board voted last week to guarantee a Gold Level director if there are Gold Members.

This PR has the new bylaws. I'm keeping it in docx because that's how our lawyer manages it and it will be easier to reconcile and find diffs than exporting it to PDF before we publish it.


ryanmurakami (approved):
LGTM

williamkapke:
How about we offer a link in both formats then? It is far easier to pop it open in the browser (PDF) than download it and open it.

Truly- I'd rather things were converted to markdown and committed to git for history diffs.
... which is what I'll be doing to diff this anyhow.

mikeal:
@williamkapke If we could get the lawyers to use GitHub I'd be the happiest person in the world. Unfortunately I don't see that happening any time soon :(

Fishrock123:
I'd prefer to also keep a `.pdf`. Docx seems less accessible.

williamkapke:
This will fix #1051

fhemberger:
-1 on DOCX.

Please provide a PDF version as well. We can check in both formats, but the PDF should be linked on the page.

fhemberger (changes requested):
-1 on DOCX.

Please provide a PDF version as well. We can check in both formats, but the PDF should be linked on the page.

fhemberger (approved):

fhemberger:
@williamkapke @mikeal I converted the .docx to .pdf - if you're okay with the change, this PR can be merged.

williamkapke (changes requested):
These need more updates.

williamkapke:
Sorry- I forgot this PR was open still.

See: https://github.com/nodejs/board/issues/53

ghost:
I'd prefer to also keep a .pdf.

williamkapke:
A markdown version now exists that is identical to the PDF version currently linked on the website. PRs have been opened to bring this MD version up-to-date.

https://github.com/nodejs/board/blob/master/by-laws.md

Ya'll can consider if we want to point to that instead of a PDF version.

mikeal:
> Ya'll can consider if we want to point to that instead of a PDF version.

Yes Please!!!

fhemberger:
Should we link directly to https://github.com/nodejs/board/blob/master/by-laws.md to avoid having two copies, or add the Markdown file to the website?

williamkapke:
@fhemberger I think we should just link to the markdown file in the board repo.

fhemberger:
@williamkapke @mikeal Updated the link, I think we're good to go.

williamkapke (approved):
Excellent!

amandeepmittal (approved):

Issue #845, opened 2016-08-02
remove advisory-board (and others)


ghost:
during translation efforts, we decided that the `advisory-board` documents were not needed anymore and can be removed. do we have any other documents like this that can be removed?

ghost:
FYI `advisory-board` is an artifact from io.js times (?) and is superseded by the foundation's board

Fishrock123:
I think it can be removed, cc @mikeal

mikeal:
Yup, can be removed.

lpinca:
Hmm there are links in some blog posts that point to https://nodejs.org/about/advisory-board/.
If we remove it the links will 404.

fhemberger:
@lpinca Then let's remove the links from the blog posts.

PR #1076, opened 2016-12-17
Remove "Advisory Board" page

lpinca:
This removes the "Advisory Board" page.

https://nodejs.org/about/advisory-board/ is redirected to https://nodejs.org/en/about/organization/
so existing links should work but the "organization" page is not present on the left navigation menu.

I think the item should be added, and the "organization" page updated.

Refs: https://github.com/nodejs/nodejs.org/issues/845


lpinca:
Or maybe the "organization" page should also be removed and https://nodejs.org/about/advisory-board/ should redirect to another document.

fhemberger:
@lpinca Updated your PR, 'about/organization' should be removed as well (superseded by the TSC page).

Added a PR to have the redirects changed.

fhemberger (approved):

lpinca:
Perfect!

PR #1082, opened 2016-12-25
Add weekly update 2016-12-25

JungMinu:
Please review this post. ❤️,@nodejs/evangelism


MylesBorins:
aren't we supposed to wait 36 hours before landing content like this?

JungMinu:
@TheAlphaNerd Ah, sorry for that. I will keep that in mind.

thefourtheye:
Also, wouldn't it be better if we get some reviews before merging?

PR #1085, opened 2017-01-02
Use one stylesheet for all locales. Fixes #985

### LINKED ISSUE #985: Use one stylesheet for all locales

diagramatics:
I compiled the styles to a new directory called `/layouts`, which I don't think makes sense. Perhaps now that we have one stylesheet for all versions we might as well just move all Stylus files into `/static/css` instead and reference there?

Also, new to Metalsmith, so hopefully this PR is not riddled with questionable code :P


lpinca (approved):

hiroppy (approved):

phillipj:
Awesome 👍

PR #1101, opened 2017-01-13
Fix typo

piperchester:


lpinca:
Not a native speaker but I think the original is also ok?
Anyway I'm fine with the change.

ryanmurakami:
The original was not correct grammatically, but using a semi colon instead just makes the sentence kind of difficult to read. Maybe a transition word like 'and' could be added to make it smoother?

piperchester:
Reading it over, I feel like even a period would be best for flow. Thoughts?

ryanmurakami:
@piperchester Yeah, I think that would be better.

piperchester:
@ryanmurakami made the change 😄

lpinca (approved):

ryanmurakami:
@piperchester Awesome! Thanks!

PR #1107, opened 2017-01-21
add `google-site-verification` meta tag

williamkapke:
To enable us to link from one of the YouTube videos to the website- we need to make get the nodejs.org site approved.

Adding this meta tag will do the trick!


williamkapke:
disregard- google is giving me the wrong id ... will need to figure this out later.

PR #1109, opened 2017-01-23
Add guide: "Backpressuring in Streams"

jalafel:
This is was an issue opened on the nodejs/node repository before guides were transferred directly to node/nodejs.org.

This PR contains a first go at a guide on backpressures, and missing some vital points that I do not feel equipped to answer. I have left some empty spaces for them! I would appreciate anyone's input on the content, structure, and any additions to the guide.

Some points to expand on would be the problems that streams pose which are solved by backpressure. I think I hit most of the points, but I don't think I'm versed enough to go into detail.

There is also space for an example app, but again, I am not familiar on the current best practices on building streams and for implementing your own backpressure. 

Thanks!

Refs: https://github.com/nodejs/node/issues/10766

### DIFF
--- a/locale/en/docs/guides/backpressuring-in-streams.md
+++ b/locale/en/docs/guides/backpressuring-in-streams.md
@@ -1,0 +1,264 @@
+---
+title: Backpressuring in Streams
+layout: docs.hbs
+---
+
+# Backpressuring in Streams
+ 
+The purpose of this guide will describe what backpressure is and the 
+problems in Streams that it solves. The first part of this guide will cover 
+these questions. The second part of the guide will introduce suggested best 
+practices to ensure your application's code is safe and optimized.
+
+We assume a little familiarity with the general definition of 
+[`backpressure`][], [`Buffer`][] and [`EventEmitters`][] in Node.js, as well as 
+some experience with [`Streams`][]. If you haven't read through those docs, 
+it's not a bad idea to take a look at the API documentation first, as it will 
+help expand your understanding while reading this guide and for following along 
+with examples.
+
+## The Problem with Streams
+
+In a computer system, data is transferred from one process to another through 
+pipes and signals. In Node.js, we find a similar mechanism and called 
+[`Streams`][]. Streams are great! They do so much for Node.js and almost every
+part of the internal codebase utilizes the `Streams` class. As a developer, you 
+are more than encouraged to use them too!
+
+```javascript
+const readline = require('readline');
+
+// process.stdin and process.stdout are both instances of Streams
+const rl = readline.createInterface({
+  input: process.stdin,
+  output: process.stdout
+});
+
+rl.question('Why should you use streams? ', (answer) => {
+  
+  console.log(`Maybe it's ${answer}, maybe it's because they are awesome! :)`);
+  
+  rl.close();
+});
+```
+
+
+There are different functions to transfer data from one process to another. In 
+Node.js, there is an internal built-in function called [`.pipe()`][]. There are
+[other packages][] out there you can use too! Ultimately though, at the basic 
+level of this process, we have two separate components: the _source_ of the 
+data and the _consumer_.
+
+In Node.js the source is a [`Readable`][] stream and the consumer is the 
+[`Writable`][] stream (both of these may be interchanged with a [`Duplex`][] 
+stream, but that is out-of-scope for this guide).
+
+## Too Much Data, Too Quickly
+
+There are instance where a [`Readable`][] stream might give data to the
+[`Writable`][] much too quickly --- much more than the consumer can handle! 
+
+When that occurs, the consumer will begin to queue all the chunks of data for 
+later consumption. The write queue will get longer and longer, and because of 
+this more data must be kept in memory until the entire process has completed.
+
+```javascript
+  var fs = require('fs');
+
+  var inputFile  = fs.createReadStream('REALLY_BIG_FILE.x');
+  var outputFile = fs.createWriteStream('REALLY_BIG_FILE_DEST.x');
+
+  // Secretly the stream is saying: "whoa, whoa! hang on, this is way too much!"
+  inputFile.pipe(outputFile);
+```
+
+
+This is why backpressure is important. If a backpressure system was not 
+present, the process would use up your system's memory, effectively slowing 
+down other processes, and monopolizing a large part of your system until 
+completion.
+
+This results in a few things:
+
+* Memory exhaustion
+* A very overworked garbage collector
+* Slowing down all other current processes
+
+## Memory Exhaustion
+
+
+
+## Garbage Collection
+
+
+
+## Overall Dip in System Performace
+
+
+
+## How Does Backpressure Resolve These Issues?
+
+The moment that backpressure is triggered can be narrowed exactly to the return 
+value of a `Writable`'s [`.write()`][] function. This return value is 
+determined by a few conditions, of course.
+
+In any scenario where the data buffer has exceeded the [`highWaterMark`][] or 
+the write queue is currently busy, [`.write()`][] will return `false`. 
+
+When a `false` value is returned, the backpressure system kicks in. It will 
+pause the incoming [`Readable`][] stream from sending any data and wait until
+the consumer is ready again.
+
+Once the the queue is finished, backpressure will allow data to be sent again.
+The space in memory that was being used will free itself up and prepare for the 
+next glob of data. 
+
+This effectively allows an fixed amount of memory to be used at any given
+time for a [`.pipe()`][] function. There will be no memory leakage, no 
+indefinite buffering, and the garbage collector will only have to deal with
+one area in memory!
+
+So, if backpressure is so important, why have you (probably) not heard of it? 
+Well the answer is simple: Node.js does all of this automatically for you. 
+
+That's so great! But also not so great when we are trying to understand how to 
+implement our own custom streams.
+
+## Lifecycle of `.pipe()`
+
+To achieve a better understanding of backpressure, here is a flow-chart on the 
+lifecycle of a [`Readable`][] stream being [piped][] into a [`Writable`][] 
+stream: 
+
+```javascript
++===============+
+|   Your_Data   |
++=======+=======+
+        |
++-------v-----------+          +-------------------+         +=================+
+|  Readable Stream  |          |  Writable Stream  +--------->  .write(chunk)  |
++-------+-----------+          +---------^---------+         +=======+=========+
+        |                                |                           |
+        |     +======================+   |        +------------------v---------+
+        +----->  .pipe(destination)  >---+        |    Is this chunk too big?  |
+              +==^=======^========^==+            |    Is the queue busy?      |
+                 ^       ^        ^               +----------+-------------+---+
+                 |       |        |                          |             |
+                 |       |        |  > if (!chunk)           |             |
+                 ^       |        |      emit .end();        |             |
+                 ^       ^        |  > else                  |             |
+                 |       ^        |      emit .write();  +---v---+     +---v---+
+                 |       |        ^----^-----------------<  No   |     |  Yes  |
+                 ^       |                               +-------+     +---v---+
+                 ^       |                                                 |
+                 |       ^   emit .pause();        +=================+     |
+                 |       ^---^---------------------+  return false;  <-----+---+
+                 |                                 +=================+         |
+                 |                                                             |
+                 ^   when queue is empty   +============+                      |
+                 ^---^-----------------^---<  Buffering |                      |
+                     |                     |============|                      |
+                     +> emit .drain();     |  <Buffer>  |                      |
+                     +> emit .resume();    +------------+                      |
+                                           |  <Buffer>  |                      |
+                                           +------------+  add chunk to queue  |
+                                           |            <--^-------------------<
+                                           +============+
+```
+
+
+_Note_: The `.pipe()` function is typically where backpressure is invoked. In 
+an isolate application, both [`Readable`][] and [`Writable`][] streams 
+should be present. If you are writing an application meant to accept a 
+[`Readable`][] stream, or pipes to a [`Writable`][] stream from another app, 
+you may omit this detail.
+
+## Example App
+
+Since [Node.js v0.10][], the [`Streams`][] class has offered the ability to 
+overwrite the functionality of [`.read()`][] or [`.write()`][] by using the
+underscore version of these respective functions ([`._read()`][] and 
+[`._write()`][]).
+
+There are guidelines documented for [implementing Readable streams][] and 
+[implementing Writable streams][]. We will assume you've read these over. 
+
+This application will do something very simple: take the data source and 
+perform something fun to the data, and then write it to another file.
+
+## Rules to Abide For Writing Custom Streams 
+
+Recall that a [`.write()`][] may return true or false dependent on some 
+conditions. Thus, when building our own [`Writable`][] stream, we must pay 
+close attention those conditions and the return value:
+
+* If the write queue is busy, [`.write()`][] will return false.
+* If the data chunk is too large, [`.write()`][] will return false (the limit 
+is indicated by the variable, [`highWaterMark`][]).
+
+_Note_: In most machines, there is a byte size that is determines when a buffer 
+is full (which will vary across different machines). Node.js allows you to set 
+your own custom [`highWaterMark`][], but commonly, the default is the optimal 
+value for what system is running the application. In instances where you might 
+want to raise that value, go for it, but do so with caution!  
+
+## Build a WritableStream
+
+Let's extend the prototypical function of [`.write()`][]: 
+
+```javascript
+const Writable = require('stream').Writable;
+
+class MyWritable extends Writable {
+  constructor(options) {
+    super(options);
+  }
+
+  _write(chunk, encoding, callback) {
+    if (chunk.toString().indexOf('a') >= 0) {
+      callback(new Error('chunk is invalid'));
+    } else {
+      callback();
+    }
+  }
+}
+``` 
+
+
+## Conclusion  
+
+Streams are a often used module in Node.js. They are important to the internal 
+structure, and for develoeprs, to expand and connect across the Node.js modules 
+ecosystem. 
+
+Hopefully, you will now be able to troubleshoot, safely code your own 
+[`Writable`][] streams with backpressure in mind, and share your knowledge with 
+colleagues and friends. 
+
+Be sure to read up more on [`Streams`][] for other [`EventEmitters`][] to help 
+improve your knowledge when building applications with Node.js.
+
+
+
+[`Streams`]: https://nodejs.org/api/stream.html
+[`Buffer`]: https://nodejs.org/api/buffer.html
+[`EventEmitters`]: https://nodejs.org/api/events.html
+[`Writable`]: https://nodejs.org/api/stream.html#stream_writable_streams
+[`Readable`]: https://nodejs.org/api/stream.html#stream_readable_streams
+[`Duplex`]: https://nodejs.org/api/stream.html#stream_duplex_and_transform_streams
+[`.drain()`]: https://nodejs.org/api/stream.html#stream_event_drain
+[`.read()`]: https://nodejs.org/docs/latest/api/stream.html#stream_readable_read_size
+[`.write()`]: https://nodejs.org/api/stream.html#stream_writable_write_chunk_encoding_callback
+[`._read()`]: https://nodejs.org/docs/latest/api/stream.html#stream_readable_read_size_1
+[`._write()`]: https://nodejs.org/docs/latest/api/stream.html#stream_writable_write_chunk_encoding_callback_1
+
+[implementing Writable streams]: https://nodejs.org/docs/latest/api/stream.html#stream_implementing_a_writable_stream
+[implementing Readable streams]: https://nodejs.org/docs/latest/api/stream.html#stream_implementing_a_readable_stream
+
+[other packages]: https://github.com/sindresorhus/awesome-nodejs#streams
+[`backpressure`]: https://en.wikipedia.org/wiki/Back_pressure#Back_pressure_in_information_technology
+[Node.js v0.10]: https://nodejs.org/docs/v0.10.0/
+[`highWaterMark`]: https://nodejs.org/api/stream.html#stream_buffering
+
+[`.pipe()`]: https://nodejs.org/docs/latest/api/stream.html#stream_readable_pipe_destination_options
+[piped]: https://nodejs.org/docs/latest/api/stream.html#stream_readable_pipe_destination_options
--- a/locale/en/docs/guides/backpressuring-in-streams.md
+++ b/locale/en/docs/guides/backpressuring-in-streams.md
@@ -298,6 +298,12 @@ Well the answer is simple: Node.js does all of this automatically for you.
 That's so great! But also not so great when we are trying to understand how to
 implement our own custom streams.
 
+Note: In most machines, there is a byte size that is determines when a buffer
+is full (which will vary across different machines). Node.js allows you to set
+your own custom [`highWaterMark`][], but commonly, the default is the optimal
+value for what system is running the application. In instances where you might
+want to raise that value, go for it, but do so with caution!
+
 ## Lifecycle of `.pipe()`
 
 To achieve a better understanding of backpressure, here is a flow-chart on the
@@ -476,13 +482,42 @@ class MyWritable extends Writable {
     callback();
 ```
 
+There are also some things to look out for when implementing [`._writev()`][].
+The function is coupled with [`.cork()`][], but there is a common mistake when
+writing:
+```javascript
+  // Using .uncork() twice here makes two calls on the C++ layer, rendering the 
+  // cork/uncork technique useless.
+  ws.cork()
+  ws.write('hello ')
+  ws.write('world ')
+  ws.uncork()
 
-Note: In most machines, there is a byte size that is determines when a buffer
-is full (which will vary across different machines). Node.js allows you to set
-your own custom [`highWaterMark`][], but commonly, the default is the optimal
-value for what system is running the application. In instances where you might
-want to raise that value, go for it, but do so with caution!
+  ws.cork()
+  ws.write('from ')
+  ws.write('Matteo')
+  ws.uncork()
 
+  // The correct way to write this is to utilize process.nextTick(), which fires
+  // on the next event loop.
+  ws.cork()
+  ws.write('hello ')
+  ws.write('world ')
+  process.nextTick(doUncork, ws)
+
+  ws.cork()
+  ws.write('from ')
+  ws.write('Matteo')
+  process.nextTick(doUncork, ws)
+
+  // as a global function
+  function doUncork (stream) {
+    stream.uncork()
+  }
+```
+
+[`.cork()`][] can be called as many times we want, we just need to be careful to 
+call [`.uncork()`][] the same amount of times to make it flow again.
 
 ## Conclusion
 
@@ -513,6 +548,10 @@ Node.js.
 [`.write()`]: https://nodejs.org/api/stream.html#stream_writable_write_chunk_encoding_callback
 [`._read()`]: https://nodejs.org/docs/latest/api/stream.html#stream_readable_read_size_1
 [`._write()`]: https://nodejs.org/docs/latest/api/stream.html#stream_writable_write_chunk_encoding_callback_1
+[`._writev()`]: https://nodejs.org/api/stream.html#stream_writable_writev_chunks_callback
+[`.cork()`]: https://nodejs.org/api/stream.html#stream_writable_cork
+[`.uncork()`]: https://nodejs.org/api/stream.html#stream_writable_uncork
+
 [push method]: https://nodejs.org/docs/latest/api/stream.html#stream_readable_push_chunk_encoding
 
 [implementing Writable streams]: https://nodejs.org/docs/latest/api/stream.html#stream_implementing_a_writable_stream
--- a/locale/en/docs/guides/backpressuring-in-streams.md
+++ b/locale/en/docs/guides/backpressuring-in-streams.md
@@ -300,8 +300,8 @@ implement our own custom streams.
 
 Note: In most machines, there is a byte size that is determines when a buffer
 is full (which will vary across different machines). Node.js allows you to set
-your own custom [`highWaterMark`][], but commonly, the default is the optimal
-value for what system is running the application. In instances where you might
+your own custom [`highWaterMark`][], but commonly, the default is set to 16kb
+(16384, or 16 for objectMode streams). In instances where you might
 want to raise that value, go for it, but do so with caution!
 
 ## Lifecycle of `.pipe()`
--- a/locale/en/docs/guides/backpressuring-in-streams.md
+++ b/locale/en/docs/guides/backpressuring-in-streams.md
@@ -4,25 +4,37 @@ layout: docs.hbs
 ---
 
 # Backpressuring in Streams
- 
-The purpose of this guide will describe what backpressure is and the 
-problems in Streams that it solves. The first part of this guide will cover 
-these questions. The second part of the guide will introduce suggested best 
-practices to ensure your application's code is safe and optimized.
 
-We assume a little familiarity with the general definition of 
-[`backpressure`][], [`Buffer`][] and [`EventEmitters`][] in Node.js, as well as 
-some experience with [`Streams`][]. If you haven't read through those docs, 
-it's not a bad idea to take a look at the API documentation first, as it will 
-help expand your understanding while reading this guide and for following along 
+There is a general problem that arises during data handling called
+[`backpressure`][]. The term is used to describe a build up of data behind a 
+buffer during data transfer. When the recieving end of the transfer has complex
+operations, or is slower for whatever reason, there is a tendency for data from 
+the incoming source to begin to accumulate.
+
+To solve this problem, there must be a delegation system in place to ensure a
+smooth flow of data from one source to another. Different communities have
+resolved this issue uniquely to their programs, Unix pipes and TCP sockets are
+good examples of this, and often times is referred to as _flow control_. In
+Node.js, streams have been the adopted solution.
+
+The purpose of this guide is to further detail what backpressure is, and how
+exactly streams address this in Node.js' source code. The second part of
+the guide will introduce suggested best practices to ensure your application's
+code is safe and optimized when implementing streams.
+
+We assume a little familiarity with the general definition of
+[`backpressure`][], [`Buffer`][], and [`EventEmitters`][] in Node.js, as well as
+some experience with [`Streams`][]. If you haven't read through those docs,
+it's not a bad idea to take a look at the API documentation first, as it will
+help expand your understanding while reading this guide and for following along
 with examples.
 
-## The Problem with Streams
+## The Problem With Data Handling
 
-In a computer system, data is transferred from one process to another through 
-pipes and signals. In Node.js, we find a similar mechanism and called 
+In a computer system, data is transferred from one process to another through
+pipes, sockets, and signals. In Node.js, we find a similar mechanism called
 [`Streams`][]. Streams are great! They do so much for Node.js and almost every
-part of the internal codebase utilizes the `Streams` class. As a developer, you 
+part of the internal codebase utilizes that module. As a developer, you
 are more than encouraged to use them too!
 
 ```javascript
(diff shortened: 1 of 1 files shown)


lpinca:
cc: @nodejs/streams

locale/en/docs/guides/backpressuring-in-streams.md
```diff
@@ -0,0 +1,264 @@
...
+
+
+There are different functions to transfer data from one process to another. In 
+Node.js, there is an internal built-in function called [`.pipe()`][]. There are
+[other packages][] out there you can use too! Ultimately though, at the basic 
+level of this process, we have two separate components: the _source_ of the 
+data and the _consumer_.
+
+In Node.js the source is a [`Readable`][] stream and the consumer is the 
+[`Writable`][] stream (both of these may be interchanged with a [`Duplex`][]
```
mcollina: or a `Transform`

locale/en/docs/guides/backpressuring-in-streams.md
```diff
@@ -0,0 +1,264 @@
...
+practices to ensure your application's code is safe and optimized.
+
+We assume a little familiarity with the general definition of 
+[`backpressure`][], [`Buffer`][] and [`EventEmitters`][] in Node.js, as well as 
+some experience with [`Streams`][]. If you haven't read through those docs, 
+it's not a bad idea to take a look at the API documentation first, as it will 
+help expand your understanding while reading this guide and for following along 
+with examples.
+
+## The Problem with Streams
```
mcollina: There is not a problem with streams. There is a problem that streams solve, which is backpressure, also called flow control in other communities. There are other solutions to backpressure. Unix pipes and TCP sockets solves this as well, and are the root of the following. Also, streams can be used for pull-based data handling.

The usual demo I do to show the problem is zipping a very large data file both in one go and with streams. As the first one crashes, the second one can work with data of any size. (the original credit goes to @davidmarkclements for that example). That shows the problem.

locale/en/docs/guides/backpressuring-in-streams.md
```diff
@@ -0,0 +1,264 @@
...
+                 ^   when queue is empty   +============+                      |
+                 ^---^-----------------^---<  Buffering |                      |
+                     |                     |============|                      |
+                     +> emit .drain();     |  <Buffer>  |                      |
+                     +> emit .resume();    +------------+                      |
+                                           |  <Buffer>  |                      |
+                                           +------------+  add chunk to queue  |
+                                           |            <--^-------------------<
+                                           +============+
+```
```
mcollina: This diagram is not exactly clear. Why everything is going back to  `pipe()`?

locale/en/docs/guides/backpressuring-in-streams.md
```diff
@@ -0,0 +1,264 @@
...
+
+## Rules to Abide For Writing Custom Streams 
+
+Recall that a [`.write()`][] may return true or false dependent on some 
+conditions. Thus, when building our own [`Writable`][] stream, we must pay 
+close attention those conditions and the return value:
+
+* If the write queue is busy, [`.write()`][] will return false.
+* If the data chunk is too large, [`.write()`][] will return false (the limit 
+is indicated by the variable, [`highWaterMark`][]).
```
mcollina: If we are _building_ a `Writable` directly, this is not needed, as it is handled by the stream state machine. However, it is needed to **use** a `Writable`  directly, see my example in the Node docs: https://github.com/nodejs/node/blob/master/doc/api/stream.md#writablewritechunk-encoding-callback

mcollina (changes requested):
The work is solid, I think the intro should be changed to explain what is the problem that streams solve.

locale/en/docs/guides/backpressuring-in-streams.md
jalafel: Thanks for this review @mcollina. So essentially there is not separation from streams & backpressure? As in ... data-handling is tricky, but streams solves it with backpressure?

locale/en/docs/guides/backpressuring-in-streams.md
jalafel: Good question. I assumed every event emitter was being found back into pipe, and then pipe would delegate what would do next. Is that incorrect?

locale/en/docs/guides/backpressuring-in-streams.md
mcollina: As in "data handling is tricky, and streams is the solution that Node has adopted".

locale/en/docs/guides/backpressuring-in-streams.md
mcollina: @jessicaquynh it is incorrect. `pipe` sets up some closures, who would keep working, but the method is not invoked anymore. The diagram should represent what `pipe` set up, but not show `pipe` at all: https://github.com/nodejs/node/blob/master/lib/_stream_readable.js#L473-L613

eljefedelrodeodeljefe:
@jessicaquynh can you incorporate Matteo's comments?

jalafel:
@eljefedelrodeodeljefe yes definitely! sorry I've been mia on this issue .. will submit changes tomorrow night!

locale/en/docs/guides/backpressuring-in-streams.md
```diff
@@ -0,0 +1,264 @@
...
+In any scenario where the data buffer has exceeded the [`highWaterMark`][] or 
+the write queue is currently busy, [`.write()`][] will return `false`. 
+
+When a `false` value is returned, the backpressure system kicks in. It will 
+pause the incoming [`Readable`][] stream from sending any data and wait until
+the consumer is ready again.
+
+Once the the queue is finished, backpressure will allow data to be sent again.
+The space in memory that was being used will free itself up and prepare for the 
+next glob of data.
```
TimothyGu: glob → blob / batch

locale/en/docs/guides/backpressuring-in-streams.md
```diff
@@ -0,0 +1,264 @@
+---
+title: Backpressuring in Streams
+layout: docs.hbs
+---
+
+# Backpressuring in Streams
+ 
+The purpose of this guide will describe what backpressure is and the
```
TimothyGu: will → is to

locale/en/docs/guides/backpressuring-in-streams.md
```diff
@@ -0,0 +1,264 @@
...
+pause the incoming [`Readable`][] stream from sending any data and wait until
+the consumer is ready again.
+
+Once the the queue is finished, backpressure will allow data to be sent again.
+The space in memory that was being used will free itself up and prepare for the 
+next glob of data. 
+
+This effectively allows an fixed amount of memory to be used at any given
+time for a [`.pipe()`][] function. There will be no memory leakage, no 
+indefinite buffering, and the garbage collector will only have to deal with
```
TimothyGu: indefinite → infinite

locale/en/docs/guides/backpressuring-in-streams.md
```diff
@@ -0,0 +1,264 @@
...
+
+_Note_: In most machines, there is a byte size that is determines when a buffer 
+is full (which will vary across different machines). Node.js allows you to set 
+your own custom [`highWaterMark`][], but commonly, the default is the optimal 
+value for what system is running the application. In instances where you might 
+want to raise that value, go for it, but do so with caution!  
+
+## Build a WritableStream
+
+Let's extend the prototypical function of [`.write()`][]:
```
TimothyGu: We are not extending `.write()` here, but extending `stream.Writable` with a custom `._write()` method, which is then _used_ internally by `.write()`.

locale/en/docs/guides/backpressuring-in-streams.md
```diff
@@ -0,0 +1,264 @@
...
+* If the data chunk is too large, [`.write()`][] will return false (the limit 
+is indicated by the variable, [`highWaterMark`][]).
+
+_Note_: In most machines, there is a byte size that is determines when a buffer 
+is full (which will vary across different machines). Node.js allows you to set 
+your own custom [`highWaterMark`][], but commonly, the default is the optimal 
+value for what system is running the application. In instances where you might 
+want to raise that value, go for it, but do so with caution!  
+
+## Build a WritableStream
```
TimothyGu: "Writable Stream" (note the space) for consistency

locale/en/docs/guides/backpressuring-in-streams.md
```diff
@@ -0,0 +1,264 @@
...
+
+## Lifecycle of `.pipe()`
+
+To achieve a better understanding of backpressure, here is a flow-chart on the 
+lifecycle of a [`Readable`][] stream being [piped][] into a [`Writable`][] 
+stream: 
+
+```javascript
++===============+
+|   Your_Data   |
```
TimothyGu: No need for `_` here

locale/en/docs/guides/backpressuring-in-streams.md
```diff
@@ -0,0 +1,264 @@
...
+[`backpressure`][], [`Buffer`][] and [`EventEmitters`][] in Node.js, as well as 
+some experience with [`Streams`][]. If you haven't read through those docs, 
+it's not a bad idea to take a look at the API documentation first, as it will 
+help expand your understanding while reading this guide and for following along 
+with examples.
+
+## The Problem with Streams
+
+In a computer system, data is transferred from one process to another through 
+pipes and signals. In Node.js, we find a similar mechanism and called
```
TimothyGu: maybe also mention sockets?

redundant "and"

locale/en/docs/guides/backpressuring-in-streams.md
```diff
@@ -0,0 +1,264 @@
...
+help expand your understanding while reading this guide and for following along 
+with examples.
+
+## The Problem with Streams
+
+In a computer system, data is transferred from one process to another through 
+pipes and signals. In Node.js, we find a similar mechanism and called 
+[`Streams`][]. Streams are great! They do so much for Node.js and almost every
+part of the internal codebase utilizes the `Streams` class. As a developer, you 
+are more than encouraged to use them too!
```
TimothyGu: `Streams` isn't really a thing. There is no `Streams` module or `Streams` class. Make it clear what exactly you are referring to for each instance of `[Streams][]`:

- the `stream` module
- streams as a concept
- `Readable` or `Writable` classes

For this paragraph, I'd say something like this:

```markdown
In Node.js, we find a similar mechanism called streams in the [`stream`
module][]. Streams are great! They do so much for Node.js and almost every part
of the internal codebase utilizes that module.
```

Ditto below.

locale/en/docs/guides/backpressuring-in-streams.md
```diff
@@ -0,0 +1,264 @@
...
+
+When that occurs, the consumer will begin to queue all the chunks of data for 
+later consumption. The write queue will get longer and longer, and because of 
+this more data must be kept in memory until the entire process has completed.
+
+```javascript
+  var fs = require('fs');
+
+  var inputFile  = fs.createReadStream('REALLY_BIG_FILE.x');
+  var outputFile = fs.createWriteStream('REALLY_BIG_FILE_DEST.x');
```
TimothyGu: `const`

locale/en/docs/guides/backpressuring-in-streams.md
```diff
@@ -0,0 +1,264 @@
...
+
+  // Secretly the stream is saying: "whoa, whoa! hang on, this is way too much!"
```
TimothyGu: It's not immediately clear that writing to the disk is slower than reading. Either mention it explicitly ("on a computer where writing is slower than reading"), or use a slower writable stream (http? zlib?)

locale/en/docs/guides/backpressuring-in-streams.md
```diff
@@ -0,0 +1,264 @@
...
+The space in memory that was being used will free itself up and prepare for the 
+next glob of data. 
+
+This effectively allows an fixed amount of memory to be used at any given
+time for a [`.pipe()`][] function. There will be no memory leakage, no 
+indefinite buffering, and the garbage collector will only have to deal with
+one area in memory!
+
+So, if backpressure is so important, why have you (probably) not heard of it? 
+Well the answer is simple: Node.js does all of this automatically for you.
```
TimothyGu: Make it be known somewhere that only `.pipe()` does this automatically (`readable.on('data', buf => writable.write(buf));` doesn't use backpressure, for example). I know in the next section you are looking more closely at `.pipe()`, but it seems to be assumed that the reader knows all this is talking about `.pipe()`.

locale/en/docs/guides/backpressuring-in-streams.md
```diff
@@ -0,0 +1,264 @@
...
+                                           |            <--^-------------------<
+                                           +============+
+```
+
+
+_Note_: The `.pipe()` function is typically where backpressure is invoked. In 
+an isolate application, both [`Readable`][] and [`Writable`][] streams 
+should be present. If you are writing an application meant to accept a 
+[`Readable`][] stream, or pipes to a [`Writable`][] stream from another app, 
+you may omit this detail.
```
TimothyGu: Not sure what "application" / "app" means here. A standalone program? A library? A function?

locale/en/docs/guides/backpressuring-in-streams.md
```diff
@@ -0,0 +1,264 @@
...
+
+## Example App
+
+Since [Node.js v0.10][], the [`Streams`][] class has offered the ability to 
+overwrite the functionality of [`.read()`][] or [`.write()`][] by using the
```
TimothyGu: I wouldn't say "overwrite", but rather "modify the behavior"

TimothyGu (commented):
Nice work, and I'm excited to read the currently unfinished sections.

One thing I noticed is the emphasis on writable streams. Writing readable streams that stop when asked to do so is just as important as teaching the writable stream to say stop when needed IMO.

locale/en/docs/guides/backpressuring-in-streams.md
mcollina: @TimothyGu we have `Stream` base class, from which each of the other inherits.

locale/en/docs/guides/backpressuring-in-streams.md
mcollina: Writing to the disk is slower than reading from it. It might not be clear for a reader though, so I'm 👍  in adding a zlib compression before writing down.

mcollina:
> One thing I noticed is the emphasis on writable streams. Writing readable streams that stop when asked to do so is just as important as teaching the writable stream to say stop when needed IMO.

As we have a backpressure (or flow-control) mechanism in place, the maximum amount of data that can flow is controlled by downstream, hence the focus on `Writable` is correct. However this is not explicit. We should also note that calling `readable.push` whenever data is ready is as bad as `readable.on('data', data => writable.write(data))`.

locale/en/docs/guides/backpressuring-in-streams.md
jalafel: Syntactically, I am still a bit confused. Wikipedia describes backpressure in IT as

> the build-up of data behind an I/O switch if the buffers are full and incapable of receiving any more data; the transmitting device halts the sending of data packets until the buffers have been emptied and are once more capable of storing information. 

In this instance .. is backpressure the problem or the I/O switch that streams implements?

locale/en/docs/guides/backpressuring-in-streams.md
jalafel: ^ This is something I'd like to clarify. Is it true to say we have streams as a few cases: 

* Streams base class (which has the standard functions in which all subclass streams share)
* Streams module (which includes Readable, Writable, etc) 
* streams as a concept (which is, in the scope of this guide, the way in which data flow is delegated)

locale/en/docs/guides/backpressuring-in-streams.md
addaleax: > In this instance .. is backpressure the problem or the I/O switch that streams implements?

Backpressure itself is the problem, and the backpressure handling mechanisms in the `streams` classes are our solution to it. :)

jalafel:
Sorry it took so long to update the guide. Hope I interpreted the change requests correctly. Ultimately, I am still not 100% what constitutes best practice for building streams. 

Would it be fair to say that best practice is also non-contradictory practice? As in, "you have a good stream if you did not do A, B, C"?

mcollina:
> Sorry it took so long to update the guide. Hope I interpreted the change requests correctly. Ultimately, I am still not 100% what constitutes best practice for building streams.

I am almost always dropping 50-70% of the "best practices" for Streams. There are certain rules to follow, all the rest can be overruled. However, I would not recommend anyone to try writing something like https://github.com/mcollina/syncthrough without a great experience in the internals. 

> Would it be fair to say that best practice is also non-contradictory practice? As in, "you have a good stream if you did not do A, B, C"?

Yes! The golden rule of streams is **to always respect backpressure**, A) never `push()` if you are not asked, and B) never call `write()` after it returns `false` but wait for `'drain'` instead. The latter holds more power for me and help me reason about things better.

The third rule: C) streams changes between different node versions, and the library you use. Be careful and test things.

locale/en/docs/guides/backpressuring-in-streams.md
```diff
@@ -0,0 +1,480 @@
...
+  ^            when queue is empty     +============+                         |
+  ^------------^-----------------------<  Buffering |                         |
+               |                       |============|                         |
+               +> emit .drain();       |  ^Buffer^  |                         |
+               +> emit .resume();      +------------+                         |
+                                       |  ^Buffer^  |                         |
+                                       +------------+   add chunk to queue    |
+                                       |            <---^---------------------<
+                                       +============+
+```
```
mcollina, line 355: This diagram is making good progress! But I think you should split `pipe()` in its own step of the diagram, and not make it part of the "loop". The event handlers are, so we should show the distinction.

The syntax for events, e.g. `drain()`, is not common throughout the docs, so we should probaly use `on('drain', cb)`  instead.

also cc @lrlna who did some visualization on this in the past on paper.

locale/en/docs/guides/backpressuring-in-streams.md
```diff
@@ -0,0 +1,480 @@
...
+
+In one scenario, we take a large file (approximately ~9gb) and compress it up
+using the familiar [`zip(1)`][] tool.
+
+```
+$ zip The.Matrix.1080p.mkv
+```
+
+While that will take a few minutes to compress, in another shell we may run
+a script that takes Node.js' module [`Zlib`][], that wraps around another compression tool, [`gzip(1)`][].
```
TimothyGu: "`zlib`" is how the documentation refers to itself

locale/en/docs/guides/backpressuring-in-streams.md
```diff
@@ -0,0 +1,480 @@
...
+                                       +------------+   add chunk to queue    |
+                                       |            <---^---------------------<
+                                       +============+
+```
+
+_Note:_ The `.pipe()` function is typically where backpressure is invoked. In
+an isolate application, both [`Readable`][] and [`Writable`][] streams
+should be present. If you are writing an application meant to accept a
+[`Readable`][] stream, or pipes to a [`Writable`][] stream from another
+component, you may omit this detail.
```
mcollina: We should add a paragraph or a section here to discuss streams pipelines. What happens when a transform is in place?

locale/en/docs/guides/backpressuring-in-streams.md
```diff
@@ -0,0 +1,480 @@
...
+    super(options);
+  }
+
+  _write(chunk, encoding, callback) {
+
+    // does something?
+
+  }
+}
+```
```
mcollina: We need an equal section on 'Write a Readable Stream'.

We should mention that [`readable-stream`][https://github.com/nodejs/readable-stream] is the best way to build a stream application across multiple versions of Node.js, and the browser too! This will make your application have a stable base, and an implementation that changes with its own lifecycle compared to node core.

locale/en/docs/guides/backpressuring-in-streams.md
```diff
@@ -0,0 +1,480 @@
...
+a script that takes Node.js' module [`Zlib`][], that wraps around another compression tool, [`gzip(1)`][].
+
+```javascript
+const gzip = require('zlib').createGzip();
+const fs = require('fs');
+
+const inp = fs.createReadStream('The.Matrix.1080p.mkv');
+const out = fs.createWriteStream('The.Matrix.1080p.mkv.gz');
+
+inp.pipe(gzip).pipe(out);
```
mcollina, line 78: we should add a note here and refer to [`pump`](https://github.com/mafintosh/pump). It's a must have, as this example does not handle error correctly.

mcollina (changes requested):
Wow, this is great work. We are getting close, thanks so much for the effort in this. There is one thing that is a must to fix, and the rest is bonus content.

locale/en/docs/guides/backpressuring-in-streams.md
```diff
@@ -0,0 +1,480 @@
...
+```
+
+While the first `zip` function will ultimately fail, Node will be able to handle
```
TimothyGu: Not sure what the "`zip` function" is referring to. The command run in the shell?

locale/en/docs/guides/backpressuring-in-streams.md
```diff
@@ -0,0 +1,480 @@
...
+down other processes, and monopolizing a large part of your system until
+completion.
+
+This results in a few things:
+
+* Slowing down all other current processes
+* A very overworked garbage collector
+* Memory exhaustion
+
+## Overall Dip in System Performance
```
TimothyGu: Remove this heading maybe. The following paragraph seems to directly follow the list above.

locale/en/docs/guides/backpressuring-in-streams.md
```diff
@@ -0,0 +1,480 @@
...
+         1  signals received
+        25  voluntary context switches
+    629566  involuntary context switches
+```
+
+The maximum byte size occupied by virtual memory turns out to be approximately
+`1.52 gb`.
+
+Without streams in place to delegate the backpressure, there is a memory space 
+with an entire degree of magnitude greater that is being allocated. That is a
```
TimothyGu: I think "order of magnitude" is the more common expression

locale/en/docs/guides/backpressuring-in-streams.md
```diff
@@ -0,0 +1,480 @@
...
+       199  block output operations
+         0  messages sent
+         0  messages received
+         1  signals received
+        25  voluntary context switches
+    629566  involuntary context switches
+```
+
+The maximum byte size occupied by virtual memory turns out to be approximately
+`1.52 gb`.
```
TimothyGu: Ditto

locale/en/docs/guides/backpressuring-in-streams.md
```diff
@@ -0,0 +1,480 @@
...
+This results in a few things:
+
+* Slowing down all other current processes
+* A very overworked garbage collector
+* Memory exhaustion
+
+## Overall Dip in System Performance
+
+In the following examples we will take out the [return value][] of the
+`.write()` function and change the it to `true`. In any reference to 'modified'
```
TimothyGu: change <del>the</del> it

So far, it hasn't been clarified that the return value of `write()` is used in the backpressure mechanism. I'd say something like "change it to `true`, which effectively disables backpressure support in Node.js core."

locale/en/docs/guides/backpressuring-in-streams.md
```diff
@@ -0,0 +1,480 @@
...
+       194  block output operations
+         0  messages sent
+         0  messages received
+         1  signals received
+        12  voluntary context switches
+    666037  involuntary context switches
+```
+
+The maximum byte size occupied by virtual memory turns out to be approximately
+`87.81 mb`.
```
TimothyGu: The monospace formatting doesn't seem to be needed

locale/en/docs/guides/backpressuring-in-streams.md
```diff
@@ -0,0 +1,480 @@
...
+event will be emitted and resume the incoming data flow.
+
+The `Readable` stream also plays a vital role in backpressure. Because the two
+processes rely on one another to communicate effectively, if the `Readable` does
+not respect when the `Writable` stream asks to stop sending data, it can be
+just as problematic.
+
+An instance of this can be unconditionally calling your `Readable` stream to
+send data using the [`push method`][] or forcing data through whenever it
+is available (signaled by the [`.data` event][]):
```
TimothyGu: "`'data'` event" is more accurate

locale/en/docs/guides/backpressuring-in-streams.md
```diff
@@ -0,0 +1,480 @@
...
+```javascript
+// This is problematic as it ignores backpressure signals
+readable.on('readable', () => {
+
+  let chunk;
+  while (null !== (chunk = readable.read())) {
+    readable.push(chunk);
+  }
+
+});
```
TimothyGu: This looks like an infinite loop, since you are pushing the chunk from `readable` back into `readable`... I feel something like this portrays the problems with unconstrained `.push`ing better:

```js
class MyReadable extends Readable {
  _read(size) {
    let chunk;
    while (null !== (chunk = getNextChunk())) {
      this.push(chunk);
    }
  }
}
```

locale/en/docs/guides/backpressuring-in-streams.md
```diff
@@ -0,0 +1,480 @@
...
+
+// Also ignores backpressure and rams data through, regardless if destination
+// stream is ready or not.
+readable.on('data', data =>
+  writable.write(data);
+);
+```
```
TimothyGu: Good examples in general, but might be misplaced. This section is about how backpressure resolves these problems, while this portrays counterexamples. Perhaps "Rules to Abide For Writing Custom Streams" will be a better place to put them.

locale/en/docs/guides/backpressuring-in-streams.md
```diff
@@ -0,0 +1,480 @@
...
+
+_Note:_ The `.pipe()` function is typically where backpressure is invoked. In
+an isolate application, both [`Readable`][] and [`Writable`][] streams
+should be present. If you are writing an application meant to accept a
+[`Readable`][] stream, or pipes to a [`Writable`][] stream from another
+component, you may omit this detail.
+
+## Example App
+
+Since [Node.js v0.10][], the [`Streams`][] class has offered the ability to
```
TimothyGu: `Stream` class

locale/en/docs/guides/backpressuring-in-streams.md
TimothyGu: As a general question, is the primarily Node.js-based `Readable` stream still the best option in the browser? In my experience, the most often cause of using Node.js `Readable` in the browser is as a shim when creating Browserify bundle. At the same time, `readable-stream` IIRC depends on additional polyfills like one for [`events` module](https://github.com/Gozala/events/blob/master/events.js). Plus, I know WHATWG has been working on [their version](https://streams.spec.whatwg.org/) of streams.

Regarding using the userland module in Node.js, definitely. You may also want to link to [@rvagg's blog post](https://r.va.gg/2014/06/why-i-dont-use-nodes-core-stream-module.html)

locale/en/docs/guides/backpressuring-in-streams.md
```diff
@@ -0,0 +1,480 @@
...
+Streams are a often used module in Node.js. They are important to the internal
+structure, and for develoeprs, to expand and connect across the Node.js modules
+ecosystem.
+
+Hopefully, you will now be able to troubleshoot, safely code your own
+[`Writable`][] streams with backpressure in mind, and share your knowledge with
+colleagues and friends.
+
+Be sure to read up more on [`Streams`][] for other [`EventEmitters`][] to help
+improve your knowledge when building applications with Node.js.
```
TimothyGu: Not sure what "for other `EventEmitters`" is trying to say...

locale/en/docs/guides/backpressuring-in-streams.md
```diff
@@ -0,0 +1,480 @@
...
+    readable.push(chunk);
+  }
+
+});
+
+// Also ignores backpressure and rams data through, regardless if destination
+// stream is ready or not.
+readable.on('data', data =>
+  writable.write(data);
+);
```
TimothyGu: Might want to add braces around `writable.write(data)` just to be clear `on` doesn't care about the return value of `write()`.

locale/en/docs/guides/backpressuring-in-streams.md
```diff
@@ -0,0 +1,480 @@
...
+
+* If the write queue is busy, [`.write()`][] will return false.
+* If the data chunk is too large, [`.write()`][] will return false (the limit
+is indicated by the variable, [`highWaterMark`][]).
+
+_Note_: In most machines, there is a byte size that is determines when a buffer
+is full (which will vary across different machines). Node.js allows you to set
+your own custom [`highWaterMark`][], but commonly, the default is the optimal
+value for what system is running the application. In instances where you might
+want to raise that value, go for it, but do so with caution!
```
TimothyGu: This section only deals with rules for writing to a Writable stream from outside the stream. We also need rules for writing to a Readable stream from inside that Readable stream (by respecting the return value of `this.push()`).

TimothyGu (commented):
Great to see how this is coming along. Apologies for the number of nits, just want to make sure those don't get forgotten when this is applied.

locale/en/docs/guides/backpressuring-in-streams.md
jalafel: Is the vital difference between chaining pipes together vs pump have to do with error handling? As in chaining pipes together can be erroneous due to the fact there is _no_ error handling? Heh, sorry for my ignorance!

locale/en/docs/guides/backpressuring-in-streams.md
jalafel: Heh this might be rhetorical question, but I will answer just in case ... when Transform is in the pipeline, it takes in the Readable output and does the `._transform()` and pipes that into the destination Writable? Do you mean to explicate this process in the guide?

Edit: Jw, do Transform Streams or Duplex Streams have special conditions to respect backpressure as well?

locale/en/docs/guides/backpressuring-in-streams.md
jalafel: Yes! I was referencing the example mcollina was describing. I hope I did it right, should this be more explicit?

locale/en/docs/guides/backpressuring-in-streams.md
jalafel: Tbh I went back to edit this many times and couldn't figure out why it sounded so off .. hehe thanks!

locale/en/docs/guides/backpressuring-in-streams.md
TimothyGu: The "function" was throwing me off a bit. And also why would it fail?

jalafel: I am not entirely sure. My layman's understanding of its failure is because of the lack of backpressure mechanism. There is a lot of data that is being compressed and the .zip file gets corrupted. But if @mcollina might be able to shed a technical light on this, it'd be much appreciated!

TimothyGu: The reason why I asked is that I'm fairly certain `zip` will not encounter this problem. Unlike Node.js, `zip` and other UNIX utilities operate synchronously. What that means is that they only `read()` as much as they need at one time, and immediately `write()` it out afterwards. Since these two system calls are synchronous, there would not be any backpressure problems since by definition they only appear in an asynchronous system.

jalafel: Interesting. So to be clear, synchronous systems do not encounter backpressure buildup like async does? 

I also did test out the `zip` tool and it does, in fact, corrupt the large file, though I am not sure why, if those are the mechanics behind data flow. (Though `gzip` on my UNIX command line does not corrupt the large file.) 

Any thoughts?  My only guess could be because the filetype I am using is video!

locale/en/docs/guides/backpressuring-in-streams.md
jalafel: Wondering if by this you meant something like:

```
                                                     +===================+
                         +-->  Piping functions  +--->   src.pipe(dest)  |
                         x     are set up during     |===================|
                         x     the .pipe method.     |  Event callbacks  |
+===============+        x                           |-------------------|
|   Your Data   |        x     They exist outside    | .on('close', cb)  |
+=======+=======+        x     the data flow, but    | .on('data', cb)   |
        |                x     importantly attach    | .on('drain', cb)  |
        |                x     events, and their     | .on('unpipe', cb) |
+-------v-----------+    x     respective callbacks. | .on('error', cb)  |
|  Readable Stream  +----+                           | .on('finish', cb) |
+-^-------^-------^-+    |                           | .on('end', cb)    |
  ^       |       ^      |                           +-------------------+
  |       |       |      |                           
  |       ^       |      |                                              
  ^       ^       ^      |    +-------------------+         +=================+
  ^       |       ^      +---->  Writable Stream  +--------->  .write(chunk)  |
  |       |       |           +-------------------+         +=======+=========+
```

Or even have more explicit steps come in?

locale/en/docs/guides/backpressuring-in-streams.md
mcollina: Yes. the problem is that in case of errors the underlining file descriptor on the other side might not get closed.

locale/en/docs/guides/backpressuring-in-streams.md
mcollina: Yes, something like that is way better.

locale/en/docs/guides/backpressuring-in-streams.md
mcollina: A `Transform`  has both an outgoing and an incoming `highWaterMark`. Most people write `Transform` to setup pipelines, so it's important to explain how it all plays out when you chain multiple streams.

locale/en/docs/guides/backpressuring-in-streams.md
mcollina: `readable-stream` is the way to go in the browser. `readable-stream` exports `Readable` & co, so all that is written here is valid there as well.

By using streams in the browser, I mean mostly to leverage code already written elsewhere for the purpose, i.e. the streams ecosystem.

locale/en/docs/guides/backpressuring-in-streams.md
jalafel: So, these are the true facts say about a `Transform` stream:

It is important because its used often to setup pipelines. In it, there are `_write()` and `_read()` calls that must be implemented with the same care as with `Readable` and `Writable` (with respect to the return values `.push()` and `.write()`. And also HWM may change along the pipeline.

mcollina: It is not correct. `Transform`  implementors should add `_transform`, as it is a `Duplex` that implement both `_read` and `_write`, i.e. `Transform` has those method implemented.

locale/en/docs/guides/backpressuring-in-streams.md
lrlna: hi! this is so so good, i love it!

in terms of the diagram, i find the loop a little weird to get around and understand, although i definitely see where you're coming from.

What about making it a bit more linear? I am not good with replicating the fancy thing you have going, but something like this on paper? 

stream-backpressure

jalafel:
Sorry for the static on this! I took a look back on this, realized some grammar was off and fixed some  formatting.

Added in a few pointers mentioned earlier in the thread, though they didn't necessarily close off all the reviews.

Skimmed out the example application because I felt that was redundant since there are a rich supply of examples found in the documentation, and I didn't want to reinvent the wheel. 

I only have one question and that's in regards to the implementation of a Writable stream. Are there any mistakes one can make in `_write()` or `_writev()` that would damage backpressure? They seem like pretty foolproof functions to me.

Should the guide extend its scope and talk about creating a custom writable wtream from scratch then? As in:

```
MyWritable.prototype.write = function(chunk, enc, cb) {
  // code that respects return value here
}

util.inherits(MyWritable, Stream);
```

Thanks for all the help! Hope to land this soon!

mcollina:
@jessicaquynh `_writev` goes in combination with `.cork()`, and it is especially useful if we are writing to a stream directly. The biggest mistake is this pattern:

```js
ws.cork()
ws.write('hello ')
ws.write('world ')
ws.uncork()

ws.cork()
ws.write('from ')
ws.write('Matteo')
ws.uncork()
```

This will cause two transfers into C++ realm, making the use of cork() a bit useless the correct pattern is:

```js
ws.cork()
ws.write('hello ')
ws.write('world ')
process.nextTick(doUncork, ws)

ws.cork()
ws.write('from ')
ws.write('Matteo')
process.nextTick(doUncork, ws)

// as a global function
function doUncork (stream) {
  stream.uncork()
}
```

`cork()` can be called as many times we want, we just need to be careful to call `uncork()` the same amount of times to make it flow again.

jalafel:
@mcollina Thank you for your input and help! I've added the changes to the writable section!

locale/en/docs/guides/backpressuring-in-streams.md
```diff
@@ -298,6 +298,12 @@ Well the answer is simple: Node.js does all of this automatically for you.
 That's so great! But also not so great when we are trying to understand how to
 implement our own custom streams.
 
+Note: In most machines, there is a byte size that is determines when a buffer
+is full (which will vary across different machines). Node.js allows you to set
+your own custom [`highWaterMark`][], but commonly, the default is the optimal
+value for what system is running the application. In instances where you might
+want to raise that value, go for it, but do so with caution!
```
mcollina: I think you should write here the defaults, it usually helps people

mcollina (approved):
LGTM with a nit

mcollina:
LGTM

fhemberger:
Great, thank you!

PR #1128, opened 2017-02-04
Update events.md

pin3da:
This is a clean PR of #998


pin3da:
#1130 automatically changed words that not were in english in events.md. That generates the conflict with this PR.

What do you suggest ?

PR #1129, opened 2017-02-05
add info about using nvm to package-manaager.md

danbev:
Adding information about a way of using nvm to use version of
Node.js built from source.

### DIFF
--- a/locale/en/download/package-manager.md
+++ b/locale/en/download/package-manager.md
@@ -291,12 +291,15 @@ pkgin -y install nodejs
 ```
 
 ## nvm
-On Unix / OS X systems Node.js built from source can be installed by
+On Unix / OS X systems Node.js built from source can be installed using
 [nvm](https://github.com/creationix/nvm) by installing into the location that nvm expects:
 
 ```bash
-$ make install DESTDIR='$(NVM_DIR)/versions/node/' PREFIX='$(FULLVERSION)-pre'
+$ make install DESTDIR=~/.nvm/versions/node/ PREFIX='$(FULLVERSION)-pre'
 ```
+Please note that this is the default location that nvm will be installed into, if you have
+changed this to be something different make sure that you also change the `DESTDIR` variable
+above.
 
 After this you can use `nvm` to switch between released versions and versions
 built from source.
--- a/locale/en/download/package-manager.md
+++ b/locale/en/download/package-manager.md
@@ -295,11 +295,8 @@ On Unix / OS X systems Node.js built from source can be installed using
 [nvm](https://github.com/creationix/nvm) by installing into the location that nvm expects:
 
 ```bash
-$ make install DESTDIR=~/.nvm/versions/node/ PREFIX='$(FULLVERSION)-pre'
+$ env VERSION=`python tools/getnodeversion.py` make install DESTDIR=`nvm_version_path v$VERSION` PREFIX=""
 ```
-Please note that this is the default location that nvm will be installed into, if you have
-changed this to be something different make sure that you also change the `DESTDIR` variable
-above.
 
 After this you can use `nvm` to switch between released versions and versions
 built from source.
--- a/locale/en/download/package-manager.md
+++ b/locale/en/download/package-manager.md
@@ -15,6 +15,7 @@ title: Installing Node.js via package manager
 * [FreeBSD and OpenBSD](#freebsd-and-openbsd)
 * [Gentoo](#gentoo)
 * [NetBSD](#netbsd)
+* [nvm](#nvm)
 * [openSUSE and SLE](#opensuse-and-sle)
 * [OSX](#osx)
 * [SmartOS and illumos](#smartos-and-illumos)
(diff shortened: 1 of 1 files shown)


locale/en/download/package-manager.md
```diff
@@ -289,6 +290,28 @@ Or install a binary package (if available for your platform) using pkgin:
 pkgin -y install nodejs
 ```
 
+## nvm
+On Unix / OS X systems Node.js built from source can be installed by
+[nvm](https://github.com/creationix/nvm) by installing into the location that nvm expects:
+
+```bash
+$ make install DESTDIR='$(NVM_DIR)/versions/node/' PREFIX='$(FULLVERSION)-pre'
```
ljharb: please use `nvm_version_path "$VERSION"` so as to not hardcode the path - i'm much more likely to break the path than I am to break that internal `nvm` function.

ljharb (changes requested):

locale/en/download/package-manager.md
danbev: @ljharb Sorry about this, I was not ignoring you suggestion but failed to get it to work for me. I'm not exactly sure about the cause for this not working yet but from within the Makefile the `nvm_version_path` function cannot be found. I'm not sure if this is because it is a function. I'll take another look tomorrow, but let me know if you have any ideas or suggestions on this. Thanks!

ljharb: It will only be found if `nvm.sh` is sourced, and `NVM_DIR` should only be set if the profile file is sourced (which sources `nvm.sh` for you). If necessary, you'll want to source `nvm.sh` in the Makefile

danbev: I don't think that we would want to source `nvm.sh` from the Makefile. I think the Makefile should stay unchanged and not have anything nvm specific. 

I've looked into exporting the function but this looks like it would not be supported by all shells. 

I'm suggesting that we change this to be `DESTDIR=~/.nvm/versions/node/` and then document that if nvm has been installed into a different location then the path must be updated to reflect that.

What do you think about that?

ljharb: The function itself is POSIX-compliant, so it should work in any POSIX-compliant shell.

I think that it depends on if your goal is correctness/robustness or convenience. Correctness requires using the same `NVM_DIR` env var that the user will use, and robustness requires not hardcoding the nvm dir's structure. Could you run a separate bash script that sources nvm.sh, and output the version path, and run that from the Makefile?

danbev: > Could you run a separate bash script that sources nvm.sh, and output the version path, and run that from the Makefile?

Would be enough just to pass a dummy version to `nvm_version_path` and use the path, without the dummy version, and then send that path to make? 

Something along the lines of : 
```shell
$ make install DESTDIR=`nvm_version_path 999|sed s/999$//` PREFIX='$(FULLVERSION)-pre'
```

ljharb: No, because `nvm_version_path` does different things depending on the version number.

danbev: > No, because nvm_version_path does different things depending on the version number.

After running `declare -f nvm_version_path` I see what you mean. Thanks

So another attempt:
```shell
$ env VERSION=`python tools/getnodeversion.py` make install DESTDIR=`nvm_version_path v$VERSION` PREFIX=""
```
Perhaps not pretty but might an acceptable solution until support for this lands in nvm like you mentioned before?

ljharb: That seems much better, thanks!

ljharb (approved):
Thanks!

danbev:
Would someone will commit rights to this repo be willing to merge and push this or grant me commit rights so I can push it?

sotayamashita:
@fhemberger Could you review and merge it, if no problem.
cc: @nodejs/website

locale/en/download/package-manager.md
```diff
@@ -289,6 +290,28 @@ Or install a binary package (if available for your platform) using pkgin:
 pkgin -y install nodejs
 ```
 
+## nvm
+On Unix / OS X systems Node.js built from source can be installed using
```
fhemberger: Could you please add a sentence or two about installing nvm in general and using it to install Node.js, not just how to add builds from source to it? Thx.

Otherwise LGTM.

fhemberger (changes requested):

locale/en/download/package-manager.md
ljharb: Please strongly stress using the sole supported installation method - the curl command in the readme - over alternative unsupported methods of installation :-)

danbev: @fhemberger @ljharb I've added some information in general (very short) about nvm and install instructions. Let me know what you think. Thanks

fhemberger (approved):

fhemberger:
LGTM!

fhemberger:
Thanks again!

PR #1131, opened 2017-02-06
guides: debugging getting started guide

joshgav:
A guide describing the basics of starting Node.js in debug/inspect mode and how to attach any client. Includes a list of popular clients.

A potential goal is to reference this from the hint text when Node.js is started with the `--inspect` flag, see https://github.com/nodejs/node/pull/8978.

/cc @nodejs/diagnostics @eugeneo

### DIFF
--- a/locale/en/docs/guides/debugging_getting_started.md
+++ b/locale/en/docs/guides/debugging_getting_started.md
@@ -1,0 +1,173 @@
+---
+title: Debugging - Getting Started
+layout: docs.hbs
+---
+
+# Debugging Guide
+
+This guide will help you get started debugging your Node.js apps and scripts.
+
+## Enable Inspector
+
+**NOTE**: The `--inspect` option and [Inspector Protocol][] are _experimental_ and may change.
+
+When started with the **--inspect** switch, a Node.js process listens via WebSockets
+for diagnostic commands as defined by the [Inspector Protocol][],
+by default at host and port 127.0.0.1:9229. Each process is also assigned a
+unique [UUID][] (e.g. `0f2c936f-b1cd-4ac9-aab3-f63b0f33d55e`).
+
+Inspector clients must know and specify host address, port, and UUID to connect
+to the WebSocket interface. The full URL is
+`ws://127.0.0.1:9229/0f2c936f-b1cd-4ac9-aab3-f63b0f33d55e`, of course dependent
+on actual host and port and with the correct UUID for the instance.
+
+Inspector also includes an HTTP endpoint to serve metadata about the debuggee,
+including its WebSocket URL, UUID, and Chrome DevTools URL. Get this metadata
+by sending an HTTP request to `http://[host:port]/json/list`.  This returns a
+JSON object like the following; use the `webSocketDebuggerUrl` property as the
+URL to connect directly to Inspector.
+
+```javascript
+{
+  "description": "node.js instance",
+  "devtoolsFrontendUrl": "chrome-devtools://devtools/bundled/inspector.html?experiments=true&v8only=true&ws=127.0.0.1:9229/0f2c936f-b1cd-4ac9-aab3-f63b0f33d55e",
+  "faviconUrl": "https://nodejs.org/static/favicon.ico",
+  "id": "0f2c936f-b1cd-4ac9-aab3-f63b0f33d55e",
+  "title": "node",
+  "type": "node",
+  "url": "file://",
+  "webSocketDebuggerUrl": "ws://127.0.0.1:9229/0f2c936f-b1cd-4ac9-aab3-f63b0f33d55e"
+}
+```
+
+A Node.js process started *without* `--inspect` can also be instructed to start
+listening for debugging messages by signaling it with `SIGUSR1` (on Linux and
+OS X). As of Node 7 this activates the legacy Debugger API; in Node 8 and later
+it will activate the Inspector API.
+
+---
+
+## Inspector Clients
+
+Several commercial and open source tools can connect to Node's Inspector. Basic
+info on these follows:
+
+#### [node-inspect](https://github.com/nodejs/node-inspect)
+
+* CLI Debugger supported by the Node.js Foundation which uses the [Inspector Protocol][].
+* A version is bundled with Node and can be used with `node inspect myscript.js`.
+* The latest version can also be installed independently (e.g. `npm install -g node-inspect`) 
+  and used with `node-inspect myscript.js`.
+    
+#### [Chrome DevTools](https://github.com/ChromeDevTools/devtools-frontend) 55+
+
+* **Option 1**: Open `chrome://inspect` in a Chromium-based
+  browser. Click the Configure button and ensure your target host and port
+  are listed.
+* **Option 2**: Copy the `devtoolsFrontendUrl` from the output of `/json/list`
+  (see above) or the --inspect hint text and paste into Chrome.
+
+#### [VS Code](https://github.com/microsoft/vscode) 1.10+
+
+* In the Debug panel, click the settings icon to open `.vscode/launch.json`.
+  Select "Node.js" for initial setup.
+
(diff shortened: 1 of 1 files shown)


locale/en/docs/guides/debugging_getting_started.md
```diff
@@ -0,0 +1,198 @@
...
+
+### --debug
+
+When started with the **--debug** or **--debug-brk** switches, Node.js listens
+for debugging commands defined by the [V8 Debugging Protocol][] on a TCP port,
+by default `5858`. Any debugger client which speaks this protocol can connect
+to and debug the running process.
+
+A running Node.js process can also be instructed to start listening for
+debugging messages by signaling it with `SIGUSR1` (on Linux and OS X).
```
bnoordhuis: Aren't we in the process of deprecating `--debug` and friends?  IIRC, the behavior of SIGUSR1 is going to change, too.

eugeneo:
LGTM from the Chrome DevTools team. Not sure I should use GitHub "Approve" as there seem to be comments from other parties.

locale/en/docs/guides/debugging_getting_started.md
```diff
@@ -0,0 +1,198 @@
...
+Each Node process is marked with its own UUID similar in form to
+`0f2c936f-b1cd-4ac9-aab3-f63b0f33d55e`, and debugger clients must specify this
+ID to connect to the debugee. The full URL is
+`ws://127.0.0.1:9229/0f2c936f-b1cd-4ac9-aab3-f63b0f33d55e`, with the specific
+UUID changing for each target.
+
+Metadata about the target debuggee including its UUID can be retrieved via HTTP
+(not websockets) from `http://[host:port]/json/list` with host and port
+properly specified (default: 127.0.0.1:9229). This returns a JSON object like
+the following; use its `webSocketDebuggerUrl` as the connection string including
```
joyeecheung: The last sentence seems to lack something? Who uses that string? (Maybe `The debugger client may use..`)

locale/en/docs/guides/debugging_getting_started.md
```diff
@@ -0,0 +1,198 @@
...
+[Inspector Protocol]: https://chromedevtools.github.io/debugger-protocol-viewer/v8/
+
+---
+
+## Debugging Clients
+
+Several commercial and open source tools can connect to Node's
+debugger and/or inspector. Info on these follows.
+
+* [Built-in Debugger](https://github.com/nodejs/node/blob/master/lib/_debugger.js)
```
joyeecheung: I think another way to organize things is:

```
### Inspector Protocol

* node-inspect
* Chrome DevTools
* VSCode

### Debugger Protocol

* Built-in Debugger
* Chrome DevTools
* VSCode
```

So people can first look for a client that uses the protocol they prefer(hopefully the inspector protocol) and look for alternatives if their current tool doesn't support the new protocol, but this works fine too.

prigara:
@joshgav we'd like to add info about WebStorm. Should I create a new PR or can you just add it to this PR? Thanks!

[WebStorm](https://www.jetbrains.com/webstorm/) 2017.1+ and other JetBrains IDEs
Create a new Node.js debug configuration and hit Debug. `--inspect` will be used by default for Node.js 7+. To disable uncheck `js.debugger.node.use.inspect` in the IDE Registry.

locale/en/docs/guides/debugging_getting_started.md
pmuellr: Since this document will appear in the version-independent "Guides" section of the doc, makes sense for it to discuss both `--debug` and `--inspect`.  But they should be in separate sections I think - inspect first, then debug. You never use both at the same time (I assume!), so having them separated seems right.  Each section should also mention which node versions support that type of debug.

naugtur: If the article would benefit from different content depending on a version, why not include version in the URL of the hint text? `...debugging_getting_started#v7`

The main "Guides" article would have no assumptions about the user's version, but if opened from a URL printed by a particular one, some content can be replaced.

Having that option, even if there's only one version that actually benefits from this, might save us some convoluted explanation while being totally optional.

locale/en/docs/guides/debugging_getting_started.md
```diff
@@ -0,0 +1,198 @@
...
+
+**NOTE**: The `--inspect` option and protocol are _experimental_ and may change.
+
+When started with the **--inspect** or **--inspect-brk** switches, Node.js
+listens for diagnostic commands defined by the [Inspector Protocol][]
+via websockets, by default at host and port 127.0.0.1:9229.
+
+Each Node process is marked with its own UUID similar in form to
+`0f2c936f-b1cd-4ac9-aab3-f63b0f33d55e`, and debugger clients must specify this
+ID to connect to the debugee. The full URL is
```
naugtur: totally ignorable hint: `debugee` might not be understandable for people less fluent in English. They might think it's yet another tool or component.

`must specify this UUID to connect to matching process` maybe?

locale/en/docs/guides/debugging_getting_started.md
naugtur: That part looks good to me, but

"connection string" is used just once, here, with no prior explanation. `ws:.../<UUID>` is called "full URL" in the paragraph above.

I suggest:
This returns a JSON object like the one below. Its `webSocketDebuggerUrl` field is the full URL a debugger needs to connect to the target process because it includes the UUID.

locale/en/docs/guides/debugging_getting_started.md
```diff
@@ -0,0 +1,198 @@
...
+            and use main process to run CLI debugger.</li>
+      </ul>
+    </td>
+  </tr>
+</table>
+
+
+## Default ports: 
+
+* Inpsector: 9229
```
richardlau: Inspector

locale/en/docs/guides/debugging_getting_started.md
joshgav: Thanks for the suggestions, I'm going to gather the --debug info in its own section toward the end.

@naugtur: the markdown headers get converted into HTML anchors ([fragment identifiers](https://www.w3.org/DesignIssues/Fragment.html), so it should be possible to refer directly to the `--debug` section. I'm not going to make the content dynamic based on the fragment here (too much work at the moment), but if you think that would be useful you could start another PR. Thanks!

locale/en/docs/guides/debugging_getting_started.md
joshgav: Fixing now. @naugtur thanks for the suggestion, I'm taking out the term "connection string."

locale/en/docs/guides/debugging_getting_started.md
joshgav: Thank you! I'll find a better term.

naugtur: I don't suppose there's a one word term other than debugee. 

How about:

> Each Node process is marked with its own UUID similar in form to
> `0f2c936f-b1cd-4ac9-aab3-f63b0f33d55e`, and debugger clients must specify this
> UUID to Indicate which process to connect to.

Note I also changed `ID` to `UUID` because ID is an undeclared reference ;)

joshgav:
Updated:

* Refactored to move legacy `--debug` info to its own section at the end and focus on Inspector and `--inspect`.
* Added WebStorm info - thanks @prigara!
* Updated instructions for some clients and did a general edit pass.

IMO this is ready to land. Could those who previously reviewed please take a look and give a GitHub approval or LGTM? Thank you!

naugtur:
Somehow after you re-arranged it seems much cleaner/shorter. LGTM, even with the debuggee.

richardlau (approved):

joyeecheung (approved):

joshgav:
Landed in https://github.com/nodejs/nodejs.org/commit/b2cdfc85a2d898931ab65e39be6d818cf6eb8f72

PR #1141, opened 2017-02-13
Making Downloads Trackable

ZibbyKeaton:
All - I updated the long form case studies and the 2016 survey report to make it so I can track on Google Analytics. Let me know if this looks good or what changes I need to make here.

### DIFF
--- a/locale/en/foundation/case-studies.md
+++ b/locale/en/foundation/case-studies.md
@@ -16,7 +16,7 @@ preferences (current and future) for front end, back end and IoT developers.
 
 * [Read the full announcement.](/en/blog/announcements/nodejs-foundation-survey/)
 * [See the infographic.](/static/documents/2016-survey-infographic.png)
-* [Read the full report.](/static/documents/2016-survey-report.pdf)
+* <a href="/static/documents/2016-survey-report.pdf" onClick="ga('send', 'event', { eventCategory: 'casestudy', eventAction: 'click', eventLabel: 'Node.js 2016 Report', eventValue: 0});">Read the full report.</a>
 
 ## Presentation - The Enterprise Case for Node.js
 The Node.js Foundation has consolidated all the best thinking around why enterprises should consider Node.js for their systems of engagement in this short presentation.
@@ -33,7 +33,7 @@ Walmart sees nearly 20,000 hits per-second on its website and mobile app during 
 
 Collin Estes is the Director of Software Engineering and Chief Architect at Mathematical Research Institute Technologies (MRI Technologies), who partners with United Technologies Corporation Aerospace Systems, the company that creates and maintains the spacesuits for extra vehicular activity (EVA) missions at NASA. He and his team took on the challenge of creating an end-to-end system for data on the full lifecycle of spacesuits. He chose Node.js to make it work.
 
-[Read the full case study.](/static/documents/casestudies/Node_CaseStudy_Nasa_FNL.pdf)
+<a href="/static/documents/casestudies/Node_CaseStudy_Nasa_FNL.pdf" onClick="ga('send', 'event', { eventCategory: 'casestudy', eventAction: 'click', eventLabel: 'NASA case study', eventValue: 0});">Read the full case study.</a>
 
 ## Node.js Drives Major Scale and Reliability at Uber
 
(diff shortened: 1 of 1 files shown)


locale/en/foundation/case-studies.md
```diff
@@ -16,7 +16,7 @@ preferences (current and future) for front end, back end and IoT developers.
 
 * [Read the full announcement.](/en/blog/announcements/nodejs-foundation-survey/)
 * [See the infographic.](/static/documents/2016-survey-infographic.png)
-* [Read the full report.](/static/documents/2016-survey-report.pdf)
+* <a href="/static/documents/2016-survey-report.pdf" onClick="ga('send', 'event', { eventCategory: 'casestudy', eventAction: 'click', eventLabel: 'Node.js 2016 Report', eventValue: 0});">Read the full report.</a>
```
fhemberger, line 19: I have a different idea here. Instead of adding this JavaScript code inline to all links, a data-attribute would be more appropriate, e.g:

```
<a href="/static/documents/2016-survey-report.pdf" data-casestudy="Node.js 2016 Report">Read the full report.</a>
```

At the end of the page, add the following snippet:
```js
<script>
if (!ga) { return; }
var caseStudies = document.querySelectorAll('a[data-casestudy]');
for (var i = 0; i < caseStudies.length; i++) {
    caseStudies[i].addEventListener('click', function(event) {
      ga('send', 'event', { eventCategory: 'casestudy', eventAction: 'click', eventLabel: this.getAttribute('data-casestudy'), eventValue: 0});
    });
}
</script>
```

This selects all links on that page that have the attribute `data-casestudy` and adds an event listener (basically the same way that you did before), which fires the Google Analytics event on click. But instead of adding the tracking code over and over for each link, it's in one place. So if you need to make adjustments for the tracking code, there is only one place to edit.

fhemberger (changes requested):

locale/en/foundation/case-studies.md
ZibbyKeaton: @fhemberger thank you. My only hesitation in doing it this way is I only know how to upload them the way that I learned... If I need to upload more will i need to create a new process here?

fhemberger: No, you just need to add a link like `<a href="filename.pdf" data-casestudy="Case Study Title">Read the full report.</a>` that's all. Everything else will happen automatically.

/cc @nodejs/website Any thoughts about this?

silverwind: How about a delegated event listener? Should work for dynamically inserted links too.

````js
if (!ga) return;
document.documentElement.addEventListener('click', function(e) {
  if(!e.target || !e.target.dataset || !e.target.dataset.casestudy) return;
  ga('send', 'event', {
    eventCategory: 'casestudy',
    eventAction: 'click',
    eventLabel: e.target.dataset.casestudy,
    eventValue: 0
  });
});
````

fhemberger: @silverwind You're right … even better.

ZibbyKeaton:
What ever you think is best that works for me! Just as long as it's simple for me to implement and I can easily how each are being downloaded.

Fishrock123:
Should this also be modified allow this to respect `do-not-track` in some way?

silverwind:
Yes I think it should honor DNT, for example:

````js
if (ga && navigator.doNotTrack !== '1') {
  document.documentElement.addEventListener('click', function(e) {
    if(!e.target || !e.target.dataset || !e.target.dataset.casestudy) return;
    ga('send', 'event', {
      eventCategory: 'casestudy',
      eventAction: 'click',
      eventLabel: e.target.dataset.casestudy,
      eventValue: 0
    });
  });  
}
````

lpinca:
~~Isn't [this](https://github.com/nodejs/nodejs.org/blob/master/layouts/partials/footer.hbs#L54-L64) sufficient?~~

lpinca:
Nvm. We can reuse [this](https://github.com/nodejs/nodejs.org/blob/master/static/js/dnt_helper.js) helper function though.

fhemberger:
@ZibbyKeaton @nodejs/website I updated the PR: #1148. Please review.

fhemberger:
Closing this in favor of #1148, which will land later.

ZibbyKeaton:
Thank you @fhemberger

PR #1148, opened 2017-02-22
Making Downloads Trackable (see #1141)

fhemberger:
See #1141 for discussion.

### DIFF
--- a/layouts/partials/footer.hbs
+++ b/layouts/partials/footer.hbs
@@ -58,7 +58,21 @@
     s=o.getElementsByTagName(e)[0];j.async=1;j.src='//www.google-analytics.com/analytics.js';
     s.parentNode.insertBefore(j,s)}(window,document,'ga','script');
 
+    if (!ga) return;
+
     ga('create', 'UA-67020396-1', 'auto');
     ga('send', 'pageview');
+
+    document.documentElement.addEventListener('click', function(e) {
+
+      // Track case studies
+      if(!e.target || !e.target.dataset || !e.target.dataset.casestudy) return;
+      ga('send', 'event', {
+        eventCategory: 'casestudy',
+        eventAction: 'click',
+        eventLabel: e.target.dataset.casestudy,
+        eventValue: 0
+      });
+    });
   }
 </script>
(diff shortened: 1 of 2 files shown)


layouts/partials/footer.hbs
```diff
@@ -58,7 +58,21 @@
     s=o.getElementsByTagName(e)[0];j.async=1;j.src='//www.google-analytics.com/analytics.js';
     s.parentNode.insertBefore(j,s)}(window,document,'ga','script');
 
+    if (!ga) return;
```
lpinca, line 61: Isn't this redundant? The above IIFE should always add the global `ga` no?

locale/en/foundation/case-studies.md
```diff
@@ -16,7 +16,7 @@ preferences (current and future) for front end, back end and IoT developers.
 
 * [Read the full announcement.](/en/blog/announcements/nodejs-foundation-survey/)
 * [See the infographic.](/static/documents/2016-survey-infographic.png)
-* [Read the full report.](/static/documents/2016-survey-report.pdf)
+* <a href="/static/documents/2016-survey-report.pdf" data-casestudy="Node.js 2016 Report">Read the full report.</a>
```
lpinca, line 19: Is this a case of study?

fhemberger: It was treated that way in the original PR. ¯\_(ツ)_/¯

layouts/partials/footer.hbs
fhemberger: If you use an ad-blocker, the GA script will not be loaded and thus the `ga` object doesn't exist.

lpinca (approved):

phillipj:
Just tested this on nodejs.org and doesn't seem to work, at least not in Chrome on my machine. There is no request sent to Google when clicking these links.

I assume that is primarily caused by this error:

screenshot1

I'll have a look and open a PR shortly.

mikeal:
This is great, thanks for cleaning this up :+1:

lpinca:
Ops, that `return` seems to be outside a function so it should be probably be changed to `if (ga) { ... }`. Sorry missed that during review.

PR #1153, opened 2017-02-22
adding a --force flag to update a release blog post

italoacasas:
Adding a flag to allow update a specific **release blog post** without the need of removing the file first

cc: @nodejs/release

### DIFF
--- a/scripts/release-post.js
+++ b/scripts/release-post.js
@@ -179,7 +179,7 @@ function writeToFile (results) {
 
   return new Promise((resolve, reject) => {
     fs.access(filepath, fs.F_OK, (err) => {
-      if (!err) {
+      if (!err && process.argv[3] !== '--force') {
         return reject(new Error(`Release post for ${results.version} already exists!`))
       }
 


phillipj:
Seems like a good addition, have been annoyed by this myself once in a while..

phillipj (approved):

MylesBorins:
So the only thing I would maybe enjoy seeing is that if the force tag is given perhaps we can use the original date of the article? This may also be completely unnecessary. Thoughts?

mikeal:
I don't know enough about this process to give an up or down, but I will note that no matter what we do for updating the content we are still sitting behind a CloudFlare CDN and as a result there will be some time between a new deploy and cache invalidation of content we update.

italoacasas:
> we can use the original date of the article? 

@MylesBorins Is that the desired behavior?

IMO, the difference is going to be a few hours, I don't think is going to affect.

MylesBorins:
@italoacasas lets do the simple implementation, easy enough to drop a specific change with git

italoacasas:
@MylesBorins can you explain me what you mean with:

> easy enough to drop a specific change with git

PR #1155, opened 2017-02-22
Fix JavaScript error related to case study tracking

phillipj:
For the tracking code just merged to work properly, we need to wrap it into an [IIFE](http://benalman.com/news/2010/11/immediately-invoked-function-expression/) because of a `return` statement which is only valid inside a function. This is what's visible in the console on nodejs.org ATM:

screenshot1

That error prevents the case study tracking scripts from attaching its click listeners, resulting in the case study tracking not being invoked at all.

Refs https://github.com/nodejs/nodejs.org/pull/1148

### DIFF
--- a/layouts/partials/footer.hbs
+++ b/layouts/partials/footer.hbs
@@ -52,27 +52,29 @@
 
 <script src="/static/js/dnt_helper.js"></script>
 <script>
-  if (!_dntEnabled()) {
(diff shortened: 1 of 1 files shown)


phillipj:
Protip when looking at the diff, add `?w=1` to the URL excluding the noisy whitespace changes: https://github.com/nodejs/nodejs.org/pull/1155/files?w=1

lpinca (approved):

fhemberger:
@phillipj Whoops, sorry for that and thanks for fixing!

PR #1157, opened 2017-02-24
Terminate ubuntu installation commands with semicolons

briandk:
If a user tried to copy these commands---with the trailing dashes on
lines 45 and 52---and paste them to a terminal, the commands would most likely
fail.

I wasn't sure whether to just delete the dashes entirely, so I replaced them with semicolons to make it clearer that these are separate statements, not a wrapped line.


chrislea:
This is an unnecessary change. People have used these instructions millions of times and we've never had any issues with people thinking it was a wrapped line. Also, the final `-` character may be needed by some shells so we don't want to take that out.

briandk:
@chrislea thanks

Issue #1169, opened 2017-03-09
Please add Enterprise Conversation #10 to the Node.js Website under "Resources"


serenam18:
Hi Can you please add this new enterprise conversation to the website? 

**Emerging Technology Advisors (Episode 10)** 
Tracy and Mikeal chat with Chris Williams, founder of  Emerging Technology Advisors, about a variety projects in which companies have adopted Node.js during different stages of transition and growth.

https://youtu.be/w9gbcZJOgPY

Thanks so much for your help!
Serena 

* **URL**:
* **Browser version**:
* **Operating system**:

serenam18:
thank you!

lpinca:
@serenam18 no problem. PR is waiting to be reviewed.

PR #1170, opened 2017-03-10
Add enterprise conversation with Emerging Technology Advisors

lpinca:
This adds enterprise conversation with Emerging Technology Advisors as requested in #1169.


fhemberger:
Thanks!

Issue #1174, opened 2017-03-20
Help adding Call for Paper to Node.js website


ZibbyKeaton:
Hey All - We are trying to get more awareness around our call for presentations (CFP) for Node.js Interactive. Given the amount of traffic to the website, we thought it would be nice to have banner on the homepage of the website to focus on the CFP and attract folks that might not normally follow us on social.

I have an .eps [file here ](https://drive.google.com/drive/folders/0BwyeYdMvqcwFRkwzRWxzT3M0Zms?usp=sharing) and can also share a psd file. The CFP closes on April 10, so we are trying to get this up as soon as possible. Thanks as always for your help here. cc: @hackygolucky and @mikeal

lpinca:
Also cc: @nodejs/website

bnb:
@ZibbyKeaton I can't access that file 🤔

ZibbyKeaton:
@bnb I just added you to the doc. Let me know if you can access now?

ZibbyKeaton:
@bnb just poking to make sure you can access and use these files. Happy to try to get something else for you here.

lpinca:
Fixed by #1179.

lpinca:
@ZibbyKeaton is it possible to make a zip with the files and upload it here?

PR #1179, opened 2017-03-23
Add 2017 NI CfP banner

hackygolucky:
Per Zibby's request #1174, this adds the Node Interactive 2017 CfP banner, with styles echoing prior banners above the fold on the homepage.


lpinca:
There is one link in the banner which should be made clickable or removed imho.
Apart from this LGTM.

lpinca (approved):

lpinca:
Hmm, it doesn't look very good live. Maybe we should separate the two layers of the image.
One layer can be used as `background-image` of the `#interactive` `div` and the other as src of the `` tag inside the same `div`.

lpinca:
@hackygolucky is it possible to have the original image without the layers flattened?

hackygolucky:
@lpinca ah shoot, let me ask creative services. I requested the assets without them being flattened and I was handed a pdf :-/

lpinca:
Thanks.

hackygolucky:
@lpinca 
https://www.dropbox.com/s/qbustyrjynonhsu/banners_node_eps.eps?dl=0

lpinca:
Saved, thanks @hackygolucky.

PR #1207, opened 2017-04-18
Correct spelling

kellytk:


lpinca:
I _think_ [debuggee](https://en.wiktionary.org/wiki/debuggee) is correct.

PR #1216, opened 2017-05-03
docs: move and update inspector help

joshgav:
Per feedback in https://github.com/nodejs/node/pull/11207, copying Inspector guide to a simpler path and removing background info.

### DIFF
--- a/locale/en/docs/inspector.md
+++ b/locale/en/docs/inspector.md
@@ -1,0 +1,112 @@
+---
+title: Inspector Help
+layout: docs.hbs
+---
+
+# Debugging Node.js Apps
+
+Many tools and libraries are available to help you debug your Node.js apps. Some
+of these are listed below.
+
+To connect manually rather than with a tool, pass the **--inspect** flag and
+connect to the printed URL.
+
+If a process was started without `--inspect`, signal it with SIGUSR1 to
+activate the debugger and print the connection URL.
+
+---
+
+## Inspector Tools & Clients
+
+These commercial and open source tools make debugging Node.js apps easier.
+
+### [node-inspect](https://github.com/nodejs/node-inspect)
+
+* A CLI debugger developed at <https://github.com/nodejs/node-inspect>.
+* Bundled with Node and invoked with `node inspect myscript.js`.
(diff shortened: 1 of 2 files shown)


joshgav:
cc @nodejs/diagnostics

hybrist (approved):

lpinca (approved):

joshgav:
Landed in https://github.com/nodejs/nodejs.org/commit/13eaf4c0fa678aa08fe5a241e8fa8b875a045c43. Thanks!

ofrobots:
I am concerned this landed without review from folks from the DevTools team. /cc @eugeneo @pavelfeldman.

hybrist:
Sorry, didn't fully realize that we now have 2 copies of the "same" guide:

* https://github.com/nodejs/nodejs.org/blob/master/locale/en/docs/guides/debugging-getting-started.md
* https://github.com/nodejs/nodejs.org/blob/master/locale/en/docs/inspector.md

There's also overlap with the debugger docs page, specifically the ["Advanced Usage" section](https://github.com/nodejs/node/blob/master/doc/api/debugger.md#advanced-usage). It might be good to do another pass to unify & cross-link to make sure each of the URLs provides value and we don't duplicate information. Imo iterating on the docs shouldn't have to block the node PR though as long as it points to a valid URL.

joshgav:
@jkrems I left the old version up since it contains more detail in the first section which perhaps we can transition into a more complete guide. Such a guide could also include upcoming JS APIs (cc @eugeneo) and the Advanced usage you mentioned. Let's keep working on that and submit PRs with suggested updates.

The doc in this PR is meant to be as short as possible and focus primarily on available tools per [earlier feedback](https://github.com/nodejs/node/pull/11207#issuecomment-295331471), to help users get started if they specify --inspect manually. In some ways it's a replacement for the chrome-devtools URL originally provided.

Also it's to engender conversation about how to make this experience as good as possible for users, so thanks for the feedback 😄 

@ofrobots would be great to get more feedback and suggestions from the DevTools and Chromium teams, please make follow-on PRs and we can get them merged quickly. Thank you!

PR #1219, opened 2017-05-10
docs: fixes broken link in debugging guide

chris--young:
There is a link in the [debugging guide](https://nodejs.org/en/docs/guides/debugging-getting-started/#legacy-debugger) that is supposed to lead to the legacy debugging documentation but is currently giving a 404. This pr will fix that by linking to to latest version of the debugger docs.

fixes: https://github.com/nodejs/node/issues/12928

### DIFF
--- a/locale/en/docs/guides/debugging-getting-started.md
+++ b/locale/en/docs/guides/debugging-getting-started.md
@@ -156,7 +156,7 @@ couple popular ones are listed below.
 
 The V8 Debugging Protocol is no longer maintained or documented.
 
-#### [Built-in Debugger](https://github.com/nodejs/node/blob/master/lib/_debugger.js)
+#### [Built-in Debugger](https://nodejs.org/dist/latest/docs/api/debugger.html)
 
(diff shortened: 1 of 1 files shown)


gibfahn:
I think it probably makes more sense to just link to the `v6.x` version of that file, which should stick around permanently.

https://github.com/nodejs/node/blob/v6.x/lib/_debugger.js

chris--young:
@gibfahn makes sense, Node 8 will get rid of the legacy debugger. What do you think about this link https://nodejs.org/dist/latest-v6.x/docs/api/debugger.html instead of linking to GitHub?

lpinca:
> What do you think about this link https://nodejs.org/dist/latest-v6.x/docs/api/debugger.html instead of linking to GitHub?

I think it makes sense.

chris--young:
cool

lpinca (approved):

refack (approved):

gibfahn (approved):
Seems reasonable to me

PR #1245, opened 2017-05-31
Update v8.0.0.md

snirad:
should be api ?


fhemberger:
"ABI" is correct in this case: "application binary interface" for compiled C++ add-on binaries.
But thanks for submitting anyway!

gibfahn:
See: https://en.wikipedia.org/wiki/Application_binary_interface

snirad:
thanks :)

Issue #1247, opened 2017-05-31
Update "Installing Node.js via package manager" to Node.js 8


bnb:
* **URL**: https://nodejs.org/en/download/package-manager/

Just took a look at the Installing Node.js via package manager guide, and noticed that it still suggests [Node.js 7](https://nodejs.org/en/download/package-manager/#debian-and-ubuntu-based-linux-distributions) for Debian/Ubuntu.

Would it be possible to update these? 🤔

vsemozhetbyt:
A try to update : https://github.com/nodejs/nodejs.org/pull/1248

PR #1248, opened 2017-05-31
doc: update current version in package-manager.md

vsemozhetbyt:
Fixes: https://github.com/nodejs/nodejs.org/issues/1247

I've just formally replaced `(v)7` with `8` where it seemed concerning Node.js. Please, correct me where I am wrong, as I am not experienced in this area.

### DIFF
--- a/locale/en/download/package-manager.md
+++ b/locale/en/download/package-manager.md
@@ -46,10 +46,10 @@ curl -sL https://deb.nodesource.com/setup_6.x | sudo -E bash -
 sudo apt-get install -y nodejs
 ```
 
-Alternatively, for Node.js v7:
+Alternatively, for Node.js 8:
 
 ```bash
-curl -sL https://deb.nodesource.com/setup_7.x | sudo -E bash -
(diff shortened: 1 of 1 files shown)


locale/en/download/package-manager.md
```diff
@@ -358,7 +358,7 @@ Using **[MacPorts](http://www.macports.org/)**:
 port install nodejs<major version>
 
 # Example
-port install nodejs7
+port install nodejs8
```
lpinca: Not sure if this is already available. The package is not listed here https://www.macports.org/ports.php?by=name&substr=nodejs

vsemozhetbyt: Should I revert this?

lpinca: I think so unless you or someone who uses macport can confirm that the port is available.

vsemozhetbyt: Done.

lpinca (approved):

gibfahn (approved):

Issue #1249, opened 2017-06-01
remove the FAQ?


Trott:
* **URL**: https://nodejs.org/en/docs/faq/
* **Browser version**: n/a
* **Operating system**: n/a

I'd like to propose that we remove the FAQ. It is low-value and out of date. I'm sure there will be champions for updating it and keeping it, and that's all well and good. I prefer removing it because there's too many options in the nav sidebar right now (seven on that page), and removing this document that almost certainly won't have the question/answer combo someone expects is a good idea. We can always put it back if people actually have questions that they can't find answers for.

PR #1250, opened 2017-06-01
remove FAQ

Trott:
Fixes: https://github.com/nodejs/nodejs.org/issues/1249


lpinca:
It should also be removed from other locales.

sotayamashita:
@lpinca I will create a pull request for other locales FAQ so please merge it.

lpinca:
@sotayamashita we should see if there is consensus for this. I'm fine with removing the FAQ but I would like to know what other collaborators think.

fhemberger (approved):

fhemberger:
LGTM, let's remove it.

PR #1252, opened 2017-06-05
Added integration for OneSignal.push

ColinMoulds:
OneSignal is a high volume and reliable push notification service for websites and mobile applications.

Added base integration as an option to use OneSignal.push


fhemberger:
Hmm, not sure we need push notifications (someone has to trigger them as well).

Opinions, @nodejs/website?

ColinMoulds:
Yes, i thought it would be useful when announcing updated version of node directly through webpush.

MylesBorins:
I'm not seeing value in this. We don't release often enough / have enough engagement to introduce web push imho

sotayamashita:
I agree with @MylesBorins. 

I would like to know why @ColinMoulds create this issue because ColinMoulds just describes what OneSignal is. Thanks in advance.

lpinca:
I'm also -1 on this unless it is really needed.

PR #1276, opened 2017-06-28
Add Social Meta Tags to Site

bnb:
This PR adds social meta tags to the site. I've shared pages/posts from nodejs.org a bunch and they're always plain URLs. These changes allow it to be more rich all around.

It also adds a 256x256 png hexagon logo in a png format to serve as the icon for the rich embeds.

This _should_ be able to be merged as-is to enable this functionality. That said, it would be *awesome* if there's a template variable we could use to add descriptions to both OpenGraph and Twitter Cards. Not required, but a nice to have. ❤️


layouts/partials/html-head.hbs
```diff
@@ -12,6 +12,19 @@
 
   <link rel="apple-touch-icon" href="/static/apple-touch-icon.png">
   <link rel="icon" sizes="32x32" type="image/png" href="/static/favicon.png">
+  
+  <meta property="og:site_name" content="Node.js Foundation">
```
fhemberger: Site name is not Node.js **Foundation**

layouts/partials/html-head.hbs
```diff
@@ -12,6 +12,19 @@
...
+  <meta property="og:title" content="{{#if title}}{{ title }} | {{/if}}{{ site.title }}">
+  <meta name="og:image" content="/static/images/logo-hexagon.png">
+  <meta name="og:image:type" content="image/png">
+  <meta name="og:image:width" content="258">
```
fhemberger: Shouldn't it be 256 instead of 258?

fhemberger (changes requested):
Is the image compressed with ImageOptim or something similar?

layouts/partials/html-head.hbs
bnb: Was trying to think of something other than just `Node.js` since that's not the name of the site either, it's the tech. Foundation seemed like the next best extension, but not perfect. I'll change to just `Node.js` for now 👍

layouts/partials/html-head.hbs
bnb: Updating. 👍

bnb:
With the image I directly exported [this file](https://github.com/nodejs/nodejs.org/blob/master/static/images/logo-hexagon.svg) from Illustrator, and it didn't turn out very cleanly 😔

Tried exporting *again*, same result. Happy to add a replacement 256x256 icon, but I'm not sure how to export it properly.

lpinca:
LGTM with the image adjusted.

bnb:
Image should be fixed, updated the OG Image Width to the correct size (the logo itself isn't 1:1, apparently).

fhemberger (approved):

fhemberger:
Cool, thank you!

PR #1290, opened 2017-07-11
Display security announcement for july security releases

lpinca:
This adds a link to the security releases announcement to the home page as we did in the past.


fhemberger:
Thanks!

PR #1292, opened 2017-07-12
Fix grammar

ivandotv:
Simple grammar fix.


fhemberger:
@ivandotv Good catch, thank you!

PR #1300, opened 2017-07-15
Fix #1299

### LINKED ISSUE #1299: Package Manager Install instructions
* **URL**: https://nodejs.org/en/download/package-manager/
* **Browser version**: Google Chrome Version 59.0.3071.115 (Official Build) (64-bit)
* **Operating system**: Windows 10 Home

CentOS Command reads: curl --silent --location https://rpm.nodesource.com/setup_6.x | bash -

Command should read: curl --silent --location https://rpm.nodesource.com/setup_6.x | sudo bash -

You need root to access the rpm directory.  It is indicated for package install, however its also necessary to download.

pierreneter:


fhemberger (approved):

lpinca:
Text says to run the command as root:

> Run as root on RHEL, CentOS or Fedora, for Node.js v6 LTS:

so I don't think this is needed. We can remove the instructions for Node.js 0.10 though.

pierreneter:
I also think like you. But, for easy copying code. Most people forget switch to root user before run this code. 
Node.js 0.10 is End of Life. I will remove the intructions.

gibfahn (approved):
+1 to copy-pastable code

lpinca:
Ok then I would replace "Run as root on RHEL..." with "On RHEL...".

gibfahn:
While you're at it maybe add a sudo to the install line below?

```
sudo yum -y install nodejs
```

gibfahn (approved):
Thanks!

silverwind:
Not sure I like this. On RHEL, attempting to sudo as root [can fail](https://access.redhat.com/solutions/674723) with a "root is not in sudoers" error.

pierreneter:
Maybe We can warn them with line: `Run as root on RHEL, CentOS or Fedora, for Node.js v6 LTS`, keep it and bold it, okay? My fix just help for easy `copy-pastable code`. If they knew they are not at root user, they did not do so, why do they need paste and run it with `sudo`? Why are so concerned about non-advanced users? Do not discuss again, why to make it up, just a command line, make it easy for people to copy it. Why does @lpinca like to reject everything?

pierreneter:
And if this line doeen't have `sudo`. Can this non-root user run it without error?

lpinca:
@pierreneter I think you are misunderstanding. Emphasizing "Run as root" while using `sudo` will make the problem worse.

> Why does @lpinca like to reject everything?

Do I? That's definitely not my intention so sorry if you feel like I'm rejecting your work. That's not the case. I suggested a way to improve it by dropping "Run as root on RHEL..." in favor of "On RHEL..." here: https://github.com/nodejs/nodejs.org/pull/1300#issuecomment-315612674.

pierreneter:
I'm a CentOS/RHEL user, and I have never had that problem before.
The user called `root` be added to sudoers file is the `default`.
If someone can remove this user from this file, they also know how to run that command with a user has full rights.

silverwind:
I checked the sudo rpm of Fedora, and root is in there, so the issue with root not being in sudoers probably can only come from custom configuration, so I guess it's fine to recommend. Just thought I'd mention this issue.

pierreneter:
@lpinca 
Or maybe the line `Run as root` could be 
> Switch to the user that you are sure it has root and must be in the sudoers file with the configuration `{user} ALL = (ALL) ALL`

right?

gibfahn:
I _think_ we're all now in agreement that this is fine as it is. If someone has manually removed `root` from the sudoers file, then they probably know what to do when they get a sudo error. This should help the average user.

---

>Why does @lpinca like to reject everything?

In my experience @lpinca is one of the most open and accepting members of our org. Looking at this thread he started by saying he didn't think it was necessary, but then withdrew his objection almost immediately (and hasn't rejected anything since). Not really sure where your impression comes from.

pierreneter:
I'm new, and there are many things I have not experienced. Sorry, ._. forgive me if I have the wrong words.
@lpinca sorry for my words.

gibfahn:
@lpinca @silverwind does this LGTY now?

lpinca:
Imho suggesting to use `root` and `sudo` together doesn't make much sense but I'm fine with it.
@pierreneter no problem.

silverwind:
> Imho suggesting to use root and sudo together doesn't make much sense

I agree. Let's drop "Run as root", then it's LGTM.

sumnerevans:
Question, does this apply to other distros or only RHEL, CentOS, and Fedora?

lpinca:
@sumnerevans use of `sudo` is suggested also for other [distros](https://nodejs.org/en/download/package-manager/#debian-and-ubuntu-based-linux-distributions). Note that we don't suggest to use `root` in that case.

FYI these instructions are directly taken from https://github.com/nodesource/distributions.

fhemberger:
Removed the "run as root" note as requested, is everybody happy now with the change?

locale/en/download/package-manager.md
```diff
@@ -111,37 +111,31 @@ Node.js is available from the [NodeSource](https://nodesource.com) Enterprise Li
...
-```
-
-Alternatively for Node.js 0.10:
-
-```bash
-curl --silent --location https://rpm.nodesource.com/setup | bash -
+curl --silent --location https://rpm.nodesource.com/setup_8.x | sudo bash -
 ```
 
 Then install, as root:
```
lpinca: I'd remove `root` for the same reason.

fhemberger:
@lpinca Done.

lpinca (approved):

fhemberger:
Merged, thank you all!

PR #1306, opened 2017-07-25
Add CVE for constant hash seeds to annoucement

mhdawson:


fhemberger:
Hmm, the CVE is only listed as "reserved", but not yet as published. Should we update the post anyway?

mhdawson:
I've submitted the request to publish and already added the CVE to the mailing list.  The goal is to have the CVE number made public close to the time the CVE details are available, but as I understand it we will generally distribute the CVE in advance of the details being available from Mitre.

fhemberger:
Okay, thanks for the clarification. 👍 
Merged.

mhdawson:
Also just checked and details have been published :).  Thanks for landing.

PR #1309, opened 2017-07-25
Add SafetyCulture as silver foundation member

hackygolucky:
Adds logo and link to new member, including translated pages that exist.


stevemao (approved):

locale/uk/foundation/members.md
```diff
@@ -96,6 +96,9 @@ Node.js Foundation відкритий для бізнесу та окремих
 <a target="_new" class="imagelink" href="//yld.io">
   <img alt="yld!" class="memberlogo" src="/static/images/foundation/yldLogo.png" />
```
stevemao: You removed the `!` in en but not here. Should we keep the `!` or remove it everywhere?

stevemao (changes requested):

locale/uk/foundation/members.md
hackygolucky: Good catch.

lpinca (approved):

PR #1312, opened 2017-07-26
Node.js Foundation 2017 User Survey

ZibbyKeaton:
All - we are trying to post this to the website tomorrow. The original one that I posted here had data that is now invalid due to a question that we asked, luckily someone caught our mistake last minute.


lpinca:
@ZibbyKeaton should this replace https://github.com/nodejs/nodejs.org/blob/master/static/documents/casestudies/Nodejs_2017_User_Survey_Exec_Summary.pdf?

ZibbyKeaton:
@lpinca yes.

lpinca:
Ok I pushed a commit on your branch which replaces the original pdf with this version.

lpinca (approved):

Issue #1031, opened 2016-11-26
[redesign] What are we going to do about the events page.


mikeal:
So, I'm starting to wonder if the events page is actually useful or not.

It's dominated by a dump of meetups from meetup.com I scraped like a year ago and organizers don't appear to be using it at all.

We should talk about what we think the events page should offer and who it should be serving. If we can reach a consensus on where we want to go I'm up for helping implement it.

ryanmurakami:
I like the existing location based information. I think the only thing that is really missing/not helpful is some sort of calendar of events. For conferences, it would be super helpful. For meetups, perhaps the Meetup API could be used?

Also, it might help to put it in a different location on the nodejs.org site. It's kind of hidden behind the "Get Involved" menu, so I wonder how many people can even find it.

mikeal:
The current data is pulled from Meetup.com over a year ago. We don't have a great way to keep it updated from that feed or to figure out which meetups are actually still happening.

edsadr:
What about adopting a system like nodeschool.io? and maybe put some effort in contacting the events organizer so it could provide useuful information about next events and involve the community to maintain this content updated

id304m:
https://runkit.io/runkit/json-endpoint-example-1/branches/master?url=http://google.com

fhemberger:
I like the idea of listing conferences and nodeschool events, but I don't think we need to list every meetup possibly related to Node.js. Many of them are tagged wrong so we include a lot of unrelated events (e.g. Ember, Angular, Meteor) but still are missing some others. Plus the meetup.com API isn't well suited for retrieving and filtering such an event list and the results are hard to maintain.

ryanmurakami:
@fhemberger Yeah, I like that idea. In addition, I think meetup has a friendly interface that makes it easy to find local meetups, so we don't really need to provide an index of them all. I'm wondering, though, if some new users may be unfamiliar with even meetups. Perhaps we could just provide a call out  and link the the meetup site. Something like "There are many local meetups that focus on Node.js. Check out meetup.com to find one near you." Then we can keep out of the weeds while sharing a local resource for users.

lpinca:
I totally agree with @fhemberger.

vdeturckheim:
related to https://github.com/nodejs/evangelism/issues/251

CharlesBeeler:
[...]

PR #1314, opened 2017-07-26
Remove events page

lpinca:
This removes the event pages as per #1031.
The diff is big but it should be easy to review :)


fhemberger:
Thought we would keep the page (for conferences and Nodeschool events) and just remove the Meetups?

amandeepmittal:
I agree with @fhemberger with removing the meetups and keeping the page.

lpinca:
I'm fine with that but then we should update them regularly. I agree with what @Trott wrote here: https://github.com/nodejs/nodejs.org/issues/1031#issuecomment-307207622.

Trott (approved):
LGTM

bnb:
Saw this issue, and logged it in https://github.com/nodejs/community-events/issues/2. That issue there shouldn't be a blocker to this PR, but did want to raise the potential there for people to start getting involved at a future date, should they be interested 👍

Trott:
Not sure how exactly the process works here these days. Can this land? Or is this blocked by objections?

fhemberger:
@Trott We are thinking about keeping some kind of events list, but without the map and the list from meetup.com. Sorry, didn't have time to look into it yet.

Trott:
> @Trott We are thinking about keeping some kind of events list, but without the map and the list from meetup.com. Sorry, didn't have time to look into it yet.

👍 Thanks.

<sup>I'll just note such a thing could always be added back later. Leaving the current not-useful events page up while that gets sorted probably isn't doing a service to website users. OK, I'll stop now.</sup>

fhemberger:
@Trott Okay, we can come up later with a better solution.

PR #1316, opened 2017-07-28
Fix Docker tutorial

fhemberger:
Example uses the Node.js 6 Docker image with npm@3, so "npm install" won't
create a `package-lock.json`. If the reader follows the tutorial step-by-step,
it would not have been created even for npm@5, as a local `npm install` step
is missing.

### DIFF
--- a/locale/en/docs/guides/nodejs-docker-webapp.md
+++ b/locale/en/docs/guides/nodejs-docker-webapp.md
@@ -96,11 +96,14 @@ WORKDIR /usr/src/app
 This image comes with Node.js and NPM already installed so the next thing we
 need to do is to install your app dependencies using the `npm` binary. Please
 note that if you are using `npm` version 5 or later you will also want to copy
-`package-lock.json`:
+`package-lock.json`, which is gernerated once you run `npm install`:
 
 ```docker
 # Install app dependencies
-COPY package.json package-lock.json .
+COPY package.json .
+# For npm@5 or later, copy package-lock.json as well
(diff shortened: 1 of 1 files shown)


lpinca (approved):

locale/en/docs/guides/nodejs-docker-webapp.md
```diff
@@ -141,7 +144,10 @@ FROM node:boron
 WORKDIR /usr/src/app
 
 # Install app dependencies
-COPY package.json package-lock.json .
+COPY package.json .
+# For npm@5 or later, copy package-lock.json as well
```
lpinca, line 148: Can we use the same style for the comment? One blank line before, no blank line below. Or change the previous one to be consistent with this one :)

fhemberger: I did this on purpose, as it still belongs to the same block (and would replace line 147)

lpinca: Ok, then how about doing the same for line 103/105? Or am I missing something?
It doesn't really matter, it just triggers my OCD 😄.

fhemberger: Whoops, you're right. Should have been the other way round.

fhemberger: Done.

lpinca: :shipit:

PR #1326, opened 2017-08-05
Fix macOS package manager anchors.

arkwright:
I noticed that the OSX link at the top of https://nodejs.org/en/download/package-manager/ was broken, due to inconsistent terminology (OSX vs macOS).

I fixed the link and normalized all localized pages to use macOS instead of OSX.


pierreneter (approved):

lpinca (approved):

lpinca:
Thank you.

PR #1331, opened 2017-08-08
Update Inspector.md with reference to rawkit debugging tool

darcyclarke:
Added rawkit to the list of available/optional means of opening the Chrome Developer Tools for debugging.

### DIFF
--- a/locale/en/docs/inspector.md
+++ b/locale/en/docs/inspector.md
@@ -34,6 +34,8 @@ These commercial and open source tools make debugging Node.js apps easier.
   are listed. Then select your Node.js app from the list.
 * **Option 2**: Install the Chrome Extension NIM (Node Inspector Manager):
   https://chrome.google.com/webstore/detail/nim-node-inspector-manage/gnhhdgbaldcilmgcpfddgdbkhjohddkj
+* **Option 3**: Install the NPM package rawkit:
+  https://github.com/darcyclarke/rawkit
 
 ### [VS Code](https://github.com/microsoft/vscode) 1.10+
 


fhemberger (approved):

gibfahn:
First of all, thanks for making this. Judging by the [GIF in your README](https://github.com/darcyclarke/rawkit#run) installing and using this is a really slick experience, so kudos for that!

However I'm not entirely sure what the aim of this tool is, is it that it saves you going to [`chrome://inspect`](chrome://inspect) and clicking the `Open dedicated DevTools for Node` button? Once you have the devtools open the connection is automatic, so you shouldn't have to do any clicking, selecting, or copy-pasting.

Not saying we shouldn't add this, it's just that we already have a bunch of options, without much differentiation to help users work out which one they should use. 

It seems like the current set are:

| Name | Purpose |
| --- | --- |
| `node-inspect` | You want command-line debugging |
| `vscode`, `Visual Studio`, `Webstorm` | It's built into the IDE you're already using |
| `Chrome Devtools` | You're already using a browser for front-end debugging |
| `chrome-remote-interface` | API to call from your JS code (as the official one isn't ready yet) |
| `NIM`, `Rawkit` | Automatically opens `chrome://inspect -> dedicated devtools` ??? |

darcyclarke:
Hey @gibfahn! Appreciate the feedback. You have it right; the point of my project is to mitigate the headache around that initial cold/slow start opening the CDT (Chrome Developer Tools) to debug. I think having options is good, especially when developer's needs vary (thus, this is an optional item/installation underneath the Chrome Devtools section of this page). This tool essentially piggy-backs/builds off anyone that already uses, or wants to use, their browser to debug.

Chrome doesn't, and likely never will, have the support for immediate prompts. Alternatively, many of the legacy debugging tools did have this baked in (ie. `node-inspector`, `devtool` etc.).

NiM (*which is already listed*) and **rawkit**, although similar, approach the above very differently. NiM utilizes long polling; actively listening for ANY Node.js debugging ports to open and is SOLELY a Chrome extension. **rawkit** is meant to function as a CLI tool first and foremost and ONLY invokes the CDT when you specifically run a process through it. I think there's enough of a difference in each tool/approach to list them both (I could be wrong though, some folks seem to be liking it: https://twitter.com/darcy/status/894967404131995648). 

Hope that gives you some clarity/makes the case.

refack:
I'm with @gibfahn. IMHO too many options get diminished returns.
I see several things I would like to edit in this guide, but to minimize scope-creep, IMHO a better way of adding a reference to `rawkit` could be something like:
```markdown
## [Chrome DevTools](https://github.com/ChromeDevTools/devtools-frontend) 55+

1. Open `chrome://inspect` in a Chromium-based browser.
2. Click the Configure button and ensure your target host and port are listed.
3. Select your Node.js app from the list.

Several tools have been made available to streamline this process, e.g.:
* [NiM (Node Inspector Manager)](https://chrome.google.com/webstore/detail/nim-node-inspector-manage/gnhhdgbaldcilmgcpfddgdbkhjohddkj) a chrome extension.
* [`rawkit`](https://www.npmjs.com/package/rawkit) and npm installable CLI wrapper.
```

P.S. I replaced the rawkit GitHub link with an npm link, since I think it's "easier" to consume, but I have no strong opinion about that.

refack (commented):
IMHO the list should be coalesced and separate the configuration of Chrome from the available 4th party tools.

fhemberger:
Hi @darcyclarke, would you be okay with the changes requested by @refack?

fhemberger:
No feedback from original poster, closing this. Feel free to re-open.

PR #1332, opened 2017-08-09
Get data from google doc for Announcements

### LINKED ISSUE #1327: Automating Announcement Section
Hello - The Node.js Foundation is now issuing all of its announcements on Medium. Given this, I thought we should update our announcement section to be more like the "In the News" section. Essentially, I just put the title and link into a google document and the "In the News" section is automatically updated.

Is this group okay with this and if so, would anyone know how to do this? I'm sure it's fairly simple, I just don't know how to go about and create it.

pierreneter:
Enable show data from Google Doc.
close #1327


layouts/announcements.hbs
```diff
@@ -11,25 +11,72 @@
...
           </ul>
-
         </div>
       </article>
-
     </div>
   </div>
+  <script type="text/javascript" src="https://cdnjs.cloudflare.com/ajax/libs/tabletop.js/1.4.2/tabletop.min.js"></script>
+  <script type="text/javascript">
+  var public_spreadsheet_url = 'https://docs.google.com/spreadsheets/d/1xwQ6ciZLADL43My_XBIWLVRuDMl8cx_omYd1FRkBZ58/pubhtml';
```
fhemberger: We use the standard.js style across the nodejs.org website, please make sure client side scripts follow the same style.

layouts/in-the-news.hbs
```diff
@@ -25,16 +26,16 @@
       window.onload = function () {
```
fhemberger: We use the standard.js style across the nodejs.org website, please make sure client side scripts follow the same style.

fhemberger (commented):
Two little nits, otherwise LGTM. 👍 
Sorry for the delayed review!

layouts/announcements.hbs
pierreneter: It was copied from [layouts/in-the-news.hbs#L22](https://github.com/nodejs/nodejs.org/master/layouts/in-the-news.hbs#L22) to use layout as news sections. If something goes wrong I'll fix it on both pages.

pierreneter: It was copied from [layouts/in-the-news.hbs#L22](https://github.com/nodejs/nodejs.org/blob/master/layouts/in-the-news.hbs#L22) to use layout as news sections. If something goes wrong I'll fix it on both pages.

fhemberger: @pierreneter The code below now has semicolons at the end, for example, which is not the case for `in-the-news.hbs` but had other issues fixed in #1352.

Could you please fix that so it validates against standard style?
See https://standardjs.com/demo.html for a quick check.

fhemberger (changes requested):

lpinca:
Can you also please remove yarn.lock? We already have package-lock.json or do we want to keep both?

layouts/announcements.hbs
pierreneter: Yes. I will push a commit tonight

pierreneter:
@fhemberger @lpinca all fix, thanks.

pierreneter:
@fhemberger @lpinca all fixed, thanks.

pierreneter (commented):
Sorry for my shortcoming, thank you.

fhemberger (approved):

fhemberger:
Great, thank you!

PR #1355, opened 2017-09-05
fix silver section head and paypal location

gtewallace:
Got some weirdness going on with the PayPal logo placement (they should be listed with Gold Members just after Nodesource) and also the Silver heading is messed up - I think we were just lacking a div

Appreciate a review. Thanks!


fhemberger:
Thanks!

PR #1357, opened 2017-09-10
Merge Website WG Members and Collaborators 🎉

fhemberger:
I don't think we need a separation of WG members and collaborators any longer. We started with that setup two years ago when the WG was just established and needed a bit more of planning and direction through WG meetings.

We gave up on the meetings pretty quickly, as there were no topics to be discussed which couldn't be handled as simple GitHub issues. Also thanks to all our **absolutely amazing and dedicated** contributors, there never was the need for the WG to resolve any disputes.

So today, there isn't really anything that sets those two groups apart. That's why I vote to officially join WG and collaborators.

Ref: https://github.com/nodejs/community-committee/issues/121 (/cc @bnb @MylesBorins)

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,13 +61,6 @@ content area.
 
 ### Website Working Group Members
 
-- Frederic Hemberger ([fhemberger](https://github.com/fhemberger)): [@fhemberger](https://twitter.com/fhemberger), `mail``@``frederic-hemberger.de`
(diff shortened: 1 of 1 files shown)


gibfahn (approved):
SGTM

phillipj:
LGTM 👍

MylesBorins:
LGTM

lpinca (approved):

PR #1359, opened 2017-09-11
Update Directors list for officer changes

hackygolucky:


fhemberger:
@hackygolucky I think you can go on and merge such small changes to Foundation content directly, as we do for release blog posts.

PR #1368, opened 2017-09-19
fix up references related to TSC/CTC merge

mhdawson:


locale/en/foundation/tsc/index.md
```diff
@@ -22,26 +22,9 @@ can be found on the project's [GitHub profile](https://github.com/orgs/nodejs/pe
...
-
-* Anna Henningsen ([addaleax](https://github.com/addaleax))
-* Ben Noordhuis ([bnoordhuis](https://github.com/bnoordhuis))
-* Bert Belder ([piscisaureus](https://github.com/piscisaureus))
-* Bryan Hughes ([nebrius](https://github.com/nebrius))
-* Chris Dickinson ([chrisdickinson](https://github.com/chrisdickinson))
-* Jeremiah Senkpiel ([Fishrock123](https://github.com/Fishrock123))
-* Myles Borins ([MylesBorins](https://github.com/MylesBorins))
+For the current list of TSC members, see the project
+[README.md](https://github.com/nodejs/node/blob/master/README.md#tsc-technical-steering-committee).
```
fhemberger, line 26: Hmm, I think it would be nice to have the members listed here instead of linking to the GitHub repository.

lpinca: The advantage of the current approach is that syncing is not needed but I'm also fine with copying the list here.

fhemberger: Yes I know, but the website is the primary entry point for people, who want to learn about the Node.js Foundation, not a GitHub repo. I know it has been challenging to keep information about the various bodies in sync, but it would be nice if we could work something out in the future.

It would be nice to have a separate `MEMBERS.md` in the repo which we could include on build.

fhemberger:
@mhdawson I'm willing to merge this, but it would really be great if we could work out a different solution here.

fhinkel:
Can we merge this so that the info is not super outdated? If not, I've prepared a PR that lists the (almost) current members - but keeping them in sync seems like unnecessary work.

fhemberger:
@fhinkel I'm merging it now, but "the info is super outdated" vs. "keeping them in sync seems like unnecessary work" is unfortunately exactly the problem. The foundation bodies seem to be forgetting that there still is this website … happened a number of times already.

fhinkel:
Sorry, I worded that pretty poorly. Thanks for merging!

mhdawson:
@fhemberger do you know enough about markdown to know if we can inline ?   If there was some way to have a file which had just the list of people as the master list and then that was inlined into the README.md and the website that would avoid duplicate work and make sure the list is consistent.

fhemberger:
@mhdawson Unfortunately, you can't inline markdown into other markdown files.
We could add a HTML comment into the `README.md` to extract the TSC members:

```
<!-- BEGIN nodejs.org/tsc/members -->
...
<!-- END nodejs.org/tsc/members -->
```

And include that section on the website.

PR #1372, opened 2017-09-20
Using curl behind proxy: for nodejs#1371

peter-lyons-kehl:


locale/en/download/package-manager.md
```diff
@@ -7,6 +7,8 @@ title: Installing Node.js via package manager
 
 ***Note:*** The packages on this page are maintained and supported by their respective packagers, **not** the Node.js core team. Please report any issues you encounter to the package maintainer. If it turns out your issue is a bug in Node.js itself, the maintainer will report the issue upstream.
 
+When running ```curl``` behind a proxy, invoke it as ```curl --proxy http://proxy.racp.edu.au:3128```.
```
fhemberger: Hmm, not sure if this change is really required, I guess people used to `curl`ing behind a proxy usually now their required config switches. Thoughts @nodejs/website?

a) Please use singleticks for inline code.
b) Maybe you should use a more generic proxy URL (something like `http://<hostname>:<port>` would be okay as well).

fhemberger (changes requested):

peter-lyons-kehl:
Hi Frederic,

This will save time even for people who have used `curl` behind a proxy. That's because of 2 stage installation, where in the first step `curl` downloads a script, which invokes `curl` again. That second step can be confusing.

a) changed to singleticks for inline code
b) made the instruction proxy-generic

silverwind:
Honestly, I see no need for this. curl honours environment variables like HTTP_PROXY, which is generally the preferred method to enable proxy usage system-wide.

fhemberger:
Personally, I agree with @silverwind, but would like to hear @nodejs/website's thoughts.

lpinca:
I also think that the usual way is to use the `HTTP_PROXY` env variable. I'm -0 on adding this.

fhemberger:
Okay, I've closed this issue. But thanks for your PR, sorry it didn't work out this time.

PR #1386, opened 2017-10-01
doc: replace http with https in link from guide

vsemozhetbyt:
Requests for http://stackoverflow.com/ are redirected to https://stackoverflow.com/ anyway.


fhemberger:
Thanks!

PR #1396, opened 2017-10-11
outreachy: rename good first contribution label

Fishrock123:
Is now `good first issue`.

Refs: https://github.com/nodejs/node/issues/16149


fhemberger:
@Fishrock123 Should this be the official label name for *all* Node.js repositories?

PR #1402, opened 2017-10-14
Remove snyk

lpinca:
We can't integrate Snyk to the org and manually updating the .snyk file
is not feasible. The last update dates back to August 2016. Dependencies
have been updated since then but the policy file remained the same
making it useless.

Refs: https://github.com/nodejs/nodejs.org/pull/841
Refs: https://github.com/nodejs/nodejs.org/issues/1094


refack:
> We can't integrate Snyk to the org

Is that still the case even after the sensitive repos were moved out?

lpinca:
@refack I don't know.

fhemberger:
Updated `package-lock.json` with latest from master (w/o snyk deps), which unfortunately broke the diff.

I'm merging this now:
- I still don't know what the status for 3rd party GitHub tools is in this org
- Removing Snyk only has a very low security impact, as we just push out static HTML and don't run any actual Node.js code in production. Problems during build time would affect developer machines and build server in the same manner, so hopefully they're caught before committing.
- If we change our mind, we can always add it back later without problems.

PR #1425, opened 2017-10-26
blog: fix indentation

maclover7:
cc @cjihrig


cjihrig (approved):
LGTM, but isn't all of this auto generated?

maclover7:
Welp, turns out the issue is that the changelog in nodejs/node for v8.8.1 is indented more than it should be... who knew two spaces could cause so much trouble... This should go in, but also looks the changelog will need to be fixed as well.

cjihrig:
Ah, my bad. I'm going to land this.

maclover7:
no worries :) opened https://github.com/nodejs/node/pull/16507 to fix this in nodejs/node

PR #1428, opened 2017-10-30
nodejs-docker-webapp - Add comment for production build

allevo:
as titled

If it is ok, I'll reproduce this changes in italian language (or using google translate in the others)

### DIFF
--- a/locale/en/docs/guides/nodejs-docker-webapp.md
+++ b/locale/en/docs/guides/nodejs-docker-webapp.md
@@ -105,6 +105,8 @@ COPY package.json .
 # COPY package.json package-lock.json ./
 
 RUN npm install
+# If you are building your code for production
+# RUN npm install --only=production
 ```
 
 Note that, rather than copying the entire working directory, we are only copying
@@ -149,6 +151,8 @@ COPY package.json .
 # COPY package.json package-lock.json ./
 
 RUN npm install
+# If you are building your code for production
+# RUN npm install --only=production
 
 # Bundle app source
 COPY . .


locale/en/docs/guides/nodejs-docker-webapp.md
```diff
@@ -105,6 +105,8 @@ COPY package.json .
 # COPY package.json package-lock.json ./
 
 RUN npm install
+# If you are building your code for production
+# RUN npm install --only=production
```
lpinca, line 109: Isn't it better to use the `NODE_ENV` env var?

allevo: Good point but an environment variable affects the following commands.
But I'm ok for changing it using `ARG NODE_ENV=production`

lpinca: It can be set when running `docker run` as per [best practices](https://github.com/nodejs/docker-node/blob/master/docs/BestPractices.md#docker-run).

allevo: Yes but `npm install` is made during the build time. The "-e" option is available only at run time.

If you prefer I can suggest to build the image using `--build-args NODE_ENV=production` in the best practise

lpinca: Yes you are right, I have no strong opinions.

lpinca (approved):

daxlab (approved):

ghost (approved):
Thank you for contributing!

PR #1441, opened 2017-11-02
doc: remove ppc BE from download matrix

mhdawson:
We removed Linux PPC **BE** from the supported list for Node Version 8.x
Remove from the download matrix now that both LTS and current are >= 8.x.
NOTE:  Linux PPC le is still supported and remains in the matrix.


mhdawson:
For reference the supported platforms for 8.x https://github.com/nodejs/node/blob/v8.x/BUILDING.md

ghost (approved):

daxlab (approved):

lpinca (approved):

PR #1444, opened 2017-11-03
Add first of the new diagnostics articles to guides

naugtur:
This is the first one of a few guides I would like to propose to add here.
There's more: https://github.com/naugtur/node-diagnostics-howtos

The goal is to make knowledge about diagnostics tools more approachable and gather it in one place. 

I'm setting up this PR to discuss how we want to do this. Please share thoughts on the format in general. 
As discussed in recent @nodejs/diagnostics meeting, we'd like to proceed with more of those guides, all of them written in a similar fashion.

Related: https://github.com/nodejs/community-committee/issues/163

### DIFF
--- a/locale/en/docs/guides/diagnostics-flamegraph.md
+++ b/locale/en/docs/guides/diagnostics-flamegraph.md
@@ -105,7 +105,15 @@ For details see:
 
 Node.js 10.x addresses the issue with Turbofan using the`--interpreted-frames-native-stack` flag.
 
-Run `node --interpreted-frames-native-stack --perf-basic-prof-only-functions` to get finction names in the flame graph regardless of which pipeline V8 used to compile your JavaScript.
+Run `node --interpreted-frames-native-stack --perf-basic-prof-only-functions` to get function names in the flame graph regardless of which pipeline V8 used to compile your JavaScript.
+
+### broken labels in flamegraph
+
+If you're seeing labels looking like this
+```
+node`_ZN2v88internal11interpreter17BytecodeGenerator15VisitStatementsEPNS0_8ZoneListIPNS0_9StatementEEE
+```
+it means the Linux perf you're using was not compiled with demangle support, see https://bugs.launchpad.net/ubuntu/+source/linux/+bug/1396654 for example
 
 
 ## Examples
--- a/locale/en/docs/guides/diagnostics-flamegraph.md
+++ b/locale/en/docs/guides/diagnostics-flamegraph.md
@@ -14,9 +14,7 @@ Flame graphs are a way of visualizing CPU time spent in functions. They can help
 You might have heard creating a flame graph for Node.js is difficult, but that's not true (anymore).
 Solaris vms are no longer needed for flame graphs on linux!
 
-1. install lldb in your system (eg. with apt-get)
-1. install llnode according to readme in [github project](https://github.com/nodejs/llnode)
-1. install linux-tools-common (eg. with apt-get)
+1. install linux-tools-common - the package containing `perf` (eg. with apt-get)
 1. try running `perf` - it might complain about missing kernel modules, install them too
 1. install stackvis `npm i -g stackvis`
 1. run node with perf enabled `perf record -e cycles:u -g -- node --perf-basic-prof app.js`
@@ -36,7 +34,7 @@ perf record -F99 -p `pgrep -n node` -g -- sleep 3
 
 Wait, what is that `sleep 3` for? It's there to keep the perf running - despite `-p` option pointing to a different pid, the command needs to be executed on a process and end with it. Just a convenient way to keep perf going for a given time.
 
-Why is `-F` (profiling frequency) set to 99? I honestly don't know, Netflix guys used that value. Results look good. You can adjust if you want.
+Why is `-F` (profiling frequency) set to 99? It's a reasonable default. You can adjust if you want. Lower values should produce less output with less precise results.
 
 After you get that 3 second perf record, proceed with generating the flame graph with last two steps from above.
 
@@ -56,8 +54,10 @@ Why do I need them at all?
 
 Well, without these options you'll still get a flame graph, but with most bars labeled `v8::Function::Call`.
 
+## Node.js 8.x V8 pipeline changes
+
+Node.js 8.x ships with new optimizations to JavaScript compilation pipeline in V8 engine which makes function names/references unreachable for perf. The result is you might not get your function names right in the flame graph. See this thread for details: https://github.com/nodejs/benchmarking/issues/168
+
 ## Examples
 
-See a real flame graph - download the file from /samples/flame-graph-example.html
-
-Or even better, practice capturing it yourself with [a flame graph exercise](https://github.com/naugtur/node-example-flamegraph)
+Practice capturing a flame graph yourself with [a flame graph exercise](https://github.com/naugtur/node-example-flamegraph)
--- a/locale/en/docs/guides/diagnostics-flamegraph.md
+++ b/locale/en/docs/guides/diagnostics-flamegraph.md
@@ -1,0 +1,63 @@
+---
+title: Diagnostics - Flame Graphs
+layout: docs.hbs
+---
+
+# Flame Graphs
+
+## What's a flame graph useful for?
+
+Flame graphs are a way of visualizing CPU time spent in functions. They can help you pin down where you spend too much time doing synchronous operations.
+
+## How to create a flame graph
+
+You might have heard creating a flame graph for Node.js is difficult, but that's not true (anymore).
+Solaris vms are no longer needed for flame graphs on linux!
+
+1. install lldb in your system (eg. with apt-get)
+1. install llnode according to readme in [github project](https://github.com/nodejs/llnode)
+1. install linux-tools-common (eg. with apt-get)
+1. try running `perf` - it might complain about missing kernel modules, install them too
+1. install stackvis `npm i -g stackvis`
+1. run node with perf enabled `perf record -e cycles:u -g -- node --perf-basic-prof app.js`
+1. disregard warnings unless they're saying you can't run perf due to missing packages; you may get some warnings about not being able to access kernel module samples which you're not after anyway.
+1. run `perf script`, but pipe it through some cleanup: `perf script | egrep -v "( __libc_start| LazyCompile | v8::internal::| Builtin:| Stub:| LoadIC:|\[unknown\]| LoadPolymorphicIC:)" | sed 's/ LazyCompile:[*~]\?/ /' > perfs.out` [explanation](#about-perf-script-filtering)
+1. run `stackvis perf < perfs.out > flamegraph.htm`
+
+Now open the flame graph file in your favorite browser and watch it burn. It's color-coded so you can focus on the most saturated orange bars first. They're likely to represent CPU heavy functions.
+
+Worth mentioning - if you click an element of a flame graph a zoom-in of its surroundings will get displayed above the graph.
+
+### Use perf to sample a running process
+
+```
+perf record -F99 -p `pgrep -n node` -g -- sleep 3
+```
+
+Wait, what is that `sleep 3` for? It's there to keep the perf running - despite `-p` option pointing to a different pid, the command needs to be executed on a process and end with it. Just a convenient way to keep perf going for a given time.
+
+Why is `-F` (profiling frequency) set to 99? I honestly don't know, Netflix guys used that value. Results look good. You can adjust if you want.
+
+After you get that 3 second perf record, proceed with generating the flame graph with last two steps from above.
+
+### About `perf script` filtering
+
+The long line of greps and seds after `perf script` call is there to remove some V8 and node internals so that it's easier for you to focus on your JavaScript calls. These are only important if you are looking for an issue with Node internals.
+
+If you read your flame graph and it seems odd, as if something is missing in the key function taking up most time, try generating your flame graph without the filters - maybe you got a rare case of an issue with Node itself.
+
+### Node's profiling options
+
+`--perf-basic-prof-only-functions` and `--perf-basic-prof` seem like the only two you might be initially interested in for debugging your JavaScript code. Other options should only be useful for profiling Node itself, which is outside the scope of this howto.
+
+`--perf-basic-prof-only-functions` produces less output, so it's the option with least overhead.
+
+Why do I need them at all?
+
+Well, without these options you'll still get a flame graph, but with most bars labeled `v8::Function::Call`.
+
+## Examples
+
+See a real flame graph - download the file from /samples/flame-graph-example.html
+
+Or even better, practice capturing it yourself with [a flame graph exercise](https://github.com/naugtur/node-example-flamegraph)
--- a/locale/en/docs/guides/diagnostics-flamegraph.md
+++ b/locale/en/docs/guides/diagnostics-flamegraph.md
@@ -32,10 +32,10 @@ Now let's get to work.
 
 1. Install `perf` (usually available through the linux-tools-common package if not already installed)
 2. try running `perf` - it might complain about missing kernel modules, install them too
-3. run node with perf enabled 
+3. run node with perf enabled (see [perf output issues](#perf-output-issues) for tips specific to node versions)
 ```bash
 perf record -e cycles:u -g -- node --perf-basic-prof app.js
-```
+``` 
 4. disregard warnings unless they're saying you can't run perf due to missing packages; you may get some warnings about not being able to access kernel module samples which you're not after anyway.
 5. Run `perf script > perfs.out` to generate the data file you'll visualize in a moment. It's useful to [apply some cleanup](#filtering-out-node-internal-functions) for a more readable graph 
 6. install stackvis if not yet installed `npm i -g stackvis`
@@ -68,7 +68,7 @@ Usually you just want to look at the performance of your own calls, so filtering
 
 ```bash
 sed -i \
-  -e "s/( __libc_start| LazyCompile | v8::internal::| Builtin:| Stub:| LoadIC:|\[unknown\]| LoadPolymorphicIC:)/d" \
+  -e "/( __libc_start| LazyCompile | v8::internal::| Builtin:| Stub:| LoadIC:|\[unknown\]| LoadPolymorphicIC:)/d" \
   -e 's/ LazyCompile:[*~]\?/ /' \
   perfs.out
 ```
@@ -89,16 +89,23 @@ Well, without these options you'll still get a flame graph, but with most bars l
 
 ### Node.js 8.x V8 pipeline changes
 
-Node.js 8.x and above ships with new optimizations to JavaScript compilation pipeline in V8 engine which makes function names/references unreachable for perf sometimes. The result is you might not get your function names right in the flame graph. 
+Node.js 8.x and above ships with new optimizations to JavaScript compilation pipeline in V8 engine which makes function names/references unreachable for perf sometimes. (It's called Turbofan) 
 
-You'll notice `ByteCodeHandler:` instead of function names in your flame graph.
+The result is you might not get your function names right in the flame graph. 
 
-[0x](https://www.npmjs.com/package/0x) has some fixes for that built in. 
+You'll notice `ByteCodeHandler:` where you'd expect function names.
 
+[0x](https://www.npmjs.com/package/0x) has some mitigations for that built in. 
 
 For details see:
 - https://github.com/nodejs/benchmarking/issues/168
 - https://github.com/nodejs/diagnostics/issues/148#issuecomment-369348961
+
+### Node.js 10+ 
+
+Node.js 10.x addresses the issue with Turbofan using the`--interpreted-frames-native-stack` flag.
+
+Run `node --interpreted-frames-native-stack --perf-basic-prof-only-functions` to get finction names in the flame graph regardless of which pipeline V8 used to compile your JavaScript.
 
 
 ## Examples
--- a/locale/en/docs/guides/diagnostics-flamegraph.md
+++ b/locale/en/docs/guides/diagnostics-flamegraph.md
@@ -9,44 +9,75 @@ layout: docs.hbs
 
 Flame graphs are a way of visualizing CPU time spent in functions. They can help you pin down where you spend too much time doing synchronous operations.
 
-## How to create a flame graph
+## How to create a flame graph 
 
 You might have heard creating a flame graph for Node.js is difficult, but that's not true (anymore).
-Solaris vms are no longer needed for flame graphs on linux!
+Solaris vms are no longer needed for flame graphs!
 
-1. install linux-tools-common - the package containing `perf` (eg. with apt-get)
-1. try running `perf` - it might complain about missing kernel modules, install them too
-1. install stackvis `npm i -g stackvis`
-1. run node with perf enabled `perf record -e cycles:u -g -- node --perf-basic-prof app.js`
-1. disregard warnings unless they're saying you can't run perf due to missing packages; you may get some warnings about not being able to access kernel module samples which you're not after anyway.
-1. run `perf script`, but pipe it through some cleanup: `perf script | egrep -v "( __libc_start| LazyCompile | v8::internal::| Builtin:| Stub:| LoadIC:|\[unknown\]| LoadPolymorphicIC:)" | sed 's/ LazyCompile:[*~]\?/ /' > perfs.out` [explanation](#about-perf-script-filtering)
-1. run `stackvis perf < perfs.out > flamegraph.htm`
+Flame graphs are generated from `perf` output, which is not a node-specific tool. While it's the most powerful way to visualize CPU time spent, it may have issues with how JavaScript code is optimized in Node 8 and above. See [perf output issues](#perf-output-issues) section below.
+
+### Use a pre-packaged tool
+
+If you want a single step that produces a flame graph locally, try [0x](https://www.npmjs.com/package/0x)
+
+For diagnosing production deployments, read these notes: [0x production servers](https://github.com/davidmarkclements/0x/blob/master/docs/production-servers.md)
+
+### Create a flame graph with system perf tools
+
+The purpose of this guide is to show steps involved in creating a flame graph and keep you in control of each step.
+
+If you want to understand each step better take a look at the sections that follow were we go into more detail.
+
+Now let's get to work.
+
+1. Install `perf` (usually available through the linux-tools-common package if not already installed)
+2. try running `perf` - it might complain about missing kernel modules, install them too
+3. run node with perf enabled 
+```bash
+perf record -e cycles:u -g -- node --perf-basic-prof app.js
+```
+4. disregard warnings unless they're saying you can't run perf due to missing packages; you may get some warnings about not being able to access kernel module samples which you're not after anyway.
+5. Run `perf script > perfs.out` to generate the data file you'll visualize in a moment. It's useful to [apply some cleanup](#filtering-out-node-internal-functions) for a more readable graph 
+6. install stackvis if not yet installed `npm i -g stackvis`
+7. run `stackvis perf < perfs.out > flamegraph.htm`
 
 Now open the flame graph file in your favorite browser and watch it burn. It's color-coded so you can focus on the most saturated orange bars first. They're likely to represent CPU heavy functions.
 
 Worth mentioning - if you click an element of a flame graph a zoom-in of its surroundings will get displayed above the graph.
 
-### Use perf to sample a running process
+### Using perf to sample a running process
+
+This is great for recording flame graph data from an already running process that you don't want to interrupt. Imagine a production process with a hard to reproduce issue.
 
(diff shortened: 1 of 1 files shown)


locale/en/docs/guides/diagnostics-flamegraph.md
```diff
@@ -0,0 +1,63 @@
...
+
+Flame graphs are a way of visualizing CPU time spent in functions. They can help you pin down where you spend too much time doing synchronous operations.
+
+## How to create a flame graph
+
+You might have heard creating a flame graph for Node.js is difficult, but that's not true (anymore).
+Solaris vms are no longer needed for flame graphs on linux!
+
+1. install lldb in your system (eg. with apt-get)
+1. install llnode according to readme in [github project](https://github.com/nodejs/llnode)
```
naugtur: I'm not 100% sure entire lldb and llnode installation is necessary, maybe a subset would suffice? I had trouble getting a flame graph without them, but maybe it's about one of the dependencies.

locale/en/docs/guides/diagnostics-flamegraph.md
```diff
@@ -0,0 +1,63 @@
...
+
+### Use perf to sample a running process
+
+```
+perf record -F99 -p `pgrep -n node` -g -- sleep 3
+```
+
+Wait, what is that `sleep 3` for? It's there to keep the perf running - despite `-p` option pointing to a different pid, the command needs to be executed on a process and end with it. Just a convenient way to keep perf going for a given time.
+
+Why is `-F` (profiling frequency) set to 99? I honestly don't know, Netflix guys used that value. Results look good. You can adjust if you want.
```
naugtur: Too honest. This part will have to change. On the other hand, I think we should provide a reasonable default here instead of a lengthy explanation of what -F really does and how to decide on a value.

locale/en/docs/guides/diagnostics-flamegraph.md
```diff
@@ -0,0 +1,63 @@
...
+
+`--perf-basic-prof-only-functions` produces less output, so it's the option with least overhead.
+
+Why do I need them at all?
+
+Well, without these options you'll still get a flame graph, but with most bars labeled `v8::Function::Call`.
+
+## Examples
+
+See a real flame graph - download the file from /samples/flame-graph-example.html
```
naugtur: How are other resources stored? This is a path from https://github.com/naugtur/node-diagnostics-howtos. This guide should have a link to an example. nodejs/nodejs.org repo doesn't seem to be a place for those.

locale/en/docs/guides/diagnostics-flamegraph.md
```diff
@@ -0,0 +1,63 @@
+---
+title: Diagnostics - Flame Graphs
+layout: docs.hbs
+---
+
+# Flame Graphs
+
+## What's a flame graph useful for?
```
naugtur, line 8: Explaining why before explaining how - part of my intended pattern for those guides.

locale/en/docs/guides/diagnostics-flamegraph.md
```diff
@@ -0,0 +1,63 @@
...
+Flame graphs are a way of visualizing CPU time spent in functions. They can help you pin down where you spend too much time doing synchronous operations.
+
+## How to create a flame graph
+
+You might have heard creating a flame graph for Node.js is difficult, but that's not true (anymore).
+Solaris vms are no longer needed for flame graphs on linux!
+
+1. install lldb in your system (eg. with apt-get)
+1. install llnode according to readme in [github project](https://github.com/nodejs/llnode)
+1. install linux-tools-common (eg. with apt-get)
```
mhdawson: It sounds like this  might be to get perf.  If so it would be good to say that specifically so that we know why we need each of the pieces that we are asking people to install.

naugtur: Right, I'll update it. thanks.

mhdawson:
We probably need to give a heads up on https://github.com/nodejs/benchmarking/issues/168 as well as it seems 8.x has issues due to the pipeline change.

locale/en/docs/guides/diagnostics-flamegraph.md
mhdawson: I agree that I don't think we need either of those 2 specifically.  I think its just the perf output that is being used.

locale/en/docs/guides/diagnostics-flamegraph.md
```diff
@@ -0,0 +1,63 @@
...
+1. try running `perf` - it might complain about missing kernel modules, install them too
+1. install stackvis `npm i -g stackvis`
+1. run node with perf enabled `perf record -e cycles:u -g -- node --perf-basic-prof app.js`
+1. disregard warnings unless they're saying you can't run perf due to missing packages; you may get some warnings about not being able to access kernel module samples which you're not after anyway.
+1. run `perf script`, but pipe it through some cleanup: `perf script | egrep -v "( __libc_start| LazyCompile | v8::internal::| Builtin:| Stub:| LoadIC:|\[unknown\]| LoadPolymorphicIC:)" | sed 's/ LazyCompile:[*~]\?/ /' > perfs.out` [explanation](#about-perf-script-filtering)
```
mhdawson: It would be good to have the part you want to cut and paste on its own line.  In this case it I included the 'explanation' part in my cut and paste and of course if failed.  So while people can figure that out making it easier to get it right the first time makes sense to me.

mhdawson:
Overall this is a great doc.  I was able to follow it and get flamegraphs out easily within 5-10 mins.  

My main suggestions would be a bit more explanation about what is going on in each of the steps.  Having said that you already have some of that later on so maybe I just needed to read further down.  Maybe a line up front that says 'if you want to understand each step better take a look at the sections that follow were we go into more detail'

locale/en/docs/guides/diagnostics-flamegraph.md
lpinca: An external link should work.

locale/en/docs/guides/diagnostics-flamegraph.md
```diff
@@ -0,0 +1,63 @@
...
+
+You might have heard creating a flame graph for Node.js is difficult, but that's not true (anymore).
+Solaris vms are no longer needed for flame graphs on linux!
+
+1. install lldb in your system (eg. with apt-get)
+1. install llnode according to readme in [github project](https://github.com/nodejs/llnode)
+1. install linux-tools-common (eg. with apt-get)
+1. try running `perf` - it might complain about missing kernel modules, install them too
+1. install stackvis `npm i -g stackvis`
+1. run node with perf enabled `perf record -e cycles:u -g -- node --perf-basic-prof app.js`
```
lpinca: I guess the number at the beginning reflect a step in the process. If so they should be incremented at every step.

lpinca:
It might make sense to mention [0x](https://github.com/davidmarkclements/0x) which makes creating flames graph very easy.

locale/en/docs/guides/diagnostics-flamegraph.md
naugtur: It's how markdown works. Generated HTML will have an ordered list here.
I could switch to incremental so the MD file is more readable, but it's not meant to be read without rendering.

naugtur:
@lpinca 0x looks great, thanks for mentioning it.

 I'm not sure if showing a one step opinionated tool is helpful for the educational purposes. Maybe a separate sections with links to all tools suits this article, but it's a maintenance cost to curate the list and keep adding new and removing deprecated. 

It's just my opinion, let's wait and see what others think.

locale/en/docs/guides/diagnostics-flamegraph.md
lpinca: I didn't know 😮. Always used `1. 2. 3. ...`. Thanks.

fhemberger:
@naugtur Hi, what's the status of this PR? Is something still missing? Do you need help/feedback?

naugtur:
Hi,

I'm going to address the minor issues above as soon as I get out of the newborn baby situation at home, sorry for the delays, there's been complications. 

From my point of view this work is ready for publication, but I'm not even a native speaker. I'm looking forward to editing/style support as discussed in nodejs/community-committee#163

Oh, and https://github.com/nodejs/benchmarking/issues/168 is a serious issue with this guide unfortunately. It will need a section added at least.

locale/en/docs/guides/diagnostics-flamegraph.md
gibfahn: I tend to spend a lot of time reading markdown files in the raw (not least in PR review), so I'd appreciate incremental numbering.

locale/en/docs/guides/diagnostics-flamegraph.md
gibfahn: +1 on a code block, and I think it's worth repeating the whole perf line:

````markdown
1. Run your script through `perf` again, removing V8 internals and redirecting output to a file.
```bash
perf record -e cycles:u -g -- node --perf-basic-prof app.js > perfs.out
```

````

I'd suggest moving the grep and sed into your paragraph below. I think it's a good idea, and worth mentioning, but it makes the initial example more complex, and as you note later it isn't _always_ useful.

**Comment by fhemberger** [comment]
@naugtur Congratulations and all the best for your kid! :tada:

File: locale/en/docs/guides/diagnostics-flamegraph.md [hunk 8]
```diff
@@ -0,0 +1,63 @@
...
+
+### About `perf script` filtering
+
+The long line of greps and seds after `perf script` call is there to remove some V8 and node internals so that it's easier for you to focus on your JavaScript calls. These are only important if you are looking for an issue with Node internals.
+
+If you read your flame graph and it seems odd, as if something is missing in the key function taking up most time, try generating your flame graph without the filters - maybe you got a rare case of an issue with Node itself.
+
+### Node's profiling options
+
+`--perf-basic-prof-only-functions` and `--perf-basic-prof` seem like the only two you might be initially interested in for debugging your JavaScript code. Other options should only be useful for profiling Node itself, which is outside the scope of this howto.
```
gibfahn: Maybe:

>`--perf-basic-prof-only-functions` and `--perf-basic-prof` seem like the only two you might be initially interested in for debugging your JavaScript code. Other options should only be useful for profiling Node itself, which is outside the scope of this howto.

to 

>`--perf-basic-prof-only-functions` and `--perf-basic-prof` are the two that are useful for debugging your JavaScript code. Other options are used for profiling Node itself, which is outside the scope of this howto.

locale/en/docs/guides/diagnostics-flamegraph.md
```diff
@@ -0,0 +1,63 @@
...
+
+Worth mentioning - if you click an element of a flame graph a zoom-in of its surroundings will get displayed above the graph.
+
+### Use perf to sample a running process
+
+```
+perf record -F99 -p `pgrep -n node` -g -- sleep 3
+```
+
+Wait, what is that `sleep 3` for? It's there to keep the perf running - despite `-p` option pointing to a different pid, the command needs to be executed on a process and end with it. Just a convenient way to keep perf going for a given time.
```
gibfahn: This wasn't immediately clear to me. Maybe something like:

>`perf` runs for the life of the command you pass to it, whether or not you're actually profiling that command. `sleep 3` ensures that `perf` runs for 3 seconds.

locale/en/docs/guides/diagnostics-flamegraph.md
```diff
@@ -0,0 +1,63 @@
...
+## What's a flame graph useful for?
+
+Flame graphs are a way of visualizing CPU time spent in functions. They can help you pin down where you spend too much time doing synchronous operations.
+
+## How to create a flame graph
+
+You might have heard creating a flame graph for Node.js is difficult, but that's not true (anymore).
+Solaris vms are no longer needed for flame graphs on linux!
+
+1. install linux-tools-common - the package containing `perf` (eg. with apt-get)
```
gibfahn: Maybe:

>Install `perf` (usually available through the `linux-tools-common` package if not already installed).

locale/en/docs/guides/diagnostics-flamegraph.md
```diff
@@ -0,0 +1,63 @@
...
+1. try running `perf` - it might complain about missing kernel modules, install them too
+1. install stackvis `npm i -g stackvis`
```
gibfahn: Might be helpful to explain why:

>Install `stackvis` with `npm i -g stackvis`, to generate html flamegraphs from the perf results.

naugtur:
I just pushed some tweaks I had laying around and a stub of the section explaining 8.x V8 issues. It's likely not good enough, but should get us started.

locale/en/docs/guides/diagnostics-flamegraph.md
```diff
@@ -0,0 +1,63 @@
...
+1. run node with perf enabled `perf record -e cycles:u -g -- node --perf-basic-prof app.js`
+1. disregard warnings unless they're saying you can't run perf due to missing packages; you may get some warnings about not being able to access kernel module samples which you're not after anyway.
```
gibfahn: Maybe simplify to:

>Disregard non-fatal warnings, e.g. those about accessing kernel module samples.

locale/en/docs/guides/diagnostics-flamegraph.md
```diff
@@ -0,0 +1,63 @@
...
+
+## How to create a flame graph
+
+You might have heard creating a flame graph for Node.js is difficult, but that's not true (anymore).
+Solaris vms are no longer needed for flame graphs on linux!
+
+1. install linux-tools-common - the package containing `perf` (eg. with apt-get)
+1. try running `perf` - it might complain about missing kernel modules, install them too
+1. install stackvis `npm i -g stackvis`
+1. run node with perf enabled `perf record -e cycles:u -g -- node --perf-basic-prof app.js`
```
gibfahn: Would be nice to break this out too:

````markdown
1. Run your script through `perf` to ensure everything is setup correctly:
```bash
perf record -e cycles:u -g -- node --perf-basic-prof app.js
```

````

File: locale/en/docs/guides/diagnostics-flamegraph.md [hunk 14]
```diff
@@ -0,0 +1,63 @@
...
+
+### Use perf to sample a running process
+
+```
+perf record -F99 -p `pgrep -n node` -g -- sleep 3
+```
+
+Wait, what is that `sleep 3` for? It's there to keep the perf running - despite `-p` option pointing to a different pid, the command needs to be executed on a process and end with it. Just a convenient way to keep perf going for a given time.
+
+Why is `-F` (profiling frequency) set to 99? It's a reasonable default. You can adjust if you want. Lower values should produce less output with less precise results.
```
gibfahn: Maybe:

>`-F99` tells `perf` to take 99 samples per second, for more precision increase the value.

locale/en/docs/guides/diagnostics-flamegraph.md
```diff
@@ -0,0 +1,63 @@
...
+
+After you get that 3 second perf record, proceed with generating the flame graph with last two steps from above.
```
gibfahn: `last two` -> `the last two`

locale/en/docs/guides/diagnostics-flamegraph.md
```diff
@@ -0,0 +1,63 @@
...
+
+### About `perf script` filtering
+
+The long line of greps and seds after `perf script` call is there to remove some V8 and node internals so that it's easier for you to focus on your JavaScript calls. These are only important if you are looking for an issue with Node internals.
```
gibfahn: As I mentioned above, I'd suggest putting the grep/sed in here:

````markdown

### Filtering out Node internal functions

Usually you just want to look at the performance of your own calls, so filtering out Node and V8 internal functions can make the graph much easier to read. You can clean up your perf file with:

```bash
sed -i \
  -e "s/( __libc_start| LazyCompile | v8::internal::| Builtin:| Stub:| LoadIC:|\[unknown\]| LoadPolymorphicIC:)/d" \
  -e 's/ LazyCompile:[*~]\?/ /' \
  perfs.out
```

````

File: locale/en/docs/guides/diagnostics-flamegraph.md [hunk 17]
```diff
@@ -0,0 +1,63 @@
...
+
+# Flame Graphs
+
+## What's a flame graph useful for?
+
+Flame graphs are a way of visualizing CPU time spent in functions. They can help you pin down where you spend too much time doing synchronous operations.
+
+## How to create a flame graph
+
+You might have heard creating a flame graph for Node.js is difficult, but that's not true (anymore).
```
AndreasMadsen, line 14: I support this a lot as a technical documentation for how to make a flame graph using only basic primitives. However, I wouldn't call this easy. I someone asked me how to make flame graphs easily I would defer to [0x](https://github.com/davidmarkclements/0x).

locale/en/docs/guides/diagnostics-flamegraph.md
```diff
@@ -0,0 +1,63 @@
...
+
+### Node's profiling options
+
+`--perf-basic-prof-only-functions` and `--perf-basic-prof` seem like the only two you might be initially interested in for debugging your JavaScript code. Other options should only be useful for profiling Node itself, which is outside the scope of this howto.
+
+`--perf-basic-prof-only-functions` produces less output, so it's the option with least overhead.
+
+Why do I need them at all?
+
+Well, without these options you'll still get a flame graph, but with most bars labeled `v8::Function::Call`.
```
AndreasMadsen, line 86: To be more detailed. They contain a map between the _code addresses_ and the JS function names, collected by walking stack. I have not tried it, but do you really get more `v8::Function::Call` frames? I would suspect that you simply don't see the JS function frames, as `pref` doesn't know how to resolve the symbols for those _code address_.

locale/en/docs/guides/diagnostics-flamegraph.md
```diff
@@ -0,0 +1,63 @@
+---
+title: Diagnostics - Flame Graphs
```
AndreasMadsen: What is this for?

locale/en/docs/guides/diagnostics-flamegraph.md
naugtur: Hehe, easy in comparison. It's a second mention of 0x, so it's time I added a mention. thx

locale/en/docs/guides/diagnostics-flamegraph.md
naugtur: I did generate a flame graph with all javascript functions labeled `v8::Function::Call` if I didn't pass `--perf-*` - that's what it does. 

`--perf-basic-prof` is probably doing more than just exposing the function names from JS, but I don't know enough about it to elaborate.

mike-kaufman:
@naugtur  - per https://github.com/nodejs/community-committee/issues/163 discussion, can you add `docs` and `help-wanted` labels to this as a start?

fhemberger:
@naugtur What's the status of this PR? Is something still missing or are we already discussing details and enhancements?

I'd love to get this merged, if there's nothing blocking, we still can iron out little nits in subsequent PRs.

gibfahn (commented):
Some thoughts/suggestions (from someone who isn't familiar with this stuff).

naugtur:
@gibfahn thanks for suggestions. I'm hoping to make another pass of editing this weekend. Will incorporate your tips.

locale/en/docs/guides/diagnostics-flamegraph.md
```diff
@@ -0,0 +1,63 @@
...
+
+`--perf-basic-prof-only-functions` and `--perf-basic-prof` seem like the only two you might be initially interested in for debugging your JavaScript code. Other options should only be useful for profiling Node itself, which is outside the scope of this howto.
+
+`--perf-basic-prof-only-functions` produces less output, so it's the option with least overhead.
+
+Why do I need them at all?
+
+Well, without these options you'll still get a flame graph, but with most bars labeled `v8::Function::Call`.
+
+## Node.js 8.x V8 pipeline changes
```
mmarchini: Maybe move this warning to the begin of the guide? Otherwise, people will read the entire guide to find out this technique may not work for them. Furthermore, they may get "weird" `BytecodeHandler::*` stuff in the Flame Graph while following the guide (see https://github.com/nodejs/benchmarking/issues/168#issuecomment-341803777)

fhemberger:
@naugtur Hey, I just wanted to ask how we should proceed with this PR. Could you please look at the comments and see if we can update the guide?

fhinkel:
ping @naugtur

fhemberger:
ping @naugtur 

@nodejs/diagnostics Could you please help landing this?

locale/en/docs/guides/diagnostics-flamegraph.md
sam-github: someone who hasn't heard its difficult now has, maybe this will make people nervous before they even started? I third the `0x` reference

locale/en/docs/guides/diagnostics-flamegraph.md
sam-github: I'm not a grammatician, but I read the "These" of the second sentence to refer to the most recently mentioned thing, "JavaScript calls", maybe be more explicit: The internals are only important if you are ....

naugtur:
Hi, I'm so sorry I couldn't get back to this. 
I'm still trying to carve out the time after becoming a father. I should be able to come back to this doc in the next 2 weeks.

fhemberger:
@naugtur Hi, no worries! Take your time. I was hoping that someone from @nodejs/diagnostics could help out a bit, as it looks like there are only some details left to be done.

naugtur:
@fhemberger there were some corner cases with node8 to add or work around and I'm not sure if they propagate to v10.

Anyway, as soon as I arrange to get back to it I'm hoping to finish this one and start another.

locale/en/docs/guides/diagnostics-flamegraph.md
naugtur: The errors you get did sound pretty fatal to me, but they only matter if they say perf is missing some dependency. I went through figuring it out and I don't think I would understand what you mean by "non-fatal" before I figured it out.

naugtur:
@fhemberger I've addressed the comments and reorganized the document a bit. Hope it's reasonable.

Next, I'll make sure the guide actually works after all those changes and report back :)

mmarchini:
Just an FYI: there's a new flag (`--interpreted-frames-native-stack`) on Node.js v10.x to address the "missing frames" issue introduced with Turbofan.

fhemberger:
@naugtur Awesome, thank you so much!

naugtur:
@mmarchini Great! I'll have to read up on that and add another edit.

naugtur:
Hi, I've added interpreted-frames-native-stack to the doc. 
It could be ready to publish, but I have some minor concerns.
- sed has different syntax on mac, maybe we should go back to egrep
- I wanted to test it on all versions of node, but now I'm getting flamegraphs without function names and with *node`_ZN...* instead. With the same commands that used to work. Looks like there's one more factor I'm not taking into account here and it changed with some updates I got.

mcollina:
@naugtur You'll need to add `--interpreted-frames-native-stack` on all commands to work and show the functions.

mmarchini:
> I wanted to test it on all versions of node, but now I'm getting flamegraphs without function names and with node`_ZN... instead.

The Linux perf you're using was not compiled with demangle support, see https://bugs.launchpad.net/ubuntu/+source/linux/+bug/1396654. IIRC this if fixed on Ubuntu 18.04.

naugtur:
> The Linux perf you're using was not compiled with demangle support, see https://bugs.launchpad.net/ubuntu/+source/linux/+bug/1396654. IIRC this if fixed on Ubuntu 18.04.

I saw that bug and:
1. I want to mention it in the guide
2. I wonder why it worked before - I'm running it on the same system as 1yr ago. It might be due to a kernel update.

Once that's done, I'm hoping to :ship: it :)

locale/en/docs/guides/diagnostics-flamegraph.md
```diff
@@ -0,0 +1,113 @@
...
+
+For details see:
+- https://github.com/nodejs/benchmarking/issues/168
+- https://github.com/nodejs/diagnostics/issues/148#issuecomment-369348961
+
+### Node.js 10+ 
+
+Node.js 10.x addresses the issue with Turbofan using the`--interpreted-frames-native-stack` flag.
+
+Run `node --interpreted-frames-native-stack --perf-basic-prof-only-functions` to get finction names in the flame graph regardless of which pipeline V8 used to compile your JavaScript.
```
mhdawson: typo finction -> function

naugtur: fixed

locale/en/docs/guides/diagnostics-flamegraph.md
```diff
@@ -0,0 +1,121 @@
...
+1. Install `perf` (usually available through the linux-tools-common package if not already installed)
+2. try running `perf` - it might complain about missing kernel modules, install them too
+3. run node with perf enabled (see [perf output issues](#perf-output-issues) for tips specific to node versions)
+```bash
+perf record -e cycles:u -g -- node --perf-basic-prof app.js
+``` 
+4. disregard warnings unless they're saying you can't run perf due to missing packages; you may get some warnings about not being able to access kernel module samples which you're not after anyway.
+5. Run `perf script > perfs.out` to generate the data file you'll visualize in a moment. It's useful to [apply some cleanup](#filtering-out-node-internal-functions) for a more readable graph 
+6. install stackvis if not yet installed `npm i -g stackvis`
+7. run `stackvis perf < perfs.out > flamegraph.htm`
```
mhdawson, line 42: I tried this today and the result does not display nicely for me in the browser.  Does it still work ok for you?

mmarchini: Maybe we should suggest using [Brendan Gregg's flamegraph scripts](http://www.brendangregg.com/blog/2014-09-17/node-flame-graphs-on-linux.html) instead? Stackvis result is quite unreadable here as well:

`stackvis perf < perfs.out > flamegraph.htm `:

image

`./FlameGraph/stackcollapse-perf.pl  < ./perfs.out  | ./FlameGraph/flamegraph.pl --color=js > ./flamegraph.svg`:

image

mhdawson: I did follow that suggestion in the demo I was putting together for my NodeConfEU presentation.  FlameGraph worked for me.  Once we close on the content we should see if we can figure out some testing to make sure our recommendation continues to work properly.

naugtur: That's exactly why I'm not saying this is ready to publish - it started showing me the same issue you have on the screenshot. It seemed an issue with perf output, so I upgraded my OS and retired, got the same thing, ran out of time. I'll check other options, but it's good to see the output from perf seems ok and it's about the visualization tool. 

I'm hoping to look for reasons in stackvis and explore other options. IMHO it'd be nice to use something one can install from npm after all.

mhdawson: @naugtur thanks for the update.

naugtur: I've compared various flamegraph generators and the difference in readabiliy is mostly due to overflow:hidden of long trace names, but the input data is still incorrect. I switched to ubuntu 18.04.1 and it still produces bad stack traces. 

Planning to check on a few systems and compare.

And BTW, 0x didn't work for me either.

mmarchini: > but the input data is still incorrect

Incorrect how? If you could share an example I can happily take a look :)

naugtur: https://naugtur.egnyte.com/fl/6GGszZUQFy 
Every tool I used to generate a flame graph on my machine had the same issues

fhemberger:
@naugtur you have been working on this for quite a while, so I'd really love to get this merged. I don't want to stress you if you're busy, but what is still needed for this PR to land? Maybe someone can help.

We could also see that we merge the current version, and add further details later if this works for you.

naugtur:
@fhemberger The last thing I was hoping to fix here is issues with displaying function names I'm experiencing. I've tried this on 2 different Ubuntu versions I use on daily basis and didn't get the nice names on either of them. 
I've failed to find time to install a clean new system and prove it works. 

If anybody can go through the steps and confirm they're seeing a nice flame graph with correct function names - I'd consider it ready.

I'm also ok with merging this as-is because the part that's not working for me seems to require a fix independent of the whole procedure.

fhemberger:
@naugtur Okay, then I'm merging it. If someone comes up with a nicer flame graph, this can be added in a separate PR.

Thank you so much for your work!

naugtur:
Thanks, sorry for the delays. 
_Noone's ever really ready for kids ;)_

ghost:
@nodejs/localization: Feel free to translate to your own languages if possible :)

AliSawari:
@Maledong thanks 
I will do my part once I have the chance, really busy these days :))

PR #1446, opened 2017-11-03
d/l: 32bit linux is experimental

refack:
Refs: https://github.com/nodejs/build/issues/885

### DIFF
--- a/layouts/partials/primary-download-matrix.hbs
+++ b/layouts/partials/primary-download-matrix.hbs
@@ -64,12 +64,18 @@
         <th>macOS Binaries (.tar.gz)</th>
         <td colspan="6"><a href="https://nodejs.org/dist/{{version.node}}/node-{{version.node}}-darwin-x64.tar.gz">64-bit</a></td>
       </tr>
-
+      {{#if versionTypeCurrent }}
       <tr>
-        <th>Linux Binaries (x86/x64)</th>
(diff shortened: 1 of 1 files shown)


seishun:
What about when 32-bit builds stop being built?

refack:
> What about when 32-bit builds stop being built?

IMHO new box should be removed.

seishun:
Which new box?

refack:
> Which new box?

The new link in `secondary-download-matrix.hbs`
image

refack:
Cross-ref: https://github.com/nodejs/node/pull/16723

seishun:
Ah, agree.

gibfahn (dismissed):
Blocking so this doesn't get landed accidentally before https://github.com/nodejs/node/pull/16723 is resolved.

gibfahn:
I think we should call out that it's experimental, I'm not even sure that it should continue to be on the main downloads page. I think we want to make sure people who are automatically downloading 32-bit builds notice the change (and come tell us that they still use 32-bit).

refack:
I completely removed the listing of the 32 bit linux binary. If people will come looking for it hope they will raise an issue that it is missing.

seishun:
@gibfahn unless it's required that Tier 1 and Tier 2 platforms must have public builds, I think this can be unblocked now.

gibfahn:
>@gibfahn unless it's required that Tier 1 and Tier 2 platforms must have public builds, I think this can be unblocked now.

I think it's reasonable to say Tier 1 platforms should have public builds where possible.

fhemberger:
Just merge it whenever you're ready.

PR #1463, opened 2017-11-09
nodejs-ja: add links to index

yuryu:
nodejs-ja
Add links and translation for the LTS schedule and the newsletter signup page to the index page.


watilde (approved):

watilde:
Found out there is a period in the template [layouts/index.hbs#L67](https://github.com/nodejs/nodejs.org/blob/master/layouts/index.hbs#L67) which should be part of locale data instead. I will arrange that after merging this.

PR #1467, opened 2017-11-13
update docker webapp guide for carbon

jbottigliero:
- updates refrences of boron for carbon
- updates express package version in example (4.16.1 is latest stable)
- updates Dockerfile `COPY` (and instructions) to assume `npm@5+` and warn about `npm@4-`
- updates various stdout examples

### DIFF
--- a/locale/en/docs/guides/nodejs-docker-webapp.md
+++ b/locale/en/docs/guides/nodejs-docker-webapp.md
@@ -36,10 +36,14 @@ create a `package.json` file that describes your app and its dependencies:
     "start": "node server.js"
   },
   "dependencies": {
-    "express": "^4.13.3"
+    "express": "^4.16.1"
   }
 }
 ```
+
+With your new `package.json` file, run `npm install`. If you are using `npm`
+version 5 or later, this will generate a `package-lock.json` file which will be copied
+to your Docker image.
 
 Then, create a `server.js` file that defines a web app using the
 [Express.js](http://expressjs.com/) framework:
@@ -78,11 +82,11 @@ touch Dockerfile
 Open the `Dockerfile` in your favorite text editor
 
 The first thing we need to do is define from what image we want to build from.
-Here we will use the latest LTS (long term support) version `boron` of `node`
+Here we will use the latest LTS (long term support) version `carbon` of `node`
 available from the [Docker Hub](https://hub.docker.com/):
 
 ```docker
-FROM node:boron
+FROM node:carbon
 ```
 
 Next we create a directory to hold the application code inside the image, this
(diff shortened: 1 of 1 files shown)


locale/en/docs/guides/nodejs-docker-webapp.md
```diff
@@ -140,15 +145,15 @@ CMD [ "npm", "start" ]
...
+FROM node:carbon
 
 # Create app directory
 WORKDIR /usr/src/app
 
 # Install app dependencies
-COPY package.json .
-# For npm@5 or later, copy package-lock.json as well
-# COPY package.json package-lock.json ./
+COPY package.json package-lock.json ./
```
jbottigliero: I considered making this a wildcard to account for both cases... but thought this was more straight-forward and consistent what was here before.

The following will _not_ fail if the `package-lock.json` file does not exist (using an older version of `npm` or if `npm install` was not run before `docker build ...`:
```
COPY package.json package-*.json ./
```

jbottigliero:
If this looks good, I can at the very least update the examples/stdout in the translation files if that is desired.

locale/en/docs/guides/nodejs-docker-webapp.md
fhemberger: At this point, `COPY .` should be enough IMHO. No need to explicitly copy `package*.json` again.

fhemberger (changes requested):

fhemberger:
Otherwise LGTM.

locale/en/docs/guides/nodejs-docker-webapp.md
jbottigliero: Just to clarify, are you suggesting to forego the explicit coping of `package*.json` files and simply `COPY . .` _before_ the `npm install` (removing the `# Bundle app source` command)? Resulting in:
```
FROM node:carbon

# Create app directory
WORKDIR /usr/src/app

# Bundle app source
COPY . .

# Install app dependencies
RUN npm install
# If you are building your code for production
# RUN npm install --only=production

EXPOSE 8080
CMD [ "npm", "start" ]
```

Or, replacing the replacing the `COPY package.json package-lock.json ./` with `COPY package*.json .`?

fhemberger: Sorry, my mistake, got a bit mixed up in the diff. I meant 

> replacing the replacing the `COPY package.json package-lock.json ./` with `COPY package*.json .`?

jbottigliero:
Updated to use `COPY pacakge*.json .`.

For the additional locale versions, is it worth updating the `boron` references and docker commands leaving the current comments/verbiage intact? 

> note that if you are using `npm` version 5 or later you will also want to copy
 `package-lock.json`, which is generated once you run `npm install`:

The above seems like it would be the only misleading copy from the original (translated) version.

fhemberger:
@jbottigliero Sounds good to me. :+1:

fhemberger (approved):

lpinca (approved):

hiroppy (approved):

jbottigliero:
I've opened #1476 adding the updates [mentioned in this comment](https://github.com/nodejs/nodejs.org/pull/1467#issuecomment-344810234) along a few URL updates.

Issue #1471, opened 2017-11-15
"Don't block the event loop": New guide on avoiding DoS


davisjam:
*We know, we know, "Don't block the event loop"*
Node.js event loop follows the event-driven architecture, with a single-threaded Event Loop supported by a small Worker Pool for expensive operations like FS and DNS.

Though "Don't block the event loop" is a good rule of thumb, developers new to the EDA could easily expose themselves to this and other Denial of Service vulnerabilities if they don't think carefully about the implications of the limited supply of threads. For example, [REDOS](https://www.owasp.org/index.php/Regular_expression_Denial_of_Service_-_ReDoS) is a very real threat but a bit of a surprise as a way to block the event loop.

There are of course a lot of blog posts describing this issue in varying levels of formality and detail, but I think the community would benefit from a guide put out by a central authority -- like nodejs.org!

*Also don't block the threadpool*
Since libuv's threadpool is tiny (default 4 threads, max 128), cavalierly offloading hugely expensive work to the threadpool is also not a good idea. Continuing with the REDOS example, unfortunately, this is done by some of the "safe" regex modules.

*More on this if you're interested*
I've discussed this a bit in an academic workshop paper ([link](http://www.sharcs-project.eu/m/documents/papers/a08-davis.pdf)), but since then I've done a larger survey of the npm vulnerabilities reported by [Snyk.io](https://snyk.io/vuln?packageManager=npm) and believe that the Node community would benefit from a public guide to this problem.

*PR*
I'm happy to prepare this guide and submit a PR if there's interest. However, I can only prepare this document in English.

davisjam:
I was hoping for some discussion here about whether or not there's interest. I haven't contributed to this project before -- am I supposed to prepare the guide first, so people can see what I have in mind?

Tiriel:
Well, TBH I think there would indeed be more to discuss if you had a draft. Though anyway, I'm probably not the most qualified to discuss the technical quality of such draft, I'd very much like to read it and the discussion that might follow.

And also no worries, globally, "no discussion" generally means "go ahead"; people tend a bit more to comment to say when they don't agree than when they agree, from what I've seen :smile:

davisjam:
>people tend a bit more to comment to say when they don't agree than when they agree

Ain't that the truth. I'll work on a draft, thank you.

davisjam:
@Tiriel and others, see #1478.

Let's continue this discussion over there.

davisjam:
Addressed by #1478.

PR #1478, opened 2017-11-19
Guide on "Don't block the event loop"

davisjam:
Towards #1471.

Not finished yet, hoping to get some initial feedback.


Tiriel (commented):
Haven't finished ready in entirety yet, but one general opinion yet: I think you should go for a more formal tone.
This sounds too much like a blog post to me, not like an "official" guide. That may not be the general feeling of everyone, but I for one would prefer a more neutral tone on a guide found on Node.js official website.

Coming soon: complete review, although I'm not as technical as some!

davisjam:
Thanks for the comment on tone. I will try a more formal approach in my next revision.

Thoughts about the technical content?

locale/en/docs/guides/dont-block-the-event-loop.md
```diff
@@ -0,0 +1,135 @@
...
+For this reason, you should be leery of using regular expressions to validate user input.
+
+There are some tools to check your regexps for safety, like
+- [safe-regex](https://github.com/substack/safe-regex)
+- [rxxr2](http://www.cs.bham.ac.uk/~hxt/research/rxxr2/)
+However, neither of these will catch all vulnerable regexps.
+
+A safer alternative is the [node-re2](https://github.com/uhop/node-re2) project, which uses Google's blazing-fast [RE2](https://github.com/google/re2) regexp engine. But be warned, RE2 is not 100% compatible with Node's regexps, so check for regressions if you swap in the node-re2 module to handle your regexps. And really complicated regexps may not be supported by node-re2.
+
+As usual, if you're trying to match something "obvious", like a URL or a file path, use an npm module!
```
Tiriel: Quick question here, mostly for my own education: 
In that case, wouldn't `stringVar.indexOf(searchVar) !== -1` or `stringVar.includes(searchVar) !== false` be more efficient, and eliminating the need of another module?

I originally come from PHP, and replacing `preg_match()` with `stripos($string, $search) !== false` is mostly preferred for the same reasons, so I thought it would be the same in JS.

locale/en/docs/guides/dont-block-the-event-loop.md
```diff
@@ -0,0 +1,135 @@
...
+- Do I feel confident enough in my C++ skills to write a [C++ addon](https://nodejs.org/api/addons.html)? If so, consider offloading this work to the Worker Pool. But see the next section for more on that.
+- Would it be better to ask a [Child Process](https://nodejs.org/api/child_process.html) to do this for me?
+
+If you're in a hurry, though, here's a two-step way to handle this situation:
+1. Do part of your task, and then...
+2. defer the rest until later, e.g. with `setImmediate`.
+
+This gives all of the other pending clients a turn.
+It's pretty easy to accomplish this by sticking whatever state you're carrying around in a closure or an object.
+And you might want to add a TODO to revisit this section of your code later.
```
Tiriel: Some example might be good here. I think most newcomers/basic users won't necessarily have a clue about what you're talking here. The notion of 'immediate' in the event loop can be misleading and some people won't understand why it would partly solve the problem IMHO.

locale/en/docs/guides/dont-block-the-event-loop.md
```diff
@@ -0,0 +1,135 @@
...
+
+## Don't block the Worker Pool
+Work in progress.
```
Tiriel: Damn, that was the piece I was most eager to read :smile:

Tiriel (commented):
As said, I'm not the most technical reviewer you could get, I'm largely a noob in Node.js and JS itself. I still hope my comments will help!

vsemozhetbyt:
@davisjam It seems your local email in git settings does not match your GitHub one. Can you add a local one to GitHub email list so that your contributions to Node.js repositories can be associated with your GitHub account?

mail

davisjam:
@vsemozhetbyt Thanks, always forget this.

locale/en/docs/guides/dont-block-the-event-loop.md
davisjam: If you're just doing a `strcmp` or `substr`-style string match, then yes, you should just use `indexOf`. You don't need a regular expression for this, nor will a regexp actually be dangerous in this context.

But sometimes people need more matching power for more complicated expressions, and a sufficiently complicated regexp can lead to REDOS.

locale/en/docs/guides/dont-block-the-event-loop.md
davisjam: Good idea. I've been thinking about some examples.

locale/en/docs/guides/dont-block-the-event-loop.md
davisjam: Soon!

locale/en/docs/guides/dont-block-the-event-loop.md
Tiriel: Okay, that's reassuring :smile: 
Maybe in that case you could specify this alternative, and that sufficiently complex regex are the real problem here? (I'm not quite sure what a sufficiently complete regex would be either; regex using escaped/reserved chars and/or logn regex?)

davisjam: Is the new REDOS section clearer? Still needs work?

davisjam:
@Tiriel:
1. Is the language up to an appropriate level of formality, or should I work on the tone further?
2. Does the regexp section make more sense now?

I haven't revised anything after the REDOS section yet.

Tiriel:
@davisjam Way better now IMHO! It's more precise, the examples add a lot, and the tone is better suited to the subject and target as I see it.

locale/en/docs/guides/dont-block-the-event-loop.md
davisjam: Alrighty, first draft of this is up.

davisjam:
The first draft of the full document is up. Comments appreciated.

ayaankazerouni:
+1 Cool guide! I learned a lot, especially from the REDOS section.

davisjam:
> I'm not the most technical reviewer you could get, I'm largely a noob in Node.js and JS itself

@Tiriel When you have a moment, can you ping some appropriate folks for technical feedback? Your thoughts on style and correctness are of course also much appreciated.

vsemozhetbyt:
Let's cc @nodejs/documentation for a start.

gibfahn:
cc/ @sam-github if you're around, seems like something you'd be interested in.

locale/en/docs/guides/dont-block-the-event-loop.md
```diff
@@ -0,0 +1,435 @@
+---
+title: Don't Block the Event Loop (or the Worker Pool)
+layout: docs.hbs
+---
+
+# Don't Block the Event Loop (or the Worker Pool)
+
+## Should you read this guide?
+If you're writing a server using Node, keep reading.
+If you're writing command-line scripts with Node, this guide might still come in handy, but it isn't crucial.
```
sam-github: cli scripts can do, for example, heavy DNS lookups at the same time as file I/O, so worker pool can be blocked.

locale/en/docs/guides/dont-block-the-event-loop.md
```diff
@@ -0,0 +1,435 @@
...
+
+## TL; DR
+Node.js runs JavaScript code in the Event Loop and handles I/O in a Worker Pool (the threadpool).
```
sam-github: No, just file I/O, network I/O is handled on the main loop.

locale/en/docs/guides/dont-block-the-event-loop.md
```diff
@@ -0,0 +1,435 @@
...
+Node scales well, sometimes better than more heavyweight approaches like Apache.
+The secret to Node's scalability is that it uses a small number of threads to handle many clients.
```
sam-github: where small number of threads is likely zero thread for a network server (no file i/o

locale/en/docs/guides/dont-block-the-event-loop.md
```diff
@@ -0,0 +1,435 @@
...
+This provides two motivations for blocking neither the Event Loop nor the Worker Pool.
+Since both the Event Loop and the Workers respond to events, we will refer to these threads generally as *Event Handlers* or *Event-Handler Threads*.
+All threads in Node are Event Handlers: one Event Loop and several Workers.
+
+1. Performance: If you regularly perform heavyweight tasks on an Event Handler, the *throughput* (requests/second) of your server will suffer.
+2. Security: If it is possible that for certain input you might block an Event Handler, a malicious client might be able to figure out the "evil input", make your code block, and keep you from working on other clients. This would be a Denial of Service attack.
+
+## A quick review of Node's architecture
+
+Node uses the Event-Driven Architecture: it has an Event Loop for orchestration and a Worker Pool for expensive tasks.
```
sam-github: file i/o isn't expensive, no more so than network i/o, it just doesn't have good completion APIs at the O/S level to allow it to be done on the main loop, so threads are needed so that i/o can be done with blocking system calls

locale/en/docs/guides/dont-block-the-event-loop.md
```diff
@@ -0,0 +1,435 @@
...
+
+## Queues and Cooperative Multitasking
+Both the Event Loop and the Worker Pool use *queues*, which are data structures with one or more producers who place items into the queue, and one or more consumers who remove items from the queue.
+
+The Event Loop's queue is full of pending events.
+When the Event Loop sees a new event, it invokes its callback, which is JavaScript code that sees your Node application.
+The Event Loop is the only consumer of this queue.
+The Node runtime produces events for this queue.
```
sam-github: the o/s produces the events, not the node runtime

locale/en/docs/guides/dont-block-the-event-loop.md
```diff
@@ -0,0 +1,435 @@
...
+Specifically, Node asks libuv to monitor file descriptors using [epoll](http://man7.org/linux/man-pages/man7/epoll.7.html)), and the Event Loop's queue consists of the ready descriptors.
```
sam-github: only on linux, its kqueue on OS X, and other mechanisms on Windows

locale/en/docs/guides/dont-block-the-event-loop.md
```diff
@@ -0,0 +1,435 @@
...
+
+The Worker Pool's queue is full of pending Tasks.
+When a Worker sees a new Task, it invokes its callback, which is C++ code that sits on the C++ side of the Node core modules.
+The consumers of this queue are the Workers in the Worker Pool, who all compete to work on the next Task.
+Most of the time, the Event Loop is the only producer of Tasks for the Worker Pool.
+
+It is crucial to understand that this queue-based approach means that the Event Loop and the Workers rely on [cooperative multitasking](https://en.wikipedia.org/wiki/Cooperative_multitasking).
```
sam-github: this isn't the common understanding of cooperative multi-tasking. coop tasking means that the running thread can suspend itself, but node doesn't do this, ever (on main loop or in pool). it runs to completion, which is not cooperative multitasking

locale/en/docs/guides/dont-block-the-event-loop.md
```diff
@@ -0,0 +1,435 @@
...
+From a Worker, you cannot manipulate a JavaScript object in the Event Loop's namespace.
+Instead, you have to serialize and deserialize any objects you wish to share.
+Then the Worker can operate on its own copy of these object(s) and return the modified object (or a "patch") to the Event Loop.
+
+For serialization concerns, see the section on JSON DOS.
+
+##### Best practices for offloading
+Follow these best practices for offloading:
+1. You should create your own Computation Worker Pool for extensive computation.
+2. The Computation Worker Pool should have no more than Workers than your machine has [logical cores](https://nodejs.org/api/os.html#os_os_cpus).
```
sam-github: what about when you have multiple addons, each with a number of threads equal to the number of logical cores?

locale/en/docs/guides/dont-block-the-event-loop.md
```diff
@@ -0,0 +1,435 @@
...
+The Workers of this Pool use the cooperative multitasking approach discussed above: each Worker completes its current Task before proceeding to the next one on the Worker Pool queue.
+
+Now, there will be variation in the cost of the Tasks required to handle your clients' requests.
+Some Tasks can be completed quickly (e.g. reading short or cached files in the I/O-WP, or computing the average of a small array in the C-WP), and others will take longer (e.g reading larger or uncached files in the I/O-WP, or doing more expensive computation in the C-WP).
+Your goal should be to *minimize the variation in Task times*, and you should use *Task partitioning* to accomplish this.
+
+### What is the Node Worker Pool used for?
+In the Node core modules, the Node Worker Pool is used for the asynchronous versions of the following classes of APIs:
+1. File system I/O
+2. DNS queries
```
sam-github: only dns.lookup, the other ones don't use the pool

locale/en/docs/guides/dont-block-the-event-loop.md
```diff
@@ -0,0 +1,435 @@
...
+But in many cases it's unclear how much an API might cost.
+
+*If you are calling an API that might do something expensive, ask the developers to document its cost, or examine the source code yourself (and submit a PR documenting the cost).*
+
+Remember, even if the API is asynchronous, you don't know how much time it might spend on an Event Handler in each of its partitions.
+For example, suppose in the `asyncAvg` example given above, each call to the helper function summed *half* of the numbers rather than one of them.
+Then this function would still be asynchronous, but the cost of each partition would be `O(n)`, not `O(1)`, making it much less safe to use for arbitrary values of `n`.
+
+## Conclusion
+Node has threads called Event Handlers, i.e. the Event Loop and the threadpool Workers, and these execute JS- and C++-callbacks.
```
sam-github: the thread in the libuv work pool aren't called Event Handlers, except here, by you! And the lib uv workers never execute JS code. I don't think its helpful to overgeneralize the main thread to describe it as a kind of worker.

sam-github (commented):
Its good to provide some information on this, particularly to provide implementation advice to users. It would be easy for someone to read this and think that network I/O occurs in the uv thread pool (which it doesn't), I think that aspect should be much more clear.

locale/en/docs/guides/dont-block-the-event-loop.md
davisjam: True. I view the server case as more concerning because of the DoS possibility. Do you want me to remove this caveat?

locale/en/docs/guides/dont-block-the-event-loop.md
davisjam: Yes. I'll fix to clarify what kinds of I/O go to the Worker Pool.

locale/en/docs/guides/dont-block-the-event-loop.md
davisjam: This depends on the env variable and on whether the server has ever used the threadpool, since libuv's threads are created the first time the threadpool is used. I don't see a reason to confuse the issue though, this is intended as the high-level summary.

locale/en/docs/guides/dont-block-the-event-loop.md
sam-github: Remove, or expand to make more clear the situations in which some things aren't problems. No need to optimize anything for a short-running app that does a one-shot HTTP request is what you are trying to say, I think.

locale/en/docs/guides/dont-block-the-event-loop.md
sam-github: I feel this confuses the issue, though, because the secret to node's scalability isn't that it uses threadpools well (so does java), the secret is that a fairly typical HTTP server making outbound DB or other network requests uses *no threads at all*.

locale/en/docs/guides/dont-block-the-event-loop.md
davisjam: My understanding of the libuv loop (which is limited to Linux) is that the trouble is not the completion APIs but rather the ability of the OS to say that the file descriptor is ready and thus not to block when you do I/O on it. libuv only sends network traffic when it will not block in doing so, but the OS will not guarantee that read/write on a file descriptor will be non-blocking.

Based on this understanding, my claim is correct - the I/O done on the event loop is always cheap *from the event loop's perspective* because the OS guarantees it to be non-blocking.

Am I missing something here?

locale/en/docs/guides/dont-block-the-event-loop.md
davisjam: Yes, via things like epoll. Where do you see the line between technical accuracy and "missing the forest for the trees" in guides like this?

locale/en/docs/guides/dont-block-the-event-loop.md
davisjam: Yes, good catch.

locale/en/docs/guides/dont-block-the-event-loop.md
davisjam: I couldn't think of a better way to describe this other than saying "queue-based". Should I stick with that? Is there a better term?

Counterpoint, though -- queue-based is "cooperative multitasking" in the sense that a callback/task yields precisely when it completes.

locale/en/docs/guides/dont-block-the-event-loop.md
davisjam: I would suggest creating a single Computation Worker Pool at the application level, similar to the Node Worker Pool that handles I/O from all points in the application. Do you have any reference applications I can look at to describe alternate patterns?

locale/en/docs/guides/dont-block-the-event-loop.md
sam-github: Maybe. The actual problem that is that you can't poll/epoll/select/kqueue a file descriptor for readiness, and so we can't do non-blocking reads or writes. And I guess you could call a blocking call "expensive", but network i/o can be done in thread pools with blocking calls, and lots of languages do exactly that. node doesn't, because network i/o can be non-blocking on unix, whereas file i/o can't (without the async i/o posix extensions, but they are poorly supported, and while windows has better async i/o for files, early tests by libuv authors showed that it would randomly block, it was to flaky to be useful).

locale/en/docs/guides/dont-block-the-event-loop.md
davisjam: Yes. I'll fix this. I believe all of the FS APIs map directly to libuv versions, though?

locale/en/docs/guides/dont-block-the-event-loop.md
davisjam: Yes, I agree that this is a weird overgeneralization. I'll take another look at this.

locale/en/docs/guides/dont-block-the-event-loop.md
davisjam: What do you think about saying "an epoll-like system call" with a link to the epoll man page? Part of my goal is to point users to resources that will give them a richer understanding of Node concepts and Node under the hood. That was also my intent with the "cooperative multitasking" link.

locale/en/docs/guides/dont-block-the-event-loop.md
sam-github: because its not really a queue, if it was, we could ask "how long is that queue?", but we can't.... because it doesn't exist. node reads from the o/s's internal queue, gets a set of ready events, processes them then blocks reading again. no queue. the whole queue thing is brought up sometimes, but I don't think introducing something that doesn' exist is helpful in understanding how node works. I never describe anything like it in https://www.youtube.com/watch?v=P9csgxBgaZ8, for example

locale/en/docs/guides/dont-block-the-event-loop.md
davisjam: Right. It uses *one* thread -- the main loop -- to handle many clients, rather than Apache-style with one thread per active client.

locale/en/docs/guides/dont-block-the-event-loop.md
sam-github: but no such thing exists, so current addons don't use it. it has been proposed that uv expose two work APIs, one for blocking non-cpu intensive calls (file i/o, dns.lookup), the other for cpu intensive activities (crypto.pbkdf for key generation, etc), but no code has been PRed, or benchmarks to show it will work. so, in the meantime, you have to consider all the internal pools of all the addons, and actually most addons will mix cpu and i/o intensive work in the same uv work pool. I'm slowly working towards getting metrics in node about the state of the uv pool, but that doesn't exist yet, so not worth mentioning here.

locale/en/docs/guides/dont-block-the-event-loop.md
davisjam: Last I checked, posix async I/O is just a threadpool in libc, no different from the libuv threadpool. I've prototyped Node with kernel AIO to see if things had improved since [2011](https://github.com/nodejs/node-v0.x-archive/issues/1446), but the O_DIRECT restriction remains and its costs are catastrophic.

Anyway, I'll flesh out this discussion a bit more.

locale/en/docs/guides/dont-block-the-event-loop.md
sam-github: https://nodejs.org/api/fs.html#fs_threadpool_usage  - except fs.FSWatcher()

locale/en/docs/guides/dont-block-the-event-loop.md
davisjam: OK. If I added such an API would it be of interest? Is anyone working on this right now?

locale/en/docs/guides/dont-block-the-event-loop.md
sam-github: aio was intended to be implemented in kernel and thus to rock... but hasn't happened yet, the saga is long, some info here: https://lwn.net/Articles/724198/. There is various user-space emulation using blocking I/O in a thread pool....  so, yeah, uv doesn't use that, there is no point.

locale/en/docs/guides/dont-block-the-event-loop.md
davisjam: Also, I'll just remove this section. It sounds like this is too much of an open problem to worry about right now.

locale/en/docs/guides/dont-block-the-event-loop.md
sam-github: I think that other than the queue to the uv work pool (which is actually a queue), I don't think queues should be mentioned. Not sure what a good replacement is, sorry! I can see the ghostly parallels, both node and co-op multitasking systems require careful consideration of how much work you do, and how you construct your app, but the analogy breaks down fast, because cooperative multi-tasking has a fairly agreed upon definition, and its not what node does, so better stay away from it, IMO.

locale/en/docs/guides/dont-block-the-event-loop.md
davisjam: From a framework developer perspective, it's a queue that you can only see looking backwards. From an application developer perspective, though, I think the concept of a queue is a fine approximation of reality. I'm envisioning the target audience to be application developers for whom I think the "queue" concept is fine.

How strongly do you feel about this?

davisjam:
Thanks for all the great comments, @sam-github! I'll get started on a revision.

locale/en/docs/guides/dont-block-the-event-loop.md
sam-github: I am OK with the whole thing, except this sentence:

>The Node runtime produces events for this queue.

I think that specific sentence should be removed. node doesn't produce the events. it asks for the o/s to produce them. I wouldn't say that I'm producing a cappucino when I order a cappucino at a cafe. Causing it to be produced? Sure, producing it? no.

Tiriel:
Just a quick semi off-topic:
Thanks @sam-github & @davisjam , just reading your comments makes me learn a huge load of new things!
I really look forward to reading the final version of this guide!

davisjam:
@sam-github e9c40e9 attempts to address your concerns.

locale/en/docs/guides/dont-block-the-event-loop.md
```diff
@@ -0,0 +1,457 @@
...
+In summary, the Event Loop executes the JavaScript callbacks registered for events, and is also responsible for fulfilling non-blocking asynchronous requests like network I/O.
+
+### What code runs on the Worker Pool?
+Node's Worker Pool is implemented in libuv ([docs](http://docs.libuv.org/en/v1.x/threadpool.html)), which exposes a general task submission API.
+
+Node uses the Worker Pool to handle "expensive" tasks.
+This includes I/O for which an operating system does not provide a non-blocking version, as well as particularly CPU-intensive tasks.
+
+These are the Node module APIs that make use of this Worker Pool:
+1. I/O-intensive
```
a0viedo, line 58: worth mentioning `net` module?

davisjam: It is my understanding that the [`net`](https://nodejs.org/api/net.html) module does *not* use the Worker Pool, thanks to good OS support for non-blocking network I/O. The module docs do not bear the usual caveats about the threadpool, and as far as I can tell from the source there is no call to `uv_queue_work`.

Am I wrong in this? Or are you suggesting I explicitly mention the `net` module as a "safe" option? If so, I attempted to do so in the section "What code runs on the Event Loop?" with the sentence "The Event Loop will also fulfill the non-blocking asynchronous requests made by its callbacks, e.g., network I/O.". Do you think I should name a few examples there, like `net` and `UDP`?

sam-github: correct, net does not use threads

sam-github:
thanks for addressing my comments @davisjam

davisjam:
Sure thing, @sam-github. Any more suggestions, or people that should be tagged?

sam-github:
@jasnell was looking at loop behaviour, he might have some comments

locale/en/docs/guides/dont-block-the-event-loop.md
```diff
@@ -0,0 +1,457 @@
...
+But in many cases it's unclear how much an API might cost.
+
+*If you are calling an API that might do something expensive, ask the developers to document its cost, or examine the source code yourself (and submit a PR documenting the cost).*
+
+Remember, even if the API is asynchronous, you don't know how much time it might spend on a Worker or on the Event Loop in each of its partitions.
+For example, suppose in the `asyncAvg` example given above, each call to the helper function summed *half* of the numbers rather than one of them.
+Then this function would still be asynchronous, but the cost of each partition would be `O(n)`, not `O(1)`, making it much less safe to use for arbitrary values of `n`.
+
+## Conclusion
+Node has two types of threads: one Event Loop and `k` Workers.
```
benjamingr, line 470: What about the JavaScript thread?

locale/en/docs/guides/dont-block-the-event-loop.md
```diff
@@ -0,0 +1,457 @@
...
+
+To avoid this, you should try to minimize variation in the length of Tasks you submit to the Worker Pool.
+While it is appropriate to treat the external systems accessed by your I/O requests (DB, FS, etc.) as black boxes, you should be aware of the relative cost of these I/O requests, and should avoid submitting requests you can expect to be particularly long.
+
+Two examples should illustrate the possible variation in task times.
+
+#### Variation example: Long-running file system reads
+Suppose your server must read files in order to handle some client requests.
+After consulting Node's [File system](https://nodejs.org/api/fs.html) APIs, you opted to use `fs.readFile()` for simplicity.
+However, `fs.readFile()` is not partitioned: it submits a single `fs.read()` Task spanning the entire file.
```
benjamingr: This is actually being fixed if I understand correctly: https://github.com/nodejs/node/pull/17054

davisjam: That's the idea, anyway. But this guide is describing things as they are, not as I hope they will be.

locale/en/docs/guides/dont-block-the-event-loop.md
davisjam: I am referring to the "JavaScript thread" as the Event Loop throughout, to contrast it with the Worker Pool (threadpool) threads, and because it does more than JavaScript (e.g. wandering into libuv to fulfill network I/O requests, timers, etc.).

See the section "What code runs on the Event Loop?"

"In summary, the Event Loop executes the JavaScript callbacks registered for events, and is also responsible for fulfilling non-blocking asynchronous requests like network I/O."

Let me know if this is not clear or if you have alternative suggestions for phrasing.

locale/en/docs/guides/dont-block-the-event-loop.md
davisjam: See also nodejs/node#17250.

gibfahn: Might be worth adding a link to nodejs/node#17054 in here to note the proposed change then.

davisjam: Added a link in 0a42816, thank you.

davisjam:
I recognize this is a pretty long document, but does anyone have additional change requests?

benjamingr (approved):

ghost (approved):

locale/en/docs/guides/dont-block-the-event-loop.md
```diff
@@ -0,0 +1,457 @@
...
+
+In a server, *you should not use the synchronous APIs from these modules*.
+These APIs are expensive, because they involve significant computation or require I/O.
+If you execute them on the Event Loop, they will take far longer to complete than a typical JavaScript instruction, blocking the Event Loop.
+
+### Blocking the Event Loop: JSON DOS
+`JSON.parse` and `JSON.stringify` are other potentially expensive operations.
+While these are `O(n)` in the length of the input, for large `n` they can take surprisingly long.
+
+If your server manipulates JSON objects, particularly those from a client, you should be cautious about the size of the objects or strings you work with on the Event Loop.
```
gibfahn, line 250: Is it worth mentioning asynchronous JSON parsing modules here (and then linking to `The risks of npm modules` section)?

locale/en/docs/guides/dont-block-the-event-loop.md
```diff
@@ -0,0 +1,457 @@
...
+### Blocking the Event Loop: Node core modules
+Node exposes synchronous versions of several expensive APIs, including:
+- [Encryption](https://nodejs.org/api/crypto.html)
+- [Compression](https://nodejs.org/api/zlib.html)
+- [File I/O](https://nodejs.org/api/fs.html)
+- [Child process interaction](https://nodejs.org/api/child_process.html)
+
+In a server, *you should not use the synchronous APIs from these modules*.
+These APIs are expensive, because they involve significant computation or require I/O.
+If you execute them on the Event Loop, they will take far longer to complete than a typical JavaScript instruction, blocking the Event Loop.
```
gibfahn: What do you do if there is no async core api?

Maybe the answer is "raise an issue in core", but this was my question on reading this, so maybe worth saying something.

locale/en/docs/guides/dont-block-the-event-loop.md
```diff
@@ -0,0 +1,457 @@
...
+
+### Conclusions
+Whether you use only the Node Worker Pool or maintain separate Worker Pool(s), you should optimize the Task throughput of your Pool(s).
+
+To do this, minimize the variation in Task times by using Task partitioning.
+
+##  The risks of npm modules
+Node developers benefit tremendously from the [npm ecosystem](https://www.npmjs.com/), with hundreds of thousands of modules offering functionality to accelerate your development process.
+
+However, the vast majority of these modules are written by third-party developers and undergo no formal verification process.
```
gibfahn: Does node core undergo a formal verification process? Mentioning it here makes it seem like it's in contrast to something else, but I'm not sure what that something is.

gibfahn (commented):
This is really helpful, thanks a lot for writing this!

LGTM, couple of questions though.

locale/en/docs/guides/dont-block-the-event-loop.md
davisjam: Good idea, will do.

locale/en/docs/guides/dont-block-the-event-loop.md
davisjam: You're right, "formal" isn't the right word. It's more of a "many eyes on Node, few eyes on most 3rd-party modules" type situation. I'll change the wording.

locale/en/docs/guides/dont-block-the-event-loop.md
davisjam: Thanks, I'll try another pass on this.

locale/en/docs/guides/dont-block-the-event-loop.md
davisjam: Addressed in ccd4f75.

locale/en/docs/guides/dont-block-the-event-loop.md
davisjam: Addressed in ccd4f75.

locale/en/docs/guides/dont-block-the-event-loop.md
davisjam: Addressed in ccd4f75.

davisjam:
@gibfahn Thank you for your review. I've tried to address your comments in ccd4f75. Let me know if you have more concerns before you'd be willing to approve.

davisjam:
@sam-github Anything else needed for your approval?

sam-github:
looks good to me

gibfahn (approved):

davisjam:
The full draft has been up for about 2 weeks and has gone through a few reviews. Ready to be landed?

Issue #1480, opened 2017-11-20
guides: use of GET method in echo server


bneevan:
* **URL**: https://nodejs.org/en/docs/guides/anatomy-of-an-http-transaction/
* **Browser version**: all
* **Operating system**: all

`Echo Server` code in _[Anatomy of an http transaction](https://nodejs.org/en/docs/guides/anatomy-of-an-http-transaction/)_ uses GET method to handle the request. But this cannot work since any additional data in the request (eg: /echo?hello) will change the route and return a 404 response. 

To handle this data processing and routing properly we should use the POST method.

lpinca:
Fixed by #1481.

PR #1481, opened 2017-11-20
Guide: Anatomy of an http…: Echo Server

bneevan:
Towards [#1480](https://github.com/nodejs/nodejs.org/issues/1480)

Changes:
* Updated GET to POST method… in echo server code and related comments.
* Guide versions of `en` and `uk` - changes at 4 places.
* Guide version of `ko` - changes at 8 places (including 4 commented `en` code used for reference).

### DIFF
--- a/locale/en/docs/guides/anatomy-of-an-http-transaction.md
+++ b/locale/en/docs/guides/anatomy-of-an-http-transaction.md
@@ -305,7 +305,7 @@ http.createServer((request, response) => {
 Now let's tweak this. We want to only send an echo under the following
 conditions:
 
-* The request method is GET.
+* The request method is POST.
 * The URL is `/echo`.
 
 In any other case, we want to simply respond with a 404.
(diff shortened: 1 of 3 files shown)


lpinca (approved):
A GET request can have a body but yes, using POST makes more sense.

bneevan:
@lpinca, thanks for that insight. I didn't know about GET requests having a body being valid. `curl` command allowed creating such a GET request and the Echo server did echo 😊.

Did some more search and found [this interesting comment](https://groups.yahoo.com/neo/groups/rest-discuss/conversations/messages/9962) from Roy Fielding on GET requests:

> any HTTP request message is allowed to contain a message body, and thus must parse messages with that in mind. Server semantics for GET, however, are restricted such that a body, if any, has no semantic meaning to the request. The requirements on parsing are separate from the requirements on method semantics.
>
>So, yes, you can send a body with GET, and no, it is never useful to do so.

[https://groups.yahoo.com/neo/groups/rest-discuss/conversations/messages/9962](https://groups.yahoo.com/neo/groups/rest-discuss/conversations/messages/9962)

bneevan:
Sorry, closed it accidentally.

lpinca:
@bneevan thank you!

PR #1484, opened 2017-11-22
Update README.md

ajmeese7:
Removed period after `etc` because there is already one after the parenthesis, which is incorrect according to [this article](http://www.chicagomanualofstyle.org/qanda/data/faq/topics/Punctuation/faq0040.html), and modified phrasing of a section because it read funny (in my opinion).


Tiriel (approved):
LGTM, phrasing clearer indeed

stevemao (approved):

PR #1503, opened 2017-12-08
Remove broken links from Get involved

fhinkel:
How-to-node and the German and Turkish sites don't work. Removing them.

Is it OK if we remove them?


fhemberger:
Thanks, good catch!

We remove broken links, projects/communities can always send a PR with their new sites, which will happily get accepted.

PR #1511, opened 2017-12-09
doc: update announce to indicate ARMv6 is avail

mhdawson:


mhdawson:
I'm heading out for the night so if anybody else can land this once its approved that would be great.

Otherwise I'll probably take a look tomorrow morning.

daxlab (approved):

PR #1512, opened 2017-12-11
chore: update card image

MylesBorins:
The current png has no margins and doesn't look great.

I've created a new png that is no longer transparent and has margins.
Should look nicer

edit:

worth mentioning this retains the exact same proportions as before, if there are better proportions for a card I can use the svg to make an entirely new image


daxlab (approved):

bnb (approved):
Thanks for this improvement, Myles <3

lpinca (approved):

ghost (approved):

MylesBorins:
For posterity this is how it looks now

screen shot 2017-12-12 at 1 26 19 am

PR #1522, opened 2017-12-19
docs: fix misspell

sotayamashita:


fhemberger:
:+1:  Thank you for spotting these!

PR #1527, opened 2017-12-24
update package-manager.md

naskapal:
fixed fedora download link description to match link version (from v6 to v8)


fhemberger:
Good catch, thank you!

PR #1533, opened 2017-12-26
Display footer from left to right

### LINKED ISSUE #1530: [Nodejs-AR] Right to left footer not properly displayed
Hello, 

i've noticed that in all web site supported languages, the footer is not translatable, However, it's also displayed in right to left :

ko

I think that the direction of footer should be ````ltr```` all time to be displayed juste like that :

ok

WDYT ?

* **URL**: nodejs.org/ar

ouadie-lahdioui:
Fixes #1530


fhemberger:
Great, thanks!

Issue #951, opened 2016-10-11
meta: update for Apple macOS branding


imyller:
- **URL**: http://nodejs.org
- **Browser version**: all
- **Operating system**: all

Apple has updated their OS branding from `OS X` to `macOS`.

```
macOS (originally named "Mac OS X" until 2012 and then "OS X" until 2016)
```

Many places such as developer documentations, marketing materials and external sites like Wikipedia have started referring to `macOS` instead of `OS X`.

I'd suggest that Node.js site drops `OS X` references and uses following generic platform terms where possible:
- `macOS`
- `Mac`
- `Macintosh`

In my opinion, `OS X` can still be used if there is need to specify old OS version, for example: `Apple OS X 10.9`

PR #1536, opened 2017-12-29
minor: update branding for macOS binaries

jedsmith:
While maintaining the trademark and heritage, Apple no longer actively
refers to any of its Mac platforms as Macintosh. Also, in a number of
places, nodejs.org refers to macOS and Macintosh on the same page.
This change updates the site for consistency, preferring macOS in all
locations. While I've updated the localized versions, the only change
is to revise a single proper noun, with one exception:
`locale/ko/download/package-manager.md` appears to spell "Macintosh" in
Korean, and I am unequipped to handle that change.

This commit does not update blog entries, which refer to the Macintosh
installer by name. Blog entries seem fixed and ephemeral, though it
would be trivial to update them all. Going forward, blog entries
should identify the installer for macOS as the "macOS Installer", as
opposed to the "Macintosh Installer".


jedsmith:
Relevant to #951, though I disagree that Macintosh should ever be used.

fhemberger (approved):
LGTM

daxlab (approved):

ghost (approved):
Thanks!

PR #1537, opened 2018-01-01
fix: changed link to lts-release-schedule page

ckotzbauer:
nodejs/lts repository was renamed to nodejs/release

### DIFF
--- a/layouts/index.hbs
+++ b/layouts/index.hbs
@@ -64,7 +64,7 @@
           </div>
         {{/if}}
         <p>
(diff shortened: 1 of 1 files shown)


daxlab (approved):

fhemberger:
Thank you!

gibfahn:
FWIW the fact that the redirect no longer works is a mistake that needs to be fixed (see https://github.com/nodejs/Release/issues/296).

This change is still correct though.

PR #1547, opened 2018-01-08
doc: provide info on impact of recent vulns

mhdawson:


mhdawson:
@nodejs/security-wg @nodejs/website, draft based on discussion in private security repo.

bnb (approved):
LGTM

mhdawson:
Pushed commit to address conflict.

mhdawson:
actually looks like I got the wrong index.md.  @fhemberger do we now need to update all of the index files ? for the banner ?

mhdawson:
Pushed commit to update the rest of the banners, am wondering if there is a more efficient way of doing it.

mcollina (approved):
LGTM

locale/en/blog/vulnerability/jan-2018-spectre-meltdown
```diff
@@ -0,0 +1,24 @@
...
+author: Michael Dawson
+---
+
+# Summary
+
+Project zero has recently announced some new attacks that have received a lot of attention: https://googleprojectzero.blogspot.ca/2018/01/reading-privileged-memory-with-side.html.
+
+The risk from these attacks to systems running Node.js resides in the systems in which your Node.js applications run, as opposed to the Node.js runtime itself. The trust model for Node.js assumes you are running trusted code and does not provide any separation between code running within the runtime itself. Therefore, untrusted code that would be necessary to execute these attacks in Node.js could already affect the execution of your Node.js applications in ways that are more severe than possible through these new attacks.
+
+This does not mean that you don't need to protect yourself from these new attacks when running Node.js applications. To do so, apply the security patches for your operating system. You do not need to update the Node.js runtime.
```
hashseed: Maybe there should be more explanation.

Both Meltdown and Spectre need to run malicious code - even if permissions are restricted - to perform an attack. However, if malicious code is allowed to run on Node.js, you've lost anyways. But only to a certain degree. That malicious code still only has user permissions. Except that with Meltdown you can even read out kernel memory, if you run on an unpatched OS on an Intel CPU.

locale/ja/index.md
```diff
@@ -1,7 +1,7 @@
 ---
 layout: index.hbs
 labels:
-  banner: Important security releases, please update now!
+  banner: Spectre and Meltdown in the context of Node.js - no current action required.
```
hashseed: The OS needs to be updated. That's more than "no current action required".

mhdawson: I was on the edge on that, I'll come up with something better.

locale/en/blog/vulnerability/jan-2018-spectre-meltdown
mhdawson: Right, but I think the point we are trying to make is that you need to patch your OS, not Node.js. I'll pull in a few more words from what you have.

locale/en/blog/vulnerability/jan-2018-spectre-meltdown
```diff
@@ -0,0 +1,24 @@
...
+category: vulnerability
+title: Meltdown and Spectre - Impact On Node.js
+slug: jan-2018-spectre-meltdown
+layout: blog-post.hbs
+author: Michael Dawson
+---
+
+# Summary
+
+Project zero has recently announced some new attacks that have received a lot of attention: https://googleprojectzero.blogspot.ca/2018/01/reading-privileged-memory-with-side.html.
```
jasnell: nit: would you please add line breaks. Makes reviewing easier.

locale/ja/index.md
jasnell: Perhaps, "No Node.js specific action required"?

mhdawson: Actually probably better to have people just read the blog so will remove the '- no current action required.'

locale/en/blog/vulnerability/jan-2018-spectre-meltdown
mhdawson: That makes it more work to paste into the email list page as I then have to go and remove them all.  I can do it but means more work on the other end.

mhdawson:
@hashseed pushed commit to address comments.

vdeturckheim (approved):

locale/en/blog/vulnerability/jan-2018-spectre-meltdown
```diff
@@ -9,16 +9,34 @@ author: Michael Dawson
...
+running trusted code and does not provide any separation between code
+running within the runtime itself. Therefore, untrusted code that
+would be necessary to execute these attacks in Node.js could already
+affect the execution of your Node.js applications in ways that
+are more severe than possible through these new attacks.
+
+This does not mean that you don't need to protect yourself from
+these new attacks when running Node.js applications. If an attacker
+manages to run malicious code on an upatched OS (whether using
+javascript or something else) they may be able to access memory and or
```
matteocontrini: Maybe "JavaScript" instead of "javascript" here?

mhdawson: Good point, thanks

mhdawson: @matteocontrini  pushed commit to fix

vdeturckheim (approved):

mhdawson:
@hashseed would be good to see if you are ok with my update.

@nodejs/website would be good to get confirmation that updating the banner in all of the index.md files is now the right way to go.

mhdawson:
Will plan to send out first thing my morning  (EST) unless there are other comments before then.

locale/en/blog/vulnerability/jan-2018-spectre-meltdown
```diff
@@ -0,0 +1,42 @@
...
+would be necessary to execute these attacks in Node.js could already
+affect the execution of your Node.js applications in ways that
+are more severe than possible through these new attacks.
+
+This does not mean that you don't need to protect yourself from
+these new attacks when running Node.js applications. If an attacker
+manages to run malicious code on an upatched OS (whether using
+JavaScript or something else) they may be able to access memory and or
+data that they should not have access to.  In order to protect yourself
+from these cases, apply the security patches for your operating
```
hashseed: does it make sense to link to these resources?

hashseed (approved):

daxlab (approved):

fhemberger:
@mhdawson Hmmm … not very happy to overwrite all local messages again. We didn't translate the security banner for exact this reason: Translation teams would never be able to make the adjustments so quickly in case of an urgent release.

But over time it was clear that the message used for the banners was always "Important security releases, please update now!" – so we decided to make it translatable as well.

---

I'm currently thinking about a "manual override" in the template instead of replacing all translations:

layouts/index.hbs#17:
```diff
- <a href="{{ project.banner.link }}">{{{ labels.banner }}}</a>
+ <a href="{{ project.banner.link }}">{{#if project.banner.text}}{{{ project.banner.text }}}{{else}}{{{ labels.banner }}}{{/if}}</a>
```
And just add the key back here: https://github.com/nodejs/nodejs.org/blob/master/build.js#L273

What do you think?

lpinca (approved):

lpinca:
👍 for @fhemberger's suggestion.

locale/en/blog/vulnerability/jan-2018-spectre-meltdown
mhdawson: @hashseed I'm not sure what all links/patches are available for OS's so far.  Maybe that is something we can add in later on.

mhdawson:
@fhemberger, the banner would go through at least a couple of steps for a security release.  First telling people the release was going to come in ~ 1 week, and then being updated to to tell them to update.

I think the override makes sense. Is that something we should include as part of this update ?  I would like to get it out today.

fhemberger:
@mhdawson Yes, please feel free to add this switch with this PR so it can land.

mhdawson:
Ok changes pushed to use override, should be good to go.

mhdawson:
Plan to wait about 1 hour for any more comments and then will land/send out on nodejs-sec mailing list.

mhdawson:
Landed as cdc41446d5cdf21fcef2eb4c86d43764826c5360

Topsmalllaxson:
request

Topsmalllaxson:
Landed as e42708c

PR #1549, opened 2018-01-10
blog: amend v9.4.0 blogpost

addaleax:
Applied from https://github.com/nodejs/node/pull/18083

---

    doc: un-mark Socket#write “removal” as notable change

    Since the method is inherited from `Writable`, usage of the method
    does not change at all after its removal as an explicit method.

    Calling it out as notable might therefore be more confusing than
    helpful.


daxlab (approved):

LinusU (approved):

addaleax:
I *think* this is my first PR here – do I/can I merge this myself?

lpinca (approved):

lpinca:
@addaleax go ahead.

PR #1567, opened 2018-02-17
Added solus to the package manager downloads page

rbnswartz:
Adds solus to the list of linux package manager install instructions


fhemberger:
Note: Found the following here: https://github.com/solus-project/package-management
> Please note that we plan to replace eopkg with sol.

Sol: https://github.com/solus-project/sol

rbnswartz:
@fhemberger Yes sol is going to replace eopkg one day but at the moment sol isn't finished. Eopkg is still the default at the moment and sol hasn't even made it into the repository yet.

fhemberger:
Ok, good to know. Merging.

PR #1576, opened 2018-03-02
blog: fix typo of v9.6.0

yous:
Ref: https://github.com/nodejs/node/pull/15752


fhemberger:
Good catch, thank you!

PR #1581, opened 2018-03-07
Better structure for guides' overview page

fhemberger:
So far guides appeared in random order which makes it hard to find
your way around. This change groups related guides. Because of that it also
disables the prev/next navigation (as order doesn't match any longer).
Also fixed some little nits.

Ref: https://github.com/nodejs/help/issues/1122


locale/ko/docs/guides/index.md
```diff
@@ -1,6 +1,26 @@
 ---
 title: 가이드
-layout: guides-index.hbs
+layout: docs.hbs
 ---
 
 # 가이드
+
+## General
```
Qard, line 8: Probably best to leave this change up to the localization teams so they can pick language-appropriate titles.

layouts/docs.hbs
```diff
@@ -13,7 +13,7 @@
       <article>
         {{{ contents }}}
 
-        {{> prev-next-navigation }}
+        {{!> prev-next-navigation }}
```
Qard, line 16: This would impact reference doc pages too. We probably shouldn't use this layout for guides and leave this line the way it was.

Qard (changes requested):

locale/ko/docs/guides/index.md
fhemberger: I'd like this change to land and let the Korean language team create a follow-up PR for the new headlines.

/cc @nodejs/nodejs-ko

layouts/docs.hbs
fhemberger: What reference doc pages are you referring to?

Qard: Oh, sorry. I misread. It'd probably be good to be able to keep the continuation feature though. Perhaps we should consider putting the guides in sub-folders and number-prefix them to keep some sort of order?

gireeshpunathil (approved):

layouts/docs.hbs
fhemberger: I thought about that as well but decided against it (for now):
- The order and categories I chose are quite arbitrary, there's no "right" flow
- This would also change URLs and we'd need to setup redirects

This section will be changed anyway in the relaunch we're currently working on, so it's a temporary fix anyway.

Qard (approved):

lpinca (approved):

amandeepmittal (approved):

daxlab (approved):

WaleedAshraf (approved):

PR #1589, opened 2018-03-13
Add security.txt file

### LINKED ISSUE #143: security.txt file on nodejs.org
security.txt files (https://securitytxt.org/) are getting popular these days and I think they are fairly useful in reducing friction in reporting security problems.

I think putting such a file on nodejs.org Web site would not hurt. What do you think about generating and uploading such a file on the Web site?

MarcinHoppe:
security.txt (https://securitytxt.org/) files are getting an accepted industry standard and it would be beneficial to adopt this standard for Node.js Web site as well.

This addition to the Web site has already been endorsed by the Security WG.

This PR resolves nodejs/security-wg#143.


fhemberger (approved):

server.js
```diff
@@ -89,6 +96,7 @@ statics.on('add', (filePath) => {
 
 // Initializes the server and mounts it in the generated build directory.
 http.createServer((req, res) => {
+  if (wellknown(req, res)) return
```
lpinca: This server is only used for development so I'm not sure if it's worth adding. In production the site is served by NGINX.

MarcinHoppe: Then in production it would be good to add `/security.txt` -> `/.well-known/security.txt` redirect as recommended by the spec and employed by a couple of bigger sites on the Web.

lpinca: Yes, it should be added here: https://github.com/nodejs/build/blob/master/setup/www/resources/config/nodejs.org

fhemberger: @lpinca `security.txt` should be in the repo, file would be served in production like our static files.

We just need to add 
```
location /security.txt {
    alias /home/www/nodejs/.well-known/security.txt;
    default_type text/plain;
}
```

in the nginx config.

lpinca: Yes exactly that's what I was suggesting.

lpinca: We don't need the build step and the addition to `server.js` though as we don't care in development, I guess. Basically only add `.well-known/security.txt` here and update the NGINX config.

fhemberger: Same build script is used in production and we need a way to place the file in the `build` directory.

lpinca: > we need a way to place the file in the build directory.

You are right.

lpinca (approved):

server.js
lpinca: @fhemberger on second thought, why should it be in `build`? `robot.txt` for example is in `static`.

```
location /robots.txt {
    alias /home/www/nodejs/static/robots.txt;
    default_type text/plain;
}
```

fhemberger: Wouldn't mind it at all. Keeps things simple. 👍

MarcinHoppe: OK, so the bottom line is to put `security.txt` into the `static` folder and handle `/security.txt` and `/.well-known/security.txt` in NGINX config and get rid of everything else in `server.js` and `build.js`?

lpinca: That works for me.

bnb (approved):

daxlab (approved):

ckotzbauer (approved):

amandeepmittal (approved):

server.js
MarcinHoppe: Thanks for all the feedback, I will make a change and submit a second PR against nodejs/build.

MarcinHoppe: Thanks for all the feedback, I will make a change and submit a second PR against nodejs/build.

MarcinHoppe:
I left the bare `security.txt` file here and updated NGINX configuration in a separate PR: nodejs/build#1181.

All feedback appreciated!

lpinca:
Thank you!

MarcinHoppe:
My pleasure!

PR #1591, opened 2018-03-14
Remove NodeJS Reactions from "Getting Started" page

bnb:
The Node.js Reactions site has some content that's generally negative toward the project and individual community members, which doesn't really promote a strong and healthy community.

Would it be beneficial to remove this from the other locales as well, or are those handled differently?


fhemberger:
@bnb Oh, good point! Didn't know we linked that page at all. If the link exists in other languages, please remove it there as well.

bnb:
@fhemberger went through each locale and removed it from the ones it existed in 👍

fhemberger (approved):

fhemberger:
Great, thank you! 👍

Issue #1618, opened 2018-04-07
Can not run


aiyadingya:
MacBook Pro (Retina, 13-inch, Early 2015)
node v8.9.1
npm v5.7.1

After running the npm start command, the following result appears:

````
aiyadingyadeMacBook-Pro:nodejs.org aiyadingya$ npm run start

> nodejs.org@ start /Users/aiyadingya/www/code/nodejs.org
> npm run serve


> nodejs.org@ serve /Users/aiyadingya/www/code/nodejs.org
> node server.js

[metalsmith] build/static started
[metalsmith] build/layouts started
http://localhost:8080/en/
[metalsmith] build/static finished: 136.254ms
[metalsmith] build/layouts finished: 1118.556ms
releases undefined
/Users/aiyadingya/www/code/nodejs.org/scripts/helpers/latestversion.js:15
  const match = releases.find((release) => !release.lts && semver.gte(release.version, '5.0.0'))
                         ^

TypeError: Cannot read property 'find' of undefined
    at Object.exports.current (/Users/aiyadingya/www/code/nodejs.org/scripts/helpers/latestversion.js:15:26)
    at loadVersions (/Users/aiyadingya/www/code/nodejs.org/build.js:268:34)
    at afterFetch (/Users/aiyadingya/www/code/nodejs.org/node_modules/node-version-data/node-version-data.js:46:14)
    at proxy (/Users/aiyadingya/www/code/nodejs.org/node_modules/node-version-data/node_modules/after/index.js:19:13)
    at /Users/aiyadingya/www/code/nodejs.org/node_modules/node-version-data/node-version-data.js:32:14
    at BufferList._callback (/Users/aiyadingya/www/code/nodejs.org/node_modules/node-version-data/node_modules/jsonist/jsonist.js:32:16)
    at BufferList.<anonymous> (/Users/aiyadingya/www/code/nodejs.org/node_modules/node-version-data/node_modules/bl/bl.js:16:14)
    at emitOne (events.js:116:13)
    at DestroyableTransform.emit (events.js:211:7)
    at ClientRequest.<anonymous> (/Users/aiyadingya/www/code/nodejs.org/node_modules/node-version-data/node_modules/hyperquest/index.js:13:19)
    at emitOne (events.js:116:13)
    at ClientRequest.emit (events.js:211:7)
    at TLSSocket.socketErrorListener (_http_client.js:387:9)
    at emitOne (events.js:116:13)
[...]

PR #1624, opened 2018-04-08
build: Don't throw when 'releases' is undefined

fhemberger:
In case `versions.json` could not be properly downloaded/parsed by
'node-version-data', avoid `TypeError: Cannot read property 'find' of undefined`
when `releases` is `undefined` and return `undefined` instead.

Error should be handled properly elsewhere.

See: #1618

### DIFF
--- a/scripts/helpers/latestversion.js
+++ b/scripts/helpers/latestversion.js
@@ -11,11 +11,11 @@ const map = (release) => release && {
 }
 
 exports.current = (releases) => {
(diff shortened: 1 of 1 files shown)


fhemberger:
Although this is not really a **fix** for #1618, at least we should not throw in this case but fail gracefully.

WaleedAshraf (approved):
LGTM

marswong (approved):

lpinca (approved):

ckotzbauer (approved):

PR #1625, opened 2018-04-08
Update dependencies, adjust build to changes in 'metalsmith-layouts'

fhemberger:
'metalsmith-layouts' has three new dependencies (handlebars-parser and auto discovery of helpers/partials). Build process has been updated.


scripts/helpers/getlistjson.js
```diff
@@ -0,0 +1,11 @@
+'use strict'
```
WaleedAshraf, line 1: Is this still being used? I think this can be removed.

scripts/helpers/json.js
```diff
@@ -0,0 +1,3 @@
+'use strict'
```
WaleedAshraf, line 1: Same, I think this can be removed.

Even if we want to keep these for future use, I think it'd be better to create something like `helpers/common.js`
And export all these methods from one file instead of creating a separate file for each method.

scripts/helpers/getlistjson.js
fhemberger: It's in all other JS files as well.

scripts/helpers/json.js
fhemberger: Doesn't work like that with metalsmith-discover-helpers/partials, which are in turn dependencies for metalsmith-layout.

scripts/helpers/getlistjson.js
```diff
@@ -0,0 +1,11 @@
...
+
+module.exports = (context) => {
+  const result = context.map(item => ({
```
lpinca: Nit: can we make this consistent? Either add `()` around `item` or remove them also from `context` above. Also why wrapping the returned object with `()`?

package.json
```diff
@@ -31,38 +31,41 @@
   "dependencies": {
     "autoprefixer-stylus": "0.14.0",
     "cheerio": "0.22.0",
-    "chokidar": "1.7.0",
-    "handlebars": "4.0.10",
+    "chokidar": "^2.0.3",
```
lpinca: Nit: We now have some pinned dependencies and some not. Can we use the same versioning scheme for all dependencies?

lpinca (approved):

scripts/helpers/json.js
WaleedAshraf: Got it. 👍

WaleedAshraf (approved):

daxlab (approved):

package.json
daxlab: IMO, we should pin the dependencies.

fhemberger: Sorry, updating …

scripts/helpers/getlistjson.js
fhemberger: First braces can be removed, second braces are required when returning an object from an Arrow function.

PR #1632, opened 2018-04-21
remove banner

Trott:
The banner for the March 2018 security updates has been up for many
weeks now. Not sure if there's a policy on how long to leave it up for,
but it seems that by having it there all the time, we encourage people
to not notice when the text changes to from "Important March 2018
security upgrades now available" to (say) "Important May 2018 security
upgrades now available". At least, that's my thinking...


fhemberger:
👍 Thanks!

PR #1636, opened 2018-04-25
Add Node 4 EOL banner

fhemberger:
I just re-used the regular banner in this case. It still can be activated as well in case of an important security release.

The message probably won't stay long, so I passed on making it translatable.

/cc @ZibbyKeaton

### DIFF
--- a/layouts/index.hbs
+++ b/layouts/index.hbs
@@ -12,6 +12,13 @@
 
         {{{ contents }}}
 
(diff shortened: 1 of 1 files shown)


ZibbyKeaton:
Thank you @fhemberger !

ckotzbauer (approved):

Tiriel (approved):
LGTM

PR #1661, opened 2018-05-21
Use bash code blocks to fix ampersand rendering

### LINKED ISSUE #1660: && displayed as &amp;&amp;
* **URL**: 
 * en: https://nodejs.org/en/download/package-manager/#freebsd
 * uk: https://nodejs.org/uk/download/package-manager/#freebsd

I guess that it could be caused by the fact that those code blocks use sh instead of bash for syntax coloring.

0mp:
If sh is used instead of bash "&&" is displayed as "&amp;&amp;".

Fix #1660 (hopefully).


WaleedAshraf (approved):
LGTM

daxlab (approved):
@fhemberger can we land this ?

fhemberger:
Awesome, thanks!

PR #1666, opened 2018-05-29
Add buffer deprecation migration guide

fhinkel:
Add guide from https://github.com/ChALkeR/safer-buffer/blob/master/Porting-Buffer.md

Fixes: https://github.com/nodejs/node/issues/19827
Refs: https://github.com/nodejs/node/issues/19079

/cc @ChALkeR @addaleax @Trott

### DIFF
--- a/locale/en/docs/guides/buffer-constructor-deprecation.md
+++ b/locale/en/docs/guides/buffer-constructor-deprecation.md
@@ -1,0 +1,271 @@
+# Porting to the Buffer.from/Buffer.alloc API
+
+<a id="overview"></a>
+## Overview
+
+This guide explains how to migrate to safe Buffer constructor methods. The migration fixes the following deprecation warning:
+
+`The Buffer() and new Buffer() constructors are not recommended for use due to security and usability concerns. Please use the new Buffer.alloc(), Buffer.allocUnsafe(), or Buffer.from() construction methods instead.`
+
+- [Variant 1: Drop support for Node.js ≤ 4.4.x and 5.0.0 — 5.9.x.](#variant-1) (*recommended*)
+- [Variant 2: Use a polyfill](#variant-2)
+- [Variant 3: Manual detection, with safeguards](#variant-3)
+
+### Finding problematic bits of code using grep
+
+Just run `grep -nrE '[^a-zA-Z](Slow)?Buffer\s*\(' --exclude-dir node_modules`.
+
+It will find all the potentially unsafe places in your own code (with some considerably unlikely
+exceptions).
+
+### Finding problematic bits of code using Node.js 8
+
+If you’re using Node.js ≥ 8.0.0 (which is recommended), Node.js exposes multiple options that help with finding the relevant pieces of code:
+
+- `--trace-warnings` will make Node.js show a stack trace for this warning and other warnings that are printed by Node.js.
+- `--trace-deprecation` does the same thing, but only for deprecation warnings.
+- `--pending-deprecation` will show more types of deprecation warnings. In particular, it will show the `Buffer()` deprecation warning, even on Node.js 8.
+
+You can set these flags using an environment variable:
+
+```console
+$ export NODE_OPTIONS='--trace-warnings --pending-deprecation'
+$ cat example.js
+'use strict';
+const foo = new Buffer('foo');
+$ node example.js
+(node:7147) [DEP0005] DeprecationWarning: The Buffer() and new Buffer() constructors are not recommended for use due to security and usability concerns. Please use the new Buffer.alloc(), Buffer.allocUnsafe(), or Buffer.from() construction methods instead.
+    at showFlaggedDeprecation (buffer.js:127:13)
+    at new Buffer (buffer.js:148:3)
+    at Object.<anonymous> (/path/to/example.js:2:13)
+    [... more stack trace lines ...]
+```
+
+### Finding problematic bits of code using linters
+
+Eslint rules [no-buffer-constructor](https://eslint.org/docs/rules/no-buffer-constructor)
+or
+[node/no-deprecated-api](https://github.com/mysticatea/eslint-plugin-node/blob/master/docs/rules/no-deprecated-api.md)
+also find calls to deprecated `Buffer()` API. Those rules are included in some presets.
+
+There is a drawback, though, that it doesn't always
+[work correctly](https://github.com/chalker/safer-buffer#why-not-safe-buffer) when `Buffer` is
+overridden e.g. with a polyfill, so recommended is a combination of this and some other method
+described above.
+
+<a id="variant-1"></a>
+## Variant 1: Drop support for Node.js ≤ 4.4.x and 5.0.0 — 5.9.x.
+
+This is the recommended solution nowadays that would imply only minimal overhead.
+
+The Node.js 5.x release line has been unsupported since July 2016, and the Node.js 4.x release line reaches its End of Life in April 2018 (→ [Schedule](https://github.com/nodejs/Release#release-schedule)). This means that these versions of Node.js will *not* receive any updates, even in case of security issues, so using these release lines should be avoided, if at all possible.
+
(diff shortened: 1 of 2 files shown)


locale/en/docs/guides/buffer-constructor-deprecation.md
```diff
@@ -0,0 +1,271 @@
...
+(node:7147) [DEP0005] DeprecationWarning: The Buffer() and new Buffer() constructors are not recommended for use due to security and usability concerns. Please use the new Buffer.alloc(), Buffer.allocUnsafe(), or Buffer.from() construction methods instead.
+    at showFlaggedDeprecation (buffer.js:127:13)
+    at new Buffer (buffer.js:148:3)
+    at Object.<anonymous> (/path/to/example.js:2:13)
+    [... more stack trace lines ...]
+```
+
+### Finding problematic bits of code using linters
+
+Eslint rules [no-buffer-constructor](https://eslint.org/docs/rules/no-buffer-constructor)
```
Trott: Nit: `ESLint`

locale/en/docs/guides/buffer-constructor-deprecation.md
```diff
@@ -0,0 +1,271 @@
...
+What you would do in this case is to convert all `new Buffer()` or `Buffer()` calls to use `Buffer.alloc()` or `Buffer.from()`, in the following way:
+
+- For `new Buffer(number)`, replace it with `Buffer.alloc(number)`.
+- For `new Buffer(string)` (or `new Buffer(string, encoding)`), replace it with `Buffer.from(string)` (or `Buffer.from(string, encoding)`).
+- For all other combinations of arguments (these are much rarer), also replace `new Buffer(...arguments)` with `Buffer.from(...arguments)`.
+
+Note that `Buffer.alloc()` is also _faster_ on the current Node.js versions than
+`new Buffer(size).fill(0)`, which is what you would otherwise need to ensure zero-filling.
+
+Enabling eslint rule [no-buffer-constructor](https://eslint.org/docs/rules/no-buffer-constructor)
```
Trott: Nit: `ESLint`

locale/en/docs/guides/buffer-constructor-deprecation.md
fhinkel: Good catch. Done.

Trott (approved):
Quick-read about-to-pack-for-a-flight LGTM!

ghost (approved):
LGTM

locale/en/docs/guides/buffer-constructor-deprecation.md
```diff
@@ -0,0 +1,271 @@
...
+Utilize [safer-buffer](https://www.npmjs.com/package/safer-buffer) as a polyfill to support older
+Node.js versions.
+
+You would take exactly the same steps as in [Variant 1](#variant-1), but with a polyfill
+`const Buffer = require('safer-buffer').Buffer` in all files where you use the new `Buffer` API.
+
+Make sure that you do not use old `new Buffer` API — in any files where the line above is added,
+using old `new Buffer()` API will _throw_. It will be easy to notice that in CI, though.
+
+Alternatively, you could use [buffer-from](https://www.npmjs.com/package/buffer-from) and/or
```
LinusU, line 99: I _think_ that does 4 additional dependencies is actually smaller in size than `safer-buffer`, maybe it could be fair to promote all three approaches more equally, and having drawbacks for each listed...

note: I'm the author of buffer-from, etc. so I'm probably very biased here 😄

Trott: @LinusU Would you be willing to draft up proposed alternate text for this part?

LinusU: Actual size of `node_modules` folder as reported by `du -sh node_modules`:

```text
safer-buffer                              60K
buffer-{alloc,alloc-unsafe,fill,from}     48K
```

LinusU: @Trott Sure thing!

Somebody should review it carefully so that I don't try to sell my own modules too much though 😄

LinusU: My proposal for the relevant part:

```markdown
<a id="variant-2"></a>
## Variant 2: Use a polyfill

There are three different polyfills available:

- **[safer-buffer](https://www.npmjs.com/package/safer-buffer)** is a drop-in replacement for the
  entire `Buffer` API, that will _throw_ when using `new Buffer()`.

  You would take exactly the same steps as in [Variant 1](#variant-1), but with a polyfill
  `const Buffer = require('safer-buffer').Buffer` in all files where you use the new `Buffer` API.

  Make sure that you do not use old `new Buffer` API — in any files where the line above is added,
  using old `new Buffer()` API will _throw_. It will be easy to notice that in CI, though.

- **[buffer-from](https://www.npmjs.com/package/buffer-from) and/or
  [buffer-alloc](https://www.npmjs.com/package/buffer-alloc)** are
  [ponyfills](https://ponyfill.com/) for their respective part of the `Buffer` API. You only need
  to add the package(s) corresponding to the API you are using.

  You would import the module needed with an appropriate name, e.g.
  `const bufferFrom = require('buffer-from')` and then use that instead of the call to
  `new Buffer`, e.g. `new Buffer('test')` becomes `bufferFrom('test')`.

  A downside with this approach is slightly more code changes to migrate off them (as you would be
  using e.g. `Buffer.from` under a different name).

- **[safe-buffer](https://www.npmjs.com/package/safe-buffer)** is also a drop in replacement for
  the entire `Buffer` API, but using `new Buffer()` will still work as before.

  A downside to this approach is that it will allow you to also use the older `new Buffer()` API
  in your code, which is problematic since it can cause issues in your code, and will start
  emitting runtime deprecation warnings starting with Node.js 10
  ([read more here](https://github.com/chalker/safer-buffer#why-not-safe-buffer)).

Note that in either case, it is important that you also remove all calls to the old Buffer
API manually — just throwing in `safe-buffer` doesn't fix the problem by itself, it just provides
a polyfill for the new API. I have seen people doing that mistake.

Enabling ESLint rule [no-buffer-constructor](https://eslint.org/docs/rules/no-buffer-constructor)
or
[node/no-deprecated-api](https://github.com/mysticatea/eslint-plugin-node/blob/master/docs/rules/no-deprecated-api.md)
is recommended.

_Don't forget to drop the polyfill usage once you drop support for Node.js < 4.5.0._
```

Rendered:

screen shot 2018-05-30 at 01 04 55

LinusU: ping @Trott @fhinkel ☝️

fhinkel: Do you mind if we merge this as is and put your changes in a separate PR to make reviewing and attribution easier?

LinusU: Sunds good 👍

fhinkel: Thank you!

ckotzbauer (approved):

locale/en/docs/guides/buffer-constructor-deprecation.md
LinusU: Here it is :)

https://github.com/nodejs/nodejs.org/pull/1674

vsemozhetbyt:
Is this style and structure difference intended?

Any other guide, say https://nodejs.org/en/docs/guides/getting-started-guide/ :




This guide (no header, no left side TOC, no code blocks styling, etc), https://nodejs.org/en/docs/guides/buffer-constructor-deprecation/ :

Trott:
> Is this style and structure difference intended?

@vsemozhetbyt Unintentional, I'm sure. PR to fix welcome. @nodejs/website

fhinkel:
Not intended, thanks for noticing.  I *think* I just fixed it.

PR #1679, opened 2018-06-03
guides: fix nits in buffer-constructor-deprecation

vsemozhetbyt:
This PR unify some formatting differences across the doc and fix some other nits.

### DIFF
--- a/locale/en/docs/guides/buffer-constructor-deprecation.md
+++ b/locale/en/docs/guides/buffer-constructor-deprecation.md
@@ -1,22 +1,22 @@
 ---
-title: Porting to the Buffer.from/Buffer.alloc API
+title: Porting to the Buffer.from()/Buffer.alloc() API
 layout: docs.hbs
 ---
 
-# Porting to the Buffer.from/Buffer.alloc API
+# Porting to the `Buffer.from()`/`Buffer.alloc()` API
 
 <a id="overview"></a>
 ## Overview
 
-This guide explains how to migrate to safe Buffer constructor methods. The migration fixes the following deprecation warning:
+This guide explains how to migrate to safe `Buffer` constructor methods. The migration fixes the following deprecation warning:
 
 `The Buffer() and new Buffer() constructors are not recommended for use due to security and usability concerns. Please use the new Buffer.alloc(), Buffer.allocUnsafe(), or Buffer.from() construction methods instead.`
 
-- [Variant 1: Drop support for Node.js ≤ 4.4.x and 5.0.0 — 5.9.x.](#variant-1) (*recommended*)
+- [Variant 1: Drop support for Node.js ≤ 4.4.x and 5.0.0 — 5.9.x](#variant-1) (*recommended*)
 - [Variant 2: Use a polyfill](#variant-2)
 - [Variant 3: Manual detection, with safeguards](#variant-3)
 
-### Finding problematic bits of code using grep
+### Finding problematic bits of code using `grep`
 
 Just run `grep -nrE '[^a-zA-Z](Slow)?Buffer\s*\(' --exclude-dir node_modules`.
 
@@ -31,9 +31,9 @@ If you’re using Node.js ≥ 8.0.0 (which is recommended), Node.js exposes mult
 - `--trace-deprecation` does the same thing, but only for deprecation warnings.
 - `--pending-deprecation` will show more types of deprecation warnings. In particular, it will show the `Buffer()` deprecation warning, even on Node.js 8.
 
-You can set these flags using an environment variable:
+You can set these flags using environment variables:
 
-```console
+```bash
 $ export NODE_OPTIONS='--trace-warnings --pending-deprecation'
 $ cat example.js
 'use strict';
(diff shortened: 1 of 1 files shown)


locale/en/docs/guides/buffer-constructor-deprecation.md
```diff
@@ -31,10 +31,11 @@ If you’re using Node.js ≥ 8.0.0 (which is recommended), Node.js exposes mult
...
 - `--pending-deprecation` will show more types of deprecation warnings. In particular, it will show the `Buffer()` deprecation warning, even on Node.js 8.
 
-You can set these flags using an environment variable:
+You can set these flags using environment variables:
 
-```console
-$ export NODE_OPTIONS='--trace-warnings --pending-deprecation'
+```bash
+$ export NODE_OPTIONS=--trace-warnings
+$ export NODE_PENDING_DEPRECATION=1
```
vsemozhetbyt: It seems [we have not](https://nodejs.org/download/nightly/v11.0.0-nightly20180530c1a87a30e7/docs/api/cli.html#cli_node_options_options) `--pending-deprecation` in `NODE_OPTIONS`.

locale/en/docs/guides/buffer-constructor-deprecation.md
```diff
@@ -31,10 +31,11 @@ If you’re using Node.js ≥ 8.0.0 (which is recommended), Node.js exposes mult
 - `--trace-deprecation` does the same thing, but only for deprecation warnings.
 - `--pending-deprecation` will show more types of deprecation warnings. In particular, it will show the `Buffer()` deprecation warning, even on Node.js 8.
 
-You can set these flags using an environment variable:
+You can set these flags using environment variables:
 
-```console
-$ export NODE_OPTIONS='--trace-warnings --pending-deprecation'
+```bash
```
vsemozhetbyt: `console` tag seems to produce [weird rendering](https://nodejs.org/en/docs/guides/buffer-constructor-deprecation/#finding-problematic-bits-of-code-using-node-js-8):

vsemozhetbyt:
I am a bit confused by [this fragment](https://github.com/nodejs/nodejs.org/blame/6aaedb203b835d7cb356bb80a0dfb2f00ea975c1/locale/en/docs/guides/buffer-constructor-deprecation.md#L163-L164) in the example code. Are the condition check and error message in some contradiction?

vsemozhetbyt:
cc @nodejs/documentation @nodejs/buffer

locale/en/docs/guides/buffer-constructor-deprecation.md
lpinca: @vsemozhetbyt https://nodejs.org/download/nightly/v11.0.0-nightly20180530c1a87a30e7/docs/api/cli.html#cli_pending_deprecation

lpinca:
> I am a bit confused by this fragment in the example code. Are the condition check and error message in some contradiction?

Yes, I think so the message should be 'The "size" argument must **not** be of type number.'

locale/en/docs/guides/buffer-constructor-deprecation.md
vsemozhetbyt: Yes, we have it as a separate cli key, but not as a `NODE_OPTIONS` env variable part option.

lpinca: I'm not sure I follow, `--pending-deprecation` is a CLI option right? If so it should work when used via `NODE_OPTIONS` which allows to specify CLI options in the environment via a space-separated list.

vsemozhetbyt: It seems we have [a whitelist of cli keys that are allowed in `NODE_OPTIONS`](https://nodejs.org/download/nightly/v11.0.0-nightly20180609ef1f130041/docs/api/cli.html#cli_node_options_options):

> Node options that are allowed are: ...

lpinca: Was it removed recently? It works on Node.js 10.

lpinca: I think it is simply not listed there.

vsemozhetbyt: Node.js 10 doc has not it as well:

https://nodejs.org/dist/latest-v10.x/docs/api/cli.html#cli_node_options_options

If it does work, maybe the API docs need updating.

lpinca: Yes it seems outdated, see https://github.com/nodejs/node/blob/master/src/node.cc#L3156-L3198

vsemozhetbyt: Thank you. I will revert this change.

vsemozhetbyt:
> Yes, I think so the message should be 'The "size" argument must **not** be of type number.'

@fhinkel Is this an appropriate fix?

locale/en/docs/guides/buffer-constructor-deprecation.md
vsemozhetbyt: PR to update the API doc: https://github.com/nodejs/node/pull/21229

fhemberger:
@vsemozhetbyt Thanks!

PR #1680, opened 2018-06-06
doc: june 2018 sec pre-announce

mhdawson:
Add blog for June 2018 pre-announce
Update header to point to pre-announce

### DIFF
--- a/build.js
+++ b/build.js
@@ -254,9 +254,9 @@ function getSource (callback) {
           lts: latestVersion.lts(versions)
         },
         banner: {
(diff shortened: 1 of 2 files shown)


mhdawson:
I'm heading off to bed, so if there are approvals it would be good if somebody who is around in the next hour or so can land, otherwise, I'll land first thing when I get in tomorrow morning.

Trott (approved):

fhemberger (approved):

PR #1689, opened 2018-06-12
[Localization] Fix a typo by adding a missing `to`

ghost:
It should be `wanted to bring` instead of `wanted bring`.


fhemberger:
Thank you, good catch!

ghost:
Nothing special :)

PR #1700, opened 2018-06-15
Use Node.js version number instead of codename in Docker guide

sonicdoe:
The use of Node.js’s release codenames isn’t very widespread and, if I had to guess, not a lot of people are aware that Node.js v8 is codenamed Carbon. Therefore, I’d argue that using the actual version number in the Docker guide would reduce confusion and make it more obvious which Node.js version is being used.


lpinca (approved):

fhemberger:
Thank you!

Issue #806, opened 2016-06-28
Language picker


Fishrock123:
Now that https://github.com/nodejs/nodejs.org/pull/490 has landed and the related build/deployment work seems to be completed, we should be able to get the language groups once again more involved with the website.

We should also start thinking of a way to choose the language.

I'm not sure io.js's way was really that great. I know we discussed the flag thing before, but to me that is the most easily recognizable way.

We don't have to have that open a country list, but we can use it as a prompt for language nonetheless.

cc @nodejs/website 
cc also @leo who may have some ideas?

leo:
That seems like something that should be implemented with the next redesign. So it would be great if you could add it to [this repo](https://github.com/nodejs/nodejs.org-new/issues). More details [here](https://github.com/nodejs/nodejs.org/issues/779):

> To keep things clear, I'd like all upcoming design-related issues to be forwarded to the design repo so that I can consider them while designing.

Fishrock123:
Sure, I guess? But we kinda need it before then. Can't hold everything up for that.

ghost:
one thing we could do if we're really getting translation efforts back up again is probe for the activeness of the i18n WGs

fhemberger:
Quoting myself here from https://github.com/nodejs/iojs.org/issues/235:

> Flags != Languages.
> We could add automatic redirection to the localized version based on the user's 'Accept-Language' header settings. There's an nginx module for that: http://wiki.nginx.org/AcceptLanguageModule

**EDIT** New link: https://www.nginx.com/resources/wiki/modules/accept_language/

I think it's the most reliable way to determine a user's language without the need for user interaction.

ghost:
we should also probably assemble a separate team to control translation efforts. the iojs way wasn't the best

leo:
@Fishrock123 Ahh, no problem! Didn't know that.

ghost:
@fhemberger i wouldn't do it automatically though. maybe a hint from somewhere within the site asking if you want to switch to the version of the site in your language

Fishrock123:
@ore https://github.com/nodejs/nodejs-ko is quite active.

---

> > Flags != Languages.

My point also, but we shouldn't have the selection be flags, only the initial, along with the Language indicator.

I.e. 

[🇩🇪 DE] is a little more visible than [DE] for such a thing.

[...]

PR #1703, opened 2018-06-16
[Fix] Fix typo errors

ghost:
[Senario]
Someone is now accessing the home page with a root path `/` ( Such as `https://nodejs.org/`)：

The current is now guiding you to an English page.

Now I guide you to a page where you like according to your local system language from `accept-language`, which looks better and more friendly.

[Notices]
Please run `npm i` to install a 3-rd party `accept-language` before you run or deploy this.

### DIFF
--- a/server.js
+++ b/server.js
@@ -36,7 +36,7 @@ const layouts = chokidar.watch(path.join(__dirname, 'layouts'), opts)
 const statics = chokidar.watch(path.join(__dirname, 'static'), opts)
 
 // Redirect mechanism meant as a fix for languages where some pages
-// has not translated yet, therefore redirect to the english equivalent,
+// have not been translated yet, therefore redirect to the english equivalent,
 // which isn't the correct language, but better than a 404-page
 function redirectToEnglishUrl (req, res) {
   return () => {


ghost (approved):
LGTM!

lpinca:
This would not work in production as the server is only used for development. Not sure how useful this feature is for development.

ghost:
@lpinca：
I had a test before submitting here, and everything goes fine with me.

For FF and Chrome, it can read out `accept-language` and guide me where I actually want to go.
For IE, some IEs cannot pre-set the `accept-Language` so the default is still `English`.

All in all, I only check when you are accessing the root without a default language, it will try to fetch your pre-set lang from the browser (Maybe according to your localization system) and guide you to a correct website of a lang. Because I didn't change any rules but just attach dynamically lang to the root 

E.g:
If you are accessing——`https://nodejs.org/ `=> It will guide to `https://nodejs.org/lang `, if some `lang` is checked.

And `This would not work in production as the server is only used for development. ` What does this mean? Have u got any special things on your server? So have you got any suggestions or ideas?
Is it possible to make the same function on the deployment?

Thanks anyway!

fhemberger:
@Maledong Files in production are served pre-gzipped through nginx (using `npm run deploy`) for performance reasons. Config can be found here: https://github.com/nodejs/build/blob/master/setup/www/resources/config/nodejs.org

We already had a discussion about this in: https://github.com/nodejs/nodejs.org/issues/806#issuecomment-229019081 (and even in the io.js days: https://github.com/nodejs/iojs.org/issues/235)

Unfortunately, there's only a 3rd party module for nginx. So this has to be addressed by the @nodejs/build team.

ghost:
Well thanks all!

I changed my commit to fix a typo error instead. Language will be discussed with all of you later after I read your discussion first.

amandeepmittal (approved):

gibfahn (approved):

tniessen (approved):

PR #1711, opened 2018-06-24
Fix extra space before period

bjb568:
> Do not introduce changes which make seamless upgrades impossible .

Fixed to:

> Do not introduce changes which make seamless upgrades impossible.


fhemberger:
Thanks!

PR #1715, opened 2018-07-01
[Fix] Make the `redirectToEnglishUrl` behaves properly for local debugging

ghost:
@fhemberger ：

According to the normal webside, we have 4 kinds of url forms:

1) https://localhost:8080/: Access the route, the default lang is `en`.
2) https://localhost:8080/zh-cn/docs/: If we don't have the proper
translations, just direct you to `en`.
3) https://localhost:8080/zh-nc/docs/: If we have a wrong language spelling, 404 will
be returned.
4) https://localhost:8080/blog: When directly accessing a folder, just add `en`
as the default lang (converted to `https://localhost:8080/en/blog`).

For other page routes, `404` will be redirected to you.

So the original codes don't meet all these cases. (e.g: https://localhost:8080/zh-cn/docs/, if you don't have doc translated files or folders, this will return you 404 instead of returning you a default `en` page).

But according to the comment, it SHOULD return you a default `en` doc files:)

Here's the fix for that.

### DIFF
--- a/server.js
+++ b/server.js
@@ -34,28 +34,34 @@ const opts = {
 const locales = chokidar.watch(path.join(__dirname, 'locale'), opts)
 const layouts = chokidar.watch(path.join(__dirname, 'layouts'), opts)
 const statics = chokidar.watch(path.join(__dirname, 'static'), opts)
+const fs = require('fs')
+// Read all the langs under `locale`
+const SUPPORTED_LANGUAGES = new Set(fs.readdirSync(path.join(__dirname, 'locale')))
 
 // Redirect mechanism meant as a fix for languages where some pages
 // have not been translated yet, therefore redirect to the english equivalent,
 // which isn't the correct language, but better than a 404-page
 function redirectToEnglishUrl (req, res) {
   return () => {
-    // Union the Url to lower case (ignore the case sensitive)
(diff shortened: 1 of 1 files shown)


ghost (approved):

server.js
```diff
@@ -35,27 +35,32 @@ const locales = chokidar.watch(path.join(__dirname, 'locale'), opts)
 const layouts = chokidar.watch(path.join(__dirname, 'layouts'), opts)
 const statics = chokidar.watch(path.join(__dirname, 'static'), opts)
 
+const OTHER_SUPPORTED_LANGS = new Set(['ar', 'ca', 'de', 'es', 'fr', 'gl', 'it', 'ja', 'ko', 'uk', 'zh-cn', 'zh-tw'])
+
```
marswong: It'd be better to use `fs` to get the directories under `locale` instead of fixed values.

marswong (approved):
LGTM

server.js
Maledong: Good idea, I forgot and I'll fix that :)

ghost: Fixed, many thanks!

ghost:
@marswong ： Fixed, thanks!

PR #1722, opened 2018-07-09
[Fix] Fix styles of `hbs` files

ghost:
Because there is something wrong in formation, and when we run `npm run test` we cannot continue：
error
This is the fixture for that：
success


fhemberger (approved):

fhemberger (approved):
Little nit: You could remove the second space (after `}}`).

ghost:
OK, done. Thanks for tips

fhemberger:
Thanks!

PR #1724, opened 2018-07-09
Link to GitHub for licensing information

sonicdoe:
The contents are the same but the [normal view](https://github.com/nodejs/node/blob/master/LICENSE) is nicer to read than the [raw view](https://raw.githubusercontent.com/nodejs/node/master/LICENSE).

### DIFF
--- a/layouts/partials/footer.hbs
+++ b/layouts/partials/footer.hbs
@@ -20,7 +20,7 @@
       <p>Linux Foundation is a registered trademark of The Linux Foundation.</p>
       <p>Linux is a registered <a href="http://www.linuxfoundation.org/programs/legal/trademark" title="Linux Mark Institute">trademark</a> of Linus Torvalds.</p>
       <p>
-        <a href="https://raw.githubusercontent.com/nodejs/node/master/LICENSE">Node.js Project Licensing Information</a>.
+        <a href="https://github.com/nodejs/node/blob/master/LICENSE">Node.js Project Licensing Information</a>.
       </p>
     </div>
   </div>


ckotzbauer (approved):

ghost (approved):

WaleedAshraf (approved):

fhemberger:
Hmm, not sure about this. Is linking to GitHub a benefit?

sonicdoe:
I’d say so, to expand a bit: The [normal view](https://github.com/nodejs/node/blob/master/LICENSE) specifies a custom monospace font and centers the whole page which makes it nicer to read. Of course it also adds GitHub’s other interface elements such as line numbers and history which arguably could be beneficial, too.

(Granted, I’m not expecting a lot of visitors to click on this link but I thought it would be a nice small change for the better.)

daxlab (approved):

refack:
I tried to look for prior art, and what I did find were projects that host the license on their site, linked to with something like:
```
Except as noted, the content of this page is licensed under the Creative Commons Attribution 3.0 License, and code is licensed under a BSD license.
```
* https://angular.io/license
* https://docs.python.org/3/license.html
* https://golang.org/LICENSE
* https://www.ruby-lang.org/en/about/license.txt

(I tried to think of projects that are GitHub based)

So I'm generally -1 on this PR's implemented, but would be +1 on serving a formatted copy from the site.

fhemberger:
Agreeing with @refack here.

I know this license document is [automatically generated](https://github.com/nodejs/node/blob/master/tools/license-builder.sh), but the formatting is horrible either way.

sonicdoe:
I’m also in favor of a nicer license page, however, the `LICENSE` file has the benefit of having one canonical location. Therefore, if there were to be a separate page, it should be generated from the `LICENSE` file (or we could generate a `LICENSE.md` instead).

Having said that, should we hold off on this pull request because of this? After all, merging this change doesn’t mean we can’t improve further later on 🙂

refack:
> Having said that, should we hold off on this pull request because of this? After all, merging this change doesn’t mean we can’t improve further later on

I'd rather keep this as is, for several reasons:
* The current link is a URI to the license, while the suggested change will direct to a _view_ of that resource
* The github view is "heavier" with all the UI resources
* Is not (trivially) machine readable
* Not that much better looking (IMHO)
* Directs to a resource that is not under our control

sonicdoe:
@refack What would be your preferred approach? As far as I can tell, your first three points would also apply to a formatted copy on our site.

Maledong:
To balance the formation style with generation isn't an easy thing，maybe we can use a auto convertor.
Here's the website for that, so you can paste our License and have a play with that:
https://cloudconvert.com/txt-to-md

And for API (Free max file size to 1GB per day, or 25 minutes, that's enough for us now：https://cloudconvert.com/pricing).

API in details：https://cloudconvert.com/api/conversions.

Hope this helps to gather all of your benifits together!

Maledong:
So can this be merged?

aymen94 (approved):

aymen94 (approved):

refack:
> @refack What would be your preferred approach? As far as I can tell, your first three points would also apply to a formatted copy on our site.

So this is a bit of a delicate point. Since the main LICENSE file for node is re-generated with each dependency change, de-facto each node release has it's own unique license file...

To the issue of what we display on the site, I'd prefer a site URL formatted with the site's template. Else I'd rather keep this as is.

tl;dr I'm -1 on this change

aymen94:
@refack I agree with you.

ghost:
Hi ：

Since this has been a long time here and because we have examples of licenses in plain text, even we want to solve the problem I think we should do some changes in the Nodejs project's LICENSE file instead of here. So I close the submit and it could be also re-opened if there's anything urgent or solution. 

Thanks all!

PR #1719, opened 2018-07-08
[Fix] Remove duplicated Translations for `index.md` with unnecessary double quotes

ghost:
1) Remove duplicated Chinese translations for `index.md`, there's ONLY
one explaination.

2) Remove double quotes for `newsletter-postfix`. We don't need them and
the result seems the same.

image

PR #1727, opened 2018-07-10
nodejs-zh-TW: Improve translation

osk2:
I made some improvements as mentioned in #1719 

1. Remove unnecessary quotation mark
2. Remove unnecessary translation


fhemberger (commented):
Not sure about removing the quotation marks. Didn't check: Is this valid YAML having two colons? Does the site build?

+1 for removing last sentence!

ghost:
Please make sure that your second ":" is in Chinese mode instead of English one., or there's no spaces between your `:` and the 2nd Chinese character. Otherwises the quotation marks should be added. I've had a test in my local machine by running `npm run test` after `npm run build`.

Compare with each other:

1) The `:` symbol in the Chinese mode, the color is yellow, which means this is of a string type, and you don't need the double quotation marks (Chinese version should match the Chinese symbol double quotations, recommanded!)：
image

The English double quotation marks with NO spaces between 2nd Chinese character should be also OK:
image

The result is：
image

But YOU CANNOT USE AN ENGLISH MODE'S double quotations with a space like this, notice the color is changed for `:`：
image

And you're getting WRONG page like this following:
image

PeterDaveHello (changes requested):
The "Merge" commits should be removed.

osk2:
It passed tests and looks no difference since I use fullwidth colon. But I will add quotation mark back to avoid confusion.

@PeterDaveHello oops, I'll open another PR soon.

ghost:
@osk2 ：BTW，In fact you can remove Merge with the help of "Reset and Rebase" and re-submit instead of closing that :)

fhemberger:
We'll squash and merge the commits anyway, so no worries about that.

osk2:
@Maledong Thanks for the tip, I'll try next time 👍

PR #1728, opened 2018-07-10
nodejs-zh-TW: Remove unnecessary translation

osk2:
As mentioned in #1719 , this PR removed unnecessary translation


fhemberger:
👍

PR #1734, opened 2018-07-14
Add hover effect to banner on home page

sonicdoe:
The download buttons have a similar hover effect, so I figured it would be nice to have something for the JS Interactive banner, too.

### DIFF
--- a/layouts/css/page-modules/_home.styl
+++ b/layouts/css/page-modules/_home.styl
@@ -40,6 +40,13 @@
   h2
     margin-bottom 0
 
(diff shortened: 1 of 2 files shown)


ghost (approved):

ckotzbauer (approved):

WaleedAshraf (approved):

fhemberger:
Hmm, not really sure if it's a good idea to make a graphic *more* transparent on hover, but you're right, at least it would behave like other links.

PR #1736, opened 2018-07-15
[Fix] Fix typo error of `Stack Overflow`

ghost:
Continue for https://github.com/nodejs/nodejs.org/pull/1735.

PS：Since `data.min.js` comes from `data.js`. So this won't be changed.


fhemberger:
Thanks!

sonicdoe:
I wasn’t sure whether to update old blog posts and external sites. Thanks for following up 👍

PR #1751, opened 2018-07-26
Fix spelling of “JavaScript”

sonicdoe:
“Javascript” → “JavaScript”

PR #1753, opened 2018-07-27
[Fix] Fix typo of "Javascript" for missing parts

ghost:
`s` should be `S`：

Ref： https://github.com/nodejs/nodejs.org/pull/1751.
Ref：https://github.com/nodejs/node/commit/c6852126fd


fhemberger:
Thanks!

PR #1755, opened 2018-07-28
[Fix] Update dnt codes

ghost:
1) Remove useless website link.
2) Simplify return value.

### DIFF
--- a/static/js/dnt_helper.js
+++ b/static/js/dnt_helper.js
@@ -1,5 +1,4 @@
 /**
- * http://schalkneethling.github.io/blog/2015/11/06/respect-user-choice-do-not-track/
  * https://github.com/schalkneethling/dnt-helper/blob/master/js/dnt-helper.js
  *
  * Returns true or false based on whether doNotTack is enabled. It also takes into account the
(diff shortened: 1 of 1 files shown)


lpinca (approved):

lpinca:
Website link is not useless, but it no longer works.

ckotzbauer (approved):

marswong (approved):

phillipj (approved):

Trott:
I wonder if the file can be removed entirely and any calls to `_dntEnabled()` replaced with `navigator.doNotTrack || window.doNotTrack`. (The `window` check is for IE 11. If you want to be REALLY accommodating and account for IE 9 and IE 10, throw in a ` || navigator.msDoNotTrack`.)

This file seems to exist mostly for operating systems and browsers we probably don't need to take particular care in accommodating. (Windows NT 6.1??!!  Firefox 31??!!) To get an idea of just how broadly the Do Not Track API is supported on current browsers: https://caniuse.com/#search=dnt and click "Show All":

screen shot 2018-08-29 at 11 13 35 pm

Note that even for browsers that show red in that chart, it's not a problem. It just means the browsers don't support Do Not Track, which just means it will never be enabled on those browsers and that's OK. We'll just get the falsy `undefined` for those browsers. If the user can't set Do Not Track in their browser, then we aren't failing to respect Do Not Track.

PR #1764, opened 2018-08-08
Update dependencies

ghost:
Update some dependencies in the package.json as a general maintance.

### DIFF
--- a/package.json
+++ b/package.json
@@ -33,7 +33,7 @@
   "dependencies": {
     "autoprefixer-stylus": "0.14.0",
     "cheerio": "0.22.0",
-    "chokidar": "2.0.3",
+    "chokidar": "^2.0.4",
     "handlebars": "4.0.11",
     "jstransformer-handlebars": "1.1.0",
     "junk": "2.1.0",
@@ -54,7 +54,7 @@
     "ncp": "2.0.0",
     "node-version-data": "1.0.1",
     "octonode": "^0.9.3",
-    "request": "2.85.0",
+    "request": "^2.87.0",
     "require-dir": "1.0.0",
     "semver": "5.5.0",
     "st": "1.2.2",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -33,12 +33,12 @@
...
     "autoprefixer-stylus": "0.14.0",
     "cheerio": "0.22.0",
-    "chokidar": "2.0.3",
+    "chokidar": "^2.0.4",
     "handlebars": "4.0.11",
     "jstransformer-handlebars": "1.1.0",
     "junk": "2.1.0",
     "lodash.defaultsdeep": "4.6.0",
-    "marked": "0.3.19",
+    "marked": "^0.4.0",
```
Trott: Careful with this one. It may break stuff.

/ping @vsemozhetbyt @rubys

ghost: Thanks! I've ran `npm run test` and everything seems going well with me. So I submitted here……
So what's wrong with that? Have you found something unstable or letting me removing "^"?

Trott: When we tried to update marked 0.3.5 to 0.4.0 in the main repo, it broke all sorts of things in our docs rendering. Granted, that's a bigger jump than is happening here, and the docs probably have more idiosyncratic and/or complicated markdown than in the website. I'm not sure if anything really breaks here. But I'm not sure it doesn't either. Maybe do a build with both 0.3.19 and 0.4.0 and use a diff tool to look at the results just to make sure nothing is obviously wrong? And/or scan through https://github.com/nodejs/node/pull/21081 to get an idea of the types of things that broke for us in the docs? For example:

* https://github.com/nodejs/node/pull/21081#issuecomment-394075360
* https://github.com/nodejs/node/pull/21081#issuecomment-394096060
* https://github.com/nodejs/node/pull/21081#issuecomment-394105036
...and more.

None of those were caught by automated test. It took @vsemozhetbyt painstakingly comparing the output from the two versions and finding these things.

Ultimately, in core, we decided to not update `marked` and instead @rubys did a tremendous job removing `marked` from core dependencies and replacing it with `remark`. That said, we're *still* fixing a few bugs there. It was not a small undertaking.

`marked` is a good lightweight package for markdown processing.  It's not anywhere near fully spec-compliant and doesn't do much in the way of parsing. It relies heavily on a lot of complicated regular expressions instead. `remark` seems to be more of a robust parser and I believe is spec-compliant (or close).

ghost: OK，revert it.

Trott: Maybe put the `marked` update in a separate pull request and people can look at just those changes carefully.

Trott (approved):
LGTM if if tests are passing

WaleedAshraf (approved):

ckotzbauer (approved):

marswong (approved):

ghost (approved):

Issue #1769, opened 2018-08-16
release-post.js script should ignore 32-bit binaries for intel and sunos


rvagg:
I'd fix this but I'm currently doing releases and am likely to be too exhausted at the end to get around to this.

./scripts/release-post.js needs to not put "Coming soon" for 32-bit binaries for Linux x86 or "sunos" x86 for Node 10+, they're deprecated. But it should still do it for 6 & 8. I think the best way might be to put metadata in scripts/helpers/downloads.js to indicate this and filter them out when we're doing a v10 release.

Maledong:
Hi：
What about trying this, if you agree with this, I'll try to submit a PR for download.js：

```javascript
module.exports = (version) => {
  let downloads = latestDownloads
  if (semver.satisfies(version, '< 1.0.0')) {
    downloads = legacyDownloads
  } else if (semver.satisfies(version, '< 8.0.0')) {
    downloads = postMergeDownloads
  } else if (semver.satisfies(version, '> 10.0.0')) {
    downloads = postMergeDownloads.filter(value =>
      !value.title.startsWith('Linux 32-bit') &&
      !value.title.startsWith('SunOS 32-bit'))
  }
  return downloads.map((item) => resolveUrl(item, version))
}
```

PS：Since your origin codes will check whether the version is <8.0.0 and then <1.0.0. Then it will download different links of files. I've optimized them.

Feel free to suggest :)

ghost:
Well, this issue is fixed and landed here: https://github.com/nodejs/nodejs.org/commit/f0818db4c8dd29df9a640ff31ea172b81a864ffd

So I closed this and for any other problems, please feel free to re-open.
Thanks anyway!

PR #1772, opened 2018-08-18
Add filters for Nodejs 10

ghost:
1) Ref: https://github.com/nodejs/nodejs.org/issues/1769
2) Optimize the logic process for module.exports of download.js.

### DIFF
--- a/scripts/helpers/downloads.js
+++ b/scripts/helpers/downloads.js
@@ -139,11 +139,14 @@ function resolveUrl (item, version) {
 
 module.exports = (version) => {
   let downloads = latestDownloads
-  if (semver.satisfies(version, '< 8.0.0')) {
-    downloads = postMergeDownloads
-  }
   if (semver.satisfies(version, '< 1.0.0')) {
     downloads = legacyDownloads
+  } else if (semver.satisfies(version, '< 8.0.0')) {
+    downloads = postMergeDownloads
+  } else if (semver.satisfies(version, '>= 10.0.0')) {
+    downloads = postMergeDownloads.filter(ver =>
+      ver.title !== 'Linux 32-bit Binary' &&
+      ver.title !== 'SunOS 32-bit Binary')
   }
   return downloads.map((item) => resolveUrl(item, version))
 }


scripts/helpers/downloads.js
```diff
@@ -139,11 +139,19 @@ function resolveUrl (item, version) {
...
 module.exports = (version) => {
   let downloads = latestDownloads
-  if (semver.satisfies(version, '< 8.0.0')) {
-    downloads = postMergeDownloads
-  }
   if (semver.satisfies(version, '< 1.0.0')) {
     downloads = legacyDownloads
+  } else if (semver.satisfies(version, '< 8.0.0')) {
+    downloads = postMergeDownloads
+  } else if (semver.satisfies(version, '> 10.0.0')) {
```
ckotzbauer: should this be `>= 10.0.0` or is it unrelated cause this will never run again for 10.0.0 ?

ghost: So are you sure that from the ref #1769, it should be >= 10.0.0? I'm not sure. I see it says `Node 10+`……?

ckotzbauer: Yes, in #1769 it seems that 10.0.0 itself is not included. But the x86 binaries are not shipped for 10.0.0 too. https://nodejs.org/dist/v10.0.0/

ghost: OK, thanks!

ckotzbauer (approved):
LGTM

scripts/helpers/downloads.js
```diff
@@ -139,11 +139,19 @@ function resolveUrl (item, version) {
...
   if (semver.satisfies(version, '< 1.0.0')) {
     downloads = legacyDownloads
+  } else if (semver.satisfies(version, '< 8.0.0')) {
+    downloads = postMergeDownloads
+  } else if (semver.satisfies(version, '>= 10.0.0')) {
+    downloads = postMergeDownloads.map(value => {
+      if (value.title === 'Linux 32-bit Binary' ||
+          value.title === 'SunOS 32-bit Binary') {
+        // Indicate that comingSoon shouldn't be shown
+        value.notShowComingSoon = true
```
lpinca: Nit: wouldn't it be easier to use `showComingSoon` instead of it's negation? Then below we can do

```js
if (binary.showComingSoon) {
  return `${binary.title}: *Coming soon*`
}
return binary.title
```

It's the same but I personally find it easier to grok.

lpinca (approved):

targos:
Will this still show both lines? They should not be there at all in the end.

scripts/helpers/downloads.js
ghost: The reason is that we DON'T wanna show commingSoon, and only those versions >= 10.0.0 will meet the requirement. So most of downloadable items won't have this symbol (They are shown in default, ONLY with this symbol won't be shown).

If I change to yours, I'll have to add the property to the original `postMergeDownloads` JSON objects and set the default value to `true`. In fact there's no need doing that. Hope you understand :)

ghost:
@targos：What do you mean? Can you elebrate it more clear?

scripts/helpers/downloads.js
lpinca: Yes, I do.

targos:
It seems to me that `postMergeDownloads.filter` should be used instead of `postMergeDownloads.map`. What we would like is to remove those lines from the blog posts entirely, not just remove the "Coming soon" part.

scripts/helpers/downloads.js
lpinca: Perhaps you can use something like `hideComingSoon` as property name then.

ghost:
@targos：If I didn't take you wrong, I think what you mean is just directly adding the filter condition to the postMergeDownloads. My 1st version is just what you think now :) (If you are interested in this, please see：https://github.com/Maledong/nodejs.org/commit/52ddf53c2755fa1ad9463fa3cd70bec5552d0602).

However there's a question：In the `urlOrComingSoon` function, we couldn't check the version directly and easily. This means we don't know when to show (or not to show) `Coming Soon`. 

And on the other hand, I'm NOT sure whether versions >10.0.0 also pass `sendRequest`'s check? If it can pass, this means we should still show the link instead of filtering them all out. Because the author seems to say: 

>./scripts/release-post.js needs to not put "Coming soon" for 32-bit binaries for Linux x86 or "sunos" x86 for Node 10+...

IMO：If we can find the link through `sendRequest` in the urlOrComingSoon, still shows the link (So I don't think we should directly do a simple filtering).

Correct me if I take you wrong :) 
And if it's right, I'll convert back to my 1st idea.

scripts/helpers/downloads.js
ghost: OK. Thanks!
And we're still considering whether there's anything better or easier way……Maybe my 1st way is the best, or am I thinking of it a bit too complicated? :D

ghost:
@targos：Maybe you are right, and I'm a little bit too complicated...
I'll fix that:)

ghost:
@targos：Fixed that in a better way by filtering out unmeeted things.

lpinca (approved):

targos:
Thanks. LGTM.

ghost (approved):

PR #1774, opened 2018-08-19
Fix 'use' instead of 'using' grammar error

tombusby:
locale/en/docs/guides/dont-block-the-event-loop.md contains a small grammar error in one of the sentences. This commit fixes it.


fhemberger:
Cool, thanks!

PR #1784, opened 2018-08-30
Remove duplicated semicolon

ghost:
It seems that in this file the very first `;` isn't a must, because we've used a self-executed function. It should be enough.

### DIFF
--- a/static/js/download.js
+++ b/static/js/download.js
@@ -1,4 +1,3 @@
-;
 (function (d, n) {
   'use strict'
 


Maledong:
/cc：@nodejs/nodejs-web

Maledong:
@nodejs/website：Please have a review or suggestions, thanks!

willin:
i think it's useful.  

i used to wrote this way years ago.

```js
;(function (){
  'use strict';
  // xxx
})();
```

ghost:
@willin：Is it a habit or some special thing?

Trott (dismissed):
It's defensive coding. If a previous JS file does not end with a semicolon and this one is concatenated after it, then the start of this file could have unintended consequences. It's a good practice and should be left as is IMO. (We're not concatenating this file as far as I can tell, but we probably should consider concatenating files if we're not, and we should probably keep all our files concatenation-ready.)

ghost:
@Trott：But we can avoid this from happening……And in fact, when I run `npm run build` and `npm run serve`. It seems everything is going well with me. So don't you think it's a little 'excessive defence'?

Trott:
> @Trott：But we can avoid this from happening……And in fact, when I run npm run build and npm run serve. It seems everything is going well with me. So don't you think it's a little 'excessive defence'?

The value of leaving the semicolon is small but non-zero while the value of removing it is zero.

The cost of leaving the semicolon is zero while the cost of removing it is small but non-zero.

Therefore, it should be left IMO.

willin:
it's excessive.

many eslint rules are excessive.  just a good habit.

Trott:
I dismissed my review because I don't feel so strongly as to block this, but I think leaving the semicolon is preferable to removing it.

Trott:
I can't believe we're having a variation on the "semicolon or no semicolon" argument. I need to find a new hobby.  😃(Yes, I know this is not the usual "semicolon or no semicolon" argument. It's just a joke.)

Maledong:
@Trott & @willin：Ah haha……I'm NOT having argmenet but just my IMO to all of you here. I think the code style isn't so in union so maybe we have diff ideas or options on this :D

AyushG3112:
Can we add a semicolon at the end though?

ghost:
@AyushG3112：This just depends because considering everyone has diff ideas on this, but according to real situation IMO：I don't think it needs here. Because we have codes controllable, and we cannot write codes that will miss `;`.

See this：https://github.com/nodejs/nodejs.org/blob/58c055d08ea53f819dea45893d736798a96a3dd3/layouts/index.hbs#L86

pierreneter:
it's just a semicolon. let it there.

pierreneter:
@Maledong, read this post for more information: https://hackernoon.com/an-open-letter-to-javascript-leaders-regarding-no-semicolons-82cec422d67d

ghost:
@pierreneter：Thanks but I cannot open your given link.
BTW：After looking at the codes and your deploy application, it seems that this file's js codes will be compressed and in order to make it clear, we have to add `;` at the very beginning. So I closed this.

PR #1794, opened 2018-09-04
add multi nodejs versions in CI

ZYSzys:
Add more nodes versions in `.travis.yml` to ensure this project can be developed and built in the lower nodejs versions .

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,10 +1,6 @@
 sudo: false
 language: node_js
-os:
-  - linux
-  - osx
 node_js:
   - "10"
   - "9"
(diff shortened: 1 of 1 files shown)


fhemberger (changes requested):
This change is not required, as we're just testing against the current production environment (which runs the build script).

ZYSzys:
screen shot 2018-09-04 at 6 04 32 pm

As travis shows, test failed in nodejs v6. So do we need to change the `engine` condition in `package.json` where shows `"node": ">=4.5.0"`

PR #1802, opened 2018-09-10
zh-CN：Fix by adding index positions to some titles

ZYSzys:
Fix index in zh-cn when users click.

[timers](https://nodejs.org/zh-cn/docs/guides/event-loop-timers-and-nexttick/#timers) and [poll](https://nodejs.org/zh-cn/docs/guides/event-loop-timers-and-nexttick/#poll) should index to the suitable position.


terry-fei (approved):

ghost:
@ZYSzys：

I just feel curious to know that since in English version we don't have that symbol, and we need to be the same as the English version. Or do you think it's better? Or English version has missed this?

What about this suggestion? Any other suggestions?

ZYSzys:
I think It should be the markdown syntax that indexing positions to some titles. 

I've tried changing titles to Chinese but it didn't work indexing. 
However it can be done by adding the symbol `<!--[titles]-->`

Maledong:
@ZYSzys：Well, you can continue to change that if free.
And what's more：If there's anyone else agree with your opinion, I'll merge this.

Maledong:
@ZYSzys：Well, Go ahead until you think you've finished modifications :)
And what's more：If there's anyone else agree with your opinion, I'll merge this.

Maledong:
@ZYSzys：Well……After your modifications I'll merge them according to 2 people's agreement with that.

willin (approved):
LGTM

ghost:
@ZYSzys: Thanks anyway!

PR #1804, opened 2018-09-12
Clarify when nextTickQueue is processed

OmarDarwish:
It was a little unclear for me when the `nextTickQueue` is processed (see [this](https://stackoverflow.com/questions/52286024/setimmediate-function-not-being-called-after-process-nexttick-function/52286418#52286418) StackOverflow question). After some [experimentation](https://repl.it/repls/UnusedIncompleteDatabases) and [more googling](https://medium.com/the-node-js-collection/what-you-should-know-to-really-understand-the-node-js-event-loop-and-its-metrics-c4907b19da4c), I now understand the mechanics of this queue.

I'm hoping to make this a little more clear for the next reader!

### DIFF
--- a/locale/en/docs/guides/event-loop-timers-and-nexttick.md
+++ b/locale/en/docs/guides/event-loop-timers-and-nexttick.md
@@ -293,8 +293,11 @@ within an I/O cycle, independently of how many timers are present.
 You may have noticed that `process.nextTick()` was not displayed in the
 diagram, even though it's a part of the asynchronous API. This is because
 `process.nextTick()` is not technically part of the event loop. Instead,
-the `nextTickQueue` will be processed after the current operation
-completes, regardless of the current phase of the event loop.
+the `nextTickQueue` will be processed after the current operation is
+completed, regardless of the current phase of the event loop. Here, 
+an *operation* is defined as a transition from the
+underlying C/C++ handler, and handling the JavaScript that needs to be
+executed.
 
 Looking back at our diagram, any time you call `process.nextTick()` in a
 given phase, all callbacks passed to `process.nextTick()` will be
@@ -394,6 +397,56 @@ When only a port is passed, the port is bound immediately. So, the
 To get around this, the `'listening'` event is queued in a `nextTick()`
 to allow the script to run to completion. This allows the user to set
 any event handlers they want.
+
+### Deduplication
+
+For the `timers` and `check` phases, there is a single transition 
+between C to JavaScript for multiple immediates and timers. This deduplication
+is a form of optimization, which may produce some unexpected side effects.
+Take this code snippet as an example:
+
+```js
+// dedup.js
+const foo = [1, 2];
+const bar = ['a', 'b'];
+
+foo.forEach(num => {
+  setImmediate(() => {
+    console.log('setImmediate', num);
+    bar.forEach(char => {
+      process.nextTick(() => {
+        console.log('process.nextTick', char);
+      });
+    });
+  });
+});
+```
+```bash
+$ node dedup.js
+setImmediate 1
+setImmediate 2
+process.nextTick a
+process.nextTick b
+process.nextTick a
+process.nextTick b
+```
+
+The main thread adds two `setImmediate()` events, which when processed
+will add two `process.nextTick()` events. When the event loop reaches
+the `check` phase, it sees that there are currently two events created by
+`setImmediate()`. The first event is grabbed and processed, which prints
+and adds two events to the `nextTickQueue`.
+
+Because of deduplication, the event loop does not transition back to the
+C/C++ layer to check if there are items in the `nextTickQueue` immediately. It
+instead continues to process any remaining `setImmediate()` events, of which
+one currently remains. After processing this event, two more events are 
+added to the `nextTickQueue` for a total of four events.
+
+At this point, all previously added `setImmediate()` events have been processed.
+The `nextTickQueue` is now checked, and events are processed in FIFO order. When
+this `nextTickQueue` is emptied, the event loop considers all operations to have 
+been completed for the current phase and transitions to the next phase.
 
 ## `process.nextTick()` vs `setImmediate()`
 


ghost:
/cc：@nodejs/documentation. Please have a careful check about this detailled describtions :)

a0viedo:
I think this is a better description but I'm no event loop expert. pinging people who can shed some light @addaleax @mcollina @Fishrock123

mcollina (approved):
LGTM

BridgeAR (approved):

sam-github:
I think you may have found a problem, the current text may not be correct for all types of operations, but I am not sure the replacement is correct either. Perhaps not all operations are the same?

The original text used "operations" because not all phases contain events. Operations is a more generic term for what phases do:

> While each phase is special in its own way, generally, when the event loop enters a given phase, it will perform any operations specific to that phase,

"events" in the sense of the event loop are read from the kernel only in the Poll phase, so the replacement shouldn't replace operations with events, they aren't the same.

I noticed what you noticed with your example code, too, when I wrote some exploratory code, but I think its particular to "immediates".  In that case, it appears the entire immediate queue is executed sequentially, and then any ticks that were queued by the immediates are executed sequentially.

In contrast, when an I/O event causes a callback to be executed, and that callback calls `nextTick()`, my understanding is that the tick queue is guaranteed to be executed *before* any other I/O callback is executed. Normally on a lightly loaded system this happens because there is never more than one event processed per Poll phase, but even on a loaded system where libuv reads more than one event from the kernel with epoll() in a single shot, I believe this guarantee holds.

mcollina:
An I/O event triggers some JavaScript to be executed. At the end of that execution, the functions in the `nextTick`  queue are executed.

sam-github:
@mcollina That is my understanding as well, so the change proposed in this PR is incorrect. It says that the nextTick queue won't be executed until the end of the entire phase.

@OmarDarwish I think that executing the entire queue of Immediates is a single "operation", from the point of view of clearing the `nextTick` queue, which isn't documented, and is what you noticed with your test code.

OmarDarwish:
>I think its particular to "immediates". In that case, it appears the entire immediate queue is executed sequentially, and then any ticks that were queued by the immediates are executed sequentially.

>executing the entire queue of Immediates is a single "operation", from the point of view of clearing the nextTick queue

I'm up for any change in the documentation that makes this particular (edge?) case well known. @sam-github what do you think is a better way to make this easier to understand for the reader?

mcollina:
> @mcollina That is my understanding as well, so the change proposed in this PR is incorrect. It says that the nextTick queue won't be executed until the end of the entire phase.

I think we should clarify the phases, and reflect the fact that _there is a single transition_ between C to JS for multiple immediates and timers. In fact, we are deduping them as a form of optimization and this is a side effect of that.

OmarDarwish:
Took a stab at defining an operation, and giving an example of deduplication.

locale/en/docs/guides/event-loop-timers-and-nexttick.md
```diff
@@ -190,8 +190,6 @@ _whose time thresholds have been reached_. If one or more timers are
 ready, the event loop will wrap back to the **timers** phase to execute
 those timers' callbacks.
 
-### check
```
sam-github: Did you remove this by accident?

OmarDarwish: Yep, lemme fix real quick 😅

locale/en/docs/guides/event-loop-timers-and-nexttick.md
```diff
@@ -20,7 +20,7 @@ this in further detail later in this topic.
 ## Event Loop Explained
 
 When Node.js starts, it initializes the event loop, processes the
-provided input script (or drops into the [REPL][], which is not covered in
+provided input script (or drops into the [REPL](https://nodejs.org/api/repl.html), which is not covered in
```
vsemozhetbyt: 1. This line exceeds 80 characters limit now.
2. It seems you can just fix the [bottom reference](https://github.com/nodejs/nodejs.org/blame/d990c746d5b9e288cf41b8d8ceaa597f49c9dae6/locale/en/docs/guides/event-loop-timers-and-nexttick.md#L492) if you prefer this URL.

locale/en/docs/guides/event-loop-timers-and-nexttick.md
```diff
@@ -303,6 +306,56 @@ situations because **it allows you to "starve" your I/O by making
 recursive `process.nextTick()` calls**, which prevents the event loop
 from reaching the **poll** phase.
 
+#### Deduplication
+
+For the `timers` and `check` phases, the there is a *single transition
```
vsemozhetbyt: the there -> there

locale/en/docs/guides/event-loop-timers-and-nexttick.md
```diff
@@ -303,6 +306,56 @@ situations because **it allows you to "starve" your I/O by making
 recursive `process.nextTick()` calls**, which prevents the event loop
 from reaching the **poll** phase.
 
+#### Deduplication
+
+For the `timers` and `check` phases, the there is a *single transition 
+between C to JS for multiple immediates and timers*. This deduplication
+is a form of optimzation, which may produce some unexpected side effects.
```
vsemozhetbyt: optimzation -> optimization

locale/en/docs/guides/event-loop-timers-and-nexttick.md
```diff
@@ -303,6 +306,56 @@ situations because **it allows you to "starve" your I/O by making
...
+Take this code snippet as an example:
+
+```js
+// dedup.js
+const foo = [1, 2];
+const bar = ['a', 'b'];
+
+foo.forEach( num => {
```
vsemozhetbyt: Unneeded space before `num`.

locale/en/docs/guides/event-loop-timers-and-nexttick.md
```diff
@@ -303,6 +306,56 @@ situations because **it allows you to "starve" your I/O by making
...
+
+Because of deduplication, the event loop does not transition back to the
+C/C++ layer to check if there are items in the `nextTickQueue` immediately. It
+instead continues to process any remaining `setImmediate()` events, of which
+one currently remains. After processing this event, two more events are 
+added to the `nextTickQueue` for a total of four events.
+
+At this point, all previously added `setImmediate()` events have been processed.
+The `nextTickQueue` is now checked, and events are processed in FIFO order. When
+these `nextTickQueue` is emptied, the event loop is considers all operations to have
```
vsemozhetbyt: these -> this
is considers -> considers
to have completed -> to have been completed?

vsemozhetbyt (commented):
Some nits.

OmarDarwish:
@vsemozhetbyt Thanks for double checking. I forgot to spell check before pushing. Fixing now!

vsemozhetbyt (approved):

OmarDarwish:
@sam-github @mcollina What do you think? If I understand the contributing policy, if/when you two approve we should be good to go to merge?

mcollina:
I think responsibility for this repo is to @nodejs/website.

dmcghan:
I just realized this doc doesn't discuss promises, which I believe have their own queue and processing timings which make them unique.

```javascript
async function doWork(char) {
  console.log('promise', char);
}

// dedup.js
const foo = [1, 2];
const bar = ['a', 'b'];

foo.forEach(num => {
  setImmediate(() => {
    console.log('setImmediate', num);
    bar.forEach(async (char) => {
      process.nextTick(() => {
        console.log('process.nextTick', char);
      });
      await doWork(char);
    });
  });
});
```

```bash
setImmediate 1
promise a
promise b
setImmediate 2
promise a
promise b
process.nextTick a
process.nextTick b
process.nextTick a
process.nextTick b
```

Maybe this PR isn't the right place to address this...

davisjam:
FWIW Here is [an academic paper on this](https://static.googleusercontent.com/media/research.google.com/en//pubs/archive/46255.pdf) by @matthewloring and others.

amiller-gh:
@mcollina, thanks for the ping. 

This does fit in well with our redesign / getting started guide work, but I think its okay for the content to live in here for now. We're hoping to keep all getting started guides alongside the source code after we've figured out the docs site content ingestion pipeline, so it'll end up back here anyway. The new content here will be helpful in the meantime. 

We'll definitely use this doc as a starting point for for the event loop page on the new site (likely take it wholesale) awesome job @OmarDarwish!

apapirovski:
@dmcghan Your example isn't quite accurate. Promises would normally only resolve after the nextTicks. In your case, you have `console.log` executing in a sync manner. Meanwhile if you were to put the `console.log` after the await, the order would be more like:

```
setImmediate 1
setImmediate 2
process.nextTick a
process.nextTick b
process.nextTick a
process.nextTick b
promise a
promise b
promise a
promise b
```

Once #22842 lands though, it will be more like the browser:

```
setImmediate 1
process.nextTick a
process.nextTick b
promise a
promise b
setImmediate 2
process.nextTick a
process.nextTick b
promise a
promise b

**Review by ryzokuken — APPROVED** [review]

File: locale/en/docs/guides/event-loop-timers-and-nexttick.md [hunk 7]
```diff
@@ -293,8 +293,11 @@ within an I/O cycle, independently of how many timers are present.
 You may have noticed that `process.nextTick()` was not displayed in the
 diagram, even though it's a part of the asynchronous API. This is because
 `process.nextTick()` is not technically part of the event loop. Instead,
-the `nextTickQueue` will be processed after the current operation
-completes, regardless of the current phase of the event loop.
+the `nextTickQueue` will be processed after the current operation is
+completed, regardless of the current phase of the event loop. Here, 
+an **operation** is defined as a transition from the
```
Trott: Nit: Italics here, not bold.

locale/en/docs/guides/event-loop-timers-and-nexttick.md
```diff
@@ -303,6 +306,56 @@ situations because **it allows you to "starve" your I/O by making
 recursive `process.nextTick()` calls**, which prevents the event loop
 from reaching the **poll** phase.
 
+#### Deduplication
+
+For the `timers` and `check` phases, there is a *single transition 
+between C to JS for multiple immediates and timers*. This deduplication
```
Trott: Nit: No italics here. We overuse italics and bold for emphasis. Just don't do it. 😄

Trott:
> Once #22842 lands though, it will be more like the browser:

So in theory, the correct content of these guides is dependent on the version of Node.js...

Trott:
(If this is an improvement over existing text, I'm totally OK with it landing with or without my nits addressed.)

dmcghan:
@apapirovski Interesting. Looking back, I'm not sure I meant to add the `await` operator. I was trying to add something to the promise queue but I wasn't concerned with it resolving before continuing execution of the current stack.

Maybe this is a bit simpler:
```javascript
const foo = [1, 2];
const bar = ['a', 'b'];

foo.forEach(num => {
  setImmediate(() => {
    console.log('setImmediate', num);
    bar.forEach(async (char) => {
      process.nextTick(() => {
        console.log('process.nextTick', char);
      });
      new Promise((resolve, reject) => {
        console.log('promise', char);
      });
    });
  });
});
```

You can reverse the order of the nextTick and Promise calls and the result is the same (expected):

```bash
setImmediate 1
promise a
promise b
setImmediate 2
promise a
promise b
process.nextTick a
process.nextTick b
process.nextTick a
process.nextTick b
```

This output surprises me. I'd have expected the output you showed:

```
setImmediate 1
setImmediate 2
process.nextTick a
process.nextTick b
process.nextTick a
process.nextTick b
promise a
promise b
promise a
promise b
```

I'm missing something. Does no

While I'd like to get a better understanding of this behavior, I was really just trying to point out that this level of detail isn't in the doc.

Do you have a link to 22842?

locale/en/docs/guides/event-loop-timers-and-nexttick.md
```diff
@@ -303,6 +306,56 @@ situations because **it allows you to "starve" your I/O by making
 recursive `process.nextTick()` calls**, which prevents the event loop
 from reaching the **poll** phase.
 
+#### Deduplication
```
thefourtheye: I am not sure if this is the correct place. The next section answers why the behaviour mentioned in the line above is allowed.

OmarDarwish: Hmm, you're right. Let me fix this.

locale/en/docs/guides/event-loop-timers-and-nexttick.md
```diff
@@ -293,8 +293,11 @@ within an I/O cycle, independently of how many timers are present.
 You may have noticed that `process.nextTick()` was not displayed in the
 diagram, even though it's a part of the asynchronous API. This is because
 `process.nextTick()` is not technically part of the event loop. Instead,
-the `nextTickQueue` will be processed after the current operation
-completes, regardless of the current phase of the event loop.
+the `nextTickQueue` will be processed after the current operation is
+completed, regardless of the current phase of the event loop. Here, 
+an *operation* is defined as a transition from the
+underlying C/C++ handler, and handling the Javascript that needs to be
```
watson:
```suggestion
underlying C/C++ handler, and handling the JavaScript that needs to be
```

locale/en/docs/guides/event-loop-timers-and-nexttick.md
```diff
@@ -395,6 +398,56 @@ To get around this, the `'listening'` event is queued in a `nextTick()`
 to allow the script to run to completion. This allows the user to set
 any event handlers they want.
 
+### Deduplication
+
+For the `timers` and `check` phases, there is a single transition 
+between C to JS for multiple immediates and timers. This deduplication
```
watson:
```suggestion
between C to JavaScript for multiple immediates and timers. This deduplication
```

ghost:
@OmarDarwish：There're too many changes here, so can you merge them by using "rebase..." in your local PC and make a push submit?

Not sure whether this post can be merged? I think it's been a long time....

OmarDarwish:
@Maledong I've rebased and squashed

ghost:
@OmarDarwish：Thanks!

@watson, @sam-github , @thefourtheye , @Trott ：Any other ideas or suggestions? Can this be merged?

sam-github (approved):

locale/en/docs/guides/event-loop-timers-and-nexttick.md
```diff
@@ -395,6 +398,56 @@ To get around this, the `'listening'` event is queued in a `nextTick()`
...
+```
+```bash
+$ node dedup.js
+setImmediate 1
+setImmediate 2
+process.nextTick a
+process.nextTick b
+process.nextTick a
+process.nextTick b
+```
```
ZYSzys, line 432: AFAIK, did the output incorrect ?
The output is always as below while my Node.js version is `v11.4.0`.
```
setImmediate 1
process.nextTick a
process.nextTick b
setImmediate 2
process.nextTick a
process.nextTick b
```

So it's indeed as @Trott said:
> So in theory, the correct content of these guides is dependent on the version of Node.js...

Maybe we also need to explain this there.

OmarDarwish: You're right. Looks like something changed for node 11:

``` bash
bash-3.2$ pbpaste
 const foo = [1, 2];
 const bar = ['a', 'b'];

 foo.forEach(num => {
   setImmediate(() => {
     console.log('setImmediate', num);
     bar.forEach(char => {
       process.nextTick(() => {
         console.log('process.nextTick', char);
       });
     });
   });
 });
bash-3.2$ nvm use 10.15.0
Now using node v10.15.0 (npm v6.4.1)
bash-3.2$ pbpaste | node
setImmediate 1
setImmediate 2
process.nextTick a
process.nextTick b
process.nextTick a
process.nextTick b
bash-3.2$ nvm use 11.0.0
Now using node v11.0.0 (npm v6.4.1)
bash-3.2$ pbpaste | node
setImmediate 1
process.nextTick a
process.nextTick b
setImmediate 2
process.nextTick a
process.nextTick b
bash-3.2$ nvm use 11.6.0
Now using node v11.6.0 (npm v6.5.0-next.0)
bash-3.2$ pbpaste | node
setImmediate 1
process.nextTick a
process.nextTick b
setImmediate 2
process.nextTick a
process.nextTick b
```

OmarDarwish: It looks like [guides](https://nodejs.org/en/docs/guides/) is not segregated by version. What would be the best way to indicate that this behavior is only applicable for Node 10 and below? 

Also, what changed in node 11?

OmarDarwish: This is what changed https://github.com/nodejs/node/pull/22842

Maledong: @OmarDarwish：Maybe this should be notified something like: 'before Nodejs v11.X, ……' and 'in Nodejs v11.X or later,……'

OmarDarwish: I'm sorry, I don't quite get the ask. What would you like me to change?

Maledong: @OmarDarwish ：Sorry I removed your comment by mistake……I removed mine first, well……what I want to ask is since diff version of Nodejs will have diff behaviours. Would you mind clearing them one by one such as: 

>In Nodejs 10, some……, however in nodejs 11, ……

ghost:
Can this be merged? It's been for ages :)

a0viedo:
I think all changes have been addressed I'm +1 on merging this.

PR #1796, opened 2018-09-04
update profiling crypto example

trevorah:
The digest parameter is required as of v6.0.0, otherwise the example code errors out like this:
```
TypeError: The "digest" argument is required and must not be undefined
    at pbkdf2 (crypto.js:694:11)
    at Object.exports.pbkdf2Sync (crypto.js:687:10)
```

Details: https://nodejs.org/api/crypto.html#crypto_crypto_pbkdf2sync_password_salt_iterations_keylen_digest

PR #1805, opened 2018-09-13
doc: Fix 'pbkdf2Sync' encoding problem

ghost:
Ref: https://github.com/nodejs/nodejs.org/pull/1796.

Since the '/newUser' request is encoded by 'sha512', and your
corresponding '/auth' must also use 'sha512' as the digest encoding
as well. Otherwise your decoding isn't right.

### DIFF
--- a/locale/en/docs/guides/simple-profiling.md
+++ b/locale/en/docs/guides/simple-profiling.md
@@ -66,7 +66,7 @@ app.get('/auth', (req, res) => {
   const { salt, hash } = users[username];
   const encryptHash = crypto.pbkdf2Sync(password, salt, 10000, 512, 'sha512');
 
-  if (hash.toString() === encryptHash.toString()) {
+  if (crypto.timingSafeEqual(hash, encryptHash)) {
     res.sendStatus(200);
   } else {
     res.sendStatus(401);
--- a/locale/en/docs/guides/simple-profiling.md
+++ b/locale/en/docs/guides/simple-profiling.md
@@ -63,9 +63,10 @@ app.get('/auth', (req, res) => {
     return res.sendStatus(400);
   }
 
-  const hash = crypto.pbkdf2Sync(password, users[username].salt, 10000, 512);
+  const { salt, hash } = users[username];
+  const encryptHash = crypto.pbkdf2Sync(password, salt, 10000, 512, 'sha512');
 
-  if (users[username].hash.toString() === hash.toString()) {
+  if (hash.toString() === encryptHash.toString()) {
     res.sendStatus(200);
   } else {
     res.sendStatus(401);
--- a/locale/ko/docs/guides/simple-profiling.md
+++ b/locale/ko/docs/guides/simple-profiling.md
@@ -114,9 +114,10 @@ app.get('/auth', (req, res) => {
     return res.sendStatus(400);
   }
 
-  const hash = crypto.pbkdf2Sync(password, users[username].salt, 10000, 512);
+  const { salt, hash } = users[username];
+  const encryptHash = crypto.pbkdf2Sync(password, salt, 10000, 512, 'sha512');
 
-  if (users[username].hash.toString() === hash.toString()) {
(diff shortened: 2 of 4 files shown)


lpinca (approved):

vsemozhetbyt:
cc @nodejs/crypto

trevorah:
😳 whoops, sorry I didnt notice that in my original PR. Good spot @Maledong!

locale/en/docs/guides/simple-profiling.md
```diff
@@ -63,7 +63,8 @@ app.get('/auth', (req, res) => {
     return res.sendStatus(400);
   }
 
-  const hash = crypto.pbkdf2Sync(password, users[username].salt, 10000, 512);
+  const salt = users[username].salt;
```
tniessen: Nit: Use destructuring.

tniessen (commented):
Unrelated: 512 is an unusual key size (note that the key size parameter is in bytes, not bits). This might have originated from the misconception that it must match the output size of the hash function?

trevorah:
@tniessen I picked sha512 for this guide because its used in the crypto examples (https://nodejs.org/api/crypto.html#crypto_crypto_pbkdf2_password_salt_iterations_keylen_digest_callback) which this guide appears to mirror.

tniessen:
@trevorah SHA512 is a good choice, just the output key size is unusual (512 bytes!). This works equally well:

```js
crypto.pbkdf2Sync('password', 'salt', 10000, 32, 'sha512')
```

PBKDF2 is usually used to derive symmetric keys or to store passwords, neither of these applications requires keys with a size of 512 bytes.

trevorah:
@tniessen hmm... weird. The only `crypto.pbkdf2*` examples that use a 512 key size are the ones that set `crypto.DEFAULT_ENCODING = 'hex'`. All the others use a 64 byte key size. Maybe the original author wanted to show some difference when setting `DEFAULT_ENCODING`?

tniessen:
Uhh `DEFAULT_ENCODING`... What a design mistake. It still doesn't really matter, if you set the `DEFAULT_ENCODING` to `hex`, the output will still be a 512-byte (1024 hex digits) sequence, just encoded. But again, it is unrelated, so doesn't need to be changed in this PR. Also, it's not wrong, just unusual.

locale/en/docs/guides/simple-profiling.md
ghost: Sorry for mistake, and I'll fix that sooooooooon :)

ghost:
@trevorah：No problem. Just found that by accident when reading lines :)

locale/en/docs/guides/simple-profiling.md
```diff
@@ -63,9 +63,10 @@ app.get('/auth', (req, res) => {
     return res.sendStatus(400);
   }
 
-  const hash = crypto.pbkdf2Sync(password, users[username].salt, 10000, 512);
+  const { salt, hash } = users[username];
+  const encryptHash = crypto.pbkdf2Sync(password, salt, 10000, 512, 'sha512');
 
-  if (users[username].hash.toString() === hash.toString()) {
+  if (hash.toString() === encryptHash.toString()) {
```
tniessen: Also unrelated: if `hash` and `encryptHash` are buffers, the best way to check for equality is `hash.equals(encryptHash)`.

LinusU: This should probably be a constant time comparison though? As not to leak the expected hash? (`crypto.timingSafeEqual`)

tniessen: Oh, right, good point, didn't look at the code as a whole to be honest.

ghost: @LinusU: Take your advice :)

tniessen (approved):

vsemozhetbyt (approved):
Doc format LGTM)

ghost:
Thanks all!

PR #1820, opened 2018-09-24
Fix HTML rendering on Docs index page

pboivin:
Hi!

This pull request is for a small fix to the 'Docs' index page:
https://nodejs.org/en/docs/

In section 'Looking for API docs of previous releases?', the block of links is currently rendered as plain text instead of HTML.

This fix changes the level of indentation of the HTML snippet within the Markdown document, from 4 spaces to 2 spaces.

#### Before:
before

#### After:
after

### DIFF
--- a/locale/ca/docs/index.md
+++ b/locale/ca/docs/index.md
@@ -22,16 +22,16 @@ relacionats al mateix. També indica quins mètodes són a l'abast per les difer
 També descriu els mòduls inclosos que proporciona Node.js, mes no documenta els mòduls que proporciona la comunitat.
 
 <div class="highlight-box">
-    <h4>Buscant la referència de l'API per a una versió anterior?</h4>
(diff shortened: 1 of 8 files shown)


fhemberger (approved):
Thanks! Seems to be related to the `metalsmith-markdown` update.
(Wow, why would indentation change the rendering of HTML at all?!)

fhemberger:
Published. Thank you!

pboivin:
Beats me. I have grepped around the `docs/` folder to compare with other HTML snippets and couldn't find any other with an indentation of 4 spaces...

Thanks for the quick reply :)

vsemozhetbyt:
> (Wow, why would indentation change the rendering of HTML at all?!)

Seems [per spec](https://daringfireball.net/projects/markdown/syntax#precode):

> To produce a code block in Markdown, simply indent every line of the block by at least 4 spaces or 1 tab.

PR #1826, opened 2018-09-29
zh-TW：Add missing translation for 'Collab Summit'

ghost:
Add this missing Traditional Chinese for 'Collab Summit'.

### DIFF
--- a/locale/zh-tw/get-involved/collab-summit.md
+++ b/locale/zh-tw/get-involved/collab-summit.md
@@ -11,10 +11,10 @@ layout: contribute.hbs
 
 ## 誰能參與？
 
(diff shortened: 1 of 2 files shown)


osk2 (commented):
I'm thinking that if it's necessary to translate "Collab Summit". if so, then we'll need to translate words in `collab-summit.md` too.

ghost:
`collab-summit.md` has been translated into zh-TW as：https://nodejs.org/zh-tw/get-involved/collab-summit/.

You see that we've translated once about "Collab Submmit" as "協作者峰會". And for the rest, there's no need to translate because they are fixed special pharses.

osk2:
I mean the word "Collab Summit" :P

I left them untranslated in `collab-summit.md`. If we decide to translate them to "協作者峰會", we should also replace "Collab Summit" with "協作者峰會" in `collab-summit.md`.

osk2 (approved):
Nice, looks much better now 👍

PR #1835, opened 2018-10-12
Replaced curl with wget for Debian and Ubuntu

Ginden:
By default, Ubuntu (checked on my Ubuntu 18 LTS) and Debian (checked on Jessie in Vagrant) come with `wget` installed, but no `curl`. Therefore, it's preferable to use `wget` in docs to make it more accessible to users.


ghost (approved):
Thanks!

PR #1854, opened 2018-10-25
update metalsmith-permalinks

Trott:
This resolves the only issue reported currently by `npm audit`.

Major version change for the dependency, but it doesn't appear to change the results for our permalinks.


ZYSzys (approved):

fhemberger (approved):
As we don't set any special slug options for the permalinks, we're not affected by the semver major bump.

PR #1862, opened 2018-10-26
Wrong variable used in documentation of hrtime()

ghost:
`hrtime()` code snippet had parameter specified as `n` which should be `before` to compute the runtime of the given method.

Issue in focus: nodejs/node#23899


ckotzbauer (approved):

aks- (approved):

fhemberger:
Thank you!

PR #1865, opened 2018-10-26
doc: Add the missing versions of Nodejs APIs' links

ghost:
1) Missing the v8, v9 links of doc.
2) Fix the site.json in `fa`, add `rtl` and `locale` attributes.

### DIFF
--- a/locale/ca/docs/index.md
+++ b/locale/ca/docs/index.md
@@ -25,7 +25,10 @@ També descriu els mòduls inclosos que proporciona Node.js, mes no documenta el
   <h4>Buscant la referència de l'API per a una versió anterior?</h4>
 
    <ul>
+    <li><a href="https://nodejs.org/docs/latest-v9.x/api/">Node.js 9.x</a></li>
+    <li><a href="https://nodejs.org/docs/latest-v8.x/api/">Node.js 8.x</a></li>
     <li><a href="https://nodejs.org/docs/latest-v7.x/api/">Node.js 7.x</a></li>
+    <li><a href="https://nodejs.org/docs/latest-v6.x/api/">Node.js 6.x</a></li>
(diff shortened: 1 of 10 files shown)


ghost:
Notice: This can be postponed until Nodejs 8.x is retired from LTS in maintance state, because I've
added Nodejs 8.x as a previous version.

lpinca (approved):

ZYSzys (commented):
How about adding v10 too ?

ghost:
@ZYSzys：

Good question!

Nodejs v10.X will be the active LTS since 2018-10-30 (See：https://github.com/nodejs/Release#release-schedule). So it shoudn't be a previous version in my mind.

But what confuses me is：Since 2018-10-30 v10 will take place of v8.X, will v8.X be the previous version? I have to put it there but as you know, v8 at that time is still in Active LTS……So how to cope with two versions of Nodejs (both v8 and v10 will be ACTIVE LTS after Oct 2018)?

together

lpinca:
Node.js 10 will be shown on the left side so it doesn't make sense to add it.

There will be 11 and 10 here:

screen shot 2018-10-28 at 11 19 18

Maledong:
@Ipinca：Thanks! do you mean both v8.X and v10.X will be shown together at `doc`?
image
If yes, there's no need to add 8.x or 10.x, but just 9.x :)

Maledong:
@lpinca : OK, I'll do changes and submit after v10.x becomes to the active LTS. Thanks again!

lpinca:
@Maledong keep it as is. It's how it should be as soon as 10 will be tagged as LTS.

ghost:
@lpinca：OK, So now I'll keep v8 in the previous list :)

ghost:
Since the latest release of Nodejs 10.x has come. So I'll push this PR :)

new

Issue #1889, opened 2018-11-06
doc: Can't find information about LTS releases.


kpm:
Hi,

I haven't been able to find info about LTS releases.  If I go to https://nodejs.org/en/about/releases/
, the page doesn't mention LTS at all.  If I go to the main docs at https://nodejs.org/en/docs/ , there's no search facility, and a quick explore through some links doesn't lead to any info I can find about LTS releases.

Sorry if I missed something obvious, I'm new to node.  I'd like to know when the next LTS release is planned and what version number it would be, or at least some page that has info that describes the LTS for nodejs.

Thanks.

fhemberger:
@Trott We could link to `schedule.png` and produce a table right from `schedule.json`. See also https://github.com/nodejs/Release/pull/380

PR #1890, opened 2018-11-07
Add Release Schedule

fhemberger:
Pulls data from https://github.com/nodejs/Release

Fixes: #1889


ZYSzys (approved):
LGTM.
But IMHO, I wonder if it's better to put this section into `about/releases.md` since this is release related ?

layouts/about-release-schedule.hbs
```diff
@@ -0,0 +1,35 @@
...
+  <div id="main">
+    <div class="container has-side-nav">
+
+      {{> navigation key='about'}}
+
+      <article>
+        <h1>{{ title }}</h1>
+
+        <p>
+          <img alt="{{ title }}" src="https://raw.githubusercontent.com/nodejs/Release/master/schedule.png">
```
richardlau, line 17: FYI the PNG might get replaced by SVG (https://github.com/nodejs/Release/pull/380). 

The image is generated from `schedule.json` (see commit message in https://github.com/nodejs/Release/commit/e82fc16b222555415948abdf728a8ee8f304dced). Would it make sense to generate it in the deploy step?

richardlau:
cc @nodejs/release

layouts/about-release-schedule.hbs
fhemberger: As soon as the PR is merged, I'll switch to the SVG instead.

IMHO it doesn't make sense to be generated on each website update. The release schedule only changes only every couple of months and is updated first in https://github.com/nodejs/Release/ anyway.

fhemberger:
@ZYSzys The "Release" page explains more what kind of changes land in version numbers. I'm currently not sure if we need this explanation anyway or if it can be replaced entirely by the release schedule.

Feedback is appreciated.

ZYSzys:
I didn't mean replacing the "Release" page with the release schedule, I'm saying can we just add a section called `schedule` into the "Release" page simply ?

Trott:
I agree with @fhemberger's suspicion that https://nodejs.org/en/about/releases/ can be deleted entirely or replaced with the schedule here. Most of it is not very informative.

fhemberger:
Ok, replaced the old "Releases" page entirely with the release schedule. The information wasn't really helpful anyway. PTAL.

Trott (approved):

marswong (approved):

locale/uk/about/releases.md
```diff
@@ -1,81 +1,18 @@
...
+  pending: 'Pending'
+columns:
+  - 'Release'
+  - 'Status'
+  - 'Codename'
+  - 'Initial Release'
+  - 'Active LTS Start'
+  - 'Maintenance LTS Start'
+  - 'End-of-life'
+schedule-footer: Dates are subject to change.
```
vsemozhetbyt:
```yaml
---
layout: about-release-schedule.hbs
title: Релізи
statuses:
  maintenance: 'Підтримуваний LTS'
  active: 'Активний LTS'
  current: 'Поточний'
  pending: 'Майбутній'
columns:
  - 'Резіз'
  - 'Статус'
  - 'Кодова назва'
  - 'Початковий реліз'
  - 'Початок активного LTS'
  - 'Початок підтримуваного LTS'
  - 'Кінець підтримки'
schedule-footer: Дати можуть змінитись.
---
```

fhemberger:
Thanks everybody!

PR #1893, opened 2018-11-09
Link version schedule to about/releases page

ZYSzys:


layouts/index.hbs
```diff
@@ -64,7 +64,7 @@
           </div>
         {{/if}}
         <p>
-          {{{ labels.version-schedule-prompt }}} <a href="https://github.com/nodejs/Release#release-schedule">{{ labels.version-schedule-prompt-link-text }}</a>
+          {{{ labels.version-schedule-prompt }}} <a href="http://nodejs.org/{{site.locale}}/about/releases/">{{ labels.version-schedule-prompt-link-text }}</a>
```
fhemberger: Good thinking (forgot we have that link there), but the release page is not yet available in all languages.

ZYSzys: It's indeed a problem, is there any solution for this ? Or can you give me some advice to solve this ?

fhemberger: Well, the easiest solution would be to create the page in every language (copying from `en` for now). But that would imply that 'about' as parent page is available as well (I think some languages are at the bare minimum at the moment).

ghost:
@fhemberger, @ZYSzys ：Hi, what about this state？Can this be merged？

Trott:
> /cc @nodejs/website What do you think of the approach of copying the English 'about' section to all languages, so we can display the release schedule? I'm not 100% sure about it, so a +1/-1 would be nice.

Potentially unpopular opinion: It might be worth considering if much of this material can simply be eliminated. For example, I don't think there's much reason to have anything about project governance on the website. The website links to the repo and the canonical information about governance is there. (The website material is already outdated in some minor ways.) Repeating information in multiple places inevitably means that someone will forget to update things.  When I go to a software project's web page, I expect to see how to download the software, documentation, where to get help, and where to get the source code. There might also be some marketing stuff about who uses the project, some testimonials, that sort of thing. 

Governance, though? Nope.  

Guidelines for logo usage??!! That may be appropriate for the Foundation page instead.

Especially-likely-to-be-an-unpopular-opinion: The Community Committee page is perhaps not terribly written, but a long way from well-written. It seems out of place, and is almost certainly unmaintained and at least somewhat out of date. It's also kind of strange. Why talk about the consensus-seeking governance of the Community Committee on the website? Why does it assume that the reader would have any idea what the significance of a "top-level committee" is? All this text is probably appropriate for the CommComm repo (which is probably where it was copied from), but is really (IMO at least) poorly targeted for the audience I would expect in our website. Get rid of it in favor of a link to CommComm in some appropriate list of resources.

Working Groups page is a duplicate of https://github.com/nodejs/TSC/blob/master/WORKING_GROUPS.md#current-working-groups and should probably link to that source material instead of duplicating it. It's not very compelling material, to be honest. It's kind of a boring list of dry charter summaries.

Resources and Trademark is probably better on the Foundation website.

Privacy Policy should likely be a link in the footer of the website rather than a nav link in the "About" section. 

Honestly, continuing my probably-unpopular-opinion thing here, after going through the whole section, I wouldn't mind seeing "About" removed entirely. The only thing I'd want to find another place for would be the release schedule. That could probably be a link in the Downloads page and/or put somewhere in the Docs.

fhemberger:
@Trott Thanks for your feedback!

Many of those resources date back to when the information didn't exist in the repositories yet. I also think some of the texts can be removed/updated.

Personally, on the other hand, I don't like the idea for the website to become a mere link list for the Working Group repositories. Then we could axe a lot of the other content as well and just link to the GitHub org. I still see the problem of the website getting out of sync with the various Working Groups. This also trickles down to our translation efforts.

/cc'ing @amiller-gh here, because this will also be an issue for @nodejs/website-redesign.

fhemberger:
Closing in favor of #2098.

PR #1896, opened 2018-11-12
fixed link parsing error in Chinese

justjavac:
Markdown made a wrong participle of the Chinese hyperlink.

as shown


stevemao:
Thanks. This looks like a bug in the parser. Should probably fix that instead.

ZYSzys (approved):

fhemberger:
Ah, we already had a ticket for it: #1612

willin (approved):
LGTM

it's a good habit to write markdown in prescriptive grammar

PR #1903, opened 2018-11-18
zh-CN：Fix the error translation of 'Twisted'

ghost:
Due to 'Twisted' is a fixed word in Python, so we don't need to translate it into Chinese.

/cc: @nodejs/nodejs-cn


Lellansin:
LGTM

PR #1913, opened 2018-11-26
Update built-in debugger link to latest

ZYSzys:
Update built-in debugger link to the latest in [guides/debugging-getting-started](https://nodejs.org/en/docs/guides/debugging-getting-started/).

### DIFF
--- a/locale/ko/docs/guides/debugging-getting-started.md
+++ b/locale/ko/docs/guides/debugging-getting-started.md
@@ -327,7 +327,7 @@ which translates the Inspector Protocol used in Chromium to the V8 Debugger
 protocol used in Node.js.
 -->
 
-#### [내장 디버거](https://github.com/nodejs/node/blob/master/lib/_debugger.js)
+#### [내장 디버거](https://nodejs.org/dist/latest-v6.x/docs/api/debugger.html)
 
 Node의 내장 명령형 디버거로 스크립트를 실행하려면 `node debug script_name.js`로 실행하세요.
 스크립트가 다른 Node 프로세스에서 `--debug-brk` 옵션으로 시작되고 원래의 Node 프로세스는
--- a/locale/ko/docs/guides/debugging-getting-started.md
+++ b/locale/ko/docs/guides/debugging-getting-started.md
@@ -313,7 +313,7 @@ TCP 포트(기본 `5858`)로 지금은 중단된 V8 디버깅 프로토콜에서
 V8 디버깅 프로토콜은 더는 관리되지 않고 문서화도 되지 않습니다.
 
 <!--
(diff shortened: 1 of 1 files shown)


fhemberger (approved):

locale/en/docs/guides/debugging-getting-started.md
```diff
@@ -241,7 +241,7 @@ couple popular ones are listed below.
 
 The V8 Debugging Protocol is no longer maintained or documented.
 
-#### [Built-in Debugger](https://nodejs.org/dist/latest-v6.x/docs/api/debugger.html)
+#### [Built-in Debugger](https://nodejs.org/docs/latest/api/debugger.html)
 
 Start `node debug script_name.js` to start your script under Node's builtin
```
refack: So if you're here, there are two other changes the should be done:
1. Remove the next section (`#### node-inspector`) [REF](https://github.com/node-inspector/node-inspector#overview)
2. replace `debug` with `inspect`
```suggestion
Run `node inspect script_name.js` to start your script under node's builtin
```

And similar in L248

P.S. This is relevant for all versions since node >= 6.3 (including all recent LTS releases)

locale/en/docs/guides/debugging-getting-started.md
```diff
@@ -241,7 +241,7 @@ couple popular ones are listed below.
 
 The V8 Debugging Protocol is no longer maintained or documented.
 
-#### [Built-in Debugger](https://nodejs.org/dist/latest-v6.x/docs/api/debugger.html)
```
richardlau: This section is specifically referring to the Legacy Debugger though, which is different from the Inspector-based one. I don't think this link should be changed (the latest docs point to the Inspector-based debugger client while the `v6.x` docs refer to the legacy one).

cc @nodejs/diagnostics

ZYSzys:
Find that it's indeed the `Legacy Debugger` section, I'm kind of confused now  whether we should change this.

fhemberger (changes requested):
@ZYSzys Please revert the changes, as it's indeed the legacy debugger.

I only noticed, that in the Korean version, links to the "Built-in Debugger" point to

- https://nodejs.org/docs/latest/api/debugger.html (comment)
- https://github.com/nodejs/node/blob/master/lib/_debugger.js

So these links should be replaced with `https://nodejs.org/dist/latest-v6.x/docs/api/debugger.html`, as in the other languages.

If I'm not mistaken, this should be the only remaining change.

ZYSzys:
@fhemberger 👌Reverted and updated the link of "Built-in Debugger" in the Korean version.

fhemberger (approved):

fhemberger:
@ZYSzys Thanks! 👍

PR #1920, opened 2018-11-30
Update blocking-vs-non-blocking.md

llwwbb:
Correct some errors


ZYSzys (approved):
LGTM, but still /cc @Maledong who is more experienced in this.

Issue #1926, opened 2018-12-05
No installation instructions for Linux downloads?


knipknap:
Are there any installation instructions for Linux for when you download the packages from nodejs.org? The only instructions I found are here:

https://nodejs.org/en/download/package-manager/

But those don't seem to relate to the downloads offered on nodejs.org.

fhemberger:
It's a tarball file. Download and extract to `/usr` or `/usr/local`, depending on your preferences.

Details: https://github.com/nodejs/help/wiki/Installation

knipknap:
That's what I have been looking for, Thanks! It would be helpful if that were linked on the website.

PR #1931, opened 2018-12-08
add link to installation instructions for Linux

ZYSzys:
Fix: #1926


ZYSzys:
@fhemberger BTW, do we need to `/cc` some teams to translate the link text ?

fhemberger:
@ZYSzys Awesome, thank you!

Regarding translation, we could either cc every translation team or active translators will pick up the change on their own. We need to come up with a better way to handle website updates for translators anyway (possibly even before the relaunch).

/cc @nodejs/i18n

ZYSzys:
@fhemberger Hi, because of the nonactive at present, I just make the text keep in English if it's not translated.

fhemberger (changes requested):
Please just add

```
    "install-on-linux": {
      "text": "Installing Node.js via binary archive on Linux"
    }
```

to all `locale/*/site.json` instead.

ZYSzys:
@fhemberger Done :)

fhemberger:
Awesome, thank you!

fhemberger (approved):

Issue #825, opened 2016-07-15
Always translate header & footer templates?


Fishrock123:
I'm not sure this is feasible, but maybe we should build all possible pages that exist from `en` in every language using the localized templates (header, footer, etc), and use those pages as the fall-backs rather than the actual english pages.

This has two obvious downsides, beyonds complexity:
1. Build times would be dramatically increased.
2. Build output size.
## 

Perhaps a bit more involved, maybe we can get nginx to serve some sort of partials where we concat the header to the body directly during serving? Is that even possible? I have no idea.

phillipj:
> This has two obvious downsides, beyonds complexity:
> 1. Build times would be dramatically increased.
> 2. Build output size.
1. Mostly an issue when developing locally. Could be less of an issue if we made it possible to specify which locale to build.
2. With the current set of locales, it would have increased the total build size from ~26MB -> 58MB. Still pretty low figure IMO.

> Perhaps a bit more involved, maybe we can get nginx to serve some sort of partials where we concat the header to the body directly during serving? Is that even possible? ..

Server Side Includes could be an option if we want to go down this route... http://nginx.org/en/docs/http/ngx_http_ssi_module.html

ZYSzys:
> I'm not sure this is feasible, but maybe we should build all possible pages that exist from en in every language using the localized templates (header, footer, etc), and use those pages as the fall-backs rather than the actual english pages.

If I'm not wrong, the PR #1940 has implemented this now.

fhemberger:
Yes, it did. Closing.

PR #1940, opened 2018-12-15
Build default language with locale (adds preserveLocale flag)

flakolefluk:
Currently when navigating to a page that does not have a locale, it will default the english one, this breaks the navigation (it doesn't really break it, but everything including links now changes locale to english)

This PR adds a preserveLocale flag that will run a second build of the files not included in the locale that exist in the "en" locale but using the current locale site.json. 
Build time is considerably longer (*en* locale has almost 700 files, compared to others with 20 or less)
The result:

Open for discussion.


ZYSzys (commented):
It's awesome for me !👍

But is there anything we can do to make a bit more optimization ? That's even better :)

flakolefluk:
As far as i know, there's no way to avoid building all the pages available in english (ignoring the ones already built for the current language) for every locale. I'm thinking of adding specific language builds, but i guess that is not part of this PR. Any advice?

flakolefluk:
Looks like it solves #825

package.json
```diff
@@ -5,10 +5,11 @@
   "homepage": "https://nodejs.org",
   "scripts": {
     "build": "node build.js",
+    "build:deploy":"node build.js --preserveLocale",
     "serve": "node server.js",
```
ZYSzys, line 9: Since we also have `preserveLocale` in `server.js`, is it more reasonable to serve as `node server.js --preserveLocale` ? Or add one more script like `build:deploy` but for serve ?

build.js
```diff
@@ -164,12 +164,124 @@ function buildLocale (source, locale) {
...
+      buildMissing(source, locale, files)
+    }
   })
 }
 
+function buildMissing (source, locale, localeFiles) {
+  console.log(`[metalsmith] build/${locale} missing files started`)
+  console.time(`[metalsmith] build/${locale} missing files finished`)
+
+  const metalsmith = Metalsmith(__dirname)
```
fhemberger: Could you please refactor this part, so we don't have the entire build chain twice?

If you want to distinguish regular/local builds, you can also use something like:

```js
function onlyLocalBuild (child) {
  if (/* whatever flag*/) {
	// return a dummy function if there's nothing to do
    return function (files, metalsmith, done) {
      done()
    }
  }
  return child
}
```

```js
.use(onlyLocalBuild(myFunction({
// ...
})))
```

fhemberger (changes requested):
If we decide to land this change, this would be a good opportunity to go over the process in general.

E.g. instead of building **all** languages for local testing/serving, setting `DEFAULT_LOCALE=en` by default and allow collaborators to overwrite that with an environment variable (e.g. `DEFAULT_LOCALE=zh-cn npm run serve`, when you're only working on the Chinese translation).

Only for deployment via CI all languages must be built.

This would save a lot of generation time for local development.

package.json
flakolefluk: I think while developing is less needed to build all the files for every language. I think it's more relevant when deploying. But a second script in package.json might help.

Current options to serve with preserveLocale are:
```
npm start -- -- --preserveLocale
npm run serve -- --preserveLocale
node server.js --preserveLocale
```

I could add a script to run:
```
npm run serve:preserveLocale
```

build.js
flakolefluk: Thanks for your feedback, i was thinking on language specific builds so i'll check how to mix everything.
So for the build options, as far as i can see:

`DEFAULT_LOCALE=zh-cn npm run serve` builds only the files present in the zh-cn locale folder (will display 404 if file is not present)
`DEFAULT_LOCALE=zh-cn npm run serve -- --preserveLocale` builds the files present in the zh-cn and adds the ones present in "en" locale 
`npm run serve` builds all languages and fallbacks to english when a file is not present in current locale
`npm run serve -- --preserveLocale` builds all languages and adds those missing that are present in "en" locale (no fallback)

the same applies to `npm run build`
does this make sense to you?

fhemberger: Sounds good to me!

flakolefluk: I added the requested changes. Only difference is that DEFAULT_LOCALE accepts comma separated languages.For example: `DEFAULT_LOCALE=es,en,it npm run serve` Let me know what you think.

server.js
```diff
@@ -104,8 +81,8 @@ http.createServer((req, res) => {
   if (req.url === '/') {
     req.url = '/en'
   }
-  mount(req, res, redirectToEnglishUrl(req, res))
+  mount(req, res)
 }).listen(port, () => console.log(`\x1B[32mServer running at http://localhost:${port}/en/\x1B[39m`))
```
fhemberger: Console output should be `http://localhost:${port}/${selectedLocales[0]}/`, I guess.

fhemberger (changes requested):
First of all: Thank you for your work! 👍

Only two little nits remaining:
a) See comment below
b) Could you add a short description of the build options (as in your comment above) to the README file?

FYI, I measured the time difference for a full build with and without `--preserveLocale`:

```
npm run build  12.83s user 1.62s system 118% cpu 12.215 total

rm -r build

npm run build -- --preserveLocale  113.75s user 9.43s system 125% cpu 1:38.04 total
```

As you can see, the latter took eight times as long. This probably isn't relevant during development, but **this will have an impact on the build time during an actual release.**

Therefore I'd love other members of @nodejs/website have a look as well.

server.js
flakolefluk: Thank you for your review. I'm about to leave on holiday with the family starting today. We'll be disconnected until the 16th, so i won't be available during this period . Hope it's ok. Hopefully you'll be able to discuss about build times, or if we should try a different approach.
Thanks again. 👍 👍

flakolefluk:
@fhemberger I added the requested changes to the readme section and when serving a different locale than english

fhemberger (approved):

fhemberger:
@flakolefluk Thanks, LGTM now.
@ZYSzys If you are okay with it, we can merge it.

build.js
```diff
@@ -164,12 +165,37 @@ function buildLocale (source, locale) {
 
   // This actually executes the build and stops the internal timer after
   // completion.
-  metalsmith.build((err) => {
+  metalsmith.build((err, files) => {
```
ZYSzys: LGTM👍, I think it almost to go.
Just a nit: was the `files` parameter added by mistake ? It seems useless here.

flakolefluk: @ZYSzys you were right. I must have added it somewhere along the way. Fixed it.

@fhemberger @ZYSzys Thank you!

ZYSzys (approved):

ZYSzys:
Thank you!👍

PR #2017, opened 2019-01-28
Fix: 'Node.js Bug Bounty Program' should be the 2nd headline

ghost:
In 'security.md', the title 'Node.js Bug Bounty Program' should be the 2nd headline instead of the 3rd one.


ZYSzys (approved):

reedloden (changes requested):
This is on purpose, as the bug bounty program applies only to Node.js core currently.

willin (approved):

ghost:
@reedloden: OK, I closed that. Thanks for telling me

PR #2028, opened 2019-02-03
Fix the links for non-English alphabets when anchors generated

ghost:
Consider that some special characters (not included in the 26 English alphabets) are ignored by md file's anchor when generated. We have to change the anchor link.

Take this as an example：'SmartOS 和 illumos', the '和' doesn't belong to an English character, so it's ignored and the generated anchor would be '#smartos-illumos' instead of '#smartos-and-illumos'.

This is the HOT fix for that. What I did now is to use the pure HTML instead of md's "##" to force the "a href"'s link to keep it the same as that in English version.

PS：For Japanese, it seems they've noticed the anchor problem so their anchor name isn't the same as the English version. What I do is just to change the title to make it the same as what we can see in the category.

【Before Fix】
image

### DIFF
--- a/locale/ja/download/package-manager.md
+++ b/locale/ja/download/package-manager.md
@@ -48,7 +48,7 @@ pacman -S nodejs npm
 ```
 
 
-## <!-- Debian and Ubuntu based Linux distributions, Enterprise Linux/Fedora and Snap packages --> Debian と Ubuntu ベースの Linux ディストリビューション、エンタープライズ Linux/Fedora と Snap packages
+## Debian と Ubuntu ベースの Linux ディストリビューション、エンタープライズ Linux/Fedora と Snap パッケージ
 
 <!-- [Official Node.js binary distributions](https://github.com/nodesource/distributions/blob/master/README.md) are provided by NodeSource. -->
 [Node.js 公式のバイナリディストリビューション](https://github.com/nodesource/distributions/blob/master/README.md)が NodeSource によって提供されています。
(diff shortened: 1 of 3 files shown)


ghost:
@nodejs/nodejs-cn：Please have a review for that.

ghost:
@nodejs/localization：Kindly tip, please notice whether your own specific language has the same problem :)

Maledong:
If this fix is OK, I'll try to fix other languages.

yous:
Did these commits disappear from master? The commits of #2026 are also gone.

Maledong:
Yes, two are missing. And there's another submit, it doesn't show in the history of "Pull Requests" but in the "Comments":

[cid:955c3764-1497-4023-9356-aecc761f2df0]

In the history of Pull Request:
[cid:c1a35dad-941c-4936-86bb-26df57118dcf]

So can you fix this? I don't think it is related to the forcely push, because it used to be good with me. And what's more——the latest submission also missed, and it's NOT a forcely submit 😃
________________________________
发件人: Chayoung You <notifications@github.com>
发送时间: 2019年2月3日 18:02
收件人: nodejs/nodejs.org
抄送: Maledong; State change
主题: Re: [nodejs/nodejs.org] Fix the links for non-English alphabets when anchors generated (#2028)


Did these commits disappeared from master? The commits of #2026<https://github.com/nodejs/nodejs.org/pull/2026> are also gone.

yous:
I don't see the images, they only show `[cid:...]`. Can you re-post them on GitHub?

ghost:
@yous：No problem, I've re-submitted and everything goes fine !

PR #2036, opened 2019-02-07
blog: fix URL of SmartOS 32-bit binary of v6.9.3

yous:
The URL points to `node-v6.9.3-sunos-x32.tar.xz`, but the actual file name contains `x86` instead of `x32`. The SHASUMS are correctly generated against the files named with x86.


willin (approved):

fhemberger:
Thanks!

PR #2041, opened 2019-02-08
remove unused static/legacy

Trott:


ghost (dismissed):

willin (dismissed):

fhemberger:
Those files are still needed for old version of the docs (< 1.0; which are basically a copy of the entire Node.js website).

Trott:
> Those files are still needed for old version of the docs (< 1.0; which are basically a copy of the entire Node.js website).

@fhemberger Can you point to where I might look to see that? On the website itself, https://nodejs.org/docs/latest-v0.10.x/api/buffer.html doesn't point to any `legacy` directories or anything like that. But I'm not sure how the docs are built, TBH. So, happy to be educated on this a bit for sure!  :-D

fhemberger:
@Trott 

Everything up to 0.10.28 uses those files:
https://nodejs.org/docs/v0.10.26/
https://nodejs.org/docs/v0.10.28/ (different design)

Those are the first versions with just the API docs and the changelog only:
https://nodejs.org/docs/v0.10.27/
https://nodejs.org/docs/v0.10.29/

Rewriting is done in the nginx config:
https://github.com/nodejs/build/blob/master/setup/www/resources/config/nodejs.org

Trott:
Thanks for the clear and complete information. Would there be support for a README file in the directory explaining that stuff?

Issue #1973, opened 2019-01-02
/about/working-groups anchors don't work when translated to Arabic


Aissaoui-Ahmed:
Hello everyone.
When we translated the about/working-groups.md to Arabic language, the anchor tags that lead to the different parts of the page stopped working.

en
url : `/en/about/working-groups/#diagnostics`
```html
<h3 id="header-diagnostics">
<a href="https://github.com/nodejs/diagnostics">Diagnostics</a>
<a name="diagnostics" class="anchor" href="#diagnostics" aria-labelledby="header-diagnostics"></a>
</h3>
```
ar
url : `/ar/about/working-groups/#التشخيص`
```html
<h3 id="header-">
<a href="https://github.com/nodejs/diagnostics">التشخيص</a>
<a name="" class="anchor" href="#" aria-labelledby="header-"></a>
</h3>
```
Are there any solutions? 

@nodejs/node-ar
@celyes @InterstellarClub

Aissaoui-Ahmed:
add regexp unicode
for Arabic language
0600-06FF --> "Arabic"

`/scripts/plugins/anchor-markdown-headings.js:6:14`

@celyes @InterstellarClub

ghost:
IMO：I also found this problem and I tended to fix that in the code, but consider there're too many non-English characters in translations (the typical is something like CJK or others...). So in the end I had to use a pair of pure Html to do with the anchor link to keep the English characters as the anchor name as well. Yours is also fixed at #2034.

Ref：https://github.com/nodejs/nodejs.org/pull/2028

Aissaoui-Ahmed:
thank you @Maledong 
IMO: easy way to change just range of Arabic Unicode not all anchors, the range so small.

ghost:
Thanks for your suggestions and in order to keep all the changes in union (Keep the anchor name in English). I take the 2nd choice. :)

celyes:
> 
> 
> IMO：I also found this problem and I tended to fix that in the code, but consider there're too many non-English characters in translations (the typical is something like CJK or others...). So in the end I had to use a pair of pure Html to do with the anchor link to keep the English characters as the anchor name as well. Yours is also fixed at #2034.
> 
> Ref：#2028

Still not working with Arabic.
@Aissaoui-Ahmed  has managed to solve the problem programmatically.
@Maledong , @InterstellarClub

ghost:
[...]

PR #2043, opened 2019-02-08
add anchor link for Arabic "working-groups.md"

Aissaoui-Ahmed:
hello 

add unicode for Arabic language
0600-06FF --> "Arabic"

### DIFF
--- a/locale/ar/download/package-manager.md
+++ b/locale/ar/download/package-manager.md
@@ -26,7 +26,7 @@ title: Installing Node.js via package manager
 
 ----------------------------
 
-## آندرويد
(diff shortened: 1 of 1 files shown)


ghost (commented):
Hi,

IMO：
This is found by me and already fixed at #2034, and we'd better keep English characters because there're too many non-English characters (the typical is 'CJK' character and others). So it's difficult to define the range one by one (See：http://www.unicode.org/Public/5.1.0/ucd/UCD.html).

Ref：https://github.com/nodejs/nodejs.org/issues/1973

PR #2045, opened 2019-02-08
Fix: Fix for the non-English character generation of Anchor Link

ghost:
I. In the recent days we've fixed the non-English character generation
of anchor link by directly using the pure HTML tags, however a better
way is:

1) Check whether the current title has the comment symbol.
2) If yes, just take the characters out of the symbol as the anchor
link's name, with the real text as the title behind it.
3) If not, just do replacements as what it used to be.

>e.g: For non-English characters such as `## <!--abc-->这是非英语的标题`

This will be processed as &lt;h2&gt;这是非英语的标题&lt;a href = '#abc'……&gt;……&lt;/h2&gt;

The advantage is that you can keep the anchor link the same as what
it is in the English version or define by yourself. What's more: It's
much more flexible when your change the level of titles, compared with
https://github.com/nodejs/nodejs.org/pull/2028.

II. Add '--fix' for local test to automatically fix some simple formation typo errors.

---
- [x] Pass unit tests


Trott:
Looks good to me, but I definitely defer to @fhemberger and others on this stuff.

willin (approved):

ghost:
@Trott：Thanks!
@fhemberger：In order to simplify and union the English anchor name, I'd like to use my way. So if free please have a review :)

scripts/plugins/anchor-markdown-headings.js
```diff
@@ -1,14 +1,42 @@
 'use strict'
 
+// Check whether we've got the comment
+const ANCHOR_COMMENTREG = /<!--[\w-]+-->/gi
```
fhemberger: `const ANCHOR_COMMENTREG = /<!--\x20?([\w-]+)\x20?-->/gi`

Matches `<!--comment-->`, `<!-- comment -->` and even `<!--  comment  -->`
(20 hex = 32 dec = space character)

scripts/plugins/anchor-markdown-headings.js
```diff
@@ -1,14 +1,42 @@
...
-    escapedText + '" aria-labelledby="header-' +
-    escapedText + '"></a></h' + level + '>'
+  let anchorTitle = null
+
+  // If we've checked the title has a comment symbol,
+  // we reguard it as the forcely-assigned anchor link name
+  // for titles with non-English characters
+  const anchorTitleArray = ANCHOR_COMMENTREG.exec(raw)
+  if (anchorTitleArray !== null) {
+    anchorTitle = anchorTitleArray[0].replace('<!--', '').replace('-->', '')
```
fhemberger: If we do the matching directly in the RegExp above, we can use `anchorTitle = anchorTitleArray[1]`.

tests/scripts/anchor-mardown-headings.test.js
```diff
@@ -59,4 +59,17 @@ test('anchorMarkdownHeadings', (t) => {
     t.plan(1)
     t.equal(output, expected)
   })
+
+  t.test('correctly pharses markdown heading with non-English characters', (t) => {
+    const text = '这是<a href="b">链接</a>的<a href="d">测试！</a>'
+    const level = 2
+    const raw = '<!--anchor-With-Non-English-Characters-->这是[链接](b)c[测试！](d)'
```
fhemberger: Now `const raw = '<!-- anchor-With-Non-English-Characters -->这是[链接](b)c[测试！](d)'` would work as well.

fhemberger (approved):
Thank you, this looks really great. I'd probably add handling of unwanted whitespaces, though. Just to be sure.

ZYSzys (approved):

ghost:
Thanks @fhemberger!
I've re-submitted one.

Aissaoui-Ahmed:
@Maledong : in Arabic anchors #1973 you can just change Unicode and the same in "CJK"

ghost:
@Aissaoui-Ahmed：Yes, it's been in this submit, thanks for tipping me :)
If you are fine with this change, you can approve it.

fhemberger (approved):

fhemberger:
Great, thank you!

Issue #2055, opened 2019-02-09
`Release.png`'s missing in the `release` page


ghost:
* **URL**: https://nodejs.org/en/about/releases/
* **Browser version**: Edge/Google/IE
* **Operating system**: Windows 10 (x86, Professional, latest).

We cannot see the 'schedule.png', so according to this submit (https://github.com/nodejs/Release/commit/2bf2ea36a162571c0aee21f813f51de790c08feb), it seems the img was removed and we've used 'schedule.svg' instead.  So the img is missing:

![default](https://user-images.githubusercontent.com/40081831/52519208-d44a5780-2c92-11e9-88cf-7ed48f76ce44.PNG)

So I tended to help to change the link in the `about-release-schedule.hbs` like this: https://raw.githubusercontent.com/nodejs/Release/master/schedule.svg

But I cannot see the img either, I'm not sure whether I should copy an img into the `static/images` and refer it? Or just remove this img link totally because there's already a table now?

ghost:
Currently my solution is:

 ```
 <img  alt="{{ title }}"
            src="https://raw.githubusercontent.com/nodejs/Release/master/schedule.svg?sanitize=true">
```
This makes my local pages run successfully:
![default](https://user-images.githubusercontent.com/40081831/52519666-d5cb4e00-2c99-11e9-844d-49e35e20ad61.PNG)

PR #2056, opened 2019-02-09
Fix: Update the img link of 'schedule.svg'

ghost:
We cannot see the 'schedule.png', so according to this submit (nodejs/Release@2bf2ea3), it seems the img was removed and we've used 'schedule.svg' instead.

The reason why I use the parameter `?sanitize=true` is that we MUST make sure that the response type should be of img (`Content-Type should `be `img/svg+xml` instead of `text/plain`).

For more about svg rendering on GitHub's server, please see these related posts:
1. https://github.com/isaacs/github/issues/316#issuecomment-67738972
2. https://stackoverflow.com/questions/13808020/include-an-svg-hosted-on-github-in-markdown

Fix for: https://github.com/nodejs/nodejs.org/issues/2055

### DIFF
--- a/layouts/about-release-schedule.hbs
+++ b/layouts/about-release-schedule.hbs
@@ -14,7 +14,7 @@
         {{{ contents }}}
 
         <p>
-          <img alt="{{ title }}" src="https://raw.githubusercontent.com/nodejs/Release/master/schedule.png">
+          <img alt="{{ title }}" src="https://raw.githubusercontent.com/nodejs/Release/master/schedule.svg?sanitize=true">
         </p>
 
         {{{ release-schedule }}}


richardlau (approved):

layouts/about-release-schedule.hbs
```diff
@@ -14,7 +14,7 @@
         {{{ contents }}}
 
         <p>
-          <img alt="{{ title }}" src="https://raw.githubusercontent.com/nodejs/Release/master/schedule.png">
+          <img alt="{{ title }}" src="https://raw.githubusercontent.com/nodejs/Release/master/schedule.svg?sanitize=true">
```
fhemberger, line 17: What is the `?sanitize=true` doing? 🤔

ZYSzys (approved):

layouts/about-release-schedule.hbs
ZYSzys: Maybe refs: https://stackoverflow.com/a/47952195/8102333

ghost: Hi,
Because if I don't use that parameter, the response type from the server would be 'text/plain', and if I use that, it will return me the right content-type.

Compare them together :)

【Before using parameter】
wrong
【After using parameter】
right

ghost: I'm not sure whether this is the problem related to the server-side, it seems that the server doesn't return me the right response of content-type until I add this parameter.

ghost: Or @fhemberger, any other better ideas? I've already tested and compared XD
Another way is copying the file to the `static/images`, but once the nodejs/Release's img is changed, I have to copy it and do replacement, so I don't think it a good way, though I've mentioned in my #2055 (1st post).

ghost: I've updated my submit commit in details to explain this :)

fhemberger: Awesome, didn't knew that trick. Let's do this!

PR #2062, opened 2019-02-10
Fix typo in about/index.md AR

celyes:
Fixed the term "blocking call" to "non-blocking call" in `about/index.md:41`
@Aissaoui-ahmed @InterstellarClub


locale/ar/about/index.md
```diff
@@ -38,7 +38,7 @@ server.listen(port, hostname, () => {
  تأخذ النود جي اس موديل الاحداث (event model) ابعد قليلا فتمثل الحلقة التكرارية الخاصة بالاحداث ([event loop][]) كمكون اساسي في وقت التشغيل (runtime construct) وليس كمكتبة في انظمة أخرى، حيث ان هنالك دائما استدعاء متزامن (blocking call) 
 للبدء في حلقة الاحداث.
 مبدئيا، يتم تحديد السلوك عبر دالة مستدعاة في بداية السكريبت في نهايتها 
-تقوم بتشغيل خادم (server) عبر استدعاءٍ غير متزامن (blocking call) مثل <span dir="ltr">`EventMachine::run()`</span>، ولكن في النود جي اس لا يوجد شيء من هذا القبيل.
+تقوم بتشغيل خادم (server) عبر استدعاءٍ غير متزامن (non-blocking call) مثل <span dir="ltr">`EventMachine::run()`</span>، ولكن في النود جي اس لا يوجد شيء من هذا القبيل.
```
fhemberger: `blocking` is correct here, see https://nodejs.org/en/about/:

> Typically behavior is defined through callbacks at the beginning of a script and at the end starts a server through a **blocking** call like `EventMachine::run()`. In Node there is no such start-the-event-loop call.

fhemberger (changes requested):

locale/ar/about/index.md
celyes: ok, Corrected the mistake.
@fhemberger

ghost:
@celyes：You don't need to close it if you had a force push and fixed it, this could be also merged if possible :)

celyes:
@Maledong : thank you, but it's basically blank lines, so i don't think it deserves to be merged IMO.

PR #2071, opened 2019-02-13
Update link to the Node.js Trademark Guidelines PDF

brianwarner:
The PDF linked previously had inactive URLs.  For increased clarity, we uploaded a new PDF with active URLs.  This is the same file linked from https://foundation.nodejs.org/about.


fhemberger:
Hmm, I'd rather update the file here.

brianwarner:
That works too.  The linked file is the corrected one, but I'm not sure if I can update that myself...

fhemberger:
@brianwarner Updated the file in #2073, closing this.

Thanks!

Issue #1977, opened 2019-01-04
Review needed for knowledge base articles


fhemberger:
The articles for the knowledge base were donated from Nodejitsu a while back but never received a proper review and thus aren't linked in the website navigation:

https://nodejs.org/en/knowledge/

Those articles need to be reviewed. Are they still relevant? Do they need to be updated for current versions of Node.js? Which articles are relevant for the ongoing website redesign content structure?

/cc @nodejs/website-redesign

keywordnew:
Thanks for surfacing this content!

Yash-Handa:
Hi @chowdhurian and @fhemberger,
I would love to contribute to **Node.js org** by reviewing these **Awesome Articles**.
Please let me know if someone else is working on it or not, So that I can Start with it :)

And also some extra questions:
1. Do all these articles need review or just some of them? [https://nodejs.org/en/knowledge/](https://nodejs.org/en/knowledge/)
2. Do I mention the changes in the articles?
   eg: if a feature, say 'XYZ' has been deprecated then should I do:
> The feature 'XYZ' has been deprecated in node version vX.X.X and use the feature 'ABC' instead

_OR_
Update the Articles as if outdated stuff was never there.

keywordnew:
@Yash-Handa you can totally get started.

To answer your questions:
1. These articles were first contributed 3 years ago. Only a few have seen further updates, which I see were more high-level rather than content updates. ([They're on github](https://github.com/nodejs/nodejs.org/tree/master/locale/en/knowledge)). I think they all need review, but I'd suggest starting by taking on a manageable amount first, say 1 article, and reviewing that.
2. Yes! And when you're unsure or want extra information, you should totally (at)mention a person for context 😄

I hope this was helpful. Let me know if I've left anything unclear.

fhemberger:
Thanks @Yash-Handa! I think if content is deprecated (technically) or no longer useful/best practice, it should be removed.

It would be good if we get a first impression how much of the content is still relevant/up to date. If you're unsure, just ask here and we can find people working on that particular topic who know the details.

Yash-Handa:
**TL;DR** : Will submit a structured Summary (with suggestions) for all the Articles of the [knowlege](https://github.com/nodejs/nodejs.org/tree/master/locale/en/knowledge) directory in **2 to 3 days**

Thanks, @fhemberger, and @chowdhurian for such prompt response :)

[...]

PR #2072, opened 2019-02-13
Update how-to-use-crypto-module.md

Yash-Handa:
In accordance with issue #1977, I have updated the Article **how-to-use-crypto-module** with the following changes:

*  Updates all links to the nodejs.org API`.
* Updated to `crypto.createCipheriv()` and `crypto.createDecipheriv()`.
* Updated the last example for `iv` support.
*  Changed `optimist` package with `yargs`.

### DIFF
--- a/locale/en/knowledge/cryptography/how-to-use-crypto-module.md
+++ b/locale/en/knowledge/cryptography/how-to-use-crypto-module.md
@@ -66,7 +66,7 @@ Ciphers allow you to encode and decode messages given a password.
 
 ### Cipher Algorithms That Work With Crypto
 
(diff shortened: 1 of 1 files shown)


fhemberger (changes requested):
I think instead of MD5 we should advocate SHA256 for hashing, and AES256 instead of 128. Just in case someone simply copies & pastes this code for something that needs to be secured properly. 😉

fhemberger:
Otherwise, great work as always!

I'd love to have a second person reviewing the crypto code example, as I'm not very familiar with it in Node.js.

Yash-Handa:
> I think instead of MD5 we should advocate SHA256 for hashing, and AES256 instead of 128. Just in case someone simply copies & pastes this code for something that needs to be secured properly. 😉

Fair enough, if we are providing an example it should be full-proof in every aspect.
Will update the example with `SHA256` and `AES256` 😎

ghost (approved):
LGTM

fhemberger (approved):

fhemberger:
Thanks!

PR #2074, opened 2019-02-13
 Update what-are-the-built-in-timer-functions.md

Yash-Handa:
In accordance with issue #1977, I have updated the Article **how-to-prompt-for-command-line-input** with the following changes:

* Added `setImmidiate()` with example

**NOTE**: Let me know if `process.nextTick` should be added as it is technically not a timing function.

### DIFF
--- a/locale/en/knowledge/javascript-conventions/what-are-the-built-in-timer-functions.md
+++ b/locale/en/knowledge/javascript-conventions/what-are-the-built-in-timer-functions.md
@@ -84,11 +84,11 @@ This only has one
 
 #### setImmediate()
 
-`setImmedaite()` is another built-in timer function which as the name suggest, runs immediately after the first iteration of the event loop is completed. In other words, `setImmedaite()` is similar to a `setTimeout()` function with a `0ms` delay. The `setImmedaite()` function can also take extra parameters that are passed when the callback is called:
+`setImmediate()` is another built-in timer function which as the name suggest, runs immediately after the first iteration of the event loop is completed. In other words, `setImmediate()` is similar to a `setTimeout()` function with a `0ms` delay. The `setImmediate()` function can also take extra parameters that are passed when the callback is called:
 
 ```js
 console.log("This will be printed first");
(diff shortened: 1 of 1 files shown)


ZYSzys (approved):
LGTM

locale/en/knowledge/javascript-conventions/what-are-the-built-in-timer-functions.md
```diff
@@ -9,60 +9,95 @@ difficulty: 1
...
 As you can see the parameters to both are the same. The number second parameter says how long in milliseconds to wait before calling the function passed into the first parameter. The difference between the two functions is that `setTimeout` calls the callback only once while `setInterval` will call it over and over again.
 
 Typically you want to be careful with `setInterval` because it can cause some undesirable effects.  If, for example, you wanted to make sure your server was up by pinging it every second, you might think to try something like this:
 
-    setInterval(ping, 1000);
+```js
+setInterval(ping, 1000);
+```
 
 This can cause problems, however, if your server is slow and it takes, say, 3 seconds to respond to the first request. In the time it takes to get back the response, you would have sent off 3 more requests - not exactly desirable!  This isn't the end of the world when serving small static files, but if you doing an expensive operation such as database query or any complicated computation this can have some undesirable results. A common solution looks like this:
```
keywordnew:
```suggestion
This can cause problems, however, if your server is slow and it takes, for example, 3 seconds to respond to the first request. In the time it takes to get back the response, you would have sent off 3 more requests - not exactly desirable! Overall, this doesn't have a large impact when serving small static files. But if you're doing an expensive operation, such as a database query or any complex computation, this can have undesirable results. A common solution looks like this:
```

keywordnew (commented):
PR looks good to me! Really good updates here 🎉

I added an update to an untouched part of the document. Piggybacking on this PR to land the correction if that's ok :)

fhemberger (approved):

ghost:
@Yash-Handa: Thanks!

PR #2080, opened 2019-02-15
Changed command for building node modules to npm ci

ddc22:
To maintain the exact tree as committed in development npm has introduced the npm ci command which is meant to provide reproducible builds.
please take a look at the [official documentation](https://docs.npmjs.com/cli/ci.html) for more information.

### DIFF
--- a/locale/en/docs/guides/nodejs-docker-webapp.md
+++ b/locale/en/docs/guides/nodejs-docker-webapp.md
@@ -110,7 +110,7 @@ COPY package*.json ./
 
 RUN npm install
 # If you are building your code for production
-# RUN npm install --only=production
+# RUN npm ci --only=production
 ```
 
 Note that, rather than copying the entire working directory, we are only copying
--- a/locale/en/docs/guides/nodejs-docker-webapp.md
+++ b/locale/en/docs/guides/nodejs-docker-webapp.md
@@ -117,7 +117,7 @@ Note that, rather than copying the entire working directory, we are only copying
 the `package.json` file. This allows us to take advantage of cached Docker
 layers. bitJudo has a good explanation of this
 [here](http://bitjudo.com/blog/2014/03/13/building-efficient-dockerfiles-node-dot-js/). 
-Furthermore, the npm ci command, specified in the comments, helps provide faster, reliable, reproducible builds for production environments. 
+Furthermore, the `npm ci` command, specified in the comments, helps provide faster, reliable, reproducible builds for production environments. 
 You can read more about this [here](https://blog.npmjs.org/post/171556855892/introducing-npm-ci-for-faster-more-reliable). 
 
 To bundle your app's source code inside the Docker image, use the `COPY`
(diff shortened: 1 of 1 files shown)


fhemberger:
Hi, thanks for your contribution!

There are a couple more instances of `npm install` which should be `npm ci` according to the definition. It would also be nice if you could add a sentence explaining the difference to `npm install`, as not everybody will be familiar with this command. A link to the npm docs would help as well.

ddc22:
Hi @fhemberger! Glad to contribute :). Please let me know if there should be any other additions.

locale/en/docs/guides/nodejs-docker-webapp.md
```diff
@@ -110,13 +110,15 @@ COPY package*.json ./
...
-# RUN npm install --only=production
+# RUN npm ci --only=production
 ```
 
 Note that, rather than copying the entire working directory, we are only copying
 the `package.json` file. This allows us to take advantage of cached Docker
 layers. bitJudo has a good explanation of this
-[here](http://bitjudo.com/blog/2014/03/13/building-efficient-dockerfiles-node-dot-js/).
+[here](http://bitjudo.com/blog/2014/03/13/building-efficient-dockerfiles-node-dot-js/). 
+Furthermore, the npm ci command, specified in the comments, helps provide faster, reliable, reproducible builds for production environments.
```
ZYSzys: Nit:
```suggestion
Furthermore, the `npm ci` command, specified in the comments, helps provide faster, reliable, reproducible builds for production environments. 
```

ZYSzys (approved):
LGTM

locale/en/docs/guides/nodejs-docker-webapp.md
```diff
@@ -110,13 +110,15 @@ COPY package*.json ./
 
 RUN npm install
```
fhemberger, line 111: Should also be changed here.

locale/en/docs/guides/nodejs-docker-webapp.md
```diff
@@ -156,7 +158,7 @@ COPY package*.json ./
...
```
fhemberger, line 159: Should also be changed here.

fhemberger (changes requested):

locale/en/docs/guides/nodejs-docker-webapp.md
ddc22: NPM CI is only required for production builds not the normal build

locale/en/docs/guides/nodejs-docker-webapp.md
ddc22: NPM CI is only required for production builds not the normal build

locale/en/docs/guides/nodejs-docker-webapp.md
ddc22: Added code formatting as suggested

willin (approved):

fhemberger (approved):

fhemberger:
Thank you!

ghost:
@nodejs/localization：Need to be translated for your own languages if possible and free for you.

PR #2085, opened 2019-02-16
Update  what-is-the-arguments-object.md

Yash-Handa:
In accordance with issue #1977, I have updated the Article **what-is-the-arguments-object** with the following changes:

* Added details about `arrow function` exception and its workaround (`rest parameter`)
* Added `Array.from()` method usage.
* Added better syntax high lighting.

### DIFF
--- a/locale/en/knowledge/javascript-conventions/what-is-the-arguments-object.md
+++ b/locale/en/knowledge/javascript-conventions/what-is-the-arguments-object.md
@@ -24,41 +24,63 @@ property of `arguments` is callee. This always refers to the function
 currently being executed. Here's an example that illustrates the
 properties of `arguments`.
 
-    var myfunc = function(one) {
-      arguments.callee === myfunc;
-      arguments[0] === one;
-      arguments[1] === 2;
-      arguments.length === 3;
-    }
+```js
+const myfunc = function(one) {
+  arguments.callee === myfunc;
+  arguments[0] === one;
+  arguments[1] === 2;
+  arguments.length === 3;
+}
 
-    myfunc(1, 2, 3);
+myfunc(1, 2, 3);
+```
 
 This construct is very useful and gives javascript functions a lot of
 flexibility. But there is an important gotcha. The `arguments` object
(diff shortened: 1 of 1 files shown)


ghost (approved):

locale/en/knowledge/javascript-conventions/what-is-the-arguments-object.md
```diff
@@ -24,41 +24,63 @@ property of `arguments` is callee. This always refers to the function
...
 
-    var myfunc = function(one) {
-      arguments.callee === myfunc;
-      arguments[0] === one;
-      arguments[1] === 2;
-      arguments.length === 3;
-    }
+```js
+const myfunc = function(one) {
+  arguments.callee === myfunc;
```
fhemberger: `arguments.callee` has been deprecated in ES5 strict mode:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/arguments/callee

locale/en/knowledge/javascript-conventions/what-is-the-arguments-object.md
```diff
@@ -24,41 +24,63 @@ property of `arguments` is callee. This always refers to the function
...
-      args.push(arguments[i]);
-    }
+```js
+//  for ES5 and below
+const args = [];
+for(let i = 0; i < arguments.length; i++) {
+  args.push(arguments[i]);
+}
+
+// for ES6 and above
```
fhemberger: Maybe we should push this as the default for current Node.js versions (and only mention the pre-ES6 variants briefly). What do you think?

fhemberger (changes requested):

locale/en/knowledge/javascript-conventions/what-is-the-arguments-object.md
Yash-Handa: Thanks, @fhemberger for the advice will update both the sections :)

fhemberger (approved):

fhemberger:
Thanks!

PR #2091, opened 2019-02-19
【zh-CN】Fix the error in debugging-getting-started.md

YFengFly:
In line 168,the word "启用" is should be “弃用”.
It might be a Pinyin select wrong.
@nodejs/node-cn


stevemao (approved):
Thanks

PR #2098, opened 2019-02-20
Link version schedule to about/releases page

fhemberger:
As we now copy the English pages where they are missing in locales, this is the remaining change needed from #1893.


ZYSzys (approved):
LGTM. Thanks👍.

PR #2099, opened 2019-02-20
Add backquotes for some function names

ghost:
In 'dependencies.md', some function names should be surrounded with backquotes to differ them from common names.


fhemberger:
LGTM, thanks!

Issue #2035, opened 2019-02-07
Does 'banner' really need translation?


ghost:
Hi all:

I notice that since we've forcely defined the banner's Text as the fixed English string in our build.js:

```js
 loadVersions((err, versions) => {
    const source = {
      project: {
       ……
        banner: {
          visible: true,
          text: 'November 2018 security releases available, upgrade now',
          link: '/en/blog/vulnerability/november-2018-security-releases/'
        }
       ……
    }
```
And then in the template, we'll check whether the 'project.banner.text' has value or not? If yes, directly use banner's text, otherwises use 'label.banner'：
```html
 {{#if project.banner.visible}}
          <p class="home-version home-version-banner">
            <a href="{{ project.banner.link }}">{{#if project.banner.text}}{{{ project.banner.text }}}{{else}}{{{ labels.banner }}}{{/if}}</a>
          </p>
        {{/if}}
```
However in the recent months for upgration, it doesn't seem we directly use 'label.banner' but directly use 'project.banner.text', something like this following (For test only at local place)：

![image](https://user-images.githubusercontent.com/40081831/52402100-581f0b00-2afe-11e9-93fb-6bc786de217a.png)

On the other hand, since we cannot forsee when we will publish a new Important relase.... So do we really need this translation for banner? There're two ways:

**1) Remove the tranlation defined in the all kinds of languages' index page of 'label.banner', but just keep the English notification.**

**2) Remove the 'project.banner.text' and use the localization defined in the all kinds of languages.**

Which seems better?

fhemberger:
So far, we avoided translation here on purpose. This banner is only used for important updates like security fixes, which need to be put out fast. 

Also, messages might differ, depending on the kind of security issue, so we can't wait until the message gets translated into every language.

(label.banner is still a relic of previous translation efforts)

ghost:
>Also, messages might differ, depending on the kind of security issue, so we can't wait until the message gets translated into every language.

Yes, that's the reason why I say that there's no need of translation but just use the project.label.text. So no need to do translations IMO. But I just wait for others' opinions on this, thanks @fhemberger.

PR #2110, opened 2019-02-22
Fix：Remove useless translations for 'labels.banner'

ghost:
According to the real senario, it doesn't seem we need to translate the 'banner' because it's frequently changed, and on the other hand, the time between notification and publishment is too short to translate to all kinds of languages.

Ref：https://github.com/nodejs/nodejs.org/issues/2035.

### DIFF
--- a/layouts/index.hbs
+++ b/layouts/index.hbs
@@ -12,11 +12,11 @@
 
         {{{ contents }}}
 
-        {{#if project.banner.visible}}
-          <p class="home-version home-version-banner">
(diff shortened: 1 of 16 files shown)


Maledong:
@nodejs/website：This seems a big change, so any other suggestions on this?

aymen94:
@Maledong I think that's okay.

ZYSzys (approved):

aymen94 (approved):

willin (approved):

PR #2111, opened 2019-02-22
Fix list item in event-loop-timers-and-nexttick.md

Hakerh400:
Double asterisk used for bold text is interpreted as list item.

**Before**

**After**


ghost (approved):
Nice!

ZYSzys (approved):

PR #2116, opened 2019-02-26
chore: Update the related npm packages

ghost:
Update these packages and make them function well. Pass all the unit tests and doc generation tests.

---
- [X] Pass Unit Tests.
- [X] Pass Doc Generation Test Locally


fhemberger:
@Maledong By the way: You don't have to explicitly ask for my review. You are doing tremendous work here and I trust your decisions.

ghost:
@fhemberger：No problem but just to avoid mistakes. So you don't need to watch immediately if you are too busy or trouble you more.... but just a notification, still say A BIG THANK YOU!

PR #2117, opened 2019-02-27
feat(events): use table for listing of events

lirantal:
Updating the events to use tabular view which is easier to manage and reads better

Example:
image

### DIFF
--- a/locale/en/get-involved/node-events.md
+++ b/locale/en/get-involved/node-events.md
@@ -11,7 +11,7 @@ This is a list of Node.js events and conferences. Please submit a PR if you'd li
 
 FORMAT
 
-- [Event Name]() | Location | Dates
+- Event Name | Location | Dates
 
 ## Events
 
--- a/locale/en/get-involved/node-events.md
+++ b/locale/en/get-involved/node-events.md
@@ -10,11 +10,15 @@ This is a list of Node.js events and conferences. Please submit a PR if you'd li
 ## Notes for adding events
 
 FORMAT
-- [Event Name](https://) - Location, Dates
+
+- [Event Name](https://) | Location | Dates
 
 ## Events
 
 ### 2019
-- [NodeConf Colombia](https://colombia.nodeconf.com) - Medellin, Colombia, June 21-22 
-- [NodeConf EU](https://www.nodeconf.eu/2019.html) - Lyrath Estate, Kilkenny, November 10-13
-- [NodeSummit](https://www.nodesummit.com) - San Francisco, TBA 
+
+|                     Event Name                     |        Location         |      Date      |
+| :------------------------------------------------: | :---------------------: | :------------: |
+| [NodeConf Colombia](https://colombia.nodeconf.com) |   Medellin, Columbia    |   June 21-22   |
+|  [NodeConf EU](https://www.nodeconf.eu/2019.html)  | Lyrath Estate, Kilkenny | November 10-13 |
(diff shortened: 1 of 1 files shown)


locale/en/get-involved/node-events.md
```diff
@@ -10,11 +10,15 @@ This is a list of Node.js events and conferences. Please submit a PR if you'd li
 ## Notes for adding events
 
 FORMAT
-- [Event Name](https://) - Location, Dates
+
+- [Event Name](https://) | Location | Dates
```
ZYSzys: Since you are here, could you please also change the `Event Name` to normal text, the link `https://` seems invalid here.

```suggestion
- Event Name | Location | Dates
```

lirantal: The idea is that the event is actually a link to the website, look at the example and the contents I added

ZYSzys: Seems reasonable. 

Then how about just remove `https://` in the scope ?

Now on our website [Node.js event](https://nodejs.org/en/get-involved/node-events/) page, clicking here(maybe accidentally) would lead to an empty page.

lirantal: True, that could be confusing. I'll remove just that reference.

Maledong: IMO：Here the 'Event Name' is ONLY a tip for you to give a link as a tip, so I don't think we need to exclipit show the link. I think those submitters will add the links for their event names as what others do,

locale/en/get-involved/node-events.md
```diff
@@ -10,11 +10,15 @@ This is a list of Node.js events and conferences. Please submit a PR if you'd li
...
 
 ## Events
 
 ### 2019
-- [NodeConf Colombia](https://colombia.nodeconf.com) - Medellin, Colombia, June 21-22 
-- [NodeConf EU](https://www.nodeconf.eu/2019.html) - Lyrath Estate, Kilkenny, November 10-13
-- [NodeSummit](https://www.nodesummit.com) - San Francisco, TBA 
+
```
ghost, line 19: For 'node-speakers' and 'node-meetups', we don't use table style. So do we really need to use the table style, or keep them in union?

locale/en/get-involved/node-events.md
ghost: What about using a pure text plain instead? Because others will see they should use link, or just use an empty link instead (don't need to open up a new window).... :)

lirantal: Agree!

locale/en/get-involved/node-events.md
lirantal: So perhaps update node-speakers and node-meetups to use table style? 😜

I like the table style more and it looks a lot more readable.
Perhaps we can try both styles and see what works better. I actually think that there would be more speakers and more meetups than conferences so it makes even more sense to really consider updating them to table style.

ghost: @lirantal：Agree with you :)

ghost (approved):

ZYSzys (approved):

Maledong:
@fhemberger：Just wonder whether we allow a table-styled content  or line one by one in union like others

ghost:
@lirantal：Thanks for your new contribution!

lirantal:
@Maledong my pleasure!

PR #2124, opened 2019-03-01
Fix：Uncomment out `moreWork()`

WofWca:
`moreWork()` is a pseudo function, so it shouldn't be commented out.


ghost:
@WofWca：Good catch, and it's really kind of you to fix it! Thanks anyway!

PR #2141, opened 2019-03-06
Refactor：Re-use the same label with the same variable

ghost:
Refactor to take the labels into the same variable to reuse it.

### DIFF
--- a/build.js
+++ b/build.js
@@ -58,7 +58,8 @@ function i18nJSON (lang) {
 // english one.
 function buildLocale (source, locale, opts) {
   console.log(`[metalsmith] build/${locale} started`)
-  console.time(`[metalsmith] build/${locale} finished`)
+  const labelForBuild = `[metalsmith] build/${locale} finished`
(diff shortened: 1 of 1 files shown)


ghost:
What makes me feel strange is：Since the other labels are registered in the function, why there's only the label not yet? I'm on windows 10 (x86, professional, Simplified Chinese, version：17143.619 with the latest version of Nodejs).

BridgeAR:
@Maledong did you run the former more than once? It looks like a race condition or something like that to me where the label would already be taken.

The added variable should not change the behavior at all.

fhemberger:
Sounds like a race condition to me as well (haven't noticed that behavior myself so far). I probably wouldn't invest _too_ much time in the build process any more, as the upcoming relaunch will have a totally new one.

But if this change fixes your issue for now, that's good.

ghost:
Thanks all, I changed the title and reguard it as a refactor of codes.

PR #2146, opened 2019-03-10
Update how-to-create-default-parameters-for-functions.md

Yash-Handa:
In accordance with issue #1977, I have updated the Article **how-to-create-default-parameters-for-functions** with the following changes:

- Added modern JavaScript (ES6) `default parameter` specification.
- Added "falsy" value anomaly in `default parameter`.
- Removed an extra example for ES5 and below.
- some other miscellaneous updates and refactors.

If there are any suggestion/enhancement/tips please feel free to tell me, I would love to incorporate those changes in this Article 😄

Sorry for such a long delay on this pull request. Student's Life ¯\__(ツ)_/¯

### DIFF
--- a/locale/en/knowledge/javascript-conventions/how-to-create-default-parameters-for-functions.md
+++ b/locale/en/knowledge/javascript-conventions/how-to-create-default-parameters-for-functions.md
@@ -32,7 +32,7 @@ In JavaScript there are many ways of indicating ["falsy"(false values)](/en/know
 
 JS PRO TIP: In `Math.pow(base, power)` if `power` is 0 or any other "falsy" value (except NaN) the result will always be 1.
 
-Unfortunately, previous versions of javascript (ES5 and below) does not support **default parameters** out of the box but over time, however, people have developed idioms to compensate.
+Unfortunately, previous versions of JavaScript (ES5 and below) does not support **default parameters** out of the box but over time, however, people have developed idioms to compensate.
 
 The first idiom is giving a default value for the last parameter. This is done by checking if the last parameter is `undefined` and setting it to a default value if it is. Sometimes people use the idiom: `optionalParameter = optionalParameter || defaultValue`. This can have some undesirable behaviour when they pass values that are equal to false such as `false`, `0`, and `""`. So a better way to do this is by explicitly checking that the optional parameter is `undefined`. Here is some code showing the two styles and the differing behaviour:
 
--- a/locale/en/knowledge/javascript-conventions/how-to-create-default-parameters-for-functions.md
+++ b/locale/en/knowledge/javascript-conventions/how-to-create-default-parameters-for-functions.md
@@ -8,17 +8,31 @@ difficulty: 1
 layout: knowledge-post.hbs
 ---
 
-Usually a function will take a set number of parameters, and require that all of them be present before it can be executed successfully. However, you will sometimes run into situations where you want to provide a default value for a parameter or take a variable number of parameters. Fortunately, from ES6/ES2015, default parameters is in the language specification:
+Usually a function will take a set number of parameters, and require that all of them be present before it can be executed successfully. However, you will sometimes run into situations where you want to provide a default value for a parameter or take a variable number of parameters. Fortunately, from ES6/ES2015 **default parameters** is in the language specification:
 
 ```js
+const pow = (base, power = 2) => {
+  return Math.pow(base, power);
+}
 
+console.log(pow(2)); // 4
+console.log(pow(2,10)); // 1024
 ```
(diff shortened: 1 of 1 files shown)


locale/en/knowledge/javascript-conventions/how-to-create-default-parameters-for-functions.md
```diff
@@ -8,61 +8,74 @@ difficulty: 1
...
+The first idiom is giving a default value for the last parameter. This is done by checking if the last parameter is `undefined` and setting it to a default value if it is. Sometimes people use the idiom: `optionalParameter = optionalParameter || defaultValue`. This can have some undesirable behaviour when they pass values that are equal to false such as `false`, `0`, and `""`. So a better way to do this is by explicitly checking that the optional parameter is `undefined`. Here is some code showing the two styles and the differing behaviour:
+
+```js
+const example = function (optionalArg) {
+  optionalArg = optionalArg || "No parameter was passed";
+  console.log(optionalArg);
+}
+
+const betterExample = function (optionalArg) {
+  if (typeof optionalArg === 'undefined') {
```
ghost: What about using `optionalArg === undefined` directly instead of using `typeof` to check the type?

locale/en/knowledge/javascript-conventions/how-to-create-default-parameters-for-functions.md
```diff
@@ -8,61 +8,74 @@ difficulty: 1
...
+console.log("\nEmpty String:");
+example("");
+betterExample("");
+```
+
+The second idiom is when the optional value is in the middle it can cause some undesired effects since all the parameters are shifted over. The optional parameter is not the `undefined` value in this case - the last parameter is the `undefined` one. So you have to check if the last parameter is `undefined` and then manually fix all the other parameters before continuing in the code. This case is also valid for modern JavaScript(ES6/ES2015). The example shows you how to do that:
+
+```js
+const example = function (param1, optParam, callback) {
+  if (typeof callback === 'undefined') {
```
ghost: What about using optionalArg === undefined directly instead of using typeof to check the type? Or there's something on purpose to use it? :D

ghost (approved):
LGTM, suggestions are in my comments, welcome your discussions :)

locale/en/knowledge/javascript-conventions/how-to-create-default-parameters-for-functions.md
```diff
@@ -8,61 +8,74 @@ difficulty: 1
...
+}
+
+console.log(pow(2, undefined)); // 4
+console.log(pow(2, null)); // 1
+```
+In JavaScript there are many ways of indicating ["falsy"(false values)](/en/knowledge/javascript-conventions/what-are-truthy-and-falsy-values/) but out of them, only `undefined` will trigger the default parameter. This is the reason why `pow(2, undefined)` returns 4 and `pow(2, null)` returns 1.
+
+JS PRO TIP: In `Math.pow(base, power)` if `power` is 0 or any other "falsy" value (except NaN) the result will always be 1.
+
+Unfortunately, previous versions of javascript (ES5 and below) does not support **default parameters** out of the box but over time, however, people have developed idioms to compensate.
```
ZYSzys: Nit: `javascript` => `JavaScript`

```suggestion
Unfortunately, previous versions of JavaScript (ES5 and below) does not support **default parameters** out of the box but over time, however, people have developed idioms to compensate.
```

locale/en/knowledge/javascript-conventions/how-to-create-default-parameters-for-functions.md
Yash-Handa: Thanks, @Maledong for that great input 🙂
Though in these examples particularly, it won't make any difference using `optionalArg === undefined` or `typeof optionalArg === 'undefined'` but in general `typeof` is slightly more reliable/predictable.
This answer [here](https://stackoverflow.com/a/4726198/9991821) on StackOverflow provides a great explanation on this.

I would love to have your suggestions/further discussion on this

ZYSzys (approved):

locale/en/knowledge/javascript-conventions/how-to-create-default-parameters-for-functions.md
ghost: @Yash-Handa：I see, but the answer means if you cannot decide whether your variable is defined or not, you can safely use `typeof somevariable === 'undefined'`, since in our senario, we already have the variable defined `optionalArg` as the input parameter, and it isn't a dynamic one. So I tend to use `optionalArg === undefined`. :)

Yash-Handa: Yeah @Maledong, In these examples it won't matter if we use `optionalArg === undefined` or `typeof optionalArg === 'undefined'`.
As you suggest, so let us go with `optionalArg === undefined` approach. It would simplify (less verbose) the code. 😎

ghost:
Thank you! @Yash-Handa

PR #2150, opened 2019-03-12
add missing semicolon

Trott:
sh_main.js uses semicolons everywhere except in one line. Add the
missing semicolon.


vsemozhetbyt (approved):

BethGriggs:
I don't know the history of this dependency, but should it also be changed over in nodejs/node? Do we maintain it in two places? 
https://github.com/nodejs/node/blob/master/doc/api_assets/sh_main.js#L523 

Also, does it need to be called out anywhere that it is licenced as `gplv3`?

fhemberger (approved):

fhemberger:
I don't think it's needed in nodejs/node any longer. It's part of the legacy docs around version 0.x.

ghost:
Thank you all!

PR #2151, opened 2019-03-13
Remove the 'copyright' and 'copyright_year.js'

ghost:
Due to the discussion at nodejs/TSC#195 (comment), we
shouldn't include an auto copyright_year and its realted token place. So
we should remove it.

### DIFF
--- a/layouts/partials/footer.hbs
+++ b/layouts/partials/footer.hbs
@@ -15,7 +15,7 @@
         </ul>
       </div>
 
-      <p>© {{copyright}} Node.js Foundation. All Rights Reserved. Portions of this site originally © {{copyright}} Joyent. </p>
+      <p>© Node.js Foundation. All Rights Reserved. Portions of this site originally © Joyent. </p>
       <p>Node.js is a trademark of Joyent, Inc. and is used with its permission. Please review the <a href="/static/documents/trademark-policy.pdf">Trademark Guidelines of the Node.js Foundation</a>.</p>
       <p>Linux Foundation is a registered trademark of The Linux Foundation.</p>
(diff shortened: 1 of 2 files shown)


Maledong:
@nodejs/website：Please have a review if possible, thanks!

Trott (dismissed):
There is no need to include the year. Please don't. It just makes for pointless annual updates that everyone forgets to do. 

EDIT 1: I am not a lawyer. Expert opinion welcome, whether it be contrary to mine or not.

EDIT 2: I see this auto-updates the year. That is also problematic, I think, although less so for sure. We should not update the year for material that isn't updated. For minor updates, it should probably be a range of years rather than the latest year. Omitting the year solves the problem.

Trott:
Resulting word from the then-Executive Director of the Node.js Foundation on this topic after previous TSC conversation: https://github.com/nodejs/TSC/issues/195#issuecomment-374443984

Third bullet point: "Adding dates is not preferable. It does not have a legal impact and would be excluded only for ease of administration."

Our current LICENSE file says: "Copyright Node.js contributors. All rights reserved."

Trott:
(Although the above was about files we distribute in the source tree and not websites.)

ghost:
@Trott：Thanks, and it's ALWAYS you that is the FIRST to help me a lot!
Now I've removed something useless and make it suitable to the discussion.

ZYSzys (approved):
LGTM

osk2 (approved):
👍

aymen94:
LGTM

aymen94 (approved):

marswong (approved):
LGTM

PR #2155, opened 2019-03-18
zh-CN：Make 'pull requests' into the uppercase

ghost:
Make 'pull requests' into an uppercase because it's a fixed phrase instead of a traditional English saying.

### DIFF
--- a/locale/zh-cn/get-involved/contribute.md
+++ b/locale/zh-cn/get-involved/contribute.md
@@ -35,9 +35,9 @@ layout: contribute.hbs
 
 协作者的影响力可扩及整个项目，成为协作者后将能审核他人的贡献内容、将问题进行分类，及策划项目未来。因重大贡献而被 TSC 认可的开发者可能会被授与项目协作者身份及享有项目的代码提交权限，重大贡献的评估标准如（包括但不限于）下方所列：
 
-* 代码提交以及 pull requests
-* 文档修订提交及 pull requests
-* 对问题的评论及 pull requests
+* 代码提交以及 Pull Requests
+* 文档修订提交及 Pull Requests
+* 对问题的评论及 Pull Requests
 * 对 Node.js 网站的贡献
 * 提供给终端用户及初学贡献者的帮助
 * 参与工作组


willin (approved):
其实已经习惯了看 `PR`

fengmk2:
-1 for this change.

I think keep "pull requests" is better than translate it to Chinese. We use "PR" or "Pull Request" in Chinese communication more than "代码审核请求".

> "代码审核请求" 用中文来说太奇怪了，没人会这样说。

laosb:
-1 too.

Pull request is the most widely-accepted way to express that meaning. Even GitLab/Gitea users say PR often. No need to translate because every one know it.

In terms of saying *代码审核请求*, it would create unnecessary misunderstanding or guess. It's not a word-by-word translation which means "code review request" in English, some logical guess is certainly involved when a non-native English speaker see that word before recognize it's pull request.

If you really need a translation, *拉取请求* will be a better one because it's direct and PR is translated so most of times.

ghost:
@fengmk2 & @laosb：

First thank you for all your patience to have a review of my translation, and here's a description of "Pull Requests" (at：https://help.github.com/en/articles/about-pull-requests )：

>Pull requests let you tell others about changes you've pushed to a branch in a repository on GitHub. Once a pull request is opened, you can discuss and review the potential changes with collaborators and add follow-up commits before your changes are merged into the base branch.

So I tend to translate it into Chinese for a pure Chinese article, and according to its meaning, I think we should explain the Chinese meaning instead of directly translating it word by word. On the other hand in my opinion, "Pull Request" => "拉取请求" isn't a proper meaning. 

BTW：Just like other translations, I'd like to put the English fixed phrase in （）, with the Chinese translation. :)

laosb:
My opinion would be just leave it as is. It's even easier to understand for most Chinese developers, so why should we bother translating it? Just doesn't make sense to use () there, they hurt the fluency.

ghost:
@laosb: OK. I've reverted it back by making it into the uppercase, because it'd be a fixed phrase.

laosb:
Great. Looks good to me now. Thank you!

PR #2158, opened 2019-03-18
remove the outdated Deduplication section from event-loop-timers-and-nexttick.md

bughit:
Deduplication section appears to be describing how things worked prior to nodejs/node#22842, so is no longer correct. The output in 11.12 is 

```
setImmediate 1
process.nextTick a
process.nextTick b
setImmediate 2
process.nextTick a
process.nextTick b
```

### DIFF
--- a/locale/en/docs/guides/event-loop-timers-and-nexttick.md
+++ b/locale/en/docs/guides/event-loop-timers-and-nexttick.md
@@ -398,56 +398,6 @@ To get around this, the `'listening'` event is queued in a `nextTick()`
 to allow the script to run to completion. This allows the user to set
 any event handlers they want.
 
-### Deduplication
-
-For the `timers` and `check` phases, there is a single transition 
-between C to JavaScript for multiple immediates and timers. This deduplication
-is a form of optimization, which may produce some unexpected side effects.
-Take this code snippet as an example:
-
-```js
-// dedup.js
-const foo = [1, 2];
-const bar = ['a', 'b'];
-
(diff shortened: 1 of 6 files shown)


Fishrock123:
This change is still pretty new... can we point out that it acts as normal since 11.2 (or whichever version it was), rather than removing it?

bughit:
Ideally guides should be versioned like api docs. Since they are not, I was assuming they apply to the latest version, as it does not seem practical to have the same document cover multiple versions.

AFAICT the Deduplication section was only recently added (b440166b214757981b5155b03d7979dccc237b99) when it was no longer valid for the latest version.

However if you want a more nuanced rewording rather than removal, perhaps the original author (@OmarDarwish) is better suited for that.

Trott:
Unpopular opinion: I'd be for removing the guides entirely. The website should be minimal. The guides are pretty much unmaintained. Other sites probably do this better.

Trott (approved):

Maledong:
@nodejs/nodejs-ko：Don't forget to update the related translation, and I'll merge this if possible. I agree with the change :)

ghost (approved):

Maledong:
@nodejs/nodejs-ko：Please check whether your file should be updated or not, and then I'll merge it soon :)

yous:
@nodejs/nodejs-ko I've just pushed to follow changes of event-loop-timers-and-nexttick.md. Please review this.

marocchino:
@yous LGTM

Issue #2165, opened 2019-04-01
Docs > how-to-use-buffers.md 


19shubham11:
I was just going through the docs and saw [how-to-use-buffers.md](https://github.com/nodejs/nodejs.org/blob/master/locale/en/knowledge/advanced/buffers/how-to-use-buffers.md), which basically explains about buffers using the `new` Buffer() constructors which is unsafe. 

Not really sure whether this is being referred anywhere, if it is, I think that documentation has to be updated.

ghost:
@19shubham11：Thanks for your quick review, and sometimes the KB files aren't published yet so there are something too old. I'll fix it by a quick description to difference the versions :)

PS：History of `new Buffer()`：https://nodejs.org/dist/latest-v10.x/docs/api/buffer.html#buffer_new_buffer_string_encoding

PR #2166, opened 2019-04-02
Fix: Formation and descriptions of 'how-to-use-buffers.md'

ghost:
1) To differ the versions between v6.0.0 and later, because there're some methods deprecated for 'Buffer' (See: https://github.com/nodejs/nodejs.org/issues/2165).

2) Add some normal js formations to make snippets of codes shown better.

cc: https://github.com/nodejs/nodejs.org/issues/1977

### DIFF
--- a/locale/en/knowledge/advanced/buffers/how-to-use-buffers.md
+++ b/locale/en/knowledge/advanced/buffers/how-to-use-buffers.md
@@ -21,7 +21,7 @@ Don't use binary strings. Use *buffers* instead!
 
 ## What Are Buffers?
 
-Buffers are instances of the `Buffer` class in node, which is designed to handle raw binary data. Each buffer corresponds to some raw memory allocated outside V8. Buffers act somewhat like arrays of integers, but aren't resizable and have a whole bunch of methods specifically for binary data. In addition, the "integers" in a buffer each represent a byte and so are limited to values from 0 to 255 (2^8 - 1), inclusive. When using `console.log()` to print the `Buffer` instance, you'll get a chain of values in hex.
+The `Buffer` class in Node.js is designed to handle raw binary data. Each buffer corresponds to some raw memory allocated outside V8. Buffers act somewhat like arrays of integers, but aren't resizable and have a whole bunch of methods specifically for binary data. The integers in a buffer each represent a byte and so are limited to values from 0 to 255 inclusive. When using `console.log()` to print the `Buffer` instance, you'll get a chain of values in hexadecimal values.
 
 ## Where You See Buffers:
 
@@ -38,9 +38,7 @@ var buffer = Buffer.alloc(8);
 // This will print out 8 bytes of zero:
 // <Buffer 00 00 00 00 00 00 00 00>
 ```
-This buffer is uninitialized and contains 8 bytes of zero, notice that `new Buffer(size)` is deprecated since v6.0.0.
-
-Besides, we've also got [Buffer.allocUnsafe(size)](https://nodejs.org/dist/latest/docs/api/buffer.html#buffer_class_method_buffer_allocunsafe_size). This makes you get a buffer quickly without initializing it with zero values, so it's sometimes unsafe.
+This buffer is initialized and contains 8 bytes of zero.
 
 ```js
 var buffer = Buffer.from([ 8, 6, 7, 5, 3, 0, 9]);
@@ -48,7 +46,7 @@ var buffer = Buffer.from([ 8, 6, 7, 5, 3, 0, 9]);
 // <Buffer 08 06 07 05 03 00 09>
 ```
 
-This initializes the buffer to the contents of this array. Keep in mind that the contents of the array are integers representing bytes. Notice that `new Buffer(array)` is deprecated since v6.0.0.
+This initializes the buffer to the contents of this array. Keep in mind that the contents of the array are integers representing bytes.
 
 ```js
 var buffer = Buffer.from("I'm a string!", "utf-8");
@@ -56,9 +54,7 @@ var buffer = Buffer.from("I'm a string!", "utf-8");
 // <Buffer 49 27 6d 20 61 20 73 74 72 69 6e 67 21>
 ```
 
-This initializes the buffer to a binary encoding of the first string as specified by the second argument (in this case, utf-8). **utf-8** is by far the most common encoding used with node, but `Buffer` also supports some other formations, please see [Supported Encodings](https://nodejs.org/dist/latest/docs/api/buffer.html#buffer_buffers_and_character_encodings).
-
-Notice that `new Buffer(string[, encoding])` is deprecated since v6.0.0.
+This initializes the buffer to a binary encoding of the first string as specified by the second argument (in this case, `'utf-8'`). `'utf-8'` is by far the most common encoding used with Node.js, but `Buffer` also supports others. See [Supported Encodings](https://nodejs.org/dist/latest/docs/api/buffer.html#buffer_buffers_and_character_encodings) for more details.
 
 ### Writing to Buffers
 
--- a/locale/en/knowledge/advanced/buffers/how-to-use-buffers.md
+++ b/locale/en/knowledge/advanced/buffers/how-to-use-buffers.md
@@ -21,7 +21,7 @@ Don't use binary strings. Use *buffers* instead!
 
 ## What Are Buffers?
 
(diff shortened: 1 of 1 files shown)


aymen94:
LGTM

aymen94 (approved):

19shubham11 (approved):
👍

Trott (dismissed):
This needs some editing.

Trott:
Wow, I didn't even know we had a supposed "Knowledge Base" section of the website. I don't think this is the type of material we should have on the website to be honest. Anything that requires active curation is a pretty bad fit for the website. We do this stuff in fits and starts at best and the results don't inspire confidence in our attention to detail.

locale/en/knowledge/advanced/buffers/how-to-use-buffers.md
```diff
@@ -21,7 +21,7 @@ Don't use binary strings. Use *buffers* instead!
 
 ## What Are Buffers?
 
-Buffers are instances of the `Buffer` class in node, which is designed to handle raw binary data. Each buffer corresponds to some raw memory allocated outside V8. Buffers act somewhat like arrays of integers, but aren't resizable and have a whole bunch of methods specifically for binary data. In addition, the "integers" in a buffer each represent a byte and so are limited to values from 0 to 255 (2^8 - 1), inclusive.
+Buffers are instances of the `Buffer` class in node, which is designed to handle raw binary data. Each buffer corresponds to some raw memory allocated outside V8. Buffers act somewhat like arrays of integers, but aren't resizable and have a whole bunch of methods specifically for binary data. In addition, the "integers" in a buffer each represent a byte and so are limited to values from 0 to 255 (2^8 - 1), inclusive. When using `console.log()` to print the `Buffer` instance, you'll get a chain of values in hex.
```
Trott:
```suggestion
The `Buffer` class in Node.js is designed to handle raw binary data. Each buffer corresponds to some raw memory allocated outside V8. Buffers act somewhat like arrays of integers, but aren't resizable and have a whole bunch of methods specifically for binary data. The integers in a buffer each represent a byte and so are limited to values from 0 to 255 inclusive. When using `console.log()` to print the `Buffer` instance, you'll get a chain of values in hexadecimal values.

File: locale/en/knowledge/advanced/buffers/how-to-use-buffers.md [hunk 2]
```diff
@@ -33,22 +33,32 @@ In the wild, buffers are usually seen in the context of binary data coming from
...
-    var buffer = new Buffer(8);
+```js
+var buffer = Buffer.alloc(8);
+// This will print out 8 bytes of zero:
+// <Buffer 00 00 00 00 00 00 00 00>
+```
+This buffer is uninitialized and contains 8 bytes of zero, notice that `new Buffer(size)` is deprecated since v6.0.0.
 
-This buffer is uninitialized and contains 8 bytes.
+Besides, we've also got [Buffer.allocUnsafe(size)](https://nodejs.org/dist/latest/docs/api/buffer.html#buffer_class_method_buffer_allocunsafe_size). This makes you get a buffer quickly without initializing it with zero values, so it's sometimes unsafe.
```
Trott: As with the constructor, maybe don't mention it at all? People can find out about it in the docs if they really want to know, and that's where they should be going anyway. (I'm honestly not sure this doc is more valuable than the Buffer ref doc.)

locale/en/knowledge/advanced/buffers/how-to-use-buffers.md
```diff
@@ -33,22 +33,32 @@ In the wild, buffers are usually seen in the context of binary data coming from
 
 There are a few ways to create new buffers:
 
-    var buffer = new Buffer(8);
+```js
+var buffer = Buffer.alloc(8);
+// This will print out 8 bytes of zero:
+// <Buffer 00 00 00 00 00 00 00 00>
+```
+This buffer is uninitialized and contains 8 bytes of zero, notice that `new Buffer(size)` is deprecated since v6.0.0.
```
Trott: The buffer is initialized with zeroes. If it were uninitialized, you could not predict what values it contains.

Trott: Maybe don't mention the constructor at all? It's deprecated in all currently supported versions of Node.js. Include prominent links to the Buffer docs in this document and people can find out about stuff like that there.

locale/en/knowledge/advanced/buffers/how-to-use-buffers.md
```diff
@@ -33,22 +33,32 @@ In the wild, buffers are usually seen in the context of binary data coming from
...
 
-    var buffer = new Buffer([ 8, 6, 7, 5, 3, 0, 9]);
+```js
+var buffer = Buffer.from([ 8, 6, 7, 5, 3, 0, 9]);
+// This will print out 8 bytes of certain values:
+// <Buffer 08 06 07 05 03 00 09>
+```
 
-This initializes the buffer to the contents of this array. Keep in mind that the contents of the array are integers representing bytes.
+This initializes the buffer to the contents of this array. Keep in mind that the contents of the array are integers representing bytes. Notice that `new Buffer(array)` is deprecated since v6.0.0.
```
Trott: Remove mention of the deprecated constructor?

locale/en/knowledge/advanced/buffers/how-to-use-buffers.md
```diff
@@ -33,22 +33,32 @@ In the wild, buffers are usually seen in the context of binary data coming from
...
 
-    var buffer = new Buffer("I'm a string!", "utf-8")
+```js
+var buffer = Buffer.from("I'm a string!", "utf-8");
+// This will print out a chain of values in utf-8:
+// <Buffer 49 27 6d 20 61 20 73 74 72 69 6e 67 21>
+```
 
-This initializes the buffer to a binary encoding of the first string as specified by the second argument (in this case, utf-8). **utf-8** is by far the most common encoding used with node, but `Buffer` also supports:
+This initializes the buffer to a binary encoding of the first string as specified by the second argument (in this case, utf-8). **utf-8** is by far the most common encoding used with node, but `Buffer` also supports some other formations, please see [Supported Encodings](https://nodejs.org/dist/latest/docs/api/buffer.html#buffer_buffers_and_character_encodings).
```
Trott:
```suggestion
This initializes the buffer to a binary encoding of the first string as specified by the second argument (in this case, `'utf-8'`). `'utf-8'` is by far the most common encoding used with Node.js, but `Buffer` also supports others. See [Supported Encodings](https://nodejs.org/dist/latest/docs/api/buffer.html#buffer_buffers_and_character_encodings) for more details.
```

locale/en/knowledge/advanced/buffers/how-to-use-buffers.md
```diff
@@ -33,22 +33,32 @@ In the wild, buffers are usually seen in the context of binary data coming from
...
 
-* **"ascii"**: This encoding is way fast, but is limited to the ascii character set. Moreover, it will convert null characters into spaces, unlike the utf-8 encoding.
-* **"ucs2"**: A two-byte, little-endian encoding. Can encode a subset of unicode.
-* **"base64"**: Base64 string encoding.
-* **"binary"**: This is the "binary string" format mentioned earlier, and is in the process of being deprecated. Avoid its use.
+Notice that `new Buffer(string[, encoding])` is deprecated since v6.0.0.
```
Trott: Remove mention of the deprecated Buffer constructor entirely?

Trott:
Left some comments/suggestions, but feel free to land this without addressing them. Getting rid of the Buffer constructor in the example code is more important than addressing my comments, which can be addressed later or not at all.

Trott:
Not sure if the YAML date should be changed after an update like this. Kind of wonder if we should either remove display of the date from the `.hbs` template or otherwise clarify if it's the original author date or the last modification date.

fhemberger:
@Trott The knowledge base isn't shown on the website at the moment, so I'd leave the date for now. +1 for removing it once the content is added to the new website.

Yash-Handa (approved):

locale/en/knowledge/advanced/buffers/how-to-use-buffers.md
ghost: I think it should be a series of torturials, so it's easy to understand than professional, compared with Buffer ref doc.
But I'll remove it because of duplication.

PR #2169, opened 2019-04-05
Update simple-profiling.md

Ivaylo-Kirov:
Fixed a typo.

PR #2170, opened 2019-04-06
Fix：Sync with the English version and change one statement

ghost:
1) Sync with the English version for the other untranslated versions (For more, please refer to https://github.com/nodejs/nodejs.org/pull/2169).

2) Now our Nodejs's version is much higher than v4.4.0. So it's not suitable to say that "recently introduced into Nodejs v4.4.0", but just change it to "since v4.4.0"……,ect instead.


Trott (approved):
English doc change looks good to me. Might leave some additional suggestions, but they are optional.

Maledong:
>English doc change looks good to me. Might leave some additional suggestions, but they are optional.

If possible, I'd like to hear from you :)

locale/en/docs/guides/simple-profiling.md
```diff
@@ -21,8 +21,9 @@ code-creation,Stub,2,0x2d5000a33e00,507,"NumberToStringStub"
 ```
 
 In the past, you needed the V8 source code to be able to interpret the ticks.
-Luckily, tools have recently been introduced into Node.js 4.4.0 that facilitate
-the consumption of this information without separately building V8 from source.
+Luckily, tools have been introduced into Node.js since v4.4.0 that
```
Trott: Before the release of Node.js 8.0.0, we decided that we would not use `v` when referring to Node.js versions. This was to avoid confusion as to whether `v8` referred to Node.js 8.0.0 or Google's V8 JavaScript engine. So, with that in mind, and also some rewriting for simplicity and active voice:

```suggestion
Node.js 4.4.0 introduced tools that
```

locale/en/docs/guides/simple-profiling.md
```diff
@@ -21,8 +21,9 @@ code-creation,Stub,2,0x2d5000a33e00,507,"NumberToStringStub"
 ```
 
 In the past, you needed the V8 source code to be able to interpret the ticks.
-Luckily, tools have recently been introduced into Node.js 4.4.0 that facilitate
-the consumption of this information without separately building V8 from source.
+Luckily, tools have been introduced into Node.js since v4.4.0 that 
+facilitate the consumption of this information without separately building
```
Trott:
```suggestion
ease the consumption of this information without building
```

ghost:
@Trott：What about changing to this:

- Luckily, tools have been introduced since Nodejs 4.4.0.

This means these tools are introduced since version of 4.4.0, and it can be also used for now. However if I take your advice "Node.js 4.4.0 introduced ……", this should be in a past tence and  I don't think it suitable in this senario.

Trott:
> @Trott：What about changing to this:

Sure, that works.

ghost:
@nodejs/nodejs-ko：Please update your translation if possible, thanks :)

locale/en/docs/guides/simple-profiling.md
sonicdoe: I’ve also read about this in the [Node.js v8.0.0 release post](https://nodejs.org/en/blog/release/v8.0.0/#long-term-support), however, all of the [recent release posts](https://nodejs.org/en/blog/) still use the `v` prefix in the title. Is the prefixless version still preferred?

yous:
@nodejs/nodejs-ko I've just pushed to follow changes of simple-profiling.md. Please review this.

taggon:
@Maledong @yous The Korean translation looks good to me. :)

outsideris:
Looks good in Korean.

locale/en/docs/guides/simple-profiling.md
Trott: Yeah, the blogs are awfully problematic in terms of following style. (Another thing they do is use "Node" where "Node.js" is preferred.) There's probably a boilerplate-generating tool that needs to be updated.

ghost:
Thanks all!

locale/en/docs/guides/simple-profiling.md
ghost: @Trott：Can you tell me where the tool is so as to let me update it when free? Thanks anyway!

PR #2176, opened 2019-04-09
AR:  Translate file - docs/guides/anatomy-of-an-http-transaction.md 

Aissaoui-Ahmed:
hello !
translate file `anatomy-of-an-http-transaction.md`
there are many different terms in domain Computer and Programing 
Languages, we trying to get exact meaning in Arabic.  I need help?

@celyes @InterstellarClub

### DIFF
--- a/locale/ar/docs/guides/anatomy-of-an-http-transaction.md
+++ b/locale/ar/docs/guides/anatomy-of-an-http-transaction.md
@@ -1,0 +1,409 @@
+---
+title: التركيبة البنيوية لمُعَامَلَة HTTP
+layout: docs.hbs
+---
+
+# التركيبة البنيوية لمُعَامَلَة HTTP
+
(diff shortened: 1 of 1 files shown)


ghost:
@nodejs/nodejs-ar：Would you mind helping or having a review of this? Thanks for your suggestions if possible.

ram-you:
@Maledong It looks like it's written by a robot, there are so many mistakes that we lose the meaning of the text.

celyes (changes requested):
This file may need a total re-write to make it more " human-readable".
@Aissaoui-Ahmed I suggest reading this web page: https://bit.ly/2MmPTDn as it has an Arabic translation for almost all of the CS Terms in general, including the most used one in the Node.js 's context

Aissaoui-Ahmed:
thank you all !

direction of changes is LTR to fix it copied in text editor and change direction RTL

celyes (approved):
LGTM

PR #2177, opened 2019-04-09
tools: update lodash dependency

Trott:
Update lodash so `npm audit` comes up clean. While we're at it, replace
`_.isNan()` with `Number.isNan()` which is all that lodash does under
the hood (if `Number.isNaN()` exists, which it does in all supported
versions of Node.js.


ZYSzys (approved):

richardlau (approved):

ghost:
Thanks Trott!

PR #2179, opened 2019-04-11
Remove security e-mail addresses

### LINKED ISSUE #123: Missing PGP key

reedloden:
The web interface is preferred for security (HTTP via TLS vs. possibly unencrypted e-mail),
and using the e-mail addresses requires that the web interface be used to finalize reports.

As such, just remove the security e-mail addresses entirely and only refer to the HackerOne
web interface.

Fixes: https://github.com/nodejs/security-wg/issues/123

### DIFF
--- a/locale/en/security.md
+++ b/locale/en/security.md
@@ -19,8 +19,7 @@ These updates will be sent at least every five days; in practice, this is more l
 ### Node.js Bug Bounty Program
 
 The Node.js project engages in an official bug bounty program for security researchers and responsible public disclosures.
-
-The program is managed through the HackerOne platform at <https://hackerone.com/nodejs> with further details.
+The program is managed through the HackerOne platform. See <https://hackerone.com/nodejs> for further details.
 
 ## Reporting a Bug in a third party module
 
--- a/locale/en/security.md
+++ b/locale/en/security.md
@@ -20,7 +20,7 @@ These updates will be sent at least every five days; in practice, this is more l
 
 The Node.js project engages in an official bug bounty program for security researchers and responsible public disclosures.
 
-The program is managed through the HackerOne platform at [https://hackerone.com/nodejs](https://hackerone.com/nodejs) with further details.
+The program is managed through the HackerOne platform at <https://hackerone.com/nodejs> with further details.
 
 ## Reporting a Bug in a third party module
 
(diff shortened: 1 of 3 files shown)


Trott:
@nodejs/security-wg

reedloden:
See also https://github.com/nodejs/security-wg/issues/454#issuecomment-481919759 and https://github.com/envoyproxy/envoy/issues/5155#issuecomment-481920455

e-jigsaw:
Added ja version https://github.com/e-jigsaw/new.nodejs.org/commit/97a6332fe30e3a03822e01645a2d3850aaaa206d

lirantal:
While I would like to provide more alternatives for researchers to report issues, as it currently stands, even if you send an email you actually need to follow-up with a login to H1 to submit the report, which is why I agree with this PR clearly promoting the H1 program instead of the e-mail which may provide less than obvious user experience.

sam-github:
Seems reasonable to me, @nodejs/security , any objections?

rvagg:
I guess not given the current situation; although I'm not thrilled about how we're forcing reporters to funnel everything through a platform they may not want to interact with.

If I look up a store in my local area and they insist on making me use Facebook to see further information about them, most of the time I'll just move on. If the Linux Foundation helpdesk wants me to use JIRA to interact with them then I'll choose not to interact with them. If I need to use Bugzilla to file a bug report on your project then I'll _nope_ on out of there. Where the transaction cost is higher than the care-factor, reporters will just opt out.

mcollina:
Considering the problems we had in the past, I think this is ok.

yous:
@nodejs/nodejs-ko Review please. Just updated to follow changes of security.md.

marocchino:
@yous LGTM

locale/en/security.md
```diff
@@ -21,13 +20,12 @@ These updates will be sent at least every five days; in practice, this is more l
 
 The Node.js project engages in an official bug bounty program for security researchers and responsible public disclosures.
 
-The program is managed through the HackerOne platform at [https://hackerone.com/nodejs](https://hackerone.com/nodejs) with further details.
+The program is managed through the HackerOne platform at <https://hackerone.com/nodejs> with further details.
```
sam-github: This text was pre-existing, but it didn't make any grammatical sense as-is. I think line 22 (blank) should be removed so this sentence becomes part of the preceeding paragraph, and should be rewritten as:

> The program is managed through the HackerOne platform, see <https://hackerone.com/nodejs> for further details.

sam-github (approved):

locale/en/security.md
reedloden: I split that into two sentences to be grammatically-correct, but good catch.

sam-github (approved):

PR #2185, opened 2019-04-13
chore: Use ES6 to write functions

ghost:
~~The 'json.js' file isn't referred, so remove it.~~

Use ES6 to write functions for 'm.read' and 'Object.keys.forEach' in 'build.js'.

### DIFF
--- a/build.js
+++ b/build.js
@@ -178,13 +178,13 @@ function withPreserveLocale (preserveLocale) {
   return (files, m, next) => {
     if (preserveLocale) {
       var path = m.path('locale/en')
-      m.read(path, function (err, newfiles) {
+      m.read(path, (err, newfiles) => {
         if (err) {
           console.error(err)
           return next(err)
         }
 
-        Object.keys(newfiles).forEach(function (key) {
+        Object.keys(newfiles).forEach((key) => {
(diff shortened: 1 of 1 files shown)


ZYSzys (commented):
Not sure exactly, maybe it's used to replace this 
https://github.com/nodejs/nodejs.org/commit/bee3f0da3c8cb2e31bd50c2cb20bdceac991fc7c#diff-efd59cd1bdc5c9ac6d0eaa368f1e149fL164 ?

fhemberger:
@ZYSzys Yes, you're right. But I can't remember where we needed the `json` Markdown helper in the first place … 🤔

ghost:
So this file shouldn't be removed? Well……I didn't find anywhere to refer…… XD

richardlau:
> 
> 
> So this file shouldn't be removed? Well……I didn't find anywhere to refer…… XD

It's indirectly referenced: 
https://github.com/nodejs/nodejs.org/blob/2ca6decd1312eb2a4b99e701bf328747ffbed082/build.js#L158-L161

ghost:
Thanks all! I've changed the topic to another submit :)

ZYSzys (approved):

PR #2186, opened 2019-04-13
Remove useless variable

arashbahreini:


fhemberger:
Function expects two parameters, with "res" being the second. Changing the number (and thus the order) of parameters breaks the function.

arashbahreini:
@fhemberger Yes you are right, I found it out and closed the pull request. 
Thanks

Issue #2204, opened 2019-04-24
ARMv6 for 12.0.0 missing


mlibby:
https://nodejs.org/en/download/current/

Current Tab selected
Linux Binaries (ARM) row - ARMv6 link (https://nodejs.org/dist/v12.0.0/node-v12.0.0-linux-armv6l.tar.xz)

Expected tarball, got 404.

fhemberger:
Hi, sometimes builds for certain platforms can take some time. To be sure, better have a look at the release notes: https://nodejs.org/en/blog/release/v12.0.0/

> ARMv6 32-bit Binary: Coming soon
> ARMv7 32-bit Binary: https://nodejs.org/dist/v12.0.0/node-v12.0.0-linux-armv7l.tar.xz

The links on the download page are auto generated from file names and the current version number, so an occasional 404 can happen. Sorry for that!

richardlau:
See https://github.com/nodejs/build/issues/1677#issuecomment-486204349.

mlibby:
Thanks to both of you. Very informative.

PR #2205, opened 2019-04-24
Remove ARMv6 from releases starting v12.0.0

fhemberger:
Ref:
- #2204
- https://github.com/nodejs/build/issues/1677#issuecomment-486204349

### DIFF
--- a/scripts/helpers/downloads.js
+++ b/scripts/helpers/downloads.js
@@ -2,7 +2,7 @@
 
 const semver = require('semver')
 
-const postMergeDownloads = [
+const allDownloads = [
   {
     'title': 'Windows 32-bit Installer',
     'templateUrl': 'https://nodejs.org/dist/v%version%/node-v%version%-x86.msi'
@@ -77,9 +77,6 @@ const postMergeDownloads = [
   }
 ]
 
-// Latest releases (v8.x and above) don't offer 'Linux PPC BE 64-bit Binary' any longer
-const latestDownloads = postMergeDownloads.filter(download => download.title !== 'Linux PPC BE 64-bit Binary')
-
 // v0.x of Node.js
 const legacyDownloads = [
   {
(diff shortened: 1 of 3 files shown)


scripts/helpers/downloads.js
```diff
@@ -147,6 +147,8 @@ module.exports = (version) => {
     downloads = downloads.filter(ver =>
       ver.title !== 'Linux 32-bit Binary' &&
       ver.title !== 'SmartOS 32-bit Binary')
+  } else if (semver.satisfies(version, '>= 12.0.0')) {
+    downloads = downloads.filter(ver => ver.title !== 'ARMv6 32-bit Binary')
```
richardlau: The filter for `>= 12.0.0` also needs to include the items filtered for `>= 10.0.0`.

richardlau (changes requested):

scripts/helpers/downloads.js
fhemberger: @richardlau Thanks, fixed.

richardlau (approved):

refack (approved):

scripts/helpers/downloads.js
```diff
@@ -147,6 +147,11 @@ module.exports = (version) => {
     downloads = downloads.filter(ver =>
       ver.title !== 'Linux 32-bit Binary' &&
       ver.title !== 'SmartOS 32-bit Binary')
+  } else if (semver.satisfies(version, '>= 12.0.0')) {
```
ZYSzys: IIUC, this condition seems unreachable since it would be caught by `semver.satisfies(version, '>= 10.0.0')` in the line 146:
https://github.com/nodejs/nodejs.org/blob/8216eeb1311c5a7056fb17b47c12113cf81b06d8/scripts/helpers/downloads.js#L146

mhdawson (approved):
LGTM

BethGriggs (approved):

scripts/helpers/downloads.js
fhemberger: You are right, of course. Stupid copy & paste mistake.
I refactored the download helper, PTAL.

rvagg (approved):

PR #2209, opened 2019-04-27
Add link to the callback pattern page

nickbouldien:
Added a link to the callback pattern page in the knowledge/getting-started section of the node docs.  

As I was reading the blocking/non-blocking documentation, I realized that some people might not know what the node callback pattern was and figured a link could be added to that page for easy access to readers.

(I did notice that the page I added a link to is from 2011, but I couldn't find a more updated one when searching (so not sure if a more updated entry on explaining callbacks exists))


fhemberger (approved):

fhemberger:
Good call, thanks!

Issue #2210, opened 2019-04-28
About "TL;DR" in the article as the title


ghost:
Just a question about this fixed phrase at (https://github.com/nodejs/nodejs.org/blob/master/locale/en/docs/guides/dont-block-the-event-loop.md#tl-dr), because this phrase explains the process of Nodejs in the event and how to improve the speed of Nodejs' work.

I don't know why "TL;DR" for this phrase?  Why does it mean "Too Long Didn't Read"？

So IMO the title should be something like **"What to notice for improvement of efficiency for Nodejs in the event loop?"**

ghost:
@nodejs/website：What do you think of this？

davisjam:
An alternative title might be "Executive summary" or "Abstract" or similar.

Trott:
Yeah, it basically means "Summary".

PR #2212, opened 2019-04-29
chore: Change the title from 'TL;DR' to 'Summary'

ghost:
In order to give a good understanding of 'TL;DR', change it to 'Summary' to make it more proper.
Ref: https://github.com/nodejs/nodejs.org/issues/2210.


davisjam:
Do you have a reason to think that "Abstract" is more understandable than "TL;DR" to the average reader?

stevemao:
I support this change. It makes the title more formal.

ghost:
@davisjam：I don't think 'TL;DR' is more formal than 'Abstract', if in a scientific article.

On the other hand, 'TL;DR' may have two meanings:

1) Too Long Don't Read.
2) Too Long Didn't Read.

This will make it too puzzled, so to make it "Abstract" just as your suggestion to make it more clear to understand (https://github.com/nodejs/nodejs.org/issues/2210#issuecomment-487388123). I think yours is better to understand.

zeke:
👍 I'm in favor of this change. TLDR is widely known among English-speaking developers, but it's not very accessible to international audiences.

I think "Abstract" is an improvement on TLDR, but "Abstract" is more likely to be recognizable to people with academic backgrounds, and could be confusing to people who don't read research papers.

I would suggest "Summary" as a more universally accessible term.

davisjam:
I like "Summary" more as well, for the reasons @zeke outlined.

ghost:
@davisjam：OK. Follow your suggestion :)

zeke (approved):
💯

davisjam (approved):
LGTM

ZYSzys (approved):

PR #2232, opened 2019-05-13
Improving SEO by addition of Meta keywords

shivaylamba:
Added meta keywords for Better SEO implementation


fhemberger:
Meta keywords don't have an impact on SEO any more nowadays. Also the selection of keywords do not really add anything new. So I'm -1 on it in its current state.

willin (commented):
also, each page should have its own keywords

PR #2233, opened 2019-05-15
Clarify release stages on "releases" page

garybernhardt:
My team recently did our first major Node version upgrade and had a bad experience. The short version is: we tried to upgrade to Node 12 (which is in "Current" status as of today, not "Active"). We couldn't get one of our dependencies to build and ended the workday without successfully upgrading.

I'd seen the [releases](https://nodejs.org/en/about/releases/) page multiple times, including during that attempted upgrade. However, that page doesn't explain what the various stages actually mean, and which ones are recommended for production use. Specifically, I didn't understand "Current" vs. "Active", and incorrectly assumed that either would be acceptable for production use, because neither of them was called "unstable", "testing", "alpha", "beta", "prerelease", or other terms that I'm familiar with for a "don't use this in production app yet!" release.

This change adds a paragraph with a brief explanation of the different statuses. I also included a bit about LTS, because I imagine a lot of people seeing this page won't be familiar with that term.

I think that a short explanation on the releases page will save a lot of other teams from this kind of multi-hour upgrade frustration. I've never contributed to this repo, so I may have put this text in the wrong file, and my explanation of the Node release statuses may be wrong in ways I don't understand yet. But I wanted to suggest a concrete change that would've helped us to avoid wasting half of a day.

(My understanding is based on this @jasnell tweet: https://twitter.com/jasnell/status/1128696625986015232.)

### DIFF
--- a/locale/en/about/releases.md
+++ b/locale/en/about/releases.md
@@ -18,3 +18,8 @@ schedule-footer: Dates are subject to change.
 ---
 
 # Releases
+
+Major Node.js versions enter _Current_ release status for six months, which gives library authors time to add support for them.
+After six months, odd-numbered releases (9, 11, etc.) become unsupported, and even-numbered releases (10, 12, etc.) move to _Active LTS_ status and are ready for general use.
+_LTS_ release status is "long-term support", which typically guarantees that critical bugs will be fixed for a total of 30 months.
+Production applications should only use _Active LTS_ or _Maintenance LTS_ releases.


ZainW:
This is a very nice addition to the docs, would help alot with people starting out and overall frustration when it comes to node versioning.

layouts/about-release-schedule.hbs
```diff
@@ -19,6 +19,13 @@
 
         {{{ release-schedule }}}
 
+        <p>
+          Major Node versions enter "Current" release status for six months, which gives library authors time to add support for them.
+          After six months, releases move to "Active" status and are ready for general use.
+          "LTS" release status is "long-term support", and guarantees that critical bugs will be fixed for a total of 30 months.
+          Production applications should only use "Active" or "LTS" releases.
```
dominykas:
```suggestion
          Production applications should only use "Active LTS" or "Maintenance LTS" releases.
```

layouts/about-release-schedule.hbs
```diff
@@ -19,6 +19,13 @@
 
         {{{ release-schedule }}}
 
+        <p>
+          Major Node versions enter "Current" release status for six months, which gives library authors time to add support for them.
+          After six months, releases move to "Active" status and are ready for general use.
```
dominykas: This is not exactly true. After six months the release will either be unsupported (odd numbers) or "Active LTS" (even numbers).

layouts/about-release-schedule.hbs
garybernhardt: Amended this in as 8aad80f5, thanks!

layouts/about-release-schedule.hbs
garybernhardt: Amended this in in 8aad80f5, thanks!

layouts/about-release-schedule.hbs
```diff
@@ -19,6 +19,13 @@
 
         {{{ release-schedule }}}
 
+        <p>
+          Major Node versions enter "Current" release status for six months, which gives library authors time to add support for them.
+          After six months, odd-numbered releases (9, 11, etc.) become unsupported, and even-numbered releases (10, 12, etc.) move to "Active LTS" status and are ready for general use.
+          "LTS" release status is "long-term support", and guarantees that critical bugs will be fixed for a total of 30 months.
```
mgol: This isn't always 30 months, I think? Some versions have shorter lifecycles, often due to EoL of an included OpenSSL version.

garybernhardt: I based this on this:

> Every major version covered by the LTS plan will be actively maintained for a period of 18 months from the date it enters LTS coverage. Following those 18 months of active support, the major version will transition into "maintenance" mode for 12 additional months.

which I found here: https://github.com/nodejs/Release#release-plan

fhemberger (changes requested):
Please don't put content directly into the **templates**, we need to be able to translate all the texts. Please use `locale/en/about/releases.md` instead.

layouts/about-release-schedule.hbs
```diff
@@ -19,6 +19,13 @@
 
         {{{ release-schedule }}}
 
+        <p>
+          Major Node versions enter "Current" release status for six months, which gives library authors time to add support for them.
```
Trott:
```suggestion
          Major Node.js versions enter "Current" release status for six months, which gives library authors time to add support for them.
```

layouts/about-release-schedule.hbs
Trott: Perhaps the solution is to add a "usually" or a "typically" in there. (Same on the source document.)

Trott:
Sorry to hear that the release info wasn't clear and caused issues for your team. Help on making that information more available and more clear is certainly appreciated.

layouts/about-release-schedule.hbs
garybernhardt: Thanks, I did exactly that.

layouts/about-release-schedule.hbs
garybernhardt: 👍

garybernhardt:
Thanks, @mgol, @fhemberger, and @Trott. I replaced "Node" with "Node.js", added the qualification that releases are "typically" LTS for 30 months, and moved the paragraph from the template to the localized md file. That makes it appear at the top of the page rather than the bottom, but that doesn't seem like a problem.

locale/en/about/releases.md
```diff
@@ -18,3 +18,11 @@ schedule-footer: Dates are subject to change.
 ---
 
 # Releases
+
+<p>
```
fhemberger: No need for the `<p>…</p>`, Handlebars adds paragraphs on its own. Otherwise LGTM.

fhemberger (changes requested):

locale/en/about/releases.md
garybernhardt: Thanks; just fixed that and force-pushed.

locale/en/about/releases.md
```diff
@@ -18,3 +18,8 @@ schedule-footer: Dates are subject to change.
 ---
 
 # Releases
+
+Major Node.js versions enter "Current" release status for six months, which gives library authors time to add support for them.
+After six months, odd-numbered releases (9, 11, etc.) become unsupported, and even-numbered releases (10, 12, etc.) move to "Active LTS" status and are ready for general use.
```
Trott:
```suggestion
After six months, odd-numbered releases (9, 11, etc.) become unsupported, and even-numbered releases (10, 12, etc.) move to _Active LTS_ status and are ready for general use.
```

locale/en/about/releases.md
```diff
@@ -18,3 +18,8 @@ schedule-footer: Dates are subject to change.
...
+"LTS" release status is "long-term support", which typically guarantees that critical bugs will be fixed for a total of 30 months.
```
Trott:
```suggestion
_LTS_ release status is "long-term support", which typically guarantees that critical bugs will be fixed for a total of 30 months.
```

locale/en/about/releases.md
garybernhardt: fixed in amend faff6881, thanks

locale/en/about/releases.md
garybernhardt: fixed in amend faff6881, thanks

fhemberger (approved):

fhemberger:
Thank you!

garybernhardt:
Thanks for merging! 👍

PR #2276, opened 2019-06-20
fix the Chinese translation error

Clark-G:
fix the Chinese translation error of the first sentence  in ES6 and beyond page


ZYSzys (approved):

ZYSzys:
@Clark-G Thanks for your first contribution to nodejs org.🎉

PR #2292, opened 2019-06-24
docs: fix typo on knowledge/javascript-conventions section

ftonato:


Trott:
The change here is fine, but I don't think we should host fundamental JavaScript info and should instead link to resources like MDN. I think files like this should be deleted and possibly replaced with links to MDN.

ftonato:
I really agree, but I'm just following the issue - https://github.com/nodejs/nodejs-pt/issues/80 :embarrassed-emoji:

Trott:
If you're looking for other corrections...

* The very first sentence of the document here is missing a word. (Missing the article "a".)
* The very first code example contains an error. (The comment says `1` is converted to a string, but that's wrong. In that piece of could, `"1"` is converted to a number.)

Missing punctuation (usually commas), wrong capitalization for _JavaScript_...

Of course, like I said, I'm for deleting the doc so if you don't want to put much time into it, I certainly understand that.

Trott:
Oooh, another argument for deleting unnecessary docs that I never considered: Reduces the translation burden!

ftonato:
Like I said, I agree with you, it makes a lot more sense not to keep basic things that can be easily clarified with some MDN links.

I voted to delete this file: +1:

Trott (approved):

ghost:
@ftonato, @Trott：
I've submitted another thing to remove this file according to the latest discussion with you. Please have a review to confirm :)

ftonato:
@MaledongGit I can't approve for not being part of the organization, but you've my written approval :laughing:

PR #2293, opened 2019-06-25
Update Dockerizing guide to adhere to best practice

kaimallea:
Hey y'all,

#### Longer version:

The current [Dockerizing a Node.js web app](https://nodejs.org/en/docs/guides/nodejs-docker-webapp/) guide links to the official [Node.js Docker Best Practices Guide](https://github.com/nodejs/docker-node/blob/master/docs/BestPractices.md) that [currently recommends](https://github.com/nodejs/docker-node/blob/master/docs/BestPractices.md#cmd) avoiding the use of `npm start` for two reasons:

1. Reduces the number of processes running inside of your container.
2. Enable exit signals such as `SIGTERM` and `SIGINT` to be received by the Node.js process instead of npm swallowing them.

There are currently [~54,000 public `Dockerfile`s on GitHub](https://github.com/search?q=%22CMD+%5B+%5C%22npm%5C%22%2C+%5C%22start%5C%22+%5D%22+filename%3ADockerfile&type=Code) not following this practice.

This PR updates the former guide to adhere to latter guide's best practice.

Additionally, the guide references "latest LTS" that points to `8`, this updates that to `10`

#### Additional Question:

How should I handle translations? The translation guide in this repo doesn't seem to apply to the guides pages, so I didn't update the corresponding paragraph of text for the locales that weren't already english.

### DIFF
--- a/locale/en/docs/guides/nodejs-docker-webapp.md
+++ b/locale/en/docs/guides/nodejs-docker-webapp.md
@@ -80,11 +80,11 @@ touch Dockerfile
 Open the `Dockerfile` in your favorite text editor
 
 The first thing we need to do is define from what image we want to build from.
-Here we will use the latest LTS (long term support) version `8` of `node`
+Here we will use the latest LTS (long term support) version `10` of `node`
 available from the [Docker Hub](https://hub.docker.com/):
 
 ```docker
-FROM node:8
+FROM node:10
 ```
 
 Next we create a directory to hold the application code inside the image, this
@@ -144,7 +144,7 @@ CMD [ "npm", "start" ]
 Your `Dockerfile` should now look like this:
 
 ```docker
-FROM node:8
+FROM node:10
 
 # Create app directory
 WORKDIR /usr/src/app
@@ -195,7 +195,7 @@ $ docker images
 
 # Example
 REPOSITORY                      TAG        ID              CREATED
-node                            8          1934b0b038d1    5 days ago
+node                            10         1934b0b038d1    5 days ago
 <your username>/node-web-app    latest     d64d3505b0d2    1 minute ago
 ```
 
--- a/locale/en/docs/guides/nodejs-docker-webapp.md
+++ b/locale/en/docs/guides/nodejs-docker-webapp.md
@@ -134,11 +134,10 @@ EXPOSE 8080
 ```
 
 Last but not least, define the command to run your app using `CMD` which defines
-your runtime. Here we will use the basic `npm start` which will run
-`node server.js` to start your server:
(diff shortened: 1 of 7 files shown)


Trott (approved):

locale/en/docs/guides/nodejs-docker-webapp.md
```diff
@@ -134,17 +134,16 @@ EXPOSE 8080
 ```
 
 Last but not least, define the command to run your app using `CMD` which defines
-your runtime. Here we will use the basic `npm start` which will run
-`node server.js` to start your server:
+your runtime. Here we will use `node server.js` to start your server:
 
 ```docker
-CMD [ "npm", "start" ]
+CMD [ "node", "server.js" ]
```
LaurentGoderre, line 140: Nitpicking perhaps but maybe this should be `script.js` or `index.js` since not every node app is a server?

kaimallea: Hi Laurent, thanks for the feedback. It should stay as-is because the guide, as it is written today, is a step-by-step guide that explicitly sets up a server and instructs the reader to create `server.js`. This PR is simply updating the pattern in the code without changing the scope of guide itself. Here's a [link to the guide for reference](https://nodejs.org/en/docs/guides/nodejs-docker-webapp/).

locale/en/docs/guides/nodejs-docker-webapp.md
```diff
@@ -80,11 +80,11 @@ touch Dockerfile
 Open the `Dockerfile` in your favorite text editor
 
 The first thing we need to do is define from what image we want to build from.
-Here we will use the latest LTS (long term support) version `8` of `node`
+Here we will use the latest LTS (long term support) version `10` of `node`
 available from the [Docker Hub](https://hub.docker.com/):
 
 ```docker
-FROM node:8
+FROM node:10
```
SimenB, line 87: this _could_ be `node:lts` instead of hard coding a version number, not sure if it should be recommended or not, though

chorrell: No it should not. It's problematic in the same way `latest` is.

kaimallea:
Is this good to go? 🚀

LaurentGoderre (approved):

locale/en/docs/guides/nodejs-docker-webapp.md
mkrawczuk: We're currently having this discussion in PR #2953. Can you elaborate on this? I clearly see there might be issues using the `latest` tag, but I don't understand in what way this might be problematic in the same way `latest` would be.

PR #2294, opened 2019-06-25
Language picker implemented

alexandrtovmach:
Implemented simple language picker #806, to make better UX on the website

How it looks you can check here:
https://www.loom.com/share/9022dba48c1d42a486df212e4acabbdf


ZYSzys:
@alexandrtovmach Thanks, this is great!👍

A little suggestion: can we avoid always redirecting to the home page when we change the default language ?

alexandrtovmach:
@ZYSzys it's kind of "overcomplexity".

Of course, I can write this, but this website very simple, and trying to avoid scripts as much as possible, so I decided to go with the same way.

alexandrtovmach:
@ZYSzys added:
https://www.loom.com/share/1cff103b50b14f0ba7d5c586f2d963f5

ZYSzys (approved):
Cool, LGTM👍

zeke:
Looking good!

Drive-by comment: is it possible to also display the English name of each language?

For example: `简体中文 (Simplified Chinese)`

alexandrtovmach:
@zeke Thanks for proposition. Do you think that's good idea? I'm agreed with [this comment](https://github.com/nodejs/nodejs.org/issues/806#issuecomment-505086687), because when user looking for his language he don't need English version. In the same time if we'd add English name, it will make string longer.

I think optimal way is to use `title` attribute, to show English name on hover. For example:
image

zeke:
> use title attribute, to show English name on hover.

👍

alexandrtovmach:
@zeke added. Thanks for help

ghost:
@nodejs/website-redesign：This is due to a web design function, so any other ideas about that? I agree with this :)

keywordnew:
This is looking good to me! Thanks for the PR @alexandrtovmach 💯

alexandrtovmach:
Thanks everyone for a review 🤗.

@MaledongGit What do you think about merge?

srl295:
I missed the original PR. 
- make sure `en` is a variable and not assumed
- you can get localized names for languages from CLDR, loadable from JSON via https://www.npmjs.com/package/cldr-data

PR #2283, opened 2019-06-22
update: Remove node-speakers

Trott:
The node-speakers page is problematic for several reasons. Remove it.

PR #2297, opened 2019-06-27
Added back speakers file..

ChellaTheCoder:
Why it was removed? if you have any issues with other languages please remove that specific language file. Let me know if you have any concerns.

### DIFF
--- a/locale/en/get-involved/node-speakers.md
+++ b/locale/en/get-involved/node-speakers.md
@@ -1,0 +1,226 @@
+---
+title: Node.js スピーカー
+layout: contribute.hbs
+---
+
+# Node.js Speakers
+
+This is a list of local Node.js speakers. Please submit a PR if you'd like to add yourself or add others! Make sure you have permission before doing so as well.
+
+## Code of Conduct
+
+If anyone is reported as malicious or making others feel uncomfortable, they may be removed from this list. 
+
+## Instructions
+
+- Submit everything alphabetically
+
+## Speakers
+
+### Brazil 🇧🇷
+
+#### Minas Gerais
+
+##### Belo Horizonte
+
+###### Hugo Iuri Cordeiro
+
+- [GitHub](https://github.com/hugoiuri)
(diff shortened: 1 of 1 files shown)


ghost:
Please see：https://github.com/nodejs/nodejs.org/pull/2283

ChellaTheCoder:
@MaledongGit Yes, I agree but we should have a process for that, how come it directly removes from source code? actual speakers will go way.

ChellaTheCoder:
@MaledongGit Please add it, back

Trott (changes requested):
Reasons for removing it were clearly stated in the PR. Several folks on the Website WG agreed. No one on the Website WG opposed it. That is our process. I'm opposed to adding it back without a compelling justification for why this content needs to be on the Node.js website and how its quality will be maintained.

ChellaTheCoder:
@Trott Agree, so what's the accuracy level for meetups? and events? did anyone is looking at their accuracy and trust? If they are not, remove them also.

Trott:
> Agree

Great. I'll close this.

>  so what's the accuracy level for meetups? and events? did anyone is looking at their accuracy and trust? If they are not, remove them also.

I agree that those should be removed as well (along with a lot of other content on the site), but the case there is not as obvious or as strong, and I expect consensus on the Website WG about that to be harder to come by. 

"Events" is probably the easier of the two as it is obviously under-maintained. It currently has one past event, two upcoming events, and one event that isn't actually happening this year. And that's it. It does not even have the Foundation's own conference (in December) listed.

"Meetups" is a bit more defensible, as one can click through in most cases and see if the Meetup is still active and so on. The website would still be better without that content in my opinion for a few reasons, but my view of what is appropriate and best for the website is probably a bit more narrow than that of the typical Website WG member.

ChellaTheCoder:
Can we show speakers list in some other areas ? My ask is why it’s suddenly removed now , we can remove all meet-up and event pages at time?

PR #2301, opened 2019-06-28
update: Remove node-events page

Trott:
The page is undermaintained and not very visible anyway. It currently
contains one past event, two upcoming events, and one event that is not
actually happening in 2019. It does not containt the Foundation's own
December conference. We could update it now, but it will simply
immediately go back to being neglected. This is not great content for
the website anyway, in my opinion. If we can get the Get Involved
section slimmed down a fair amount, it might make sense to include a
link or two to external well-maintained sources of information, whether
it be a community-maintained page somewhere or a search on Lanyrd or
similar service. But for now, let's remove this largely unused and
unmaintained page.


ghost:
I agree, the file is frequently updated and changed, maybe we need something outside the website, and then refer the link....

ghost (approved):

ghost:
@fhemberger Any other clues or ideas?

fhemberger:
I think you made some valid points. Do what you think is best. 😉

Maybe someone else from the website team has some input. If there are objections, discuss the pros and cons. If there's no feedback after 1-2 days, proceed.

alexandrtovmach:
We can make some public Google Calendar, and provide link to it on the website:
Advantages:
- easy to manage
- no more frequent updates in this repository
- we also can try to use Calendar API to show events near current date on the website

Disadvantages:
- require the Google account to create and manage this calendar
- require Google accounts to make request to add/modify event 
- localization: we should use universal (English) language to describe event details

PR #2319, opened 2019-07-04
RU : Getting Started Guide

unstoo:
A human translation of 'Getting started' article to Russian.

PR #2321, opened 2019-07-07
Ru: Update getting-started-guide.md

pwrshi:


alexandrtovmach:
@stasonuke thanks for contributing, good translation but it's already in progress here #2319 Maybe you can take a look to other docs?

PR #2342, opened 2019-07-18
AR : Fix translate 'hello world' & 'server running at' ....

3imed-jaberi:
1.  `hello world` ===> `أهلا بالعالم`
2.  `server running at ...` ===> `الخادوم يعمل في ...`


celyes (changes requested):
what's inside the code should be kept as it is and not translated.

3imed-jaberi:
> what's inside the code should be kept as it is and not translated.

@celyes ,I understand your words well and thank you for your review..

I have translated words that do not affect the code but they affect the view ..
I mean the shift from English content to Arabic content ..

PR #2353, opened 2019-07-23
Blog: add #28630 to "notable" section of v12.7.0 release

boneskull:
@targos I don't know if it's typical to update this after-the-fact, but here it is in case you want to merge it.  I think it's worth drawing attention to, anyway, because it will break people using the experimental "reports" API.


targos:
Could you revert the other changes? LGTM otherwise

boneskull:
oops. blame prettier

boneskull:
@targos this is updated

Trott:
LGTM but maybe include what it used to be? It's returning an object instead of....

Trott (approved):

targos:
I updated the release entry: https://github.com/nodejs/node/releases/tag/v12.7.0
I modified the text to "Modify process.report.getReport() to return an Object instead of a JSON string"

boneskull:
LGTM

lpinca (approved):

PR #2354, opened 2019-07-25
chore: Add Markdownlint

nschonni:
- Add [Markdownlint](https://github.com/DavidAnson/markdownlint/blob/master/doc/Rules.md) checking for Markdown files
- Add .markdownlint.json to disable all non-passing rules to start
- Add markdownlint run to the test:lint target


.markdownlint.json
```diff
@@ -0,0 +1,36 @@
+{
+  "default": true,
+  "MD001": false,
```
lpinca, line 3: I wonder if it makes sense to use yaml and add comments to explain what the rules do.

lpinca (approved):

ghost (approved):

.markdownlint.json
nschonni: I don't think YAML is supported in the Node version, but they do have "aliases" for the rules that are more readable. I just stuck with the rule names for the initial pass to disable the failing ones, but if you want to do a styleguide, the aliases are better

PR #2360, opened 2019-08-05
fix: MD005/list-indent

nschonni:
Inconsistent indentation for list items at the same level

PR #2362, opened 2019-08-05
fix: MD038/no-space-in-code

nschonni:
Spaces inside code span elements


locale/en/blog/release/v4.5.0.md
```diff
@@ -121,7 +121,7 @@ Semver Patch:
 * [[`b08ff33c01`](https://github.com/nodejs/node/commit/b08ff33c01)] - **doc**: add `added:` information for https (Anna Henningsen) [#7392](https://github.com/nodejs/node/pull/7392)
 * [[`1d99059bb1`](https://github.com/nodejs/node/commit/1d99059bb1)] - **doc**: add `added:` information for dns (Julian Duque) [#7021](https://github.com/nodejs/node/pull/7021)
 * [[`a0ca24b798`](https://github.com/nodejs/node/commit/a0ca24b798)] - **doc**: add `added:` information for console (Adrian Estrada) [#6995](https://github.com/nodejs/node/pull/6995)
-* [[`eb08c17a20`](https://github.com/nodejs/node/commit/eb08c17a20)] - **doc**: add `added: ` data for cli.md (Rich Trott) [#6960](https://github.com/nodejs/node/pull/6960)
+* [[`eb08c17a20`](https://github.com/nodejs/node/commit/eb08c17a20)] - **doc**: add `added:` data for cli.md (Rich Trott) [#6960](https://github.com/nodejs/node/pull/6960)
```
Trott: Maybe it's best to omit `blog/release/*` from the linting. The text being changed here is directly from the commit log. Even if there are typographical errors in the commit log, I think it's best that the text match.

(Alternatively, maybe we can add an HTML comment that instructs the linter to skip this line? No idea if the tool we're using permits that but remark-lint does, so maybe this one does too?)

nschonni: It can be ignored from the command line glob, but that would still show up in IDEs. A disable command could be added to the files/template to disable it in the files by adding `<!-- markdownlint-disable-->` or for specific rules

Trott: Just to get this PR landed, can you add a comment to disable it for these lines in the changelog rather than altering the lines?

nschonni: OK, I guess a different approach might be needed for #2360

nschonni: I've disabled it at the file level

nschonni:
Arg, GitHub decided to close the PR on a force push. Looks like a new service glitch/bug, so I'll need to resubmit

PR #2378, opened 2019-08-10
Add "server-side"

Lotharyx:
Complete Node.js neophytes (like myself) might not immediately understand that Node.js is a server-side tool, since the lion's share of JavaScript uses are client-side (i.e. browser-based).  Added "server-side" to the initial sentence to clear that possible misunderstanding up right away.

### DIFF
--- a/locale/en/about/index.md
+++ b/locale/en/about/index.md
@@ -5,10 +5,11 @@ trademark: Trademark
 ---
 # About Node.js&reg;
 
-As an asynchronous event driven JavaScript runtime, Node is designed to build
-scalable network applications. In the following "hello world" example, many
-connections can be handled concurrently. Upon each connection the callback is
-fired, but if there is no work to be done, Node will sleep.
+As an asynchronous event driven JavaScript server-side runtime, Node is 
+designed to build scalable network applications. In the following "hello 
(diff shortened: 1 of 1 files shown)


Trott:
I see how it would help to clarify that it's outside the browser. But it's not just server-side. It's pretty much anywhere outside the browser. Electron apps, command-line tools, etc.

That's why it's called a "JavaScript runtime". But while that technically means that it's a standalone thing that runs JavaScript without needing a browser, that implication is awfully easy to overlook.

Not sure what the best solution is here. Maybe a separate sentence explaining something like "Because Node.js is a JavaScript *runtime*, it does not run inside a web browser. It can run on a server, or embedded in desktop application, or as a script interpreter invoked at the command line." That sentence will need some editing, but something like that?

MylesBorins (changes requested):
I too am -1 on this change. This description is used fairly often to describe Node.js and I don't think adding "serverside" is necessary or worth the amount of ecosystem churn we would see in the description.

Trott:
I'm going to close this, but feel free to re-open or comment if you intend to work on this further. (And anyone else looking to clarify things can certainly also open a different PR.)

PR #2379, opened 2019-08-10
fix: MD014/commands-show-output

nschonni:
Dollar signs used before commands without showing output

### DIFF
--- a/locale/fa/docs/guides/backpressuring-in-streams.md
+++ b/locale/fa/docs/guides/backpressuring-in-streams.md
@@ -60,7 +60,7 @@ In one scenario, we will take a large file (approximately ~9gb) and compress it
 using the familiar [`zip(1)`][] tool.
 
 ```
-$ zip The.Matrix.1080p.mkv
+zip The.Matrix.1080p.mkv
 ```
 
 While that will take a few minutes to complete, in another shell we may run
--- a/.markdownlint.json
+++ b/.markdownlint.json
@@ -5,7 +5,6 @@
   "MD009": false,
   "MD012": false,
   "MD013": false,
-  "MD014": false,
   "MD022": false,
   "MD024": false,
   "MD025": false,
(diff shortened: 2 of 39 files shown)


XhmikosR:
While at it, I'd add the dollar sign to the CSS; this way it doesn't affect copy pasting :)

See https://github.com/twbs/bootstrap/blob/3130288ba694451365bf7aa1b78a2a4e3ae67740/site/static/docs/4.3/assets/scss/_syntax.scss#L76-L79

You might need to add the prefixes too, not sure if you guys have autoprefixer set up.

nschonni:
I think what the rule is suggesting to not show a dollar sign if a line doesn't show console output on the following line, so adding it back through CSS would just be undoing that

XhmikosR:
Yeah but it's because it doesn't output anything. So it makes copy pasting harder. With the CSS solution you can have the dollar sign and copy pasting won't be affected.

https://github.com/DavidAnson/markdownlint/blob/master/doc/Rules.md#md014---dollar-signs-used-before-commands-without-showing-output

locale/fa/docs/guides/backpressuring-in-streams.md
```diff
@@ -60,7 +60,7 @@ In one scenario, we will take a large file (approximately ~9gb) and compress it
 using the familiar [`zip(1)`][] tool.
 
 ```
```
XhmikosR, line 62: Personally I'd mark those who lack a lang in this patch as `bash` so that it's consistent and so that they can later be targeted with CSS. Now it's plain text.

Trott:
> Yeah but it's because it doesn't output anything.

It's because the output is omitted. Commands like `npm install` and `node --inspect server.js` definitely have output.

IMO, this PR is correct to remove the `$` prompts from these lines and replacing it with a CSS `$` would be misguided. The prompt should not be there.

Trott (approved):

locale/fa/docs/guides/backpressuring-in-streams.md
nschonni: Yeah, those would get flagged if/when MD040 gets enabled again

PR #2384, opened 2019-08-12
Compress SVGs with svgo 1.3.0.

XhmikosR:
The config used can be found on https://gist.github.com/XhmikosR/f0ead200a2006e6924998e664ead3663


alexandrtovmach:
Hello, thanks for contributing.

What is the functional reason of this changes?

XhmikosR:
Smaller size :P

The output should be the same.

silverwind:
Did you compare the rendering? GitHub does not seem to want to show a rendered diff here.

Also, I'd check what `--precision` argument was used, would suggest at least `3`. Can't seem to find what the default value is in SVGO's docs.

XhmikosR:
Yeah, it looks fine AFAICT. Default is 3, I can change it, but I didn't notice any issues visually.

The only SVG file that seems to have been using a different precision is lfcp.svg, which BTW seems to be unused currently.

alexandrtovmach (approved):

XhmikosR:
For transformations it seems precision is 5, in the rest of the cases it's 3. I had to search their codebase for this.

silverwind (approved):

Trott (approved):

PR #2386, opened 2019-08-12
Losslessly compress PNG files.

XhmikosR:
Before: 4,97 MB (5.215.076 bytes)
After:  4,12 MB (4.324.203 bytes)

Used `optipng -o7`, `zopflipng -m --lossy_transparent -y --iterations=500`, and `pngout /kt /mincodes0` with various blocksizes


MylesBorins (approved):
LGTM

PR #2391, opened 2019-08-12
Enforce LF.

XhmikosR:
Previously on Windows with `autocrlf=true` which is the default, this led to CRLF.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,2 @@
+# Enforce Unix newlines
+* text=auto eol=lf


XhmikosR:
BTW this requires Git >= 2.10 IIRC. An alternative would be to specify each extension but that's more verbose.

XhmikosR:
Can we get this merged?

Trott:
@nodejs/website This could use some reviews.

lpinca (approved):

lpinca:
LGTM. We use the same on ws https://github.com/websockets/ws/blob/master/.gitattributes

Trott (approved):

PR #2392, opened 2019-08-12
Fix IE and minor tweaks

XhmikosR:
Before:

1-before

After:

2-after

### DIFF
--- a/layouts/css/_base.styl
+++ b/layouts/css/_base.styl
@@ -51,6 +51,7 @@ p
 img
   display block
   max-width 100%
+  border none
 
   .logos &
     // Fixes logo size on Firefox, see:
--- a/layouts/css/page-modules/_header.styl
+++ b/layouts/css/page-modules/_header.styl
@@ -140,6 +140,5 @@ header
   img
     display block
     margin 0.5em auto 0
-    border none
     max-height 75px
     max-width 180px
--- a/layouts/css/page-modules/_header.styl
+++ b/layouts/css/page-modules/_header.styl
@@ -29,7 +29,8 @@ header
     position absolute
     right 0
     bottom 0
-    height 20%
+    width 25px
+    height 25px
     fill $light-gray2
     cursor pointer
     padding 12px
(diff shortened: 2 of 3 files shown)


XhmikosR:
BTW I can always rebase and generally make any needed changes. I just happened to notice the IE issues and maybe the last patch could be split to a separate PR.

MaledongGit:
Any special changes？I don't see anything perticular……？

ghost (commented):
Any special changes？I don't see anything perticular……？

alexandrtovmach (commented):
Looks good for me, but I haven't sure about these changes. I can't believe that is some developer using old version of IE. In the same time you refactored JS code to ES5 style, that's not very readable (especially `Array.prototype` stuff).

alexandrtovmach:
> Any special changes？I don't see anything perticular……？

@MaledongGit 
Take a look on language switcher, it wasn't show before in IE

XhmikosR:
>Looks good for me, but I haven't sure about these changes. I can't believe that is some developer using ol version of IE. In the same time you refactored JS code to ES5 style, that's not very readable (especially Array.prototype stuff).

Well, this is the only place in frontend that ES6 was used. And that's just one line, we can all live with it or with `[].slice.call` :)

ghost:
> > Any special changes？I don't see anything perticular……？
> 
> @MaledongGit
> Take a look on language switcher, it wasn't show before in IE

Thanks for telling me, it's right.
I've no doubt about this change, but I'll see others' opinions about this change. I'm OK with it.

ghost (approved):

silverwind:
That `var` change is kind of pointless thought. IE11 supports `const` and supporting anything older is also pointless.

XhmikosR:
@silverwind: it's the only place ES6 is used. And IE doesn't support arrow functions. I went with consistency here, which is better in the long-term IMO.

silverwind (approved):

silverwind:
Ok, whatever. Would prefer we stop supporting IE11 soon and compile to ES6.

XhmikosR:
Fine by me but it should be done properly is all I'm saying. And this was easier for me at the time :P

PR #2397, opened 2019-08-13
Add Berlin meetup

vladimino:
There's an active Node.js community in Berlin and I am surprised to not find it in the Meetup's list. Should be fixed now! :)


MylesBorins (approved):
LGTM

PR #2401, opened 2019-08-14
Fix generating anchors with empty IDs and links.

XhmikosR:
Previously this resulted in duplicate IDs like `header-` and empty `name` attributes in translations.


XhmikosR:
Current behavior without the patch https://validator.w3.org/nu/?doc=https%3A%2F%2Fnodejs.org%2Far%2Fget-involved%2F Also the anchors are broken naturally.

ghost (approved):

PR #2402, opened 2019-08-14
Add new favicons.

XhmikosR:
Generated with https://realfavicongenerator.net/ and compressed losslessly.


silverwind:
` static/images/favicons/safari-pinned-tab.svg` looks odd without color.

XhmikosR:
Yeah that's how is supposed to be :p

silverwind:
Oh, I see, it's probably meant to be monochrome.

silverwind:
Would not move `/static/favicon.ico`. Some tools rely on the favicon to be present at `/favicon.ico` to retrieve a site's icon and nginx has a rewrite from `/favicon.ico` to `/static/favicon.ico`.

silverwind:
Rewrites could be adjusted in [here](https://github.com/nodejs/build/blob/4021ba59f8680b4d87df48e4a2d078fc9b9eea43/ansible/roles/nginx/templates/nodejs.org.conf.j2#L341) I suppose. I'm not sure if that is the file currently deployed, @rvagg surely knows. Given that these are `permanent` redirects they are cached.

XhmikosR:
Hmm I actually thought since it's not in root it wouldn't matter. I guess
we need to adapt the rewrite config. Or I could move it back, whatever you
guys prefer. I just thought I'd put all favicon related files into the same
folder.

silverwind:
Either move those rewrite targets back or adjust the rewrites shortly after landing this, but keep in mind it may never work for some clients because they can theoretically cache a 301 redirect forever.

BTW [here](https://github.com/sindresorhus/is-reachable/blob/bc6e064e6ec9ff2748336b569fe0eb8e336ad545/browser.js#L17) is one instance of `/favicon.ico` being used. GitHub Traffic stats itself is another.

silverwind:
Thought favicon is currently broken on traffic stats, I don't know why:

XhmikosR:
I created https://github.com/nodejs/build/pull/1897 for the path change.

MylesBorins (approved):
LGTM

@rvagg can you please land this PR when you are going to deploy https://github.com/nodejs/build/pull/1897

silverwind (approved):

rvagg:
I don't have access to land this, does it need to be done at precisely the same time? Maybe someone who's going to land this can ping me on IRC and we can sync? (remember I'm in AU so timing will likely be awkward).

MaledongGit:
@rvagg：I'm free to land this, but I cannot load into IRC.  So if it's OK, I'll land you when your response is OK, and if there's something else requested, you can also write down here.

Trott:
@rvagg I've added you as a Collaborator to this repo so you can land this yourself rather than have to coordinate with anyone.

rvagg:
seems to be working, please confirm

XhmikosR:
Yes, all good 👍

silverwind:
Yes, working. The issue mentioned in https://github.com/nodejs/nodejs.org/pull/2402#issuecomment-523584881 is likely because GitHub (incorrectly) does not follow the redirects on the `/favicon.ico` request. Could probably be fixed by removal of the `permanent` keyword.

PR #2403, opened 2019-08-15
Feature: Multi-Language Switcher Improvements

ghost:
1. Do not show the current choosen language unless your want to choose for a new one.
2. Show tooltip when moving your mouse onto the language symbol for the choosen language.

### DIFF
--- a/layouts/partials/header.hbs
+++ b/layouts/partials/header.hbs
@@ -29,14 +29,16 @@
     </svg>
 
     <div class="lang-picker hidden">
-      <ul>
-        {{#each locales}}
-          <li
-            data-lang="{{locale}}"
-            title="{{languageEnglishVersion}}"
-          >
-            {{language}}
-          </li>
+      <ul
+        id="language-items">
+          {{#each locales}}
+            <li
+                data-lang="{{locale}}"
+                title="{{languageEnglishVersion}}"
+                language="{{language}}"
+            >
+                {{language}}
+            </li>
         {{/each}}
       </ul>
     </div>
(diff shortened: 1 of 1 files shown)


XhmikosR:
FYI this will conflict with #2392

ghost:
@XhmikosR：Thanks, either you or I will meet with the conflict and fix it later, whatever submit is merged.

XhmikosR:
BTW I was thinking the language switcher should be static in the templates.
Right now even if a page doesn't have a translation, the switcher shows all
of them.

Something to consider for the future maybe

ghost:
> Right now even if a page doesn't have a translation, the switcher shows all of them.

Yes, but it would be nicer if we can show the current language as a tooltip for all of the users to see which language he/she is viewing with, and what's more, the picked language doesn't need to be shown in the menu list, unless you choose another one. 

That's my opition. For futher changes, maybe we can get together to change it little by little.

layouts/partials/header.hbs
```diff
@@ -29,14 +29,13 @@
...
-            data-lang="{{locale}}"
-            title="{{languageEnglishVersion}}"
-          >
-            {{language}}
-          </li>
+        <li data-lang="{{locale}}" 
+        title="{{languageEnglishVersion}}" 
+        language="{{language}}">
+          {{language}}
+        </li>
```
alexandrtovmach: Could you prettify the code

ghost: ok, what about my changes? I've changed that.

ghost: 捕获

alexandrtovmach (approved):

3imed-jaberi (approved):

Aissaoui-Ahmed (approved):

PR #2405, opened 2019-08-15
Split npm test:lint scripts.

XhmikosR:
This way we can run them individually too.

### DIFF
--- a/package.json
+++ b/package.json
@@ -14,7 +14,11 @@
     "load-schedule": "curl -sS https://raw.githubusercontent.com/nodejs/Release/master/schedule.json -o source/schedule.json",
     "start": "npm run serve",
     "test": "npm run test:lint && npm run test:unit && npm run test:smoke",
-    "test:lint": "standard --fix && htmllint **/*.hbs && stylint layouts/css && markdownlint \"**/*.md\" -i \"node_modules/\"",
+    "htmllint": "htmllint **/*.hbs",
+    "markdownlint": "markdownlint \"**/*.md\" -i \"node_modules/\"",
+    "standard": "standard --fix",
+    "stylint": "stylint layouts/css",
+    "test:lint": "npm run standard && npm run htmllint && npm run stylint && npm run markdownlint",
     "test:unit": "tape tests/**/*.test.js | faucet",
     "test:smoke": "tape tests/*.smoketest.js | faucet"
   },


package.json
```diff
@@ -14,7 +14,11 @@
     "load-schedule": "curl -sS https://raw.githubusercontent.com/nodejs/Release/master/schedule.json -o source/schedule.json",
     "start": "npm run serve",
     "test": "npm run test:lint && npm run test:unit && npm run test:smoke",
-    "test:lint": "standard --fix && htmllint **/*.hbs && stylint layouts/css && markdownlint \"**/*.md\" -i \"node_modules/\"",
+    "htmllint": "htmllint **/*.hbs",
```
MylesBorins, line 17: I think a naming convention like this would be more consistent

```suggestion
    "test:lint:html": "htmllint **/*.hbs",
```

XhmikosR: TBH I prefer the shorter one, but it's your call. We can keep the current scheme for scripts that run other scripts.

MylesBorins: stick with the shorter one then... nbd

XhmikosR:
Can we land this?

Trott:
@nodejs/website This needs some reviews?

package.json
alexandrtovmach: I agree with @MylesBorins ― consistency is more preferable.

XhmikosR: That's just extra characters to type without any real gain. Similarly with the other individual scripts, they are not all prefixed.

I strongly suggest that you guys reconsider. Typing less is better.

silverwind: While we're bikeshedding this, I suggest scheme `lint-xx` like we have in the Node.js repo and `lint` for the main task:

https://github.com/nodejs/node/blob/e4608381683de487c3c5844875e9adb692bb8725/Makefile#L1227

3imed-jaberi, line 17:
```suggestion
    "test:html-lint": "htmllint **/*.hbs",
```

package.json
```diff
@@ -14,7 +14,11 @@
...
+    "markdownlint": "markdownlint \"**/*.md\" -i \"node_modules/\"",
```
3imed-jaberi, line 18:
```suggestion
    "test:markdown-lint": "markdownlint \"**/*.md\" -i \"node_modules/\"",
```

package.json
```diff
@@ -14,7 +14,11 @@
...
+    "standard": "standard --fix",
```
3imed-jaberi, line 19:
```suggestion
    "test:standard-lint": "standard --fix",
```

package.json
```diff
@@ -14,7 +14,11 @@
...
+    "stylint": "stylint layouts/css",
```
3imed-jaberi, line 20:
```suggestion
    "test:style-lint": "stylint layouts/css",
```

package.json
```diff
@@ -14,7 +14,11 @@
...
+    "test:lint": "npm run standard && npm run htmllint && npm run stylint && npm run markdownlint",
```
3imed-jaberi, line 21:
```suggestion
    "test:lint": "npm run test:standard-lint && npm run test:html-lint && npm run test:style-lint && npm run test:markdown-lint",
```

3imed-jaberi (commented):
@XhmikosR , Thanks for your contribution ..
I agree with @MylesBorins here ..
So, I think the script key should be clearer more than short ( consistency ) ..
I have made some suggestions .. You can check 🚀 ..

XhmikosR:
Sorry, I don't like this approach. You are welcome to make that change yourself. I don't see why I need to type the extra characters.

PR #2394, opened 2019-08-13
Minor CSS cleanup.

XhmikosR:
* simplify padding shorthand
* remove units from 0
* use `property-name | duration | timing function` for transition shorthand
* remove trailing spaces
* add missing EOF

PR #2406, opened 2019-08-15
Tighten stylint config.

### LINKED ISSUE #2404: Tighten stylint config
While looking at #2394 I notice that stylint's config is very lax.

I'll make a PR later to tighten things a bit

XhmikosR:
WIP, I'll try to find the config that matches the current style best so that we don't need to make many changes in the source files.

Fixes #2404

### DIFF
--- a/.stylintrc
+++ b/.stylintrc
@@ -1,10 +1,117 @@
 {
+  "blocks": false,
+  "brackets": {
+    "expect": false,
+    "error": true
+  },
   "colons": {
     "expect": "never",
     "error": true
   },
+  "colors": false,
+  "commaSpace": {
+    "expect": "always",
+    "error": true
+  },
+  "commentSpace": {
+    "expect": "always",
+    "error": true
+  },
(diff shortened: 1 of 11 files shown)


XhmikosR:
This is ready. First #2394 should be merged and I'll rebase this.

.stylintrc
```diff
@@ -1,10 +1,117 @@
...
+  "extendPref": "@extends",
+  "globalDupe": false,
+  "groupOutputByFile": true,
   "indentPref": {
     "expect": 2,
     "error": true
-  }
+  },
+  "leadingZero": {
+    "expect": "always",
```
XhmikosR, line 42: Personally I'd use `never` here, but most cases were using a leading 0 so I went with `always`.

alexandrtovmach (approved):

XhmikosR:
Please note that #2394 should be merged first.

Aissaoui-Ahmed (approved):

PR #2412, opened 2019-08-15
blog: aug 2019 security post-release announcement

sam-github:
Waiting on actual release to update timestamps and merge.

### DIFF
--- a/locale/en/blog/vulnerability/aug-2019-security-releases.md
+++ b/locale/en/blog/vulnerability/aug-2019-security-releases.md
@@ -1,5 +1,5 @@
 ---
-date: 2019-08-13T21:30:00.000Z
+date: 2019-08-15T22:43:22.000Z
 category: vulnerability
 title: August 2019 Security Releases
 slug: august-2019-security-releases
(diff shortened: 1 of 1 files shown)


richardlau:
It looks like some of the fixes are revertible (via `--security-reverts=`). Do we document those?

richardlau:
> 
> 
> It looks like some of the fixes are revertible (via `--security-reverts=`). Do we document those?

(Also it looks like the flag is `--security-revert=` in `8.x` and `--security-reverts=` in `10.x` and `12.x` which is an annoying inconsistency.)

sam-github:
In these cases, I would rather not doc the sec reverts. They are there if anyone reports problems, but there weren't any concerns raised about the limits during PR review, they seem pretty generous and unlikely to be hit by normal apps.

The inconsistency is not good. I think it was a typo introduced in https://github.com/addaleax/node/blob/be4a3b658a692d390f4fa9a187189c2ee993de1c/src/node_options.cc#L137 but probably not noticed because there were no revertible CVEs at the time.

I don't think this should be fixed now, but maybe it is worth fixing on master for next time, if there is one? EDIT or perhaps it was intentional because the argument is an array? Not sure

BethGriggs (approved):

addaleax (approved):

richardlau (approved):

Trott (approved):

PR #2428, opened 2019-08-17
Remove English commented out code from the Korean files.

XhmikosR:
This is too much duplication and slows down the build.

### DIFF
--- a/locale/ko/about/governance.md
+++ b/locale/ko/about/governance.md
@@ -6,19 +6,6 @@ layout: about.hbs
 
 ## 기술 결정 위원회(TSC, Technical Steering Committee)
 
-<!--
-The project is jointly governed by a Technical Steering Committee (TSC)
-which is responsible for high-level guidance of the project.
-
-The TSC has final authority over this project including:
-
-* Technical direction
-* Project governance and process (including this policy)
-* Contribution policy
-* GitHub repository hosting
-* Conduct guidelines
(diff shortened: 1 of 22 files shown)


Aissaoui-Ahmed (approved):

yous:
Don't remove the comments. It's for making diffs by translators when the original (en) articles are updated.

XhmikosR:
It's useless and you should use the original files doe this purpose. See
how many lines there are which slow down the builds and/or introduce false
markdown errors.

yous:
Translators make diffs of the original files and the comments. So they can update the translation where it needs update. Re-translating the whole file is time wasting, and translations are not updated real-time to reflect changes of every single commit on `locale/en` directory.

Using HTML comments are [valid](https://daringfireball.net/projects/markdown/syntax#html) in Markdown. Also I see that [every build of master branch](https://travis-ci.com/nodejs/nodejs.org/builds) runs around just one minute. I don't think HTML comments slow down the builds dozens of seconds.

XhmikosR:
You do know there is git available?

Anyway I don't want to engage in this. I'll let the maintainers decide.

To me it's totally useless and bloats the source code with so much
duplication.

yous:
There are translators not used to git.

ghost:
@XhmikosR, I know what you mean, but since there are still some people loving to keep the original content to match what they've translated. So there's no need to remove them.

PR #2430, opened 2019-08-17
Fix a couple of Markdownlint issues.

XhmikosR:


Trott (changes requested):
These changes mean that the commit messages listed in the change logs will be different from the commit message in the git logs. I'd prefer we tolerate the minor errors and keep git history aligned with the changelog.

Trott:
Er...blog posts rather than changelog but the point still stands.

XhmikosR:
NP, I was just going through the inline suppressions. As it is now, those files are ignored. Not a big deal, I just happened to notice.

PR #2432, opened 2019-08-20
Remove the npm gzip script.

### LINKED ISSUE #2429: is npm gzip script still useful?
Being that now Cloudflare is used, which means we have Brotli compression available, is the npm gzip script still useful and/or used at all in practice?

https://github.com/nodejs/nodejs.org/blob/229e9f07febf2f877b8a2a2a4bbdc5500f12b1b9/package.json#L11

Actually judging from the response headers, it seems Brotli isn't enabled on Cloudflare. Any special reason for not doing this and getting rid of the whole gzip script and serving precompressed files?

XhmikosR:
Since Cloudflare is being used, Brotli compression is used.

Closes #2429

Draft until @rvagg feels the Brotli switch is fine


rvagg:
👍 up to the website team, it _looks_ right to me:

```
$ curl -IH 'Accept-encoding: gzip,br,deflate'  https://nodejs.org/static/images/logo.svg
HTTP/2 200
date: Tue, 20 Aug 2019 11:53:16 GMT
content-type: image/svg+xml
...
content-encoding: br
```
I guess that's all good. Don't let me hold it up.

silverwind:
Which compression level does CF use? If it's not 11/9 for brotli/gzip, there could still be gains had by precompressing assets (I do have [a module](https://github.com/silverwind/precompress) for that).

XhmikosR:
That I don't know TBH. I don't think they use the best compression, though, since it's slow.

I tried https://tools.paulcalvano.com/compression.php and it seems it's 5.

silverwind:
@rvagg can you link the `nginx.conf` in use? I think it might use `gzip_static` which is generally preferable for performance compared to on-the-fly compression like CF does, e.g. gzip static level 9 ought to beat brotli level 5 on-the-fly.

Also, if nginx has a brotli module loaded, we could use `brotli_static`, but it's not part of the regular nginx distribution. On Ubuntu, I use [this](https://github.com/cryptofuture/nginx-hda-bundle) nginx ppa which brings `ngx_http_brotli_static_module.so`.

XhmikosR:
For what is worth, the diff in TTFB is ~100ms AFAICT. **EDIT:** Which might be due to something else, I can't tell for sure myself.

https://gtmetrix.com/compare/csIu6DJG/s4fDLhlk

So, if the server conf supports serving precompressed assets, we should drop this PR and instead use your solution @silverwind to generate both br and gzip assets.

silverwind:
The server uses `gzip_static` as seen [here](https://github.com/nodejs/build/blob/4021ba59f8680b4d87df48e4a2d078fc9b9eea43/ansible/roles/nginx/templates/nodejs.org.conf.j2#L26). Ideally, we would introduce `ngx_http_brotli_static_module.so` and then adjust the gzip script to also generate `.br` files.

Now I do wonder how Cloudflare handles the case of the source server serving brotli, I think it needs to be checked, not that the work is in vain if Cloudflare compresses again regardless.

XhmikosR:
@silverwind how do you want to proceed with this?

XhmikosR:
@silverwind friendly ping

XhmikosR:
Just so that we are safe, @rvagg is the `gzip` script used in any deployment scripts which are not part of this repo?

rvagg:
Yes, it's used by the next line, you need to remove it from there too. `npm run deploy` is what's executed.
You can push to master and watch the log at https://nodejs.org/github-webhook.log to see how it goes. If it fails it'll let you know there and you can fix with a new push to master.

XhmikosR:
Oops, that was a rebase error, fixed now.

Thanks for the webhook info, I wasn't aware it existed.

It seems we are building node-sass each time :/ I guess #2724 might help in this regard.

XhmikosR:
Thanks for the webhook info, I wasn't aware it existed.

It seems we are building node-sass each time :/ I guess #2724 might help in this regard.

XhmikosR:
@silverwind ping

silverwind:
I'm unsure here. If we keep gzip it looks like this (assuming cloudflare supports retrieving gzip):

origin (gzip) -> cloudflare (un-gzip + brotli) -> user

If we remove gzip:

origin (uncompressed) -> cloudflare (brotli) -> user

The first method saves bandwidth, while the second one saves cpu time. Ideal case would of course be serving brotli+gzip from the source, but that needs a [non-default nginx module](https://github.com/google/ngx_brotli).

silverwind (approved):
On second thought, I guess with proper caching, these origin -> cloudflare transfers probably only happen when files change, so it's probably better overall to remove it and shave of a few seconds from the deploy runtime.

XhmikosR:
Thanks @silverwind. @rvagg your thoughts?

rvagg:
fine by me I suppose

XhmikosR:
Alright, let's merge this and if there's anything unusual @rvagg we can revert it.

PR #2435, opened 2019-08-20
Remove unused Stylus variables.

XhmikosR:


silverwind:
Can you set stylelint to warn on unused vars like this?

XhmikosR:
In this project we use stylint, not stylelint. But even Stylelint doesn't have such a check :P

For Sass I've created https://github.com/XhmikosR/find-unused-sass-variables

silverwind:
`stylint`, wtf 🤣

Well, if `stylelint` can't do it by itself, I guess we better wait til it can 😉

silverwind (approved):

3imed-jaberi (approved):

PR #2441, opened 2019-08-21
anchor-markdown-headings.js: fix IDs to be unique.

### LINKED ISSUE #2440: Duplicate heading IDs
[anchor-markdown-headings.js](https://github.com/nodejs/nodejs.org/blob/master/scripts/plugins/anchor-markdown-headings.js) generates duplicate IDs in many cases causing breakage since IDs are supposed to be unique.

An example is https://nodejs.org/en/download/package-manager/#alternatives but if you go to Windows https://nodejs.org/en/download/package-manager/#windows, you will see it also has a header with the same ID.

XhmikosR:
Works fine, but I can't figure out the tests.

Any help is welcome.

Fixes #2440

### DIFF
--- a/scripts/plugins/anchor-markdown-headings.js
+++ b/scripts/plugins/anchor-markdown-headings.js
@@ -12,7 +12,7 @@
  * `<!---->` to quote your English anchor name inside, with your
  * own title beside it.
  */
-module.exports = function anchorMarkdownHeadings (text, level, raw) {
+module.exports = function anchorMarkdownHeadings (text, level, raw, slugger) {
   // Check whether we've got the comment matches
   // <!--comment-->, <!--comment --> and even <!-- comment-->
   // (20 hex = 32 dec = space character)
@@ -42,8 +42,8 @@ module.exports = function anchorMarkdownHeadings (text, level, raw) {
 
   anchorTitle = anchorTitle.toLowerCase()
 
-  return '<h' + level + ' id="header-' + anchorTitle + '">' + text + '<a name="' +
-    anchorTitle + '" class="anchor" href="#' +
-    anchorTitle + '" aria-labelledby="header-' +
-    anchorTitle + '"></a></h' + level + '>'
+  const anchorId = `${slugger ? slugger.slug(anchorTitle) : anchorTitle}`
+  const headerId = `header-${anchorId}`
+
+  return `<h${level} id="${headerId}">${text}<a id="${anchorId}" class="anchor" href="#${anchorId}" aria-labelledby="${headerId}"></a></h${level}>`
 }
--- a/locale/it/docs/guides/buffer-constructor-deprecation.md
+++ b/locale/it/docs/guides/buffer-constructor-deprecation.md
@@ -5,7 +5,6 @@ layout: docs.hbs
 
 # Porting to the `Buffer.from()`/`Buffer.alloc()` API
 
-<a id="overview"></a>
 ## Overview
 
 This guide explains how to migrate to safe `Buffer` constructor methods. The migration fixes the following deprecation warning:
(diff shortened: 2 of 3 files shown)


XhmikosR:
I think the proper solution would be to hook marked into the tests.

I might have a take at it tomorrow, but any help is welcome.

XhmikosR:
Alright, I think we got it. Basically I only use marked in the duplicate IDs test and I check for `slugger` in our custom function.

The behavior should be the same, I just swapped `name` with `id` on the anchor so the generated anchors should be the same, except for cases with duplicate IDs where they didn't work anyway.

If anyone has any suggestions please let me know.

scripts/plugins/anchor-markdown-headings.js
```diff
@@ -42,8 +42,8 @@ module.exports = function anchorMarkdownHeadings (text, level, raw) {
 
   anchorTitle = anchorTitle.toLowerCase()
 
-  return '<h' + level + ' id="header-' + anchorTitle + '">' + text + '<a name="' +
-    anchorTitle + '" class="anchor" href="#' +
-    anchorTitle + '" aria-labelledby="header-' +
-    anchorTitle + '"></a></h' + level + '>'
+  const anchorId = `${slugger ? slugger.slug(anchorTitle) : anchorTitle}`
+  const headerId = `header-${anchorId}`
```
XhmikosR: Maybe this logic is wrong and I should use `slugger` here too. I'll try to verify again later.

Trott:
> I think the proper solution would be to hook marked into the tests.
> 
> I might have a take at it tomorrow, but any help is welcome.

@phillipj @nodejs/website Anyone able to give @XhmikosR a hand with this?

scripts/plugins/anchor-markdown-headings.js
```diff
@@ -42,8 +42,8 @@ module.exports = function anchorMarkdownHeadings (text, level, raw) {
 
   anchorTitle = anchorTitle.toLowerCase()
 
-  return '<h' + level + ' id="header-' + anchorTitle + '">' + text + '<a name="' +
-    anchorTitle + '" class="anchor" href="#' +
-    anchorTitle + '" aria-labelledby="header-' +
-    anchorTitle + '"></a></h' + level + '>'
+  const anchorId = `${slugger ? slugger.slug(anchorTitle) : anchorTitle}`
```
phillipj, line 45: Reading the [marked docs](https://github.com/markedjs/marked/blob/master/docs/USING_PRO.md) I get the impression that the `slugger` argument will _always_ be provided when our custom header renderer function (`anchorMarkdownHeadings()`) is called by marked. Or have I misunderstood?

If it is indeed always provided, could we invoke `slugger.slug()` directly, without having logic checking if `slugger` is indeed provided? Primarily to help new contributors and our future selfs when trying to understand what this code does later, not having to graps unnecessary logic is valuable.

phillipj:
Seems like you've found a decent way of testing the duplicate IDs by using marked in the tests 👍 

Other than the raised question about invoking `slugger.slug()`, this LGTM.

XhmikosR:
Thanks for the comments! The only case that slugger was not set was in
tests.

Feel free to push to my branch or suggest an edit.

XhmikosR:
@MaledongGit: this wasn't supposed to be squashed either. I keep the patches separate on purpose in my PRs, when it makes sense to do so. Please do not squash my patched blindly.

silverwind:
I think the default UI merge action is set to squash. Maybe add a note to the initial comment on PRs that should not be squashed.

XhmikosR:
Yeah, the thing with the merge button, is that it remembers your last choice.

Issue #2443, opened 2019-08-21
watch is broken


XhmikosR:
Could be I'm on Windows, but whenever I edit a markdown file, the server crashes for me:

```
C:\Users\xmr\Desktop\nodejs.org>node -v
v12.9.0

C:\Users\xmr\Desktop\nodejs.org>npm start

> nodejs.org@ start C:\Users\xmr\Desktop\nodejs.org
> npm run serve


> nodejs.org@ serve C:\Users\xmr\Desktop\nodejs.org
> node server.js

[metalsmith] build/static started
[metalsmith] build/layouts started
Server running at http://localhost:8080/en/
(node:1704) TimeoutOverflowWarning: 4294967296000 does not fit into a 32-bit signed integer.
Timer duration was truncated to 2147483647.
(node:1704) TimeoutOverflowWarning: 4294967296000 does not fit into a 32-bit signed integer.
Timer duration was truncated to 2147483647.
(node:1704) TimeoutOverflowWarning: 4294967296000 does not fit into a 32-bit signed integer.
Timer duration was truncated to 2147483647.
(node:1704) TimeoutOverflowWarning: 4294967296000 does not fit into a 32-bit signed integer.
Timer duration was truncated to 2147483647.
[metalsmith] build/static finished: 93.953ms
[metalsmith] build/layouts finished: 505.888ms
[metalsmith] build/ar started
[metalsmith] build/ca started
[metalsmith] build/de started
[metalsmith] build/en started
[metalsmith] build/es started
[metalsmith] build/fa started
[metalsmith] build/fr started
[metalsmith] build/gl started
[metalsmith] build/it started
[metalsmith] build/ja started
[metalsmith] build/ko started
[metalsmith] build/pt-br started
[metalsmith] build/ru started
[metalsmith] build/uk started
[metalsmith] build/zh-cn started
[metalsmith] build/zh-tw started
[metalsmith] build/gl finished: 6020.478ms
[metalsmith] build/de finished: 6286.983ms
[metalsmith] build/ca finished: 7431.933ms
[metalsmith] build/es finished: 7393.092ms
[metalsmith] build/zh-tw finished: 7292.250ms
[metalsmith] build/fr finished: 7363.862ms
[metalsmith] build/ar finished: 9901.110ms
[metalsmith] build/ko finished: 10096.795ms
[metalsmith] build/it finished: 10130.890ms
[metalsmith] build/ja finished: 10118.911ms
[...]

PR #2444, opened 2019-08-21
Update chokidar to v3.0.2.

XhmikosR:
Tested it with layouts and stylus editing locally and seems to work fine.

I cannot test it with Markdown files due to #2443


silverwind (approved):

Trott:
Does this fix a bug/issue?

XhmikosR:
>Does this fix a bug/issue?

No, I was just trying to keep the deps up to date and see if it'll help with my watch issues. But judging from the changelog it should be a good change. I don't see any API changes mentioned.

PR #2461, opened 2019-08-25
Remove unused Prism CSS files.

XhmikosR:
These aren't used, probably left for historical reasons?

PR #2462, opened 2019-08-25
Convert prism-tomorrow.css to stylus and include it in our main CSS.

XhmikosR:
Requires #2461!

The idea is to combine the Prism CSS with our main CSS and thus run autoprefixer and minifier like our main CSS file.

I know it's a 3rd-party file and we probably shouldn't touch it, but this way we end up with less CSS, which has the same transformations applied as our main CSS file.

Let me know your thoughts. I tried just `@import`ing the Prism CSS file in our styl file, but then it wasn't being minified. I'll give that another go later, because this solution would be the least intrusive one.


XhmikosR:
Not sure how to proceed with this. If we go with #2530, the stylus conversion is moot, but #2530 is built on top of this PR. Also, I'm not sure if or when #2530 will land, so perhaps we should move with this in the meantime?

XhmikosR:
Since I have based my PRs on this, I think it makes sense to move on with this for the time being.

When the Sass PR is merged, the prism CSS file will be restored.

Trott:
@nodejs/website

alexandrtovmach (approved):

layouts/css/vendor/_prism-tomorrow.styl
```diff
@@ -0,0 +1,99 @@
+/**
+ * prism.js tomorrow night eighties for JavaScript, CoffeeScript, CSS and HTML
+ * Based on https://github.com/chriskempson/tomorrow-theme
+ * @author Rose Pritchard
+ */
+
+code[class*='language-'],
+pre[class*='language-']
+  color #ccc
```
Aissaoui-Ahmed, line 9: $light-gray2

Aissaoui-Ahmed (changes requested):

layouts/css/vendor/_prism-tomorrow.styl
XhmikosR: It's a 3rd-party file, I'd rather keep it close to the original.

Aissaoui-Ahmed (approved):

PR #2465, opened 2019-08-25
footer.hbs: tweak scrollToTop.

XhmikosR:
* use `addEventListener`
* use `event.preventDefault()`
* remove unneeded parentheses
* remove `$` from the variable name

### DIFF
--- a/layouts/partials/footer.hbs
+++ b/layouts/partials/footer.hbs
@@ -29,16 +29,18 @@
 
 <link rel="stylesheet" href="/static/css/prism-tomorrow.css" media="all">
 <script>
-  var $scrollToTop = document.getElementById('scroll-to-top');
+  var scrollToTop = document.getElementById('scroll-to-top');
+
   (window.onscroll = function() {
     window.requestAnimationFrame(function() {
-      $scrollToTop.style.display = (window.pageYOffset > window.innerHeight) ? 'block' : 'none';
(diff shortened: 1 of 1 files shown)


layouts/partials/footer.hbs
```diff
@@ -29,16 +29,18 @@
 
 <link rel="stylesheet" href="/static/css/prism-tomorrow.css" media="all">
 <script>
-  var $scrollToTop = document.getElementById('scroll-to-top');
+  var scrollToTop = document.getElementById('scroll-to-top');
+
   (window.onscroll = function() {
```
XhmikosR, line 34: If I change this to use `addEventListener` then the link shows on page load. Maybe this can be improved further, so any ideas welcome.

lpinca (approved):
Mostly a cosmetic change but LGTM anyway.

XhmikosR:
@lpinca: how about changing `onscroll` to use `addEventListener`? I think it fails because `onscroll` overwrites any events while addEvenetListener adds a new one. But I feel it would be better if we were consistent here.

lpinca:
I'm fine with that if it works.

XhmikosR:
It doesn't work, hence why I'm asking for ideas :P

Trott:
Since the current version works, will land this for now so it doesn't stall. The additional troubleshooting/discussion can certainly happen subsequently.

PR #2468, opened 2019-08-26
Fixture: Change the comment

ghost:
'/en' should be the default language of 'mainLocale'.

### DIFF
--- a/server.js
+++ b/server.js
@@ -78,7 +78,7 @@ const mainLocale = (selectedLocales && selectedLocales[0]) || 'en'
 
 // Initializes the server and mounts it in the generated build directory.
 http.createServer((req, res) => {
-  // If we are accessing the root, it should be redirected to `/en` instead.
+  // If we are accessing the root, it should be redirected to the default language,
   // We shouldn't get a 404 page.
 
   if (req.url === '/') {


server.js
```diff
@@ -78,7 +78,7 @@ const mainLocale = (selectedLocales && selectedLocales[0]) || 'en'
 
 // Initializes the server and mounts it in the generated build directory.
 http.createServer((req, res) => {
-  // If we are accessing the root, it should be redirected to `/en` instead.
+  // If we are accessing the root, it should be redirected to the default language,
```
XhmikosR, line 81: Unrelated to this patch, I just noticed it here. Maybe this should be `res.redirect`?

In another PR, just thinking out loud :)

ghost: IMO, I just mean that when DEFAULT_LOCALES is null or multiple languages, ONLY the first one will be taken instead of 'en'. So that's the reason why we change the comment.

BTW, we don't have 'redirect' function for 'res' here, so change 'url' is also fine (See：https://nodejs.org/dist/latest-v10.x/docs/api/http.html#http_class_http_serverresponse), Maybe we can use something like `res.writeHead` instead, but here for 'url' setting is enough  ;)

Issue #2459, opened 2019-08-24
htmllint issues on Windows


XhmikosR:
It seems httmllint gets confused by the missing space in the handlebar templates, but this issue only appears on Windows hence why Travis CI builds just pass.

```
C:\Users\xmr\Desktop\nodejs.org>npm t

> nodejs.org@ test C:\Users\xmr\Desktop\nodejs.org
> npm run test:lint && npm run test:unit && npm run test:smoke


> nodejs.org@ test:lint C:\Users\xmr\Desktop\nodejs.org
> standard --fix && htmllint **/*.hbs && stylint layouts/css && markdownlint "**/*.md" -i "node_modules/"

layouts/partials/footer.hbs: line 3, col 20, value must match the format: dash
layouts/partials/footer.hbs: line 3, col 1, tag is not closed
layouts/partials/header.hbs: line 10, col 9, tag is not closed
layouts/partials/navigation.hbs: line 6, col 11, tag is not closed
```

I don't think we should change the code to add the space back. In fact I'm not sure if htmllint offers any value... The website was full of serious HTML validation errors that were caught by the official HTML validator which I plan to add here eventually.

Let me know how to proceed.

XhmikosR:
https://travis-ci.org/XhmikosR/nodejs.org/jobs/576205825

If Windows on Travis CI wasn't so slow, I'd say we added it, but it's too slow.

Are there any plans to switch to GitHub Actions CI?

ghost:
@XhmikosR , It seems this is also checked in CI, just like what you said in https://travis-ci.org/XhmikosR/nodejs.org/jobs/576205825, and I've fixed it in my submit ;)
![image](https://user-images.githubusercontent.com/52018749/64055294-2ebeed00-cbbe-11e9-87f8-dc1f01f6f129.png)

BTW: Do you mean the time in windows or Linux is a huge difference when using Htmllint? I'd say that we should keep it because a pretty code needs time to be checked with. As for GitHub Action CI, this may be done in the future. Until now we have to use Htmllint instead.

XhmikosR:
@MaledongGit: yeah I meant Windows, but Travis is a no go for this currently, unfortunately.

About your PR, I commented there. These aren't real issues, and for example, it hasn't warned us for indentation issues. We shouldn't be using htmllint IMO.

ghost:
@XhmikosR：You mean in Linux it skipped the check……？This is really strange....

XhmikosR:
[...]

PR #2437, opened 2019-08-21
Convert a few posts to Markdown

XhmikosR:
Fixes a few HTML validation errors along the way.

Also remove the 404 images from /multi-server-continuous-deployment-with-fleet.md.

For npm's posts, I kept the old IDs for backward compatibility.

Issue #2422, opened 2019-08-17
Wrong path when there's no previous yearly archive


XhmikosR:
There's a bug somewhere that results in the wrong link in https://nodejs.org/en/blog/year-2011/

Older: https://nodejs.org/en/blog/index.md

I haven't had the time to look into it yet.

PR #2469, opened 2019-08-26
Add vnu.jar support

### LINKED ISSUE #2398: Fix HTML validation errors
It seems there are plenty of errors, with some of them important IMO.

For example https://validator.w3.org/nu/?doc=https%3A%2F%2Fnodejs.org%2Fen%2Fblog%2F

We could automate the checks if you agree, most should be easy to fix. Here's a first attempt to use the official HTML validator: https://travis-ci.org/XhmikosR/nodejs.org/builds/572662096

XhmikosR:
This is WIP because there are some things that need to be addressed:

1. #2441 needs to be merged
2. #2437 probably needs to be merged too
3. There's a bug in vnu.jar, I have already discussed with sideshowbarker (not CC'ing right now), that causes the build to exit with non-zero.
4. I'd prefer it if we could actually fix the table errors, because these are real issues:
    ```
    "file:/home/travis/build/XhmikosR/nodejs.org/build/en/download/index.html":276.50-277.32: error: Table column 3 established by element "td" has no cells beginning in it.
    "file:/home/travis/build/XhmikosR/nodejs.org/build/en/download/index.html":277.113-278.32: error: Table column 7 established by element "td" has no cells beginning in it.
    ```
5. The license of the new file might need to be adapted for this repo. It's just a simple script so it's OK to change it, especially since I also adapted it to fit our needs

Note that this is how I noticed and fixed many of the issues so far, like https://github.com/nodejs/nodejs.org/commit/bae664a98a772781314b05eae4cdcc57efafd2b0, https://github.com/nodejs/nodejs.org/commit/b533fcddc0deffbf086d229b5f79155d8a21868d, https://github.com/nodejs/nodejs.org/commit/a714afcc90df0283d0cd9006d049a78ce60361fc, https://github.com/nodejs/nodejs.org/commit/ceee600686ae067f439a7ad9230499c442081e12 etc.

IMO this should definitely land as soon as we sort out the above points. Personally I'd remove htmllint too; I'm [having issues with it on Windows](#2459) and it just doesn't seem to catch any real issues.

Marking it as draft for further discussion and tweaks.

vnu.jar: <https://github.com/validator/validator>

Closes #2398

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -16,6 +16,7 @@ jobs:
 
       - run: node --version
       - run: npm --version
+      - run: java -version
 
       - name: Install npm dependencies
         run: npm ci
@@ -27,6 +28,9 @@ jobs:
 
       - name: Run tests
         run: npm test
+
+      - name: Run HTML validator
+        run: npm run test:html
 
       - name: Run linkinator
         run: npm run test:linkinator
--- a/package-lock.json
+++ b/package-lock.json
@@ -1876,7 +1876,6 @@
       "integrity": "sha1-lEvs00zEHuMqY6n68nrVpl/Fl3c=",
       "dev": true,
       "requires": {
-        "cross-spawn": "^5.0.1",
         "get-stream": "^3.0.0",
         "is-stream": "^1.1.0",
         "npm-run-path": "^2.0.0",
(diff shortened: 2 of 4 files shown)


XhmikosR:
I think we should try fixing the table errors. I'm gonna revert the ignore so that we see the errors, and someone might be able to help fix those. I believe the wrong exit code comes from this for some reason.

XhmikosR:
Alright, I don't know how to fix the table errors, so I have them ignored for now.

This is ready for review. Note that on Windows there will be a failure due to #2422. This isn't triggered on nix, but I've yet to figure out the root cause (like I never managed to figure it out for the slashes, hence the workaround we have).

XhmikosR:
After #2511 CI will be green. Generally that is why I insist that branches need to be up to date and we have the button.

XhmikosR:
Can we move with this?

XhmikosR:
Can I get some reviews please?

This is the only proper way to make sure we are valid and we don't have issues like duplicate IDs and so on. Note that I intentionally skip the test if Java isn't available, and that I haven't added the script in `npm t` like I haven't done it with linkinator, because these tests are a little slow.

On CI Java is present so this will run there.

Trott:
@nodejs/website

phillipj (approved):
LGTM

PR #2470, opened 2019-08-26
Update dependencies.

XhmikosR:
This streamlines the deps. It removes the caret operator from `0.0.x` packages since it's moot and regenerates package-lock.json from scratch, which flattens the deps tree a bit.


MaledongGit:
It seems this conflicts with https://github.com/nodejs/nodejs.org/pull/213, so I have to invite @phillipj to have a review of it clearly :)

ghost (commented):
It seems this conflicts with #213, so I have to invite @phillipj to have a review of it clearly :)

XhmikosR:
NP, it's just that it is currently inconsistent.

I personally do this once in a while in my projects so that the deps tree is flattened and from then on I just update each dependency individually.

silverwind (approved):

phillipj:
Although late to reply, I'll chime in here anyways since I got mentioned.

This looks good IMO. It looks to me as if #213 was done in an effort of pinning dependency version, so that we had better confidence in which package @ version got installed at any given time -- i.e. reproducible builds. In hindsight, pinning version in `package.json` only controlled our direct dependencies, which is only the tip of the iceberg in reality. Usually the majority of dependencies aren't direct dependencies, but rather dependencies of other dependencies.

Now that we've got `package-lock.json` instead which does in fact lock down the entire dependency tree, the initial reason for pinning versions in `package.json` no longer exists 🎉

silverwind:
Yes, carets have no significance with a lockfile. I personally prefer to not use them even without lockfiles, but that requires one to update manually from time to time.

PR #2471, opened 2019-08-26
Remove trailing spaces.

XhmikosR:
Also enable the markdowlint rule to check for this.

I split the patches by locale so that it's easier to review.

The easiest way to review this is to check the rich diff of each file. I already checked most files and everything seems good, but more eyes is always better.

An alternative way, would be to move each patch to each own PR and when we are done, we enable the Markdownlint rule to enforce this.


ghost (approved):
LGTM for formation changes only.

ghost:
@XhmikosR , forcely push your submit, there're conflicts ;)

PR #2476, opened 2019-08-27
Specify `word-break: break-word` for article links.

XhmikosR:
Without this, the links were not fully showing on mobile if they didn't fit the screen.

Before:

| Before | After |
| ------- | ------ |
| 1-before | 2-after |

### DIFF
--- a/layouts/css/styles.styl
+++ b/layouts/css/styles.styl
@@ -20,6 +20,9 @@
 @import 'page-modules/_release-schedule'
 @import 'page-modules/_resources'
 
+article a
+  word-break break-word
+
 .intro
   margin-top 140px
   font-size 38px


silverwind:
Any downside applying this to all elements (`*`)? I can't think of any.

XhmikosR:
Well, I personally don't like using the universal selector, so I won't do it myself.

In fact, I still have this PR in draft because I want to double check the effect in more pages.

XhmikosR:
Preview: <https://frosty-almeida-4843d4.netlify.com/en/>

It seems fine so far, but I'd like more people to check.

Trott (approved):

Aissaoui-Ahmed (approved):

PR #2479, opened 2019-08-29
Update dont-block-the-event-loop.md

MariahWine:
Added broader description on DNS.

### DIFF
--- a/locale/en/docs/guides/dont-block-the-event-loop.md
+++ b/locale/en/docs/guides/dont-block-the-event-loop.md
@@ -360,7 +360,7 @@ A CPU-intensive task only makes progress when its Worker is scheduled, and the W
 If you have 4 logical cores and 5 Workers, one of these Workers cannot make progress.
 As a result, you are paying overhead (memory and scheduling costs) for this Worker and getting no return for it.
 
-I/O-intensive tasks involve querying an external service provider (DNS, file system, etc.) and waiting for its response.
+I/O-intensive tasks involve querying an external service provider ([DNS](https://hosting.review/web-hosting-glossary/#9), file system, etc.) and waiting for its response.
 While a Worker with an I/O-intensive task is waiting for its response, it has nothing else to do and can be de-scheduled by the operating system, giving another Worker a chance to submit their request.
 Thus, *I/O-intensive tasks will be making progress even while the associated thread is not running*.
 External service providers like databases and file systems have been highly optimized to handle many pending requests concurrently.


XhmikosR:
This should link to Wikipedia instead of a random site...

Trott:
HI, @MariahWine! Welcome, and thanks for the pull request.

First, I agree with @XhmikosR's comment that we ought to link to a _de facto_ standard reference like [Wikipedia](https://en.wikipedia.org/wiki/Domain_Name_System) or [MDN](https://developer.mozilla.org/en-US/docs/Glossary/DNS) instead. (I personally think the MDN definition may be the better choice here, but either one would be fine.)

I'd also add that it should be at the first occurrence of the term in the document. This is not the first occurrence. The first occurrence already links to the [Node.js docs for the `dns` module](https://nodejs.org/api/dns.html), so this might already be OK. I'm not sure it's terribly user-friendly to have one instance of the term link to one page and the next instance of the term link to a different page.

So if we want to do this, it probably involves changing the earlier link rather than adding a link here? But I'm not sure we want to do that?

The argument for doing nothing would be this: _DNS_ is a standard and easily-searchable term so I'm not sure a link is really required here. For example, the [Node.js docs for the `dns` module](https://nodejs.org/api/dns.html) do not define _DNS_, probably assuming that people using the module will either know what it means or else can do a web search for _DNS_.

On the other hand: It's not as if this guide has too many links or anything, so adding one here should be fine, I think. The question is finding a way to add it that doesn't introduce other issues.

XhmikosR:
This looks awfully a lot like spam, BTW... It might not be, but:

1. the account was registered the day this PR was made
2. the only contribution so far is this change

Trott:
> This looks awfully a lot like spam, BTW... It might not be, but:
> 
> 1. the account was registered the day this PR was made
> 2. the only contribution so far is this change

That's also consistent with someone creating an account specifically to make this pull request because they thought the link would help.

MariahWine:
Hi @XhmikosR , @Trott !

Really appreciate your time while reviewing my pull request! 
I see there is a discussion happening, so I'd like to explain myself better. 

I was studying this document with my student for a few days in a row. We both have agreed that some terms require in-depth analysis for juniors/middle to be understood for sure. I thought that this "fun-informative" balanced article would be a good match.

As I'm new in Github community, my opinion may be wrong. And yes, as @Trott noticed, I've created this account particularly for this documentation, so I don't know the exact rules of contributions. 

Hope now my pull request looks more logical to you. 
Thank you again for your time and for sharing this useful docs with us!

Trott:
@nodejs/website Thoughts on this one? I'm OK with landing it as-is if there's consensus that it's an improvement, regardless of whether or not it's ideal. I'm also OK with a "Hmmm, thanks, but I don't think we want to do this." But it would be great to get a decision, one way or the other.

PR #2481, opened 2019-08-30
fixture: change hbs rules to allow template class name formation without dashes and some typos

ghost:
Now if you run with `npm run test` locally, we've got some errors of
page checkings:

> 1. layouts/partials/footer.hbs: line 3, col 20, value must match the
> format: dash

=> This problem is caused by the special formation in {{...}}, because
this is an hbs file, just like Angular or some template pages, we
SHOULDN'T check its variable name, so ignore it by setting:
id-class-ignore-regex": "{{.*?}}" in the '.htmllintrc'.

> 2. layouts/partials/footer.hbs: line 3, col 1, tag is not closed

=> Add spaces between some tag names.

> 3. layouts/partials/navigation.hbs: line 6, col 11, tag is not closed

=> Add spaces between some tag names.

> 4. layouts/partials/header.hbs: line 10, col 9, tag is not closed

=> Add spaces between some tag names.

Now after this fixture, it looks nice:
image

### DIFF
--- a/layouts/partials/footer.hbs
+++ b/layouts/partials/footer.hbs
@@ -1,6 +1,6 @@
 <a href="#" id="scroll-to-top">&uarr; <span>{{site.scrollToTop}}</span></a>
 
-<footer{{#if className}} class="{{className}}"{{/if}}>
+<footer {{#if className}} class="{{className}}"{{/if}}>
 
   <div class="container">
     <div class="linuxfoundation-footer">
--- a/.htmllintrc
+++ b/.htmllintrc
@@ -5,6 +5,7 @@
   "attr-quote-style": false,
   "attr-req-value": false,
   "id-class-style": "dash",
+  "id-class-ignore-regex": "{{.*?}}",
   "img-req-alt": false,
   "indent-style": "spaces",
   "indent-width": 2,
--- a/layouts/partials/header.hbs
+++ b/layouts/partials/header.hbs
@@ -7,7 +7,7 @@
 
     <nav>
       <ul class="list-divider-pipe">
-        <li{{#equals path ''}} class="active"{{/equals}}>
+        <li {{#equals path ''}} class="active" {{/equals}}>
           <a href="/{{site.locale}}/">{{site.home.text}}</a>
         </li>
 
--- a/layouts/partials/navigation.hbs
+++ b/layouts/partials/navigation.hbs
@@ -3,7 +3,7 @@
     {{#each nav}}
       {{#equals @key ../key}}
         {{#each this}}
-          <li{{#equals ../../path link}} class="active"{{/equals}}>
+          <li {{#equals ../../path link}} class="active" {{/equals}}>
             {{#startswith link '/'}}
               <a href="{{link}}">{{text}}{{#if subtext}} <span class="small color-lightgray">{{subtext}}</span>{{/if}}</a>
             {{else}}
--- a/package.json
+++ b/package.json
@@ -14,7 +14,7 @@
     "load-schedule": "curl -sS https://raw.githubusercontent.com/nodejs/Release/master/schedule.json -o source/schedule.json",
     "start": "npm run serve",
     "test": "npm run test:lint && npm run test:unit && npm run test:smoke",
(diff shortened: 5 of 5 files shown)


alexandrtovmach (approved):

yous:
In my environment, `npm run test` runs htmllint against only 18 files. Can you also correct the script?
``` diff
index f1da085d..6d806381 100644
--- a/package.json
+++ b/package.json
@@ -14,7 +14,7 @@
     "load-schedule": "curl -sS https://raw.githubusercontent.com/nodejs/Release/master/schedule.json -o source/schedule.json",
     "start": "npm run serve",
-    "test:lint": "standard --fix && htmllint **/*.hbs && stylint layouts/css && markdownlint \"**/*.md\" -i \"node_modules/\"",
+    "test:lint": "standard --fix && htmllint \"**/*.hbs\" && stylint layouts/css && markdownlint \"**/*.md\" -i \"node_modules/\"",
     "test:unit": "tape tests/**/*.test.js | faucet",
     "test:smoke": "tape tests/*.smoketest.js | faucet"
   },
```

layouts/partials/footer.hbs
```diff
@@ -1,6 +1,6 @@
 <a href="#" id="scroll-to-top">&uarr; <span>{{site.scrollToTop}}</span></a>
 
-<footer{{#if className}} class="{{className}}"{{/if}}>
+<footer {{#if className}} class="{{className}}" {{/if}}>
```
yous: As you added `id-class-ignore-regex`, the latter change isn't needed. But it's your choice.
```suggestion
<footer {{#if className}} class="{{className}}"{{/if}}>
```

ghost: >In my environment, npm run test runs htmllint against only 18 files. Can you also correct the script?

I ran on window 10 and I didn't see this mistake……Well, I changed according to your ideas, maybe you can fetch the latest version of codes and have a try;)

PS：Don't forget to run with `npm i` first.

yous: Yes, maybe wildcard is treated differently based on OSes. Also, other two files can be updated, too!

XhmikosR:
Please do not add the spaces just because a random tool complains.

This was intentionally changed in https://github.com/nodejs/nodejs.org/commit/71488cdc7cb2a2c7b4c65574c0fc15535e63733c

XhmikosR:
I don't want to sound too harsh, but htmllint is pretty much useless IMO. It doesn't catch real issues. See also #2459 and #2469 for a proper HTML checking tool.

Trott:
> Please do not add the spaces just because a random tool complains.
> 
> This was intentionally changed in [71488cd](https://github.com/nodejs/nodejs.org/commit/71488cdc7cb2a2c7b4c65574c0fc15535e63733c)

In a way, this comes down to values. With no difference in rendering or user experience, do you value pretty formatting of the final HTML output (in which case you get rid of the space) or do you value readabaility/maintainability of the template files for developers (in which case you add the space)?

In this particular case, I don't care much, but I suppose if forced to make a choice, I would say the template readability is more important than avoiding an extra unsightly-but-harmless space in the HTML tag. So I'd be for adding the space. But I don't care enough to argue the point if others disagree.

ghost:
> Please do not add the spaces just because a random tool complains.
> 
> This was intentionally changed in [71488cd](https://github.com/nodejs/nodejs.org/commit/71488cdc7cb2a2c7b4c65574c0fc15535e63733c)

Why? What do you mean by 'random tool complains'? Any other errors happen?

XhmikosR:
All I'm saying is that we shouldn't do what we are doing.

Even if httmllint was a better tool (which in our case it's proven it's not), we aren't using it properly. These are handlebars templates, not HTML files.

Anyway, you guys do whatever you prefer. For me, what we have on master is fine and readable.

ghost:
> All I'm saying is that we shouldn't do what we are doing.
> Even if httmllint was a better tool (which in our case it's proven it's not), we aren't using it properly. These are handlebars templates, not HTML files.
> Anyway, you guys do whatever you prefer. For me, what we have on master is fine and readable.

Ah ha, I'm afraid you've misunderstood me, so please don't be worried or angry. What I mean doesn't challenge you but just wanna know your real reason——If you have enough reason to tell us and that looks good, maybe we can dismiss this submit and follow your ideas, this is only the place where we can chat happily ;)

PR #2483, opened 2019-09-01
Reindent header.hbs

XhmikosR:


XhmikosR:
Note that the indentation error wasn't caught by htmllint. Just pointing it out.

~~As for the end tags, they are not needed.~~

layouts/partials/header.hbs
```diff
@@ -23,11 +23,11 @@
       </ul>
     </nav>
 
-  <svg id="lang-picker-toggler" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 512 512">
-    <rect id="lang-rect" width="512" height="512" x="0" y="0" fill="none" pointer-events="fill"/>
```
osk2, line 27: I don't think slash of self-close tag can be omitted. 🤔

XhmikosR: Of course it can :)

XhmikosR: Actually, you are right. Let me try something and I'll sort this out.

XhmikosR: Ok fixed. Not sure how I got confused here :/

osk2: 👍

osk2 (approved):
LGTM

Issue #2463, opened 2019-08-25
Split npm test scripts


XhmikosR:
I had made #2405, but my patch wasn't accepted as it was.

That being said, I find the current situation inconvenient. We can't run just `standard` without running the whole batch of scripts in `npm run test:lint`.

I believe this should be addressed.

PR #2484, opened 2019-09-02
Remove extra semicolons.

XhmikosR:
'build.js' has extra semicolons, remove them together.


Trott:
> `standard --fix` which runs on `npm t` doesn't fail if it can fix the issues. This is another reason to separate the npm scripts...

Wand to include the fix to `package.json` here as a second commit so that more things like this don't creep in?

XhmikosR:
@Trott I believe this should be handled in a separate PR, see #2463. It's already pretty troublesome developing, let alone that we have `npm t` in a precommit hook which slows down things a lot...

Trott (approved):

PR #2487, opened 2019-09-03
fix: MD012/no-multiple-blanks

nschonni:
Remove multiple consecutive blanks.
Find:\n\n\n
Replace:\n\n
Filter:.md


3imed-jaberi (approved):

locale/en/blog/npm/npm-1-0-link.md
```diff
@@ -107,7 +107,9 @@ I highly doubt that a native Windows node will ever have comparable symbolic lin
 
 When there is a native windows port of Node, if that native windows port has \`fs.symlink\` and \`fs.readlink\` support that is exactly identical to the way that they work on Unix, then this should work fine.
 
-<h2 id="aside_credit_where_credit8217s_due">Aside: Credit where Credit&#8217;s Due</h2>
+But I wouldn't hold my breath. Any bugs about this not working on a native Windows system (ie, not Cygwin) will most likely be closed with `wontfix`.
```
nschonni: Don't think the conflict resolution was right here. I'll rebase this to fix it properly

locale/en/blog/release/v0.8.0.md
```diff
@@ -96,14 +96,12 @@ Requests per second:    3317.24 [#/sec] (mean)
 # 0.8.0
 Requests per second:    3795.34 [#/sec] (mean) +14.41%
 
-
```
XhmikosR, line 99: This changes line in `pre`s too.

ghost (approved):
Just a confirm：locale/en/blog/announcements/interactive-2016-ams.md, it seems that we should keep the last line as a newline (L84), but this change will remove it? If we don't have such problems, this LGTM for me.

locale/en/blog/release/v0.8.0.md
nschonni: Good catch!
Looking at the other blocks in this file, the double line in the other calls isn't consistent, so i don't think it matters in this case

nschonni:
@MaledongGit in that file, it had 2 trailing space instead of the single trailing line before the EOF

PS: If both trailing lines were removed, then MD047 would have gotten triggered

PR #2492, opened 2019-09-04
fix: Unescape backtick

nschonni:


XhmikosR:
There are a few more cases like this, for `\_` when inside code. I've been meaning to search the docs about it but I forget.

XhmikosR:
For example `FIXED\_ONE\_BYTE\_STRING`. We need a regex for this so that we find all these instances.

XhmikosR:
Alright, I figured this out. I'm gonna make a PR and we review it thoroughly.

XhmikosR (approved):

Trott (approved):

PR #2493, opened 2019-09-04
Fix wrong escapes.

XhmikosR:
When we use a code block, we don't need to escape underscores or other special characters.

/CC @nschonni


XhmikosR:
BTW my regex is probably not perfect, so I might have missed some cases. Any suggestions welcome.

```
`(.*)?\\(.*)?`
```

XhmikosR:
Alright, this is ready. It includes #2492 so it might need a rebase after that is merged.

I'm not sure how to check for this automatically. Would make total sense to have an automated check for extraneous escapes.

nschonni (approved):

nschonni:
Looks like it didn't cause a conflict with the other one landing

XhmikosR:
Rebased and fixed 12.10.0 too.

We need an automated check for this, otherwise it'll keep happening. Maybe after switching to remark-lint we could do something about it.

BridgeAR:
@XhmikosR I see that they are not necessary but I am not sure if they might break anything. AFAIK they are just not needed? If that's the case, I would not really bother.
If they indeed break something, I would include a check in our commit message [linter](https://github.com/nodejs/node/blob/76faccccbb51bb477a88ebfa3fa2a8254da37696/tools/lint-pr-commit-message.sh) in the [core](https://github.com/nodejs/node) repository. That would prevent most of these ending up in the changelog.

XhmikosR:
@BridgeAR: it changes rendering completely, so they are not just redundant, they are wrong.

I don't have any solution for the linter, at least not in this repo. But we definitely need something to check for this.

BridgeAR:
@XhmikosR linting for the commit message in core would prevent the issue in this repository in the future. Currently it seems to only apply to the release posts and those include the changelog. The changelog is build by our tooling and includes all individual commit messages. The commit messages some times include obsolete underscores and that could be checked for.

Besides that: this repository uses pre-commit hooks to verify a couple of things. It should be relatively straight forward to include a check for such cases in the hook as well.

XhmikosR:
I'm just not sure if it can be automated. I mean, how can we distinguish between a wrong escape and an escape on purpose? We can't just disallow `\`.

Anyway, that's something that should be discussed in #2501, I'm not sure I can work on this too myself :)

Trott (approved):

PR #2495, opened 2019-09-04
helpers/slashes.js: remove the process check.

### LINKED ISSUE #2494: False canonical link on your pages
* **URL**: https://nodejs.org/en/blog/ (your docs, download & other pages are affected)
* **Browser version**: Google Chrome Version 76.0.3809.132 (Official Build) (64-bit)
* **Operating system**: macOS Mojave Version 10.14.5 (18F132)

Hi,

I have noticed that a lot of pages on node.js website have a canonical link https://nodejs.org/en/false/. This link is a 404 page.

Issue is that your organic traffic is tanking on numerous pages throughout the Node.js website that has this false canonical set.

This can mainly be seen on your /guides/ directory:
Screen Shot 2019-09-04 at 11 13 58 AM

I'm reporting this as a bug because I haven't thought of any reason to canonicalize your pages like this that would yield positive results.

XhmikosR:
Sorry for the regression, my initial patch didn't include the process check, but due to a review comment I added it, which I shouldn't have.

/CC @Trott @MaledongGit please try on nix and merge ASAP.

Fixes #2494

### DIFF
--- a/scripts/helpers/slashes.js
+++ b/scripts/helpers/slashes.js
@@ -1,5 +1,5 @@
 'use strict'
 
 module.exports = function (str) {
-  return process.platform === 'win32' && str.replace(/\\/g, '/')
+  return str.replace(/\\/g, '/')
 }


scripts/helpers/slashes.js
```diff
@@ -1,5 +1,5 @@
 'use strict'
 
 module.exports = function (str) {
-  return process.platform === 'win32' && str.replace(/\\/g, '/')
+  return str.replace(/\\/g, '/')
```
XhmikosR, line 4: Alternatively, if you still want the check I can change it to a ternary.

That being said, this way is consistent with one more case we have to replace the slashes where we don't check for process.

Trott (approved):

Trott:
I'm going to go ahead and merge this right away. If it causes problems, it should be easy enough to back out.

PR #2496, opened 2019-09-04
Update devDependencies.

XhmikosR:
* nock 11.3.2 (see https://github.com/nock/nock/releases/tag/v11.3.2)
* handlebars 4.2.0 (see https://github.com/wycats/handlebars.js/blob/master/release-notes.md#v420---september-3rd-2019)

Please confirm that the nock changes don't require changes on our side; I checked the `reply()` calls and we don't seem to pass a function.


ghost (approved):

Trott (approved):

Trott:
`npm test` fails for me but in the same way it does on the master branch, so 
¯\\_(ツ)_/¯

Trott:
`npm run test:unit` and `npm run test:smoke` both pass so I guess the `nock` update is OK?

ghost:
@Trott , I did a test locally by fetching the latest version of codes, and then change according to the post, run `npm i` and see the result. Everything seems fine with me, maybe you can remove node_modules and have another try ;)

image
image

ghost:
> `npm run test:unit` and `npm run test:smoke` both pass so I guess the `nock` update is OK?

Everything seems fine with me. But since yours is  of failure, maybe we want another to have a try? I'm running on windows 10, x86, latest version of upgration.

Trott:
Looks like my problem is with a subdirectory's node_modules folder:

```console
scripts/contributor-list/node_modules/winston/README.md: 840: MD001/heading-increment/header-increment Heading levels should only increment by one level at a time [Expected: h3; Actual: h4]
```

I can delete the `node_modules` folder to get it to pass but perhaps we ought to add an exclusion for `node_modules`?

Either way, this is not an error that's introduced in this PR, so this can land.

ghost:
@Trott, node_modules seems to be excluded... Take this in '.gitignore'.

image

XhmikosR:
@Trott: node_modules should already be excluded, this is the command we have:

```
markdownlint \"**/*.md\" -i \"node_modules/\"
```

Trott:
> @Trott, node_modules seems to be excluded... Take this in '.gitignore'.
> 
> image

@MaledongGit It's excluded in git, but that's not relevant here. That just means that someone (like me) who goes and builds that subdirectory from time to time will still have those files lying around and markdownlint still flags them.

Trott:
> @Trott: node_modules should already be excluded, this is the command we have:
> 
> ```
> markdownlint \"**/*.md\" -i \"node_modules/\"
> ```

@XhmikosR Apparently not. Try this:

* Clean checkout of current master
* `npm ci`
* `(cd scripts/contributor-list && npm install )`  or `cd scripts/contributor-list && npm install && cd ..` or Windows equivalent
* `npm run test:lint` (or Windows equivalent)

Errors...  My guess is that without explicit globbing, the `-i` option assumes the path is relative to PWD

XhmikosR:
We should make it a glob then. Personally, I wasn't aware of this case.

Trott:
> We should make it a glob then. Personally, I wasn't aware of this case.

Testing and globbing isn't fixing it. Might be a straight up bug. Or maybe `-i` needs to precede the main argument/target.

Trott:
This fixes it:

```
markdownlint  -i "**/node_modules/**" "**/*.md"
```

XhmikosR:
LGTM, I was about to check it out myself.

Trott:
PR: https://github.com/nodejs/nodejs.org/pull/2507

PR #2497, opened 2019-09-04
Blog: v12.10.0 release post

BridgeAR:
Refs: https://github.com/nodejs/node/pull/29429


XhmikosR:
Note that it's wrong to escape special characters in code, for example `SLICE\_START\_END()`.

I'll fix this in #2493, but unless we have an automated way to check for this, it will keep happening...

PR #2500, opened 2019-09-05
Update node-version-data to v1.1.0.

### LINKED ISSUE #2442: TimeoutOverflowWarning when building the site
When building the site with `npm run build` I get:

```
TimeoutOverflowWarning: 4294967296000 does not fit into a 32-bit signed integer.
Timer duration was truncated to 2147483647.
```

I'm on Node.js 12.8.1 on Windows 10 64-bit but it happens on Travis too. It's just not shown because currently the build script is inside tap. You can see it on my fork though https://travis-ci.org/XhmikosR/nodejs.org/builds/574910707#L216

IMO we should run `npm run build` in an npm script and not inside tap.

XhmikosR:
@rvagg is this tested? I don't know how to test this, personally, so better be careful :)

Fixes #2442.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -346,31 +346,11 @@
       "integrity": "sha512-Phlt0plgpIIBOGTT/ehfFnbNlfsDEiqmzE2KRXoX1bLIlir4X/MR+zSyBEkL05ffWgnRSf/DXv+WrUAVr93/ow=="
     },
     "bl": {
-      "version": "1.0.3",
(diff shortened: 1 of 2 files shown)


rvagg:
I guess it's not tested. node-version-data is and the API should be stable so I don't foresee any problems here. :thumbsup:

rvagg (approved):

Trott (approved):

PR #2504, opened 2019-09-05
Split npm scripts and run them in parallel

### LINKED ISSUE #2463: Split npm test scripts

XhmikosR:
This is a first approach at addressing #2463.

Other changes:

1. removed `--fix` from `standard` because on CI this didn't make the build fail. One can still do `npm run test:lint:js -- --fix` locally if they want to
2. runs `test:lint` scripts in parallel

I still think we should move `npm run build` outside of tape and just run this before test. Build warnings aren't shown in [tape](https://github.com/nodejs/nodejs.org/issues/2442#issue-483485909)

Fixes #2463


ghost (approved):

Trott:
Seems fine to me.

XhmikosR:
@Trott how about moving `build` in scripts and not in tape? See https://github.com/nodejs/nodejs.org/issues/2442#issue-483485909

Trott:
> @Trott how about moving `build` in scripts and not in tape? See [#2442 (comment)](https://github.com/nodejs/nodejs.org/issues/2442#issue-483485909)

👍

XhmikosR:
Cool, I'll make this change in another PR.

PR #2506, opened 2019-09-05
Respect the user's preference about reduced motion.

XhmikosR:
See https://developer.mozilla.org/en-US/docs/Web/CSS/scroll-behavior#Accessibility_concerns


XhmikosR:
Can we get this merged? It's a simple change.

XhmikosR:
@Trott can we move with this? It's a pretty simple change. @nschonni seems to also approve.

Trott (approved):

PR #2514, opened 2019-09-06
Convert evolving-the-node-js-brand.md to Markdown.

XhmikosR:
Also add its images to repo since they were 404.

I had to use the wayback machine to get the images. I also compressed them losslessly while at it.


bnb (approved):
LGTM.

PR #2517, opened 2019-09-07
zh-CN: Add the missing translation for 'All Versions'

ghost:
In 'release.md' at zh-CN, 'All Versions' isn't translated yet. This is the fixture for it.


MaledongGit:
@nodejs/nodejs-cn , Please have a review, thanks!

PR #2518, opened 2019-09-07
doc: add the missing versions for Nodejs 11.x and Nodejs 10.x

ghost:
For the previous API, Nodejs 11.x and Nodejs 10.x are missing up. This is the fixture for it.

### DIFF
--- a/locale/ar/docs/index.md
+++ b/locale/ar/docs/index.md
@@ -22,6 +22,8 @@ labels:
 
 ### تبحث عن توثيق لنسخة سابقة ؟?
 
+* [Node.js 11.x](https://nodejs.org/docs/latest-v11.x/api/)
+* [Node.js 10.x](https://nodejs.org/docs/latest-v10.x/api/)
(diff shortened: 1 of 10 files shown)


Trott:
Should we add 10.x as well? 8.x is currently supported and it's in the list.

MaledongGit:
@Trott, Not sure, but I see that Nodejs the LTS now at the website...So I should add it?

ghost:
@Trott, Yes, I think LTS is just in the state of maintance, so I've added v10.x in the doc.

richardlau (approved):

XhmikosR:
Just a note, this should be automated.

ghost:
@XhmikosR , Until now we have to do it manually. Do you mean auto generate this?

XhmikosR:
Yup, because there's just so many things one can remember :p

So, automating this later, is the best option IMO.

ryzokuken (approved):

XhmikosR:
BTW I mean something like a partial. I'm not an expert in handlebars but
should be doable.

aymen94 (approved):

PR #2522, opened 2019-09-07
Replace `i` with `em`.

XhmikosR:
Mostly for consistency and being more schematically correct.

### DIFF
--- a/.markdownlint.json
+++ b/.markdownlint.json
@@ -18,7 +18,7 @@
       "div", "span", "p",
       "table", "th", "tr", "td",
       "code", "pre",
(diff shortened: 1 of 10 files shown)


Aissaoui-Ahmed (approved):

Trott (dismissed):
TBH, I think `i` is more correct in most of these instances. Most of these are not about emphasis but about setting something apart or denoting it as a string that needs to be changed or "words as words".

Trott (approved):

XhmikosR:
Yeah, the plan is to clean up the elements as we can.

PR #2526, opened 2019-09-07
Remove Ukrainian English files.

XhmikosR:
Only the first 3 files had their page title translated and nothing else.

I don't think it's worth maintaining another set of the English files just for the title. A full translation should be made.


ghost (approved):
If there's no-one translating these files, no need to be kept here.

Aissaoui-Ahmed (approved):

PR #2473, opened 2019-08-27
Ru: translation of simple-profiling.md

ghost:

PR #2527, opened 2019-09-07
Russian locale: remove untranslated files.

XhmikosR:


ghost (approved):
If there's no-one translating these files, no need to be kept here.

Trott:
@nodejs/nodejs-ru

Trott:
(I agree these can be removed if there is no translation going on. The ping was in case anyone wanted to translate one or more of these--in case they didn't realize there was lots of untranslated content.)

XhmikosR:
I'm just trying to clean up the locales. When a page has been translated, then it should be added, not before.

XhmikosR:
I see one ongoing translation in #2473. I can exclude the removal of this file assuming that PR will land.

XhmikosR:
So, how should we proceed with this?

ghost:
Until  #2473 is merged

XhmikosR:
I don't follow? I can just skip this file from this PR, if #2473 is going to be merged.

Note that this removes obsolete/old files too.

ghost:
@XhmikosR , What I mean is you can revert ONLY this file 'locale/ru/docs/guides/simple-profiling.md' but just remove the others. But I still see your file is in the state of remove.
image

Issue #2466, opened 2019-08-25
Set up Netlify for Pull Request previews?


XhmikosR:
Or something else that has the same effect.

Currently one has to pull down the changes locally to see them in action which is far from optimal.

XhmikosR:
@Trott can we do this please? It will really help with PRs. I can help setting it up after you add it to the repo.

Trott:
> @Trott can we do this please? It will really help with PRs. I can help setting it up after you add it to the repo.

You can either open a separate issue in the nodejs/admin repo, or I can transfer this issue there.

XhmikosR:
@Trott yeah, please move it where appropriate. I thought it belonged to this repo.

To clarify, I don't mind whatever solution we go with, we just need this to make our lives easier :)

PR #2531, opened 2019-09-07
fix: MD041/first-line-heading/first-line-h1

nschonni:
First line in file should be a top level heading


XhmikosR:
This definitely changes the style. IMO it's not worth it.

Screen Shot 2019-09-07 at 20 38 02-fullpage

Shouldn't be merged without the appropriate CSS changes if the maintainers decide to go with it. Also note that's why I said we need #2466.

Trott (changes requested):
I appreciate the intention, and we should definitely do something to get initial level-1 headers in the docs, but making that first sentence a header isn't the way to do it, I don't think. Open to suggestions, obviously.

XhmikosR:
We can tweak the CSS to match the current style, and keep it as a header.

Alternatively, we could just ignore these lines in markdownlint.

XhmikosR:
Alternatively, we could just ignore these lines in markdownlint.

Trott:
> We can tweak the CSS to match the current style, and keep it as a header.

That seems like an accessibility problem. It's not a header. It's just a sentence. So making it a header but rendering it to look like it's not a header could have unanticipated negative impact on assistitve-reading technology (which will almost certainly treat it as a header).

> Alternatively, we could just ignore these lines in markdownlint.

If it's only the index.md files, then that may be an acceptable alternative.

Python's pages seem to use the Python logo for the H1 header and supply alt-text, which seems like an OK approach. But our logo is inserted on all pages and isn't part of the markdown, so that may not be the way to go here.

nschonni:
Yeah, since these pages use their own template, the H1 could be figured out there if you think that line doesn't make sense as a title https://github.com/nodejs/nodejs.org/blob/master/layouts/index.hbs

XhmikosR:
After thinking about it, I think it's a good change for the document heading levels structure.

Without this we have `h2` inside the buttons and no `h1`, which isn't right.

We just need to make sure we fix the style. I can make another PR based on this branch tomorrow if you agree.

alexandrtovmach (approved):

XhmikosR:
@alexandrtovmach are you blindly approving PRs? :P

This is definitely not supposed to go in like this.

MylesBorins:
Is this still being worked on?

XhmikosR:
It shouldn't land as is unless we want the style to change that is.

https://github.com/nodejs/nodejs.org/pull/2531#issuecomment-529128957

PR #2532, opened 2019-09-07
translations: fix missing header ID `enabling-remote-debugging-scenar…

XhmikosR:
…ios`.

Note that I didn't update the files I have open PRs to remove. Also, there are probably more cases like this, I just happened to stumble upon this one.


Aissaoui-Ahmed (approved):

ghost (approved):

Trott:
HTML comments become header ids for linking? How did I not already know this??!!

ghost:
@Trott, This is a feature done by me, because as you know, non-English characters WON'T generate any id in English alphabets, so we should support non-English characters to generate their IDs.
Here's the link：https://github.com/nodejs/nodejs.org/pull/2045, Maybe you have forgotten ;)

Trott:
Now I'm curious: What was the reason for processing HTML comments rather than putting `a` tags directly in the markup? `<a name="foo"></a>`

Trott:
> Now I'm curious: What was the reason for processing HTML comments rather than putting `a` tags directly in the markup? `<a name="foo"></a>`

Answering my own question: Maybe it's to generate `id` attributes in the header elements rather than using `a` elements.

XhmikosR:
For what is worth name is deprecated. An ID should be used.

That being said, my guess would be so that the anchor targets directly the
header element and not add empty elemements for this purpose.

PR #2534, opened 2019-09-07
Fixed "or" vs. "and" confusion.

SMcCandlish:
Typo fix.

### DIFF
--- a/locale/en/about/index.md
+++ b/locale/en/about/index.md
@@ -40,7 +40,7 @@ If some of this language is unfamiliar, there is a full article on
 ---
 
 Node.js is similar in design to, and influenced by, systems like Ruby's
-[Event Machine][] or Python's [Twisted][]. Node.js takes the event model a bit
+[Event Machine][] and Python's [Twisted][]. Node.js takes the event model a bit
 further. It presents an [event loop][] as a runtime construct instead of as a library. In other systems there is always a blocking call to start the
 event-loop.
 Typically behavior is defined through callbacks at the beginning of a script


locale/en/about/index.md
```diff
@@ -40,7 +40,7 @@ If some of this language is unfamiliar, there is a full article on
 ---
 
 Node.js is similar in design to, and influenced by, systems like Ruby's
-[Event Machine][] or Python's [Twisted][]. Node.js takes the event model a bit
+[Event Machine][] and Python's [Twisted][]. Node.js takes the event model a bit
```
Trott, line 43: I think both conjunctions are idiomatic here. I don't really see much benefit to changing it, unless it makes it more understandable for people for whom English is not their primary language. I won't block this if other's think it's a good change.

Honestly, I suspect it would be best to rewrite this paragraph to eliminate the references to Ruby and Python. Those probably made more sense when those technologies were vastly more well-known than Node.js. Right now, the references to Ruby and Python don't clarify anything. It's just extra stuff. I very much doubt that there are many readers who see EventMachine or Twisted and then say, "Oh, I see, I understand how those technologies are designed and now Node.js makes a lot more sense to me."

Perhaps an indication of how not-particularly-illuminating the references are, there's a typo: Ruby's is `EventMachine` and not `Event Machine`, but no one has apparently thought to say something until right now.

Trott:
Welcome @SMcCandlish  and thanks for the pull request! I'm neutral on this one, but there's 100+ other people who can comment on pull requests, so let's see what others say.

SMcCandlish:
Folded that into a larger copyedit. As for removing the references to both previous coding contexts, I don't have a strong opinion either way (and did not catch the Ruby-related typo, either). On the or/and matter, coders in particular tend to be clear on the difference. "Or" would not be idiomatic here except in the most informal register, e.g. among teenagers. It sharply changed the meaning [in formal/technical English] from "influenced by X and Y" to "influenced by either X or Y; I don't know which". I would suspect the error got in by way of the text originally just giving a comparison (in which an "or" would work) and later being revised to claim influence.

ghost:
@SMcCandlish , Please repush your files, because there're some conflicts.

ghost (approved):
I think we should use 'or' here, for syntax, if we use 'and', 'takes' should be 'take', on the other hand, it means either of the ways is OK.

SMcCandlish:
> @SMcCandlish , Please repush your files, because there're some conflicts.

This minor "and" versus "or" bit was folded into a later copy-editing push, and can be closed.

SMcCandlish:
> I think we should use 'or' here, for syntax, if we use 'and', 'takes' should be 'take', on the other hand, it means either of the ways is OK.

The "and" (ex-"or") in question has nothing to do with the sentence containing "takes", and has no effect on that word.

PR #2535, opened 2019-09-07
General copypediting.

SMcCandlish:
Hyphenate compound adjectives.  Use commas and periods (full points) properly. Clarify confusing sentence.  Replace slangish use of "where" with "in which".  If using typical informal-English contractions, do it consistently (at least in the same sentence).  Also includes the previous "and" vs. "or" fix. [Edited the commit comment to fix an ironic/inevitable typo, ha ha.]

### DIFF
--- a/locale/en/about/index.md
+++ b/locale/en/about/index.md
@@ -6,9 +6,9 @@ trademark: Trademark
 
 # About Node.js&reg;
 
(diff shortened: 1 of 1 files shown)


ghost (commented):
> This is in contrast to today's more common concurrency model, in which OS threads
are employed. 

Here 'in which' = where, I don't think it needs changing ;)

Trott (approved):

PR #2537, opened 2019-09-08
feat: minify HTML on build

nschonni:

PR #2540, opened 2019-09-08
Stop minifying CSS on dev.

### LINKED ISSUE #2538: Don't minify CSS on dev
I was thinking to use an env var for this in `npm start` so that this helps with debugging locally.

Also, this should be made for cases like #2537

XhmikosR:
Fixes #2538.

#2537 should be adapted to make use of this too along with the other requested changes.

### DIFF
--- a/build.js
+++ b/build.js
@@ -253,7 +253,7 @@ function buildLayouts () {
         // file anyways.
         .use(filterStylusPartials())
         .use(stylus({
-          compress: true,
+          compress: process.env.NODE_ENV !== 'development',
           paths: [path.join(__dirname, 'layouts', 'css')],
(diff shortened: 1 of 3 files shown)


Fishrock123:
I'm not really sure why I am being assigned to review since I haven't done anything here in years but this seems fine I guess.

I've never used `cross-env`.

XhmikosR:
cross-env is just a package which makes setting env vars cross platform a breeze, that's all.

This change doesn't affect production, only dev. BTW I also tried enabling sourcemaps but I couldn't make it work, so I settled with just not minifying the CSS for now.

lpinca (approved):

PR #2543, opened 2019-09-08
Move inline JS to static/js/main.js.

### LINKED ISSUE #2542: Move inline JS outside of HTML
This will make it cached, linted and should reduce the HTML size. Since we have HTTP/2, it shouldn't make a noticeable difference assuming we load it asynchronously.

Just to be clear, I personally prefer to inline small JS. But in this case here, we have ~11K HTML files.

XhmikosR:
~~We should probably finish #2465, #2462 and look into #2523 before this.~~

Fixes #2542

Preview: <https://elastic-wing-9ce26c.netlify.com/en/>


XhmikosR:
How about we move all JS (maybe not the IE custom modernizr load) to a file named main.js instead of 2 files? I know we have http/2, but these are small files and this way we keep things simple.

XhmikosR:
I will go with the one file solution. This way we won't have multiple JS files and we won't need to look into concatenating them on build.

I think it's a good compromise given the nature of the project.

XhmikosR:
Marked it as ready for review.

I only left the IE modernizer check in the HTML. Let me know what you guys think. (Preview https://github.com/nodejs/nodejs.org/pull/2543#issue-315254946)

PS. I hope no one links to download.js from other Node.js sites. If so, we'll need to rethink this.

XhmikosR:
I moved everything to main.js. Now the HTML files are free from inline JS.

Trott:
@nodejs/website This could use some reviews.

alexandrtovmach (approved):

aymen94 (approved):

PR #2544, opened 2019-09-08
Output CSS in `/static/css` and switch to using stylus directly

### LINKED ISSUE #2467: Move CSS/Stylus into static folder
I have a WIP patch, I'll make a PR in the next days because it requires a couple of my other PRs to be sorted first.

If we go with this, I'm not sure if we need metalmisth-stylus anymore. We'll see how it'll go.

XhmikosR:
~~Draft because it requires #2461 and #2462.~~

Fixes #2467


Trott:
@nodesj/website This could use reviews.

Trott:
@nodejs/website This could use reviews.

alexandrtovmach (approved):

build.js
```diff
@@ -238,33 +237,37 @@ function githubLinks (options) {
...
+            .set('paths', [path.join(__dirname, 'layouts/css')])
+            .use(autoprefixer())
+            .render((error, css) => {
+              if (error) {
+                throw error
+              }
 
-      // This actually executes the build and stops the internal timer after
-      // completion.
-      metalsmith.build((err) => {
```
phillipj, line 265: This is probably a stupid question since I haven't worked with this project & metalsmith in a good while; is it deliberate to stop invoking `metalsmith.build()`?

This line removes it, but doesn't re-introduce it, right?

XhmikosR: That's one of the things this patch solves. We don't need metalsmith in order to build the CSS, we can just use stylus directly.

Not sure how things were in the past, but as it is now, this was redundant.

build.js
```diff
@@ -238,33 +237,37 @@ function githubLinks (options) {
...
 // This function builds the layouts folder for all the Stylus files.
-function buildLayouts () {
-  console.log('[metalsmith] build/layouts started')
-  const labelForBuild = '[metalsmith] build/layouts finished'
+function buildCSS () {
+  console.log('[stylus] static/css started')
+  const labelForBuild = '[stylus] static/css finished'
   console.time(labelForBuild)
 
   fs.mkdir(path.join(__dirname, 'build'), () => {
```
phillipj, line 245: Did you consider the `{ recursive: true }` option of `fs.mkdir()` to avoid the nested callbacks below? As described by docs;

```
// Creates /tmp/a/apple, regardless of whether `/tmp` and /tmp/a exist.
fs.mkdir('/tmp/a/apple', { recursive: true }, (err) => {
  if (err) throw err;
});
```

build.js
phillipj: Huh, alrighty! I gotta take a deeper dive to see how that actually works then I guess.

build.js
XhmikosR: Yeah, but I thought this change should be made separately after all the other changes have settled :)

XhmikosR: Also, this requires Node.js 10, another reason I didn't make this change yet.

phillipj: Is that requirement a problem though?

As far as I see it, we're running Node.js 10 in CI and $latest when building on the hosting server [nodejs/build/setup/www/resources/scripts/build-site.sh](https://github.com/nodejs/build/blob/master/setup/www/resources/scripts/build-site.sh#L45).

XhmikosR: No problem for me, it's just that such change should be a clear patch IMO and not sneaked in in other patches, that is all.

phillipj:
Really cool to get rid of seemingly unnecessary abstraction to use stylus 💯

PR #2539, opened 2019-09-08
Remove smoke test.

XhmikosR:
Instead run the `build` script before `npm test` on Travis. This allows us to see any warnings that may happen.

Refs https://github.com/nodejs/nodejs.org/pull/2504#issuecomment-529164515 and https://github.com/nodejs/nodejs.org/issues/2442#issue-483485909

Issue #2524, opened 2019-09-07
Lint inline Markdown JS


XhmikosR:
This will ensure consistency. We should try doing this after #2523 and use the same config as upstream.

nschonni:
Looks like it can be done with the markdown plugin https://github.com/eslint/eslint-plugin-markdown
I tried it with a --fix locally, but didn't PR anything since I find the "standard" config kind of makes a mess of the code

XhmikosR:
Yup, this should go after #2523

carterbancroft:
What's the status of this issue, is it still something that needs to be worked on?

If so, can you share an example or two of what needs linting?

XhmikosR:
IMO yeah. Every .md file needs to be linted.

carterbancroft:
Roger. I'm new here and just trying to get more context so please bear with me.

So, is there no linting on markdown files at all right now? From the code (and your PR here: https://github.com/nodejs/nodejs.org/pull/2557/files) it looks like there are markdown linting rules in place and linting scripts in package.json. Is this not happening?

Or, is it just the inline javascript within .md files that needs linting? If so any chance you could link me to a specific code block of something that needs linting as an example.

XhmikosR:
This is issue is for linting JS code in the markdown files. You can search the locale files for the following:

```
```js
#or
```javascript
```

There are hundreds of instances. Ideally we should use the same linting used in the main repo so that we are consistent, so this might need to be done along with #2523

carterbancroft:
OK gotcha, thanks.

So am I understanding correctly that there are still more linting rules from the node repo that should be implemented in this repo? I made the (probably faulty) assumption that this was the last step since this is the only remaining issue referenced in https://github.com/nodejs/nodejs.org/issues/2523. The rest are merged PRs.

XhmikosR:
TBH, I'm not aware of the internals in the upstream repo. I know they use eslint while here we use standard, which is also using eslint under the hood.

carterbancroft:
I see, so it sounds like https://github.com/nodejs/nodejs.org/issues/2523 should be figured out before there's any movement on this one. Maybe I should move conversation over there to try to get some context on that?

XhmikosR:
Not sure which one should be tackled first.

/CC @Trott @MylesBorins

Trott:
I don't know that it matters which one is done first, although I guess I'd default to doing #2523 first.

carterbancroft:
[...]

PR #2557, opened 2019-09-08
fix: MD046/code-block-style

nschonni:
Code block style of fenced to match nodejs style


locale/en/blog/feature/streams2.md
```diff
@@ -109,7 +109,9 @@ feedback.
 
 # Stream
 
-    Stability: 2 - Unstable
+```
+Stability: 2 - Unstable
```
Trott: This isn't code.

nschonni: I agree, but it's currently rendered as a code block because of the indentation, so I didn't want to change the rendering

XhmikosR:
@Trott please don't merge this!

@nschonni did you run `npm start` locally? This throws, or at least should throw, when a prism lang is specified that doesn't exist.

For example:

```
Failed to load prism syntax: console
Error: Cannot find module 'prismjs/components/prism-console.js'
Require stack:
- C:\Users\xmr\Desktop\nodejs.org\node_modules\metalsmith-prism\lib\index.js
- C:\Users\xmr\Desktop\nodejs.org\build.js
    at Function.Module._resolveFilename (internal/modules/cjs/loader.js:772:15)
    at Function.Module._load (internal/modules/cjs/loader.js:677:27)
    at Module.require (internal/modules/cjs/loader.js:830:19)
    at require (internal/modules/cjs/helpers.js:68:18)
    at requireLanguage (C:\Users\xmr\Desktop\nodejs.org\node_modules\metalsmith-prism\lib\index.js:42:11)
    at Object.<anonymous> (C:\Users\xmr\Desktop\nodejs.org\node_modules\metalsmith-prism\lib\index.js:86:11)
    at initialize.exports.each (C:\Users\xmr\Desktop\nodejs.org\node_modules\cheerio\lib\api\traversing.js:300:24)
    at C:\Users\xmr\Desktop\nodejs.org\node_modules\metalsmith-prism\lib\index.js:62:17
    at C:\Users\xmr\Desktop\nodejs.org\node_modules\lodash\lodash.js:4905:15
    at baseForOwn (C:\Users\xmr\Desktop\nodejs.org\node_modules\lodash\lodash.js:2990:24) {
  code: 'MODULE_NOT_FOUND',
  requireStack: [
    'C:\\Users\\xmr\\Desktop\\nodejs.org\\node_modules\\metalsmith-prism\\lib\\index.js',
    'C:\\Users\\xmr\\Desktop\\nodejs.org\\build.js'
  ]
}
```

XhmikosR:
We really need #2539, although to be fair, in this case these are wrapped in try/catch so it doesn't throw. But at least we would see them in our console.

nschonni:
OK, just removed the "console" ones with plain blocks

locale/en/knowledge/advanced/buffers/how-to-use-buffers.md
```diff
@@ -61,19 +61,25 @@ This initializes the buffer to a binary encoding of the first string as specifie
 
 Given that there is already a buffer created:
 
-    > var buffer = new Buffer(16);
+```
+> var buffer = new Buffer(16);
```
XhmikosR, line 65: Can you try with `bash` in such cases? Is it highlighted properly?

XhmikosR: Actually scratch that, this should happen in another PR. Just make sure any languages you specify exist in prism.

nschonni: No, bash wouldn't work since this is the Node REPL

XhmikosR:
@Trott: do you know which library is used upstream for highlighting? It seems [metalsmith-prism](https://github.com/Availity/metalsmith-prism) is archived, so it might be worth looking into switching to the same way it's done upstream in another PR.

locale/en/blog/feature/streams2.md
nschonni: Found that the main docs were using blockquotes for these, so I swapped that

Trott:
> @Trott: do you know which library is used upstream for highlighting? It seems [metalsmith-prism](https://github.com/Availity/metalsmith-prism) is archived, so it might be worth looking into switching to the same way it's done upstream in another PR.

Looking at https://github.com/nodejs/node/blob/6723169097759502da11c47389ffb3d650404417/tools/doc/html.js, I guess we're using https://github.com/remarkjs/remark-rehype.

XhmikosR:
This isn't ready IMO.

When we use fenced code blocks, we need to remove any indented code otherwise we'll end up with extra spaces in the snippets. It seems there are some cases this is missed like https://github.com/nodejs/nodejs.org/pull/2557/files#diff-74f34b44a0285dc0c6cd1f2eeed2e2b5

I didn't go through all of the changes, so I don't know if there are more.

locale/pt-br/knowledge/errors/what-are-the-error-conventions.md
```diff
@@ -14,42 +14,46 @@ No node.js, considera-se uma prática padrão lidar com erros em funções assí
...
+
+var callback = function (error, retval) {
+  if (error) {
+    console.log(error);
+    return;
+  }
+  console.log(retval);
+}
+
+<!--
```
XhmikosR: These comments will show up in the code block

XhmikosR:
LGTM except for my comment above

locale/en/knowledge/getting-started/control-flow/how-to-write-asynchronous-code.md
```diff
@@ -17,19 +17,21 @@ Many of the functions in Node.js core have both synchronous and asynchronous ver
...
+var fs = require('fs');
 
-    fs.readFile('example.file', 'utf8', function (err, data) {
-        if (err) {
-          return console.log(err);
-        }
-        console.log(data);
-    });
+fs.readFile('example.file', 'utf8', function (err, data) {
+    if (err) {
```
XhmikosR: Maybe change this to 2-space indentation. There might be more cases like this

locale/pt-br/knowledge/errors/what-are-the-error-conventions.md
```diff
@@ -14,42 +14,46 @@ No node.js, considera-se uma prática padrão lidar com erros em funções assí
...
-      type: undefined,
-      message: 'O valor não era verdadeiro!' }
-    O valor era verdadeiro.
+```javascript
+var isTrue = function(value, callback) {
+  if (value === true) {
+    callback(null, "O valor era verdadeiro.");
+  }
+  else {
+    callback(new Error("O valor não era verdadeiro!;
```
XhmikosR: Please add the missing 2 `)`

nschonni:
@Trott I think all the major stuff is resolved now. The indentation stuff is as-is for the existing examples

Trott:
> @Trott I think all the major stuff is resolved now. The indentation stuff is as-is for the existing examples

This is a big change set. @nodejs/website Anyone up for trying to give this a review?

locale/en/knowledge/getting-started/control-flow/how-to-write-asynchronous-code.md
XhmikosR: This still has 4-space indentation

XhmikosR:
@Trott we should move forward with #2523 and #2524 so that we don't have to review the snippets manually and that we are consistent. I've left a comment on #2523.

As far as this PR goes, it's definitely an improvement. I'd only go through the snippets and fix the indentation while at it. Assuming we don't use any code block languages that prism doesn't support (like `console`) this should be OK. It also fixes a broken example in the pt_BR translation.

nschonni:
Easiest way to review it is with the "whitespace ignored" option set, then you just see the code fence lines and the example fix

XhmikosR:
Yeah, but in this case whitespace does matter. So, the only safe way is viewing the rich preview of each file which takes time though.

locale/en/knowledge/intermediate/how-to-log.md
```diff
@@ -53,37 +59,45 @@ The advantage here is that the behavior of our logging mechanisms can now be mod
...
-      warn: Hull Breach Detected on Deck 7!
+```
+warn: Hull Breach Detected on Deck 7!
+```
 
 Because of the file transport we set up, winston also logged the warning to 'somefile.log'.  After the `logger.warn` call we just used, the log file, `somefile.log`, would contain the following output:
 
-      $ cat somefile.log
-      {'level':'warn','message':'Hull Breach Detected on Deck 7!'}
+```
```
XhmikosR, line 97: This (in all languages) could be `bash`

locale/en/knowledge/getting-started/control-flow/what-are-callbacks.md
```diff
@@ -12,42 +12,48 @@ layout: knowledge-post.hbs
...
+  // ... lots of hard work ...
+  if ( /* an error occurs */ ) {
+    return callback(new Error("An error has occurred"));
+  }
+  // ... more work ...
+  callback(null, d, e, f);
+}
+
+asyncOperation ( params.., function ( err, returnValues.. ) {
+    //This code gets run after the async operation gets run
```
XhmikosR: This still has 4-space indentation

XhmikosR:
This looks good to me. There are 3 unresolved comments and then it can be merged IMO.

locale/en/knowledge/intermediate/how-to-log.md
XhmikosR: Maybe just leave it as is.

alexandrtovmach (approved):

locale/en/knowledge/intermediate/how-to-log.md
Trott: Yeah, please don't use <code>```bash</code> for this!

Use that for bash scripts only. It does confusing highlighting for command-line shell stuff. Use <code>```console</code> for that instead if you really want to specify a language/formatter/highlighter.

XhmikosR: @Trott `console` is not a valid prism language thus this will fail.

nschonni: Yeah, we figured out that "console" chokes out metalsmith, unless some of the dependency updates done recently fix that

XhmikosR: See https://github.com/nodejs/nodejs.org/pull/2557#issuecomment-529292779

Trott: Ah, OK, yeah, in that case, yes, specifying nothing is 👍.  Sorry I'm having difficulty keeping up.

XhmikosR: Nope, this is how metalsmith-prism works. If we want to change this, we need to fork the project.

XhmikosR: which has been archived BTW

XhmikosR:
@nschonni can you just address the 2 indentation comments while you change the code snippets?

Otherwise someone with push rights could fix them and then just squash.

PR #2558, opened 2019-09-09
fix: MD004/ul-style

nschonni:
Unordered list style


XhmikosR:
Is this the same as upstream?

nschonni:
The docs hare primarily asterisks, but it hasn't been normalized on way or the other

XhmikosR:
Alright, it would make sense to make this change upstream and update the remark-lint config too.

ghost (approved):

XhmikosR:
I'm all for this assuming this is enforced in the upstream Markdown config.

We need to try to be as close to the main repo lint config so that developers coming from there don't need to adapt to yet another config IMO.

3imed-jaberi (approved):

nschonni:
Rebased for conflict

PR #2559, opened 2019-09-09
backpressuring-in-streams: fix wrong syntax highlighting.

XhmikosR:
These aren't JavaScript snippets.

### DIFF
--- a/locale/en/docs/guides/backpressuring-in-streams.md
+++ b/locale/en/docs/guides/backpressuring-in-streams.md
@@ -185,8 +185,7 @@ and instead with the replaced `return true;`.
 Let's take a look at a quick benchmark. Using the same example from above, we
 ran a few time trials to get a median time for both binaries.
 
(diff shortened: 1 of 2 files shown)


locale/en/docs/guides/backpressuring-in-streams.md
```diff
@@ -186,7 +186,7 @@ Let's take a look at a quick benchmark. Using the same example from above, we
 ran a few time trials to get a median time for both binaries.
 
 <!-- eslint-skip -->
```
XhmikosR: If we add linting for JS in Markdown, will this even be checked? If not I should also remove the `eslint` comments.

XhmikosR: To answer my question, they are not linted, so we can safely remove the eslint comments.

ghost (approved):

Trott (approved):

PR #2562, opened 2019-09-09
ru: Fixture of formations

ghost:
Since the MD file rule is changed, and the previous version https://github.com/nodejs/nodejs.org/pull/2473 has something wrong with
itself, this is the fixture package for it.


MaledongGit:
@vasbaz, You can review this and then continue your translation from L126 on my base ;)

PR #2563, opened 2019-09-09
Assorted link fixes.

XhmikosR:
There's probably a lot more...

But in order to speed things up, we need to make any doc links relative to the server root when possible, and use the absolute URLs only for files that are not generated in this repo.

For example, `https://nodejs.org/en/blog/` should become `/en/blog/`. Being that this change will be a big one, I first need confirmation before moving with it.

I've managed to scan for internal links for the English version in my [linkinator branch](https://github.com/nodejs/nodejs.org/compare/master...XhmikosR:master-xmr-linkinator). Travis CI build: https://travis-ci.org/XhmikosR/nodejs.org/builds/582621043
There's one failure only due to #2422, which I'm unsure how to solve.

That being said, if we agree on the relative vs absolute links, I'll make a PR with linkinator so that others can chime in and help.


XhmikosR:
Can we move forward with this?

Trott:
It's not clear to me that `https://nodejs.org/en/blog/` is better than `https://nodejs.org/en/blog`. I know old-school HTML thinking was that you should always make it explicit that it's a directory when it's not a file. That avoids an extra redirect for the end user. But this isn't necessarily a filesystem--I mean, it *is* in our case and there is a redirect--but I don't know about all of these? And supposedly `/blog` will give better SEO than `/blog/`?

Anyway, not opposing this or anything. Just explaining why I haven't offered an opinion. I'm not sufficiently informed about current best practices etc.

Let's ping @nodejs/website and see if people can approve or explain why not.

Feel free to open a smaller PR with just a few of these if there are some that are obvious slam-dunk must-change links so that they aren't held up by other links that perhaps need discussion.

XhmikosR:
The missing trailing slash is a useless redirect.

XhmikosR:
I fixed more links. There's probably even more but I want to do it in batches and also we need #2565 eventually.

XhmikosR:
OK, this is done. I don't want to push more stuff to keep it easier to review.

The only thing that I'm not sure is what's the policy to linking to translated files. I mean, I know in production we generate all the missing files from translations, but linking to them isn't consistent AFAICT.

So, for the time being, I only link to translated pages.

Trott:
Lots of files changed, but fewer than 300 lines. Anyone up for reviewing this? @nodejs/website (I'd hate to ask @XhmikosR to split this into multiple PRs to make reviewing easier but if that's what we gotta do to get some folks to review it sufficiently, then so be it.)

locale/ca/docs/index.md
```diff
@@ -43,7 +43,6 @@ La [secció d'ES6](/en/docs/es6/) descriu l'arbre dels grups de les característ
 característiques són activades per defecte en Node.js, juntament amb enllaços explicatius. També mostra com trobar
 quina versió de V8 fa servir una versió particular de Node.js.
 
-## Preguntes freqüents
+## Guides
```
alexandrtovmach, line 46: Why do you changed translation?

alexandrtovmach (approved):

locale/ca/docs/index.md
XhmikosR: Because there's no `/faq` link and I don't speak `ca`.

locale/ko/about/resources.md
```diff
@@ -8,21 +8,21 @@ title: 로고와 그래픽
 
 ## Logo Downloads
 
- Please review the <a href="/about/trademark/">trademark policy></a> for information about permissible use of Node.js&reg; logos and marks.
+Please review the [trademark policy](/en/about/trademark/) for information about permissible use of Node.js&reg; logos and marks.
 
- Guidelines for the visual display of the Node.js mark are described in
- the <a href="/static/documents/foundation-visual-guidelines.pdf">Visual Guidelines></a>.
+Guidelines for the visual display of the Node.js mark are described in
+the <a href="/static/documents/foundation-visual-guidelines.pdf">Visual Guidelines></a>.
```
yous: This can be synced.
```suggestion
the [Visual Guidelines](/static/documents/foundation-visual-guidelines.pdf).
```

PR #2564, opened 2019-09-09
Work around the blog/index.md in the first year.

### LINKED ISSUE #2422: Wrong path when there's no previous yearly archive

XhmikosR:
This is quite hacky, but I don't know the root cause, so I couldn't think of a better way.

Regardless, this fixes the issue. I just point to the blog index page when year is 2011. Alternatively we could hide the older button completely, but the `|` separator makes it complex.

Requires #2563.

Fixes #2422.

### DIFF
--- a/layouts/blog-index.hbs
+++ b/layouts/blog-index.hbs
@@ -31,11 +31,17 @@
         {{/if}}
 
         {{#if pagination.prev.path}}{{#if pagination.next.path}}
+        {{#equals (slashes pagination.next.path) 'blog/index.md'}}
+        {{else}}
         |
+        {{/equals}}
         {{/if}}{{/if}}
 
         {{#if pagination.next.path}}
+        {{#equals (slashes pagination.next.path) 'blog/index.md'}}
+        {{else}}
         <a href="/{{ site.locale }}/{{ pagination.next.path }}/">Older &gt;</a>
+        {{/equals}}
         {{/if}}
       </nav>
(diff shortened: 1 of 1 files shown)


XhmikosR:
@phillipj: could this be a bug in metalsmith-yearly-pagination?

phillipj:
Good question!

I wouldn't be surprised if this was metalsmith-yearly-pagination' fault and plausibly should be fixed there. I'm more than happy to push a patch release if anyone finds the culprit in [phillipj/metalsmith-yearly-pagination](https://github.com/phillipj/metalsmith-yearly-pagination).

XhmikosR:
Unfortunately, I'm not very familiar with that codebase. :/

I'll make a PR with other tweaks there, though.

XhmikosR:
Alright, I made an upstream issue. Maybe @Trott could help us out there so that we don't have to use this workaround here.

XhmikosR:
I changed the workaround to hide completely the link. I think it makes more sense and it's what I'd expect if we didn't hit this bug.

I'd still like to fix this upstream, but I had no luck so far there.

XhmikosR:
Closing this since the issue should be fixed upstream and we'll get it sorted here with a dep update.

PR #2565, opened 2019-09-09
Add linkinator.

XhmikosR:
Currently only scans the English files only.

I think we'll need to move this a script and use linkinator's API to loop through all languages.

Note that this solution works on Windows too hence why I don't use `&` in the script. Also, note that I had to add a delay since the `npm run serve` scripts was starting after linkininator.

Upstream discussion: <https://github.com/JustinBeckwith/linkinator/issues/82>

Any help is welcome.


richardlau:
Just for reference, I abandoned a PR in core that ran linkinator on the API docs: https://github.com/nodejs/node/pull/27267

There were some false positives (mainly links to Google's V8 docs, https://github.com/JustinBeckwith/linkinator/issues/24).

XhmikosR:
Yeah, this is for internal links, mostly. Hence why I suggest to clean up absolute vs relative links so that we manage to at least scan successfully the internal links properly.

XhmikosR:
OK, this only fails due to #2564.

XhmikosR:
This should pass now. It should be better than nothing, and assuming we extend it later to scan all locales, it should be a good step forward.

Still doesn't scan 3rd-party links, nor `nodejs.org` links but we can tweak it further.

XhmikosR:
Forgot to emphasize, it's definitely not a bulletproof solution. I mean, for starters, we shouldn't need to run our server, but it doesn't work with the current dist folder structure.

So, we definitely need to do this via the API later.

XhmikosR:
For example, I don't get why the broken links I fixed on #2604 weren't reported.

EDIT: could be because they point outside of `/en` maybe

Aissaoui-Ahmed:
Hi @XhmikosR 

you can use configuration in file `linkinator.config.json`

alexandrtovmach:
@XhmikosR Why we can't use this for all languages?

XhmikosR:
It doesn't work with our folder structure, so doing `linkinator build/` just doesn't work.

We need to switch to using the API later for each folder, see https://github.com/JustinBeckwith/linkinator/issues/82 and a very first patch https://github.com/nodejs/nodejs.org/compare/master...XhmikosR:master-xmr-linkinator-2

alexandrtovmach (approved):
Let's try this tool. Sounds optimistic

PR #2570, opened 2019-09-12
fixture: make the logic of both non-English and English characters filtering with the same logic

ghost:
1. We should make sure that either your English anchor name or non-English anchor name is filtered with the same logic, for non-English characters, we should also include mid-spaces filtered out.
2. Move the ANCHOR_COMMENTREG out of the scope as the global-based one.
3. Remove the 'g', because it will remember the lastIndex where it began to search for the last time.

### DIFF
--- a/scripts/plugins/anchor-markdown-headings.js
+++ b/scripts/plugins/anchor-markdown-headings.js
@@ -12,12 +12,19 @@
  * `<!---->` to quote your English anchor name inside, with your
  * own title beside it.
  */
-module.exports = function anchorMarkdownHeadings (text, level, raw, slugger) {
-  // Check whether we've got the comment matches
-  // <!--comment-->, <!--comment --> and even <!-- comment-->
-  // (20 hex = 32 dec = space character)
-  const ANCHOR_COMMENTREG = /<!--\x20?([\w|-]+)\x20?-->/gi
 
+// Check whether we've got the comment matches
+// <!--comment-->, <!--comment --> and even <!-- comment-->
+// (20 hex = 32 dec = space character)
+// Only need to find one that matches the Regex, we cannot use
+// 'g' here because it will continue searching for the rest string,
+// and you have to reset the lastIndex, which isn't a necessary for
+// our current situation.
+// For more, you can see:
+// https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/lastIndex#Description
+const ANCHOR_COMMENTREG = /<!--\x20?([\w\x20-]+)\x20?-->/
+
+module.exports = function anchorMarkdownHeadings (text, level, raw, slugger) {
   let anchorTitle = null
 
   // If we've checked the title has a comment symbol,
@@ -27,14 +34,14 @@ module.exports = function anchorMarkdownHeadings (text, level, raw, slugger) {
   if (anchorTitleArray !== null) {
     anchorTitle = anchorTitleArray[1]
   } else {
-    // For others, directly replace all non-English characters
-    // to '-' in the middle, only keep the English characters
     anchorTitle = raw
-      .replace(/(\[([^\]]+)]\([^)]+\))/g, '$2')
-      .replace(/[^\w]+/g, '-')
-      .replace(/-{2,}/g, '-')
-      .replace(/(^-|-$)/g, '')
   }
+
+  anchorTitle = anchorTitle.replace(/(\[([^\]]+)]\([^)]+\))/g, '$2')
+    .replace(/[^\w]+/g, '-')
+    .replace(/[\x20]+/g, '-')
+    .replace(/-{2,}/g, '-')
+    .replace(/(^-|-$)/g, '')
 
   if (!anchorTitle) {
     return `<h${level}>${text}</h${level}>`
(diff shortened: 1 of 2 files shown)


XhmikosR (approved):

Aissaoui-Ahmed (approved):

scripts/plugins/anchor-markdown-headings.js
```diff
@@ -12,12 +12,13 @@
...
-module.exports = function anchorMarkdownHeadings (text, level, raw, slugger) {
-  // Check whether we've got the comment matches
-  // <!--comment-->, <!--comment --> and even <!-- comment-->
-  // (20 hex = 32 dec = space character)
-  const ANCHOR_COMMENTREG = /<!--\x20?([\w|-]+)\x20?-->/gi
 
+// Check whether we've got the comment matches
+// <!--comment-->, <!--comment --> and even <!-- comment-->
+// (20 hex = 32 dec = space character)
+const ANCHOR_COMMENTREG = /<!--\x20?([\w|-]+)\x20?-->/gi
```
XhmikosR: BTW any reason you don't use a space here?

Trott: Is there a bug in the regexp? The `|` seems like an error. Don't we want to allow combinations of `\w` and `-`? Shouldn't it be this instead?

```suggestion
const ANCHOR_COMMENTREG = /<!--\x20?([\w-]+)\x20?-->/gi
```

As it is, the regexp will not match `<!-- a-b-c -->` but the above suggested change will.

ghost: @Trott , yes, I've fixed it, and what I meant was `\w` or `-`, so here '|' means 'or', but we can also ignore it.

ghost: > BTW any reason you don't use a space here?

@XhmikosR , You mean add '\s' inside? OK, at the very beginning, I didn't think of that ;)

Trott (approved):

MaledongGit:
@Trott and @XhmikosR , In fact I intended to put the global-based 'ANCHOR_COMMENTREG ' out of the scope, however it CANNOT satisfy my meet of the newly-added unit test UNTIL I put it into the function body, and I don't know why it happens like this? But actually this has fixed this issue for now.

XhmikosR:
I don't have any ideas about the issue you are hitting, but makes me think something isn't right in the plugin code or the tests.

tests/scripts/anchor-mardown-headings.test.js
```diff
@@ -76,6 +76,21 @@ test('anchorMarkdownHeadings', (t) => {
     t.equal(output, expected)
   })
 
+  t.test('correctly parses markdown heading with empty spaces', (t) => {
+    const text = '这是<a href="b">链接</a>的<a href="d">测试！</a>'
+    const level = 2
+    const raw = '<!-- empty  spaces - id -->这是[链接](b)c[测试！](d)'
```
Trott, line 82: This comment will not match `ANCHOR_COMMENTREG`. Is it supposed to?

Trott: Oh, wait, never mind, you updated `ANCHOR_COMMENTREG` to account for spaces.

MaledongGit: @Trott, I'm doing an experiment on win 10, x86. You can have a try by moving the `ANCHOR_COMMENTREG` out of the function as a global one, you will find it very strange not to pass the newly-added unit test (I'm NOT sure why it will add '-c' at the end of 'empty-spaces-id', but it won't if I use it directly in the function).

ghost: @Trott, I found the problem, it's that if I use 'g', it will continue searching according to the lastIndex. But what makes me feel strange is that if so, only one unit test should pass, while the others fail. This problem is still in pending and disscussion, but this problem is fixed.

scripts/plugins/anchor-markdown-headings.js
```diff
@@ -12,12 +12,19 @@
...
+// Check whether we've got the comment matches
+// <!--comment-->, <!--comment --> and even <!-- comment-->
+// (20 hex = 32 dec = space character)
+// Only need to find one that matches the Regex, we cannot use
+// 'g' here because it will continue searching for the rest string,
+// and you have to reset the lastIndex, which isn't a necessary for
+// our current situation.
+// For more, you can see:
+// https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/lastIndex#Description
+const ANCHOR_COMMENTREG = /<!--\x20?([\w\x20-]+)\x20?-->/i
```
XhmikosR: I still wonder why this can't be

```js
const ANCHOR_COMMENTREG = /<!-- ?([\w -]+) ?-->/
```

ghost: This is also OK, but I tend to use \x20 or \s, something like this as a symbol.

XhmikosR: I also removed `i` since we don't care about the case here. `\s` is fine too, but it also allows newlines apart from spaces and tabs.

Anyway just thinking out loud.

ghost: Ah I see, \s is more strict than what I meant.

XhmikosR: Which we don't want in our case since we don't use newlines in the comment. But removing `i` seems fine.

ghost: OK, I'll submit again. Thanks for your patience

hg-pyun (approved):

PR #2575, opened 2019-09-15
zh-CN: fixture of 'evangelism' in the simplified Chinese

ghost:
'evangelism' is a more regional word, so it'd be better to translate it into '宣传'.


ghost:
@nodejs/nodejs-cn : Please have a review, thanks!

ZYSzys (approved):

PR #2576, opened 2019-09-15
zh-CN: fixture of some translations

ghost:
To translate the 'zh-CN' as the simplified Chinese, while the 'zh-tw' as the traditional Chinese, it's the normal behavior to do something like this.


ghost:
@nodejs/nodejs-cn, please have a review, thanks!

ZYSzys (approved):

3imed-jaberi (approved):
LGTM 💯

PR #2460, opened 2019-08-25
chore: Swap Markdownlint for remark-lint

nschonni:
Node org has styleguide defined in remark-preset-lint-node so makes sense to leverage here

PR #2580, opened 2019-09-16
chore: Swap Markdownlint for remark-lint

nschonni:
Re-opening #2460 because GitHub closed it on a force-push


.markdownlint.json
```diff
@@ -1,31 +0,0 @@
...
-  "default": true,
-  "MD004": false,
-  "MD013": false,
-  "MD022": false,
-  "MD024": false,
-  "MD025": false,
-  "MD026": false,
-  "MD027": false,
-  "MD032": false,
-  "MD033": {
```
ghost, line 14: Are you sure if this should be also removed or kept in your change?

nschonni: Technically this repo could run both, but I think the idea would be to clean them up with one or the other

XhmikosR:
I'm definitely in favor of this assuming https://github.com/nodejs/remark-preset-lint-node/pull/21 lands, released and this PR makes the minimum possible overwrites.

The more closer we are to the upstream rules there better. My main concern is the different markdown renderer we use here, but hopefully it won't cause any issues.

XhmikosR:
@nschonni are the current rules the minimum needed ones? Do we cover most of what we covered before?

Also, I wonder, does remark-lint have a cache feature?

nschonni:
I don't think this actually covers everything that Markdownlint does, but it has some more that it doesn't. A PR landed to added the "recommended" preset to the node baseline, but it needs a new release to include it.
Not sure what you mean by caching in this case

Trott:
> A PR landed to added the "recommended" preset to the node baseline, but it needs a new release to include it.

Would it be useful for me to publish a release at this time?

nschonni:
Sure, maybe land the other minor dependency cleanup one first though.

nschonni:
For now, I've submitted a few PRs so some of the rules can be re-enabled

Trott:
> Sure, maybe land the other minor dependency cleanup one first though.

Published.

nschonni:
Thanks, bumped the package and disabled newly failing rules

XhmikosR:
Re: caching, I meant like stylelint or eslint has, so that consequent runs are faster. Maybe it doesn't have this feature, no big deal for now.

So, to summarize:

1. I personally care about covering most of the things we covered before
2. We need to get the suppressions as close to zero eventually

That being said, this shouldn't block us from merging this. We can tackle the rules as we go.

MylesBorins (approved):
RSLGTM

package.json
```diff
@@ -71,10 +71,12 @@
     "delay-cli": "^1.1.0",
     "faucet": "0.0.1",
     "linkinator": "^1.6.0",
-    "markdownlint-cli": "^0.18.0",
     "nock": "^11.4.0",
     "pre-commit": "^1.2.2",
     "proxyquire": "^2.1.3",
+    "remark-cli": "^7.0.0",
+    "remark-frontmatter": "^1.3.2",
+    "remark-preset-lint-node": "^1.10.1-0",
```
Trott: Are we good to land this once we bump to 1.10.1?

nschonni: I think so, I've re-installed and rebased

PR #2585, opened 2019-09-18
Broken link solved.

effy-coding:
/ko/docs/es6/

==>

/ko/docs/es6.md/

### DIFF
--- a/locale/ko/docs/index.md
+++ b/locale/ko/docs/index.md
@@ -66,7 +66,7 @@ The [ES6 section](/en/docs/es6/) describes the three ES6 feature groups, and det
 
 ## ES6 기능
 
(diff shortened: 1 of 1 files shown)


ghost:
First many thanks for your 1st contributions!

Sorry, this isn't a bug actually, because we can convert a normal md into an HTML through build. If you see the normal English and other version, it can directly go to the right page, as well as the Korean version. 

If I change like what you said, it cannot work properly, you can have a play by using something like:

image


I have to close your request, but if you have any other requests, feel free to continue ;)

Trott:
(See resulting HTML https://nodejs.org/ko/docs/. The link works there. This change would break it there.)

PR #2587, opened 2019-09-18
Fix `githubLinks()` to work on Windows too.

XhmikosR:
Not sure if there's a better solution, but I think this is the last thing I notice differently on Windows.

### DIFF
--- a/build.js
+++ b/build.js
@@ -213,7 +213,7 @@ function withPreserveLocale (preserveLocale) {
 // This middleware adds "Edit on GitHub" links to every editable page
 function githubLinks (options) {
   return (files, m, next) => {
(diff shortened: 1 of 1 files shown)


ghost (commented):
It seems on windows I'm not effected by the original code...? So is this really a bug? Can we reproduce it?

XhmikosR:
Definitely happens here. Only the security page has the Edit on GitHub button without this patch.

XhmikosR:
See this for example https://frosty-almeida-4843d4.netlify.com/en/docs/

It should have the button, but it doesn't.

ghost (approved):

PR #2597, opened 2019-09-23
fix(guide): dont-block-the-event-loop.md

objectisundefined:
incorrect example code

### DIFF
--- a/locale/zh-cn/docs/guides/dont-block-the-event-loop.md
+++ b/locale/zh-cn/docs/guides/dont-block-the-event-loop.md
@@ -256,19 +256,19 @@ app.get('/redos-me', (req, res) => {
 var obj = { a: 1 };
 var niter = 20;
 
-var before, res, took;
+var before, str, pos, res, took;
 
 for (var i = 0; i < niter; i++) {
   obj = { obj1: obj, obj2: obj }; // 每个循环里面将对象 size 加倍
 }
 
 before = process.hrtime();
-res = JSON.stringify(obj);
+str = JSON.stringify(obj);
 took = process.hrtime(before);
 console.log('JSON.stringify took ' + took);
 
 before = process.hrtime();
-res = str.indexOf('nomatch');
+pos = str.indexOf('nomatch');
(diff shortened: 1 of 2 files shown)


XhmikosR:
This will need to be applied to translations too, if it's approved.

objectisundefined:
@XhmikosR ok

locale/zh-cn/docs/guides/dont-block-the-event-loop.md
```diff
@@ -256,19 +256,19 @@ app.get('/redos-me', (req, res) => {
 var obj = { a: 1 };
 var niter = 20;
 
-var before, res, took;
+var before, str, pos, res, took;
 
 for (var i = 0; i < niter; i++) {
-  obj = { obj1: obj, obj2: obj }; // 每个循环里面将对象 size 加倍
+  obj = { obj1: obj, obj2: obj }; // Doubles in size each iter
```
XhmikosR: The comments could stay since they are translated.

objectisundefined: Fixed.

Trott:
It's not clear to me that the existing example code is incorrect. This seems to use three different variables for no reason. (Or maybe it's more clear that way? Probably makes sense to rename `res` if that's the case?)

If you don't like the variable re-use, another option is to get rid of the variable assignments entirely. The point of the code is the impact of calling `JSON` methods. The variable assignments are irrelevant, although I guess they make the code seem less contrived.

Trott:
(My comments above are non-blocking. If everyone else thinks this is an improvement, then that's fine by me. I don't think this change makes it any worse or anything like that.)

objectisundefined:
@Trott variable `str` in the existing example code was not declared.

Trott:
> @Trott variable `str` in the existing example code was not declared.

Ah! I missed that! Thanks! Yes, LGTM!

Trott (approved):

ghost (approved):
Good catch!

PR #2602, opened 2019-09-25
Add initial .browserslistrc file

### LINKED ISSUE #2464: Supported Browsers
As things are right now, I can't find what the supported browsers are.

I think we should first start by adding a .browserslistrc file in the repo root, which will be used by autoprefixer.

There's also some ancient code like a `document.querySelectorAll` polyfill https://github.com/nodejs/nodejs.org/blob/e8e7d3192c418282626083189463cdcfc3370947/static/js/download.js#L5-L28

which I'm pretty confident we can get rid of nowadays.

XhmikosR:
This is a very basic patch, mostly opened for discussion.

To visualize what's supported with this patch please check: <https://browsersl.ist/?q=%3E%3D+0.2%25%2C+last+2+major+versions%2C+not+dead%2C+Android+%3E%3D+6%2C+Chrome+%3E%3D+45%2C+Edge+%3E%3D+12%2C+Explorer+%3E%3D+10%2C+Firefox+%3E%3D+38%2C+iOS+%3E%3D+9%2C+Safari+%3E%3D+9>

and this is with the defaults: <https://browsersl.ist/?q=defaults>

Notes:
* for Firefox we should probably use `Firefox ESR`
* for Chrome, I'm not sure what the minimum version should be
* for iOS/Safari I'm not sure either

Refs #2464


XhmikosR:
Alright, I changed the patch to be defaults + IE >= 10, see https://browsersl.ist/?q=defaults%2C+Explorer+%3E%3D+10

This should be the safest approach since without a .browserslistrc file the defaults are used.

XhmikosR:
Can we get a review about this? The solution I went with is as close as it gets to the current master.

IE9 does not work due to `requestAnimationFrame`, although the content is fine. Maybe I should add IE9 too?

This is the diff with IE9 added vs the current patch (which is defaults + IE>=10): https://gist.github.com/XhmikosR/0ffdac274b042e3e89f2c8c45d0943dd/revisions

XhmikosR:
This is the diff with IE9 added too https://gist.github.com/XhmikosR/0ffdac274b042e3e89f2c8c45d0943dd/revisions

XhmikosR:
BTW if we decide to sort of support IE9, we should tweak the scroll to top script to not use `requestAnimationFrame` if it's not available. Everything else works AFAICT.

lpinca (approved):

lpinca:
I think this is fine and can be revisited later if it turns out to be too restrictive.

nschonni:
Is this used by any other tools besides the `.use(autoprefixer())` line?

XhmikosR:
Currently no AFAICT, but it can be used by other tools.

XhmikosR:
So, how about IE 9?

nschonni:
Not sure if it matters much if this setup will be eventually replaced with https://github.com/nodejs/nodejs.dev/. Whatever browsers it's working with today is probably OK

XhmikosR:
Well, IE9 isn't currently specified in master. But apart from the JS error due to `requestAnimationFrame`, which we could work around, the display is good enough on IE9 AFAICT.

MylesBorins:
@XhmikosR I think we can move forward with this as is right now and in a follow up PR get a solution to fix IE9 and update the support table. Thoughts?

MylesBorins (approved):
LGTM

XhmikosR:
SGTM. I'll leave #2464 open for discussion

PR #2606, opened 2019-09-26
Replace a few HTML entities with their Unicode symbol

XhmikosR:
Since we use UTF-8 and these changes are in Markdown, this should be fine.

Draft because we need #2469 to make sure everything is still OK.


XhmikosR:
Preview: <https://angry-brahmagupta-2a359f.netlify.com/en/>

Aissaoui-Ahmed (approved):

PR #2608, opened 2019-09-26
blog/release/v12.11.0.md: fix wrong escape

XhmikosR:
I'm not sure how these keep being added, but anyway this was wrong because it's wrapped in code.


richardlau:
> 
> 
> I'm not sure how these keep being added, but anyway this was wrong because it's wrapped in code.

I believe the changelogs are partially generated by https://github.com/nodejs/changelog-maker (cc @nodejs/releasers) and the escaping is done in https://github.com/nodejs/changelog-maker/blob/867cf90110b3539d33f503360df1a94840bfd868/commit-to-output.js#L5-L8.

XhmikosR:
Perfect, then we need that tweaked to take into account backticks and not escape those then.

codebytere (approved):

XhmikosR:
I made https://github.com/nodejs/changelog-maker/issues/58 to track this upstream

PR #2612, opened 2019-09-28
Update dependencies list WIP

juanarbol:
`http-parser` was removed [see](https://github.com/nodejs/node/pull/29589), I just updated dependencies file


locale/en/docs/meta/topics/dependencies.md
```diff
@@ -39,13 +39,13 @@ asynchronously at the operating system level.
 
 - [Documentation](http://docs.libuv.org/)
 
-### http-parser
+### llhttp
 
-HTTP parsing is handled by a lightweight C library called http-parser. It is
+HTTP parsing is handled by a lightweight C library called llhttp. It is
```
Trott:
```suggestion
HTTP parsing is handled by a lightweight TypeScript and C library called llhttp. It is
```

You'll need to word-wrap to 80 chars if you accept this change.

juanarbol: Thanks for reviewing! Fixed at 6be662d

Trott:
This LGTM but a 👍 from @indutny to make sure we've got it exactly right (especially the removing http-parser part) would be great.

MylesBorins (approved):
LGTM

PR #2615, opened 2019-09-28
CI: switch to Node.js 12.

XhmikosR:


.travis.yml
```diff
@@ -1,6 +1,6 @@
 cache: npm
 language: node_js
 node_js:
-  - 10
+  - 12
```
Trott: Might it make sense to always use the latest stable?

```suggestion
  - node
```

Trott (approved):

.travis.yml
XhmikosR: TBH that's your call entirely. I don't mind either way.

XhmikosR: I'm not sure if `node` is an alias for the latest version, or the latest LTS. Probably the latest version judging from the docs. https://github.com/nvm-sh/nvm#long-term-support

My intention with this PR was to move to the stable 12.x version because it's going to be LTS soon.

Trott: `lts/*` will always provide the most recent LTS branch, so that's also an option.

XhmikosR:
Yeah but that doesn't work yet with actions AFAICT

PR #2620, opened 2019-09-29
CSS tweaks.

XhmikosR:
* switch to `rem`s (mostly for font-size for now)
* move a few selectors outside of styles.scss
* add `margin-bottom` in header logo so that there's more room for tap targets

I still think many `font-size`s are too small, but this should be tackled separately. In this PR, I tried to keep things consistent with what it was.

Preview: <https://youthful-nightingale-b83864.netlify.com/en/>

Note that this will conflict with #2530 but the Sass switch should probably come first.

### DIFF
--- a/layouts/css/_utils.scss
+++ b/layouts/css/_utils.scss
@@ -6,12 +6,32 @@
   background-color: $node-gray !important;
 }
 
+.color-lightgray {
+  color: $light-gray;
+
+  a:hover & {
+    color: $white;
+  }
+}
+
 .table-no-border-no-padding {
   border-spacing: 0;
 
   td {
     padding: 0;
   }
+}
+
+.full-width {
+  width: 100%;
+}
+
+.small {
+  font-size: .625rem;
+}
+
+.no-padding {
+  padding: 0;
 }
 
 // borrowed from Bootstrap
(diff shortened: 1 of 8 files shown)


layouts/css/_utils.scss
```diff
@@ -14,6 +22,18 @@
   }
 }
 
+.full-width {
+  width: 100%;
+}
+
+.small {
+  font-size: 10px;
```
nschonni: Should this be rem too?

nschonni (commented):
I'm not familiar enough with Stylelint to tell if rem can be enforced https://github.com/stylelint/stylelint/blob/master/docs/user-guide/configuration.md#rules

layouts/css/_utils.scss
XhmikosR: Good catch, I'll fix this. This is due to the rebase after the switch to Sass.

XhmikosR:
I could only find https://github.com/meowtec/stylelint-no-px but I don't know if it's worth bothering with it now. You are welcome to try it after this lands :)

XhmikosR:
I'm gonna drop this for now. We are still blocked by https://github.com/nodejs/build/pull/1947

I might do some of the changes that don't affect the final CSS in another PR.

PR #2621, opened 2019-09-29
server.js: only build the CSS when CSS changes, not the layouts 

XhmikosR:
~~Draft because it requires #2544 merged, but most importantly because there's [a bug](https://github.com/paulmillr/chokidar/issues/888) in chokidar 3.1.1 that globs don't work on Windows so I can't test properly yet.~~

But the idea is that since the CSS is separated completely, we shouldn't rebuild everything when just the CSS changes.


XhmikosR:
I believe there are a few cases this won't be perfect, but I think it's better this way because it's a lot faster for development.

ghost (approved):

Trott:
@nodejs/website This could use some more reviews.

aymen94 (approved):

PR #2626, opened 2019-09-29
package.json: require Node.js 10.

XhmikosR:
So that we use `mkdir`'s recursive option later.

Or should I make this `>=10.12.0` since that's when it was added? https://nodejs.org/api/fs.html#fs_fs_mkdir_path_options_callback

/CC @phillipj

### DIFF
--- a/package.json
+++ b/package.json
@@ -25,7 +25,7 @@
   "author": "Node.js Website Working Group",
   "license": "MIT",
   "engines": {
-    "node": ">=8"
+    "node": ">=10"
   },
   "standard": {
     "ignore": [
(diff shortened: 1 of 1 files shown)


phillipj:
Your suggestion of being even more specific with `>=10.12.0` sounds reasonable to me. This is mainly for documenting our requirements, as it doesn't have anything to do with what's used in either CI or the hosting server (as mentioned in https://github.com/nodejs/nodejs.org/pull/2544#discussion_r329349604).

phillipj (approved):
LGTM

PR #2627, opened 2019-09-29
build.js: make use of `mkdir`'s `recursive` option

XhmikosR:
~~Draft because it'll needs some of my other PRs merged first.~~

Better check the non-whitespace diff: https://github.com/nodejs/nodejs.org/pull/2627/files?w=1


phillipj:
Great, thx for following up 👍

PR #2628, opened 2019-09-29
Revert "Update dont-block-the-event-loop.md (#2479)"

XhmikosR:
This reverts commit 5455032a9727aad9194eb101d6c2316d94d68a5f.

My reasoning is simple (I think).

1. first of all DNS isn't only mentioned in this page, so why add it specifically here? If we go with this, then we'd need to link to hundreds of things per page
2. secondly and most important, I don't think we should link to such sites. I don't have anything against this specific site, I just believe we should only link to sources everyone (or most people) trust like Wikipedia

I commented in #2479 but I don't think this was discussed enough, especially with #1971 open and pending.

/CC @phillipj @Trott @fhemberger


Trott:
Yeah, that was merged with no approvals and unaddressed comments. Should be reverted.

Trott (approved):

phillipj (approved):

osk2 (approved):

PR #2629, opened 2019-09-29
category-index.hbs: fix generation on Windows

XhmikosR:
I would seriously like to get this fixed where it's broken, but it's either metalsmith-collections or metalsmith itself AFAICT...

Anyway, I hope this is the last place :/


XhmikosR:
BTW I've pinpointed the root cause of the Windows backslashes on metalsmith/permalinks. I have submitted upstream patches, and there's one Windows test failure which is causing exactly these issues we hit here.

XhmikosR:
This was noticeable in `http://localhost:8080/en/blog/announcements/`:

Screenshot_2019-10-09 Announcements Node js

On `*nix` the behavior should be the same, but someone should try just to be safe.

ghost (approved):

PR #2633, opened 2019-10-01
Limit locales build

XhmikosR:
This is something I've been experimenting for some time. Feel free to benchmark it and share the results.

The idea is that since we have many locales and thousands of files, running the forEach loop at once saturates the machines. I experimented with `threads + 2` and `threads * 1.5` and I find the latter a better solution, especially for machines with 4 cores and 8 threads; machines with CPUs with 6 cores and 12 threads don't have this issue.

This is an experiment that seems to improve things for me, at least for development. I'm opening this for discussing and any further tweaks.


XhmikosR:
After further investigation, I'm not sure if limiting the cores is the right solution. I believe the root cause is that we open too many files simultaneously.

This patch does help a bit for development, though.

Issue #2113, opened 2019-02-24
Use bot for dependency management


aymen94:
Many repos have management of automatic dependence,
Why not use a bot like:
[dependabot](https://dependabot.com/) or others.

otherwise, we have to do it manually.
#2112 

what do you think?
@nodejs/website

JustinBeckwith:
FWIW, we use [renovate](https://renovatebot.com) in the google npm modules and love it.

ZYSzys:
I’m +1 on renovate.

fhemberger:
I've only been using greenkeeper so far, but I'm fine with everything that works.

It highly depends on the rights needed for the bot, as this is a security concern for the entire node.js github org. **So this must be approved by the org admins.**
(This has also been the reason so far that we still do this manually.)

silverwind:
There is a non-zero risk that a dependency includes breaking changes in minor and patch releases. For that reason, I personally prefer to audit every dependency update manually. Tools like [updates](https://github.com/silverwind/updates) make this process simple and fast.

bnb:
FWIW, I'm hugely in favor of Greenkeeper. I've used it personally, and the maintainers are long-standing (trusted?) members of the Node.js community.

Regardless of the tool we choose, this kind of automation is _a security feature_. It helps keep dependencies updated, which reduces the surface area that is available to attackers + reduces the burden on maintainers of code to resolve security issues that are introduced via stale dependencies.

aymen94:
@bnb I thinks we should use it on many repos of the organization.
So we'll add it?

zeke:
I've used Greenkeeper, Renovate, and Dependabot. My favorite is Dependabot, because it has lots of controls for things like automatically merging PRs that pass checks, or auto-merging a specific set of whitelisted (trusted) modules.

XhmikosR:
I think we should move with dependabot. But before that we should definitely enable automated security fixes in the repo: https://help.github.com/en/articles/configuring-automated-security-fixes

Trott:
> I think we should move with dependabot. But before that we should definitely enable automated security fixes in the repo: https://help.github.com/en/articles/configuring-automated-security-fixes

Automated security fixes are turned on for the repository (as of a few days ago, I think). IIUC, GitHub purchased dependabot and that's how automated security fixes are done. So I guess this can be closed?

XhmikosR:
Nope, one still needs to add the dependabot app for normal dependency
updates and configure it optionally

[...]

PR #2653, opened 2019-10-09
Update dependencies.

XhmikosR:
* handlebars to v4.4.3
* nock to v11.4.0


Aissaoui-Ahmed (approved):

Trott:
Should we open an issue in the admin repo about installing dependabot for the org so we can enable it on this repo (and probably several others)?

XhmikosR:
@Trott See https://github.com/nodejs/nodejs.org/issues/2113

Just note that I personally regenerate the lock file from time to time to update the transitive dependencies too and flatten the deps tree. But using dependabot is definitely a good thing.

The only thing that is not tested on CI in the server used for development.

PR #2662, opened 2019-10-11
scroll to top: specify `display: none`

XhmikosR:
This is so that when JavaScript is disabled, the button doesn't show up.

BTW the button could work without JS if we targeted an `id` in `body` maybe. It's just that it would always show up regardless of the pageYOffset.

### DIFF
--- a/layouts/css/page-modules/_scrollToTop.styl
+++ b/layouts/css/page-modules/_scrollToTop.styl
@@ -18,6 +18,7 @@ html
   min-width 20px
   text-align center
   padding 0 5px 1px
+  display none
 
   // Prevent button from flashing on page load
   opacity 0


XhmikosR:
Now, on master, when JS is disabled, the button shows up and it does "work" in the sense it scrolls up to an empty anchor `#`.

So, anyway, I don't know what it better.

alexandrtovmach:
I'm still not sure about adopting this website for disabled JS cases, but it looks okay for me

alexandrtovmach (approved):

nschonni:
@XhmikosR should this be ported to Sass now?

XhmikosR:
Yeah, but I'm not sure yet about this. I mean, it does sort of work fine without JS too, the scroll to top button just shows all the time.

ghost:
@XhmikosR: There's a conflict....

PR #2672, opened 2019-10-12
Introduce linting for package-lock.json

phillipj:
Because double checking each and every line changed in `package-lock.json` can be daunting for humans. That makes it quite easy for bad actors to slip in unoriginal packages as part of a PR that looks well intended.

Refs https://snyk.io/blog/why-npm-lockfiles-can-be-a-security-blindspot-for-injecting-malicious-modules/

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -34,3 +34,6 @@ jobs:
 
       - name: Run linkinator
         run: npm run test:linkinator
+
+      - name: Run package-lock.json linting
+        run: npm run test:lockfile
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -34,6 +34,3 @@ jobs:
 
       - name: Run linkinator
         run: npm run test:linkinator
-
-      - name: Run package-lock.json linting
-        run: npm run test:lockfile
--- a/package-lock.json
+++ b/package-lock.json
@@ -283,6 +283,12 @@
         "@types/unist": "*"
       }
     },
+    "@yarnpkg/lockfile": {
+      "version": "1.1.0",
+      "resolved": "https://registry.npmjs.org/@yarnpkg/lockfile/-/lockfile-1.1.0.tgz",
+      "integrity": "sha512-GpSwvyXOcOOlV70vbnzjj4fW5xW/FdUF6nQEt1ENy7m4ZCczi1+/buVUPAqmGfqznsORNFzUMjctTIp8a9tuCQ==",
+      "dev": true
+    },
     "abbrev": {
       "version": "1.1.1",
       "resolved": "https://registry.npmjs.org/abbrev/-/abbrev-1.1.1.tgz",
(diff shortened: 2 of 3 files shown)


nschonni:
Maybe this is more of a CI thing. If it's added as a lint target, it'll run as part of the pre-commit too.
What about using an "npx" call in the GitHub action?

phillipj:
Good point indeed! I'm more than happy to move into being a part of the GitHub Actions flow instead.

To me it still sounds reasonable to have it as dev dependency though, primarily to ensure we have control of which lockfile-lint *version* we're using. Don't hesitate to shout if anyone disagrees.

package-lock.json
```diff
@@ -3204,7 +3403,7 @@
         "debug": "^4.1.1",
         "moment": "^2.24.0",
         "slugify": "^1.3.4",
-        "substitute": "https://github.com/segmentio/substitute/archive/0.1.0.tar.gz"
+        "substitute": "https://registry.npmjs.org/substitute/-/substitute-0.1.0.tgz"
```
nschonni: This isn't the same package and will revert when the metalsmith package that depends on it updates

.github/workflows/test.yml
```diff
@@ -34,3 +34,6 @@ jobs:
 
       - name: Run linkinator
         run: npm run test:linkinator
+
+      - name: Run package-lock.json linting
+        run: npm run test:lockfile
```
nschonni:
```suggestion
        run: npx lockfile-lint@2.1.2 --type npm --path package-lock.json --allowed-hosts npm github.com
```

nschonni (dismissed):
Tested the NPM way locally and provided the string that allows the "substitute" package to load from GitHub. The NPM package you pointed to doesn't look to be the same package

.github/workflows/test.yml
phillipj: My rationale for *not* adding github.com to allowed hosts is that packages can be overwritten/republished, which counteracts what we should be aiming for, right?

package-lock.json
phillipj: Good catch, thx! And since that `substitute` package doesn't seem like it has ever been published to npm, it looks like we don't have much choice but to allow packages from github.com afterall 😞

.github/workflows/test.yml
phillipj: Aside from adding github.com as allowed host, don't you consider it valuable to (1) lock down `lockfile-lint`s dependency tree by installing it as dev dependency and (2) aim for consistency with other similar npm scripts (`test:linkinator | test:html`)?

XhmikosR:
This was in my TODO list so definitely I like it. What I don't like is not adding the package in our lock file.

Any reason for that?

EDIT:

Ah, scratch that, you did it like I would. Definitely don't do it through `npx`.

XhmikosR:
Ah, scratch that, you did it like I would. Definitely don't do it through `npx`.

XhmikosR:
As for adding it in `test:lint*`, it's your call. I specifically didn't add linkinator and test:html there because they can be slow.

Also, personally I'm against the precommit hook because it's just slow.

package.json
```diff
@@ -20,6 +20,7 @@
     "test:lint:md": "markdownlint \"**/*.md\" -i \"**/node_modules/**\"",
     "test:lint:stylint": "stylint layouts/css",
     "test:lint": "npm-run-all --parallel test:lint:*",
+    "test:lockfile": "lockfile-lint --type npm --path package-lock.json --allowed-hosts npm github.com",
```
XhmikosR: Does this accept wildcards? It seems we have scripts/contributor-list too

XhmikosR (approved):

package.json
XhmikosR: Also, shouldn't we specify `--validate-https`?

package-lock.json
XhmikosR: Yeah. Feel free to make an upstream issue for this :)

XhmikosR:
I pushed a few changes, feel free to revert. This is how I'd do it personally.

package-lock.json
XhmikosR: Opened https://github.com/segmentio/metalsmith-permalinks/issues/109

XhmikosR:
Rebased to keep my patch separate and updated to v2.1.4.

XhmikosR (approved):

XhmikosR:
@Trott can you review? @phillipj please confirm you are OK with my patch too.

Trott:
@nschonni Does this look good to you now or still needs changes?

nschonni:
I'm still not sure if it adds much. EX: it didn't pick up when the package was replaced manually in the lock file

XhmikosR:
You need to see its limitations. https://github.com/lirantal/lockfile-lint/tree/master/packages/lockfile-lint#cli-command-options

This is definitely a good addition, even if it lacks some features.

XhmikosR:
And [here](https://github.com/bryanbraun/anchorjs/pull/113/files#diff-32607347f8126e6534ebc7ebaec4853dL250) is a recent example where I hit issues that would have been caught by lockfile-lint.

phillipj:
> I'm still not sure if it adds much. EX: it didn't pick up when the package was replaced manually in the lock file

Myeah, guess that's a valid point.. To be the devils advocate, since I was able to manually temper with the lockfile and switch out a package, bad actors could do the exact same thing, just as long as the new package that gets slipped in is either from npmjs.org or github.com.

What I'm assuming you're looking for is that `lockfile-lint` would go through our entire dependency tree and compare it against the upstream package's list of dependencies. Although that would be more time consuming, it would indeed make the result from the linter more valuable.

@lirantal any thoughts on the above?

Trott:
Easy solution is to remove the lock file and live without it. I think it adds value, but if the tradeoff is risk unless we meticulously review lock file changes (which can be large), then I'm OK with giving it up.

Another solution would be to add dependabot or whatever and only allow it to update dependencies, etc. Although I guess that doesn't do a whole lot for new dependencies, so that's a partial solution only.

XhmikosR:
We should add dependabot regardless, see #2113. Still having this isn't bad IMO. And assuming the missing features will be added, this can only get better.

lirantal:
@nschonni  
> I'm still not sure if it adds much. EX: it didn't pick up when the package was replaced manually in the lock file

There's no reference of what was attempted so I'm not sure but I guess that maybe you changed one package for another, both from npm or such? Currently, the tool doesn't test that yet but there' active development that I won't be surprised if we get that supported soon. PRs welcome :)

alexandrtovmach (approved):

nschonni:
@lirantal sorry, it's collapsed up in https://github.com/nodejs/nodejs.org/pull/2672#discussion_r334255699

lirantal:
No worries. So this is something that definitely lockfile-lint catches
image

XhmikosR:
It would have caught it if we didn't allow github.com.

I think we should move forward with this and lockfile-lint will keep improving AFAICT.

phillipj:
> No worries. So this is something that definitely lockfile-lint catches

@lirantal really? Does that mean we are missing any arguments to trigger that behaviour?

I pushed the initial version of this branch temporary, to let you see the build succeeds and how the lockfile was tempered with: [lockfile-lint-tempered](https://github.com/nodejs/nodejs.org/commits/lockfile-lint-tempered) (see those two latest commit).

lirantal:
@phillipj in:

```
    "test:lint:lockfile": "lockfile-lint --allowed-hosts npm github.com --validate-https --type npm --path package-lock.json",
```

you are allowing github and npm with the above command options so any of them will pass just fine:

image

Otherwise, I think I may be missing what was the attempt to try and bypass the lockfile-lint. Can you explain to me what trust policy you would like to support and how/where did the lint miss it?

phillipj:
@lirantal sure! In the article you wrote that I mentioned when opening this PR, I got really excited about the thought that we could ensure no one would be able to slip-in their own package deep inside our dependency tree (via `package-lock.json`).

I naively wanted to get rid of the only dependency we had from github.com to ensure _all_ dependencies originated from npm at least. But as @nschonni spotted, the old & new package was not the same, although they have the exact same name.

That means I (undeliberately) in practise did what we want to enforce no one could possibly do, right? Or did I misunderstood the message of your article?

lirantal:
if I understood right, you are referring to this change:

image

You were indeed correct to try and find the npm package for `substitute` instead of the github link to try and enforce all resources from the official npm registry. lockfile-lint will not prevent you from using the wrong package, right? that's not up to the linting task and this is more of a human error. 

> That means I (undeliberately) in practise did what we want to enforce no one could possibly do, right? Or did I misunderstood the message of your article?

Correct. If you had enforced only npm as the allowed hosts and someone had tried to replace the source of packages with something in github, then lockfile-lint will have caught it. It is also a very easy thing to check, you can manually edit the lockfile in your dev environment to change the host to github and run the lockfile-lint task to assert it.

Let me know if there's anything else I can help with ✨

phillipj:
Alrighty, bottom line; `lockfile-lint` will enforce **where** packages are downloaded from, not **which** packages are installed.

> @Trott said:
> Easy solution is to remove the lock file and live without it. I think it adds value, but if the tradeoff is risk unless we meticulously review lock file changes (which can be large), then I'm OK with giving it up.

My biggest concern without the lock file is what will end up being installed when deployed on our production server. As far as I know, not many people can get a hold of the logs from that server and if [`$ npm run deploy`](https://github.com/nodejs/build/blob/0a1b2e9ad35ba7aab215e799529781a3522e00ce/ansible/www-standalone/resources/scripts/build-site.sh#L24) fails there, it will be challenging (1) knowing it failed and (2) what failed. Personally been bitten a few times due to slightly different dependency trees when there is a lack of lock file. So with the lock file in place and we run `npm run build` in CI, we're more confident it will succeed on the production server as well.

@XhmikosR appreciate your good intentions. In future PRs I hope we can discuss and agree before you amend and force push tweaks though.

XhmikosR:
@phillipj I didn't change your patch. I just pushed my patch separately. If you didn't like it, you could just revert it.

PR #2677, opened 2019-10-14
fix: Replace implicit blockquotes

nschonni:
This render correctly, but are flagged by remark-lint.
Sometimes they may be unitentional


XhmikosR:
I assume this rule is enabled in the upcoming remark-lint config and also upstream?

XhmikosR:
Also, is there any markdownlint rule we should have enabled? Maybe https://github.com/DavidAnson/markdownlint/blob/master/doc/Rules.md#md027---multiple-spaces-after-blockquote-symbol or md0028?

nschonni:
No, and that's already enabled

XhmikosR:
We should really stick to one base config until we make the switch. It gets too complicated.

XhmikosR (approved):

alexandrtovmach (approved):

PR #2678, opened 2019-10-14
fix: Consistent emphasis and strong

nschonni:
- Some had a space in the emphasis marker
- Escape some elements in the changelogs
- remove double asterisks order list style


XhmikosR:
The thing is that the changelog files aren't supposed to change, that's what I was told :P

So, if we go with this we'll need to make these changes upstream too maybe?

nschonni:
Most of these fixes appear to have been made in the upstream repo, but I did see one or 2 where they used backticks instead of escaping

XhmikosR:
Let's make it consistent with upstream, otherwise LGTM.

XhmikosR:
I think markdowlint has some related rules, which I thought were enabled by default. Maybe enable them so that we are safe until the remark-lint change happens. Note that this comment applies for all of your PRs; just enable any related markdownlint rules.

Aissaoui-Ahmed (approved):

XhmikosR:
@nschonni I assume markdowlint doesn't have rules for these?

nschonni:
Not that I'm aware of

XhmikosR (approved):

PR #2679, opened 2019-10-14
fix: remark-lint list-item-bullet-indent rule

nschonni:


locale/en/docs/guides/event-loop-timers-and-nexttick.md
```diff
@@ -174,7 +174,7 @@ either the queue has been exhausted, or the system-dependent hard limit
 is reached.
 
 * _If the **poll** queue **is empty**_, one of two more things will
-happen:
+  happen:
```
XhmikosR: I think you can move this to the previous line; it will be 75 characters

XhmikosR (approved):

XhmikosR:
Don't we need to enable any markdownlint rules for these?

nschonni:
No, this is something remark catches, but Markdownlint doesn't

Issue #2684, opened 2019-10-15
Add link to WSL2 install to download page


mattwojo:
* **Node.js Version**: ALL
* **OS**: Windows
* **Scope**: Install

Windows is recommending that Node.js be installed on WSL2 (the subsystem for Linux, running the latest version of Ubuntu) for an optimal experience doing Node.js development.

At the bottom of the [Node.js download page](https://nodejs.org/en/download/) there is a list of links:

Signed SHASUMS for release files (How to verify)
All download options
Installing Node.js via package manager
Previous Releases
Nightly builds
Node-ChakraCore Nightly builds
Building Node.js from source on supported platforms
Installing Node.js via binary archive on Linux

Similar to [this request](https://github.com/nodejs/unofficial-builds/issues/3), it would be helpful if this list of links included a link to the instructions to install on WSL2 (now the recommendation from Windows). Maybe something like this:

Signed SHASUMS for release files (How to verify)
All download options
Installing Node.js via package manager
[Installing Node.js via the Windows subsystem for Linux (WSL2)](https://docs.microsoft.com/windows/nodejs/setup-on-wsl2) 
Previous Releases
Nightly builds
Node-ChakraCore Nightly builds
Unoffical builds
Building Node.js from source on supported platforms
Installing Node.js via binary archive on Linux

This is a MUCH more happy path for anyone using Node.js on Windows!

Thanks!!

nschonni:
> now the recommendation from Windows

Where are you seeing this?

mattwojo:
https://docs.microsoft.com/windows/nodejs/
Instructions > https://docs.microsoft.com/en-us/windows/nodejs/setup-on-wsl2
Recommendation from the Beginners page > https://docs.microsoft.com/en-us/windows/nodejs/beginners

![image](https://user-images.githubusercontent.com/7566797/66873070-65797880-ef5c-11e9-92fe-3bdf46e7d6bb.png)

nschonni:
OK, i've opened an issue over there, since suggesting that beginners change to the "Insiders" ring of Windows and install WSL2 that isn't finalized doesn't seem like great advice IMO.

mattwojo:
Thanks @nschonni. Try it -- it will change your mind. : )
Do you have a link to the issue you opened?

[...]

PR #2692, opened 2019-10-16
Add WSL to list of install options

mattwojo:
Discussion about this here: #2684


mattwojo:
Discussion about this here: https://github.com/nodejs/nodejs.org/issues/2684

MylesBorins (approved):
LGTM

MylesBorins:
One though... the instructions listed here show how to setup Windows Subsystem Linux and then point towards using nvm

https://docs.microsoft.com/en-us/windows/nodejs/setup-on-wsl2#install-nvm-nodejs-and-npm

Do we want to include this documentation when it is a bit of a duplication?

Trott:
@nodejs/platform-windows @nodejs/website This could use a few more eyes. It's a pretty simple change, but I don't have any WSL expertise so someone else indicating that this is a good idea (or that it's not a good idea) would be great.

XhmikosR:
I'm on Windows but I don't use WSL so I don't have a personal opinion. But since the link points to Microsoft's own article, it should be OK.

nschonni:
:-1: for the reasons on the linked issue. I'm also not sure where this link would show up when being added to the JSON file.

Trott:
Yeah, without a corresponding string in a layout file somewhere, I'm not sure what this accomplishes.

Trott:
Thanks for the PR, @mattwojo! I'm going to close this, but feel free to comment or re-open if you think that's being hasty.

mattwojo:
re-open

PR #2695, opened 2019-10-17
fix: MD022/blanks-around-headings/blanks-around-headers

nschonni:
Headings should be surrounded by blank lines


XhmikosR:
Does remark-lint have a similar rule?

tests/scripts/CHANGELOG.fixture.md
```diff
@@ -45,6 +45,7 @@
 will be supported actively until April 2017 and maintained until April 2018.
 
 <a id="4.4.4"></a>
+
```
XhmikosR, line 48: We should exclude `tests/**/*.md` from this IMO.

XhmikosR:
The English changes are fine. The translations are mostly because they are using HTML comments with the English text. So, it should be OK.

Also, I'm not sure if the test files should keep getting modified; they are historical files.

tests/scripts/CHANGELOG.fixture.md
Trott: Ideally, this file should probably be in a `fixtures` dir that all linting should ignore. But that doesn't have to happen in this PR necessarily.

XhmikosR:
@nschonni does remark-lint have a rule for this?

nschonni:
I think https://github.com/remarkjs/remark-lint/blob/master/doc/comparison-to-markdownlint.md is out of date, but it isn't a rule that is in the current baseline node preset. It's good because stuff like VS Code's markdown parser also trips up on rendering stuck headers. The upstream rule will need some cleanup because of the current stuck HTML comments used for generation

XhmikosR:
We juts need a rule because this is going to happen again. My suggestion would be to not merge this until we have a check for it.

wooorm:
Is [`remark-lint-no-missing-blank-lines`](https://github.com/remarkjs/remark-lint/tree/master/packages/remark-lint-no-missing-blank-lines#remark-lint-no-missing-blank-lines) what you’re looking for?

XhmikosR:
Thanks @wooorm! @nschonni can you add this rule to this repo's rules?

Chasdaspaz:
pull request

Chasdaspaz:
pull request

Chasdaspaz:
pull request

XhmikosR:
This needs a rebase and the rule https://github.com/nodejs/nodejs.org/pull/2695#issuecomment-544114526 added

XhmikosR:
@nschonni can you rebase this and add the remark-lint rule please?

PR #2704, opened 2019-10-21
Update handlebars to v4.4.5.

XhmikosR:
Let's wait a bit in case there's another patch release. v4.4.4 broke some stuff.


XhmikosR:
It's include in #2716

PR #2707, opened 2019-10-21
feature: hide current release when LTS release is more recent

MylesBorins:
This is an edge case we hit today. The latest Current release is 12.12.0,
the latest LTS release is 12.13.0. This leads to a confusing landing page.

This adds an additional semver comparison and hides the current release
from the main page on days like today when it shouldn't be shown.

Refs: https://github.com/nodejs/Release/issues/495

### DIFF
--- a/layouts/index.hbs
+++ b/layouts/index.hbs
@@ -42,7 +42,7 @@
         </div>
 
         {{#if project.latestVersions.current.node}}
-        {{#if project.latestVersions.current.hideCurrent}}
+        {{#unless project.latestVersions.hideCurrent}}
           <div class="home-downloadblock">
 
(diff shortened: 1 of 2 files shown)


layouts/index.hbs
```diff
@@ -42,6 +42,7 @@
         </div>
 
         {{#if project.latestVersions.current.node}}
+        {{#if project.latestVersions.current.hideCurrent}}
```
ljharb: do you mean this to render when `hideCurrent` is true? or the inverse, when hideCurrent is false?

MylesBorins: fixed.

ljharb:
LGTM!

nschonni (approved):

watson (approved):

XhmikosR:
@MylesBorins shouldn't we do something similar for https://nodejs.org/en/download/ ?

PR #2709, opened 2019-10-22
zh-cn: correct a grammar flaw

kidonng:


nschonni:
Ping @nodejs/nodejs-cn

willin:
from my point of view, i vote previous version

Lellansin:
Can't get the necessary reason to make this modification.

kidonng:
Sorry for the mistake 😞, I intended to correct a grammar flaw:
`Download for {{ labels.download }}` is translated into `下载为 {{ labels.download }}` which sounds unnatural, `为 {{ labels.download }} 下载` would be more accurate, since we can't modify the template, a sole `下载 {{ labels.download }}` will be better.
I wasn't meant to make any further modifications. I've reverted the unnecessary changes, please check if it is appropriate now.

PS: In Chinese Traditional translation `目前版本：{{ labels.download }}` is used, thus avoids the problem.

ghost:
@kidonng：

We may have different kinds of versions for download, so this means Nodejs will give you a choice to choose which to be choosen according to your real senario. So here 'download for' means 'we download something **for** a specific kind of platform'.

However, according to the Chinese language meaning,  I agree with what you said, but this cannot show the meaning of 'for'.

So I'd tend to make a change for you as '下载平台为：' instead of '下载为', which seems better.

kidonng:
@MaledongGit Thanks for your understanding! It seems I did forget that a simple "Download" can be frustrating. There's one way to catch that though: "为以下平台下载" ("Download for the following platform"), but it may seem superfluous. "下载平台为" still sound not so natural but definitely better than the previous one, so I'm guessing leave it as it is now it's fine.

PR #2711, opened 2019-10-22
chore: Remove pre-commit

nschonni:
Although run in parallel, these lints aren't quick and are tested on CI before merging.
This was done on the nodejs.dev as well.


XhmikosR:
Totally agree, I always delete the hook locally.

XhmikosR (approved):

PR #2723, opened 2019-10-25
Update dependencies

XhmikosR:
* autoprefixer 9.7.0
* postcss 7.0.21
* lockfile-lint 2.1.6
* nock 11.6.0

@Trott how is it going with enabling dependabot? We could set it up to check daily or weekly and assign specific people for review

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -586,16 +586,16 @@
       "dev": true
     },
     "autoprefixer": {
-      "version": "9.6.5",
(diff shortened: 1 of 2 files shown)


Trott:
> @Trott how is it going with enabling dependabot? We could set it up to check daily or weekly and assign specific people for review

I just opened https://github.com/nodejs/admin/issues/430. Let's give TSC/CommComm/observers 72 hours to comment and then we'll see where we are.

Trott (approved):

PR #2728, opened 2019-10-26
Add `.netlify` in .gitignore

XhmikosR:


.gitignore
```diff
@@ -9,3 +9,6 @@ npm-debug.log
 .DS_Store
 .AppleDouble
 .LSOverride
+
+# Netlify
+.netlify
```
nschonni, line 14: Not really familiar with Netifly, but is this a file or folder that gets ignored?

XhmikosR: Yeah,

nschonni: Sorry, I think part of your response got cut off

nschonni: Oh, I guess my question might not have been clear. Is this a file or a folder? If it's a folder it should have a trailing slash

XhmikosR: It's a folder but due a bug in their client, it just adds `.netlify` in .gitignore even if one has `/.netlify/`

lpinca (approved):

PR #2730, opened 2019-10-27
Update getting-started-guide.md

sathya-tt:
The output of console.log was like http://${host}:${port}/, without the variables being printed.

### DIFF
--- a/locale/en/docs/guides/getting-started-guide.md
+++ b/locale/en/docs/guides/getting-started-guide.md
@@ -21,7 +21,7 @@ const server = http.createServer((req, res) => {
 });
 
 server.listen(port, hostname, () => {
-  console.log(`Server running at http://${hostname}:${port}/`);
+  console.log(`Server running at http://%s:%d/', hostname, port);
 });
 ```
 


locale/en/docs/guides/getting-started-guide.md
```diff
@@ -21,7 +21,7 @@ const server = http.createServer((req, res) => {
 });
 
 server.listen(port, hostname, () => {
-  console.log(`Server running at http://${hostname}:${port}/`);
+  console.log(`Server running at http://%s:%d/', hostname, port);
```
Trott, line 24: This introduces a syntax error. This has changed one of the backquotes to a single quote. In other words, this breaks the template string.

Trott (changes requested):

Trott:
> The output of console.log was like http://${host}:${port}/, without the variables being printed.

The current code is correct. The results you describe are what will happen if you replace the backquotes with single quotes. That will change the template literal into a string. See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals for information on template literals.

I'd also be reluctant to make this change because format strings like this are not as idiomatic in JavaScript.

Thanks for the pull request. I'm going to close this. However, that's not to say that the text and code can't stand to be improved and additional submissions would be welcome.

sathya-tt:
Got it, As I did not copy paste, but typed myself the sample program to get started and got into that problem. I had used single quote and the sample code had back quote enclosure for the string.

The existing code works well with back-quote.  Thanks.

Issue #2735, opened 2019-10-30
Anchor link is incorrectly referenced in Flame Graph documentation


hysan:
* **URL**: https://nodejs.org/en/docs/guides/diagnostics-flamegraph/
* **Browser version**: All known browsers
* **Operating system**: All known OSes

The anchor link to _Filtering out Node.js internal functions_ on the Flame Graph documentation page is referenced incorrectly here:

> `It's useful to [apply some cleanup](#filtering-out-node-internal-functions) for a more readable graph`

Based on the title of the subsection, the anchor link should instead be: `#filtering-out-node-js-internal-functions`

This affects the en, ja, and zh-cn versions of `diagnostics-flamegraph.md`.

**Fix**

I've already went ahead and make a fix for the en version as that is straightforward and will submit a PR. For the ja and zh-cn versions though, the fix isn't as simple. The anchor links for the documentation is auto-generated. From my observations, non-ASCII characters are ignored and because of that, the anchor link for these sections and up being:

* ja docs - _Node.js の内部関数を除外する_ == `#node-js`
* zh-cn docs - _过滤掉 Node.js 的内置函数_ == `#node-js`

Subsequent sections that similarly are stripped down to just `#node-js` have auto-numbering that looks like `#node-js-1`. So if the link is fixed by linking to `#node-js`, future versions of the docs might break the link again if a new section is added that precedes that section.

A better fix might be to define the name of the anchor link in the ja and zh-cn version of the docs, but before I proceed to do so, I'd like to know what approach is desired. Thoughts?

**Sidenote**

In searching for the affected file, I also noticed that the ko index links to a non-existent ko version of the `diagnostics-flamegraph.md` file. The link should either be removed or changed to point to the doc in a different language. Should that be done in this bug fix or in a separate issue?

Trott:
Fixed in https://github.com/nodejs/nodejs.org/pull/2736

PR #2736, opened 2019-10-30
Fix the anchor link referenced in the Flame Graph docs.

hysan:
This initial PR will fix the en version of the docs for issue #2735. Awaiting feedback for desired solutions for other languages.


lpinca (approved):

locale/en/docs/guides/diagnostics-flamegraph.md
```diff
@@ -39,7 +39,7 @@ Now let's get to work.
     ```
 
 4. disregard warnings unless they're saying you can't run perf due to missing packages; you may get some warnings about not being able to access kernel module samples which you're not after anyway.
-5. Run `perf script > perfs.out` to generate the data file you'll visualize in a moment. It's useful to [apply some cleanup](#filtering-out-node-internal-functions) for a more readable graph
+5. Run `perf script > perfs.out` to generate the data file you'll visualize in a moment. It's useful to [apply some cleanup](#filtering-out-node-js-internal-functions) for a more readable graph
```
nschonni, line 42: Looks like the fix for the built/metalsmith version isn't the same as GitHub (which has it as `nodejs` with no dash).
Not a big deal since I think it's just broken for both ways right now, but might be an issue if adding https://github.com/remarkjs/remark-validate-links

PR #2738, opened 2019-10-31
Switch to stylelint-config-twbs-bootstrap.

XhmikosR:
It's what I based off the initial config and reduces the rules.

EDIT: BTW we could also make our own config, but I'm not sure it's worth it.

### DIFF
--- a/.stylelintrc
+++ b/.stylelintrc
@@ -1,122 +1,12 @@
 {
   "extends": [
-    "stylelint-config-standard",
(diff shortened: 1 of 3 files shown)


XhmikosR:
BTW we could also make our own config, but I'm not sure it's worth it.

aymen94 (approved):

Trott (approved):

PR #2755, opened 2019-11-08
[DNM] test CODEOWNERS on English page

nschonni:


nschonni:
@Trott doesn't look like the "*" for the website team is working. Did you hits something similar in the core repo? I can see "nodejs/website" in the Reviewers dropdown, so I think the access is right for the team

Trott:
> @Trott doesn't look like the "*" for the website team is working. Did you hits something similar in the core repo? I can see "nodejs/website" in the Reviewers dropdown, so I think the access is right for the team

That team has write privileges to the repository so they should get notified to the best of my knowledge. I'm not sure why that's not happening.

Trott:
I'm going to close this to minimize the chance someone merges it by accident, but re-open it if/when you need to experiment more.

PR #2757, opened 2019-11-10
update schedule.json

ghost:
1. According to our plan in https://nodejs.org/en/about/releases/, we've now published the 13th version of Nodejs, add this to the schedule for the maintance.

2. Add the missing coding name for v12.


csabapalfi (approved):

osk2 (approved):

Trott:
Looks good to me but I never even noticed this file before (that I can remember anyway) so I'll leave the actual approving to others (which has already happened anyway).

XhmikosR:
IMO this should be autogenerated somehow. Unfortunately, I don't have the time to do it myself.

Trott:
> IMO this should be autogenerated somehow. Unfortunately, I don't have the time to do it myself.

Maybe open an issue and label it appropriately as `help wanted` and/or `good first contribution`? Then we can land this as-is but not lose the recommendation you're making.

XhmikosR:
Oops sorry, I meant it for #2758

nschonni:
Should the build just curl this file from https://github.com/nodejs/Release/blob/master/schedule.json instead of keeping a separate copy?

ghost:
@nschonni: Not sure yet, but according to your description, it seems this file here is duplicated?
@nodejs/release: It seems this file has been defined in the https://github.com/nodejs/Release/blob/master/schedule.json, so do we have to maintain this copy here? Just for a confirm.
If you think this file can be removed, I'll remove it.

richardlau:
I think the file in the repository is a placeholder that gets overridden during build by `load-schedule` (called by `deploy`):
https://github.com/nodejs/nodejs.org/blob/575a47ec75132585404c2f19bd80a084405f723a/package.json#L12-L14 (see https://github.com/nodejs/nodejs.org/pull/1890)

XhmikosR:
Can't we remove the folder and the file completely?

ghost:
> I think the file in the repository is a placeholder that gets overridden during build by `load-schedule` (called by `deploy`):
> https://github.com/nodejs/nodejs.org/blob/575a47ec75132585404c2f19bd80a084405f723a/package.json#L12-L14
> 
> (see #1890)

So this file should be totally removed, or just run the script to update the file, just like what I submitted?

aymen94 (approved):

XhmikosR:
This patch is moot since this is overwritten on build time. We should instead see if we can just remove the folder and the file.

PR #2763, opened 2019-11-12
chore: Remove Prev/upcomming code and learn headings

nschonni:
Removes the need to shuffle them as they occur


nschonni:
Alternate mid-way to this would be to add headings for years and/or months to add a little structure

Trott (approved):
Yes, thank you.

PR #2697, opened 2019-10-18
fix: no-shortcut-reference-link

nschonni:

PR #2769, opened 2019-11-13
Fix broken link in docs to «How to publish N-API package» article

EgorGumin:
Current look of the page:

image


XhmikosR:
@nschonni that's why we need to finish #2697

XhmikosR (approved):

PR #2688, opened 2019-10-16
Remove the weekly updates script and the unused collection

XhmikosR:
Refs #2682 

Needs a PR in nodejs/build to redirect these. Maybe to the blog index page?

PR #2771, opened 2019-11-13
Replace request with got release-post script

### LINKED ISSUE #2689: Replace `request` with another package?
We use request in two places, one is download.js which will be removed in #2688 and the other one is in https://github.com/nodejs/nodejs.org/blob/master/scripts/release-post.js.

Note that this won't remove any deps, since request is a node-sass dependency.

MrBenJ:
Fixes #2689


scripts/release-post.js
```diff
@@ -33,15 +33,14 @@ function sendRequest (opts) {
       headers: { 'User-Agent': 'nodejs.org release blog post script' }
     }, opts)
 
-    request(options, (err, res, body) => {
-      if (err) {
-        return reject(new Error(`Error requesting URL ${options.url}: ${err.message}`))
+    const { url, ...gotOptions } = options
```
XhmikosR: Is the destructuring really needed here?

XhmikosR:
Need a rebase also :)

XhmikosR:
BTW would be nice to use the same for download.js and thus remove request completely from the project dependencies. Although, I have a PR which removes the file in #2688

scripts/release-post.js
MrBenJ: It isn't. I was just trying to be safe with the options if url was passed in twice. I took it out in another commit, along with a rebase to master :). Let me know what you think!

MrBenJ:
@XhmikosR - I was thinking of that too, but knowing the file is about to be deleted, I think it would be better just to leave that in. Maybe in your previous PR you can delete `request` from `package.json` since it's still in here? PR race? ;)

If that PR gets merged in first, I'll be happy to rebase again and remove request completely from the dependencies.

Let me know what's up!

MrBenJ:
@XhmikosR  - any news?

XhmikosR:
It's not my decision to make only :)

/CC @Trott @MylesBorins for review. Assuming we move forward with #2688, this is a good step forward.

Trott:
@nodejs/website

XhmikosR:
@MrBenJ since #2688 might not even be merged or take a long time, can you update download.js to use got too? BTW I've rebased the branch, so be careful not to overwrite it.

silverwind:
I guess it's mostly down to personal preference, but I prefer a `fetch` implementation over `got`:

https://www.npmjs.com/package/make-fetch-happen
https://www.npmjs.com/package/node-fetch

The former offers proxy support just like `request`, so it is my go-to fetching library currently.

XhmikosR:
FYI I don't mind what we use, as long as we do it for all scripts and thus replace request :)

XhmikosR:
@MrBenJ can you rebase this and also update to the latest got version? The other script that was using request is now removed, so you can remove request completely from the repo.

XhmikosR:
Alternatively, if someone has time, we could just use the native https to do the same and not add a dependency.

MrBenJ:
Hey @XhmikosR ! I'm on holiday right now but have the time off tomorrow to probably dig into this. I think I'll make another fork of the repo and just start fresh with all this new info :).

XhmikosR:
Sound good to me!

PR #2800, opened 2019-11-24
Revert "feat: minify HTML on build (#2537)"

tniessen:
This reverts commit 0cc3648df2741afbfdc29e43247e3dc0e6e44b66. Based on my estimations and calculations, the benefit is minimal, while making our build pipeline more complex, more time consuming, adding a dependency and making the output less readable. Contrary to assumptions in the original PR, the patch does not affect the docs.

(Also, @XhmikosR suggested that there might be a better alternative provided by Cloudflare.)

Refs: https://github.com/nodejs/nodejs.org/pull/2537

cc @nschonni @MylesBorins @aymen94 @bnb @Trott


tniessen:
And to be clear, I am fine with closing this and leaving the commit as it is if that is what people want, but I believe that the original PR was landed prematurely.

XhmikosR:
I agree, this should be reverted.

Now, I understand if we don't want to do this via Cloudflare, but the slowdown from this isn't justified by the gains IMHO.

MylesBorins (approved):
RSLGTM, we should turn it on with cloudflare imho

XhmikosR (approved):

bnb:
SGTM. +1 to turning it on through CloudFlare, since that’s where the majority of our traffic is served from anyway AFAIK.

aymen94 (approved):

aymen94:
if there is a better alternative, then so be it.

PR #2813, opened 2019-12-04
Add image dimensions and update resources page to use SVGs

XhmikosR:
Live Preview:

* <https://thirsty-kepler-e18e83.netlify.com/en/about/releases/>
* <https://thirsty-kepler-e18e83.netlify.com/en/about/resources/>

This aims to help browsers calculate the image dimensions thus prevent unnecessary reflows.

TODO:

- [ ] Add image dimensions in the resources SVGs
- [x] Check if everything is still fine on mobile and IE >= 10


XhmikosR:
**DO NOT MERGE**

I just noticed we actually don't set `height: auto` for images and this results in the wrong ratio. I will need to test things and look into specifying `height: auto`.

layouts/about-release-schedule.hbs
```diff
@@ -14,7 +14,7 @@
         {{{ contents }}}
 
         <p>
-          <img alt="{{ title }}" src="https://raw.githubusercontent.com/nodejs/Release/master/schedule.svg?sanitize=true">
+          <img src="https://raw.githubusercontent.com/nodejs/Release/master/schedule.svg?sanitize=true" alt="{{ title }}" width="960" height="500">
```
nschonni, line 17: Should the SVG get pulled down as part of the build like the schedule.json?

XhmikosR: I guess it could, but this is clearly a lot easier and shouldn't really break anyway :)

I still need to fix the height auto issue.

XhmikosR:
/CC @MartijnCuppens so that we get the about resources fixed in IE and simplify things

XhmikosR:
@MartijnCuppens the only thing pending is that the SVGs in the about/resources page now are smaller than what they were.

Also, IE >= 10 seems to be fine with these patches, can you also confirm? I like the removal of modernizr, so if everything is good in all browsers, this is a win win.

MartijnCuppens:
> @MartijnCuppens the only thing pending is that the SVGs in the about/resources page now are smaller than what they were.

Does it matter? It's vector anyway.

> Also, IE >= 10 seems to be fine with these patches, can you also confirm? I like the removal of modernizr, so if everything is good in all browsers, this is a win win.

I've never used modernizr, but it does look good afaict.

XhmikosR:
I think this is as good as it gets for now and it's definitely an improvement IMHO.

@Trott @MylesBorins for review: **Please do not squash the patches**

Trott:
> I think this is as good as it gets for now and it's definitely an improvement IMHO.

@nodejs/website This is ready for review.

lpinca (approved):

XhmikosR:
Only thing left is to add dimensions to the SVGs in about/resources/, but I guess we could it in another PR if we want this to land sooner. Or we can hold off this PR a bit

PR #2815, opened 2019-12-05
Switch from png to svg & fix IE whitespace

MartijnCuppens:
Switched from png to svg files which are smaller, sharper and fix the whitespace IE has. Also removed `nodejs-new-white-pantone.png` which was not used.

CC @XhmikosR


XhmikosR:
Thanks! I'm gonna have a look tomorrow

XhmikosR:
I wonder, can't we get rid of modernzr and just use `html { height: 100% }` for all browsers? From my tests this seems to be working fine.

MartijnCuppens:
Idk, never used modernizr. This PR only fixes the image issue btw.

XhmikosR:
https://github.com/nodejs/nodejs.org/blob/master/layouts/css/layout/_sticky-footer.scss

Basically it only adds `height: 100%` in `body` and `html` for IE. But from my testing I don't see any issue with other browsers having it set too. I haven't tested Apple browsers though.

XhmikosR:
LGTM, thanks! I will merge it to my branch and see if I can get rid of modernizr completely there.

XhmikosR (approved):

PR #2827, opened 2019-12-16
fix: remove misleading note section

szhigunov:
Section misleading about setImmediate using in different environments.

Fixes: https://github.com/nodejs/node/issues/30801


szhigunov:
If that PR will be accepted as appropriate, should I update other locales too?

ghost:
@Fishrock123：It's fine with me, what about your ideas?

ghost (approved):

locale/en/docs/guides/event-loop-timers-and-nexttick.md
```diff
@@ -412,9 +412,7 @@ added, which means every day we wait, more potential breakages occur.
 While they are confusing, the names themselves won't change.
 
 *We recommend developers use `setImmediate()` in all cases because it's
-easier to reason about (and it leads to code that's compatible with a
-wider variety of environments, like browser JS.)*
-
+easier to reason about.*
```
Trott, line 415: I think the blank line between the paragraph and the header might be required? Regardless, it makes the raw markdown easier to read. Let's keep it.

```suggestion
easier to reason about.*

```

ghost: I've helped to add it, thanks for your tip ;)

PR #2830, opened 2019-12-16
Remove unicode line seperator characters

sam-github:
U+2028, https://www.fileformat.info/info/unicode/char/2028/index.htm, do not display well in many tools, and were not likely intentional.


sam-github:
They are showing up as undisplayable unicode characters on the blog: https://nodejs.org/en/blog/vulnerability/december-2019-security-releases/

PR #2845, opened 2019-12-19
lang-switcher: fix history following

### LINKED ISSUE #2541: lang-switcher: make it follow history
It seems currently `window.location.replace` is used, which doesn't follow history.

IMO we should change this to https://developer.mozilla.org/en-US/docs/Web/API/Location/assign

szhigunov:
Changed location.replace to location.assign

Fixes: https://github.com/nodejs/nodejs.org/issues/2541


Trott (approved):

MylesBorins (approved):
LGTM

PR #2857, opened 2019-12-26
Regenerate package-lock.json

XhmikosR:
Flattens the deps tree and saves us 50 packages.

Maybe this is something that could be improved on npm? I have an issue in the dependabot repo and they say that's how npm installs packages. This issue is something that keeps happening, but I'm not an expert in npm code so I don't know if it's normal for this to happen.

For reference https://github.com/dependabot/feedback/issues/776 just in case anyone has any ideas

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -272,9 +272,9 @@
       "dev": true
     },
     "@types/node": {
-      "version": "12.7.12",
-      "resolved": "https://registry.npmjs.org/@types/node/-/node-12.7.12.tgz",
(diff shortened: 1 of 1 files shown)


Trott (approved):
LGTM. Large. package-lock.json changes are difficult to review thoroughly. I make sure all "resolved" entries point to the npm registry. Would be interested to know what others check for when reviewing changes like this.

XhmikosR:
Since we have lockfile-lint, we are somehow covered. BTW it seems we can specify more options to lockfile-lint, not sure if everything is on by default. But that's for another PR.

PR #2858, opened 2019-12-26
Fix the remaining table HTML validation errors.

XhmikosR:
The second patch removes the now unused semver-gte.js helper script, but perhaps we want to keep it for the future? If so let me know and I'll revert the removal.


Trott:
> The second patch removes the now unused semver-gte.js helper script, but perhaps we want to keep it for the future? If so let me know and I'll revert the removal.

I imagine it's inevitable that we'll change support/provided builds on some future release again, at which point that functionality will be useful again, IIUC. The problem is that if it's not already in the hbs partial, then no one will know it's there and will reinvent the wheel, possibly/probably. Maybe leave it in but add a comment in the partial showing how to use it?

Trott (approved):
LGTM with or without my comment/suggestion.

XhmikosR:
>Maybe leave it in but add a comment in the partial showing how to use it?

It's moot adding a comment because none of the other hbs scripts have one, and they all follow the same pattern.

I will just revert the removal for now, or we can just remove it and bring it back if needed. We have git log after all :)

Trott:
> I will just revert the removal for now, or we can just remove it and bring it back if needed. We have git log after all :)

I'm fine either way.

XhmikosR:
Let's merge it without squashing then. Or whatever you think is better for history

nschonni (approved):

PR #2860, opened 2019-12-26
Update lockfile-lint flags

XhmikosR:
Add `--allowed-schemes https:` and `--empty-hostname false`


Trott (approved):

Trott:
@lirantal

lirantal:
LGTM, except you could probably remove `--validate-https` since the https: scheme is now added as the only allowed source.

XhmikosR:
Removed `--validate-https` as per @lirantal's suggestion :)

PR #2864, opened 2019-12-27
Clarify that 'Linux on Power' is Little Endian

sxa:
I was perusing the node.js website and in order to verify that the Linux on Power download was little-endian and not big-endian (unlike AIX it can be either) I had to hover over the download link to check the filename. This change will make it more visible on the download site. With this PR I am assuming that changing `Power` to `Power LE` as an architecture name will be ok across languages and not need further translation. Someone else can challenge that assumption if desired ;-)

Signed-off-by: Stewart Addison <sxa@uk.ibm.com>


Trott (approved):

nschonni (approved):
I think it was both, but looking back through https://nodejs.org/dist I don't think BE has been supported since v6

lpinca (approved):

Aissaoui-Ahmed (approved):

sxa:
@nschonni Yeah I remember us doing BE originally and I suspect the pages were just never updated ... Seemed a bit overdue ;-)

sxa:
@Trott Should this have gone live by now?

Trott:
> @Trott Should this have gone live by now?

I would have thought so, but yeah, I'm not seeing it. Maybe this file only gets regenerated when there's a release or something like that?

nschonni:
Think the last one failed on installing fibers (WARNING, large link ~200MB) https://nodejs.org/github-webhook.log

PR #2867, opened 2019-12-29
Replace request with node-fetch

### LINKED ISSUE #2689: Replace `request` with another package?

MrBenJ:
Fixes #2689 

@XhmikosR - I tried using `http` but trying to figure it out was taking a bit, so I went back to using what I know (`isomorphic-fetch`). Running the script locally is producing the same output as with using `request` previously, so without knowing the usage of this script, it seems to be working.

If there are any concerns, please let me know. I am happy to edit and change things to your guidelines and standards :).

Best,
Ben Junya


MrBenJ:
Hey! I noticed there were CI tests added for this. Yay!

Didn't realize it until I started. I've fixed the failing tests. Please let me know if you need anything else

package.json
```diff
@@ -41,6 +41,7 @@
     "cheerio": "^0.22.0",
     "fibers": "^4.0.2",
     "handlebars": "^4.5.3",
+    "isomorphic-fetch": "^2.2.1",
```
lpinca: `isomorphic-fetch` uses `node-fetch` on Node.js so why not using `node-fetch` directly?

MrBenJ: Hey, thanks! Thanks a great call and a good catch. I've updated the code and rebased it to the latest commit on `master`

Trott:
Looks good to me although I would prefer to defer to others who use that code more often than I do.

scripts/release-post.js
```diff
@@ -32,16 +32,21 @@ function sendRequest (opts) {
     const options = Object.assign({
       headers: { 'User-Agent': 'nodejs.org release blog post script' }
     }, opts)
-
-    request(options, (err, res, body) => {
+    fetch(options.url, options).then(resp => {
```
lpinca: `fetch()` returns a `Promise` so there is no need to wrap it in `new Promise()`, you can make `sendRequest` return the fetch promise directly.

scripts/release-post.js
```diff
@@ -32,16 +32,21 @@ function sendRequest (opts) {
...
+      if (resp.status !== 200) {
+        return reject(new Error(`Invalid status code (!= 200) while retrieving ${options.url}: ${resp.status}`))
```
lpinca:
```suggestion
        throw new Error(`Invalid status code (!= 200) while retrieving ${options.url}: ${resp.status}`)
```

scripts/release-post.js
```diff
@@ -32,16 +32,21 @@ function sendRequest (opts) {
...
+      }
+      if (options.json) {
```
lpinca: I would simplify this as follow:

```js
return options.json ? resp.json() : resp.text()
```

The added benefit is that rejections will be handled.

scripts/release-post.js
```diff
@@ -32,16 +32,21 @@ function sendRequest (opts) {
...
+        resp.json().then(json => resolve(json))
+      }
+      resp.text().then(text => resolve(text))
+    }).catch(err => {
+      if (err.status !== 200) {
```
lpinca: Please remove this `if` statement. This will always evaluates to `true` because afaik there is no status `property` on [`node-fetch errors`](https://github.com/bitinn/node-fetch/blob/master/ERROR-HANDLING.md) and errors thrown in above suggestions.

scripts/release-post.js
```diff
@@ -32,16 +32,21 @@ function sendRequest (opts) {
...
+        return reject(new Error(`Invalid status code (!= 200) while retrieving ${options.url}: ${err.status}`))
+      }
       if (err) {
         return reject(new Error(`Error requesting URL ${options.url}: ${err.message}`))
```
lpinca:
```js
throw new Error(`Error requesting URL ${options.url}: ${err.message}`)
```

and remove the wrapper `if` statement.

Note that `err.message` could be different from current error messages because `err` can be a `FetchError`. If it doesn't break anything I would just remove the whole `catch()` branch instead of using a new `Error`.

scripts/release-post.js
```diff
@@ -23,26 +23,20 @@
...
-    }, opts)
-
-    request(options, (err, res, body) => {
-      if (err) {
-        return reject(new Error(`Error requesting URL ${options.url}: ${err.message}`))
-      }
-      if (res.statusCode !== 200) {
-        return reject(new Error(`Invalid status code (!= 200) while retrieving ${options.url}: ${res.statusCode}`))
-      }
+  const options = Object.assign({
```
XhmikosR: Maybe use spread instead?

XhmikosR: NVM, that's something we can change later since it seems Object.assign is used in this whole file.

MrBenJ: Pushed!

scripts/release-post.js
```diff
@@ -23,26 +23,21 @@
...
-        return reject(new Error(`Error requesting URL ${options.url}: ${err.message}`))
-      }
-      if (res.statusCode !== 200) {
-        return reject(new Error(`Invalid status code (!= 200) while retrieving ${options.url}: ${res.statusCode}`))
-      }
-
-      resolve(body)
-    })
+  const options = {
+    ...opts,
```
lpinca: Nit: can you please swap this with `headers` so it works like the original?

lpinca (approved):

scripts/release-post.js
MrBenJ: Done! Pushed up

PR #2869, opened 2020-01-02
Allow LTS and 'Other downloads' link to work locally

### LINKED ISSUE #2865: Many links on the website are absolute instead of relative
I had a quick look through the issues and couldn't see anything relating to this, but is there a reason why many of the links on the website explicitly include `https://nodejs.org` as a prefix? This makes it harder to test changes locally as you can't click on things on the main page without it going through to the main nodejs.org, not your local one.

I'm happy to take a look at changing this, but wanted to float the idea first in case there was a good reason why it was done like this.

sxa:
I believe this fixes #2865 adequately.

I only indentified two of them as it happens (although they were the first two I clicked when I spotted the issue)

This will allow these links to be clickable when testing changes locally

Signed-off-by: Stewart Addison <sxa@uk.ibm.com>


layouts/index.hbs
```diff
@@ -52,7 +52,7 @@
 
             <ul class="list-divider-pipe home-secondary-links">
               <li>
-                <a href="https://nodejs.org/{{site.locale}}/download/current/">{{ labels.other-downloads }}</a>
+                <a href="download/current/">{{ labels.other-downloads }}</a>
```
nschonni: nit: since the others are site absolute
```suggestion
                <a href="/download/current/">{{ labels.other-downloads }}</a>
```

sxa: Needs to include the language `{{site.locale}}` otherwise the path doesn't resolve (i.e. https://nodejs.org/download/current doesn't work). Have pushed that up to this commit too now.

nschonni (approved):

XhmikosR (approved):

PR #2870, opened 2020-01-02
disable banner

Trott:
Time to take down the security banner. Don't want to leave it up too
long lest people get used to not noticing it when it appears again.

### DIFF
--- a/build.js
+++ b/build.js
@@ -283,7 +283,7 @@ function getSource (callback) {
           lts: latestVersion.lts(versions)
         },
         banner: {
-          visible: true,
+          visible: false,
           text: 'New security releases now available for all release lines',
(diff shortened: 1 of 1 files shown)


nschonni (approved):

antsmartian (approved):

Trott:
Looks like this failed to deploy similar to https://github.com/nodejs/nodejs.org/pull/2864#issuecomment-570371948. 

/ping @nodejs/build I guess?

XhmikosR:
The last entries in https://nodejs.org/github-webhook.log are:

>stdout: HEAD is now at af0ce5c blog: v13.5.0 release post

which is very old

BTW we should rotate the log file more often since it becomes very big, very fast

nschonni:
I've opened up https://github.com/nodejs/build/issues/2123

PR #2911, opened 2020-01-24
Fix: change 'js' to 'bash'

ghost:
1. The change is to change the sample from 'js' to 'bash'.
2. Some other typo changes.

### DIFF
--- a/locale/en/knowledge/command-line/how-to-get-colors-on-the-command-line.md
+++ b/locale/en/knowledge/command-line/how-to-get-colors-on-the-command-line.md
@@ -77,7 +77,7 @@ console.log(colors.bold.grey.black.blue(stringThree));
 
 **Note**: Unlike the `String.prototype` approach, the chained methods on the `colors` instance are executed left to right i.e., the method closest to the string is finally applied. In the last `console.log` you can think of your terminal saying to itself, "Make this grey. Now, make this black. Now, make this blue. No more coloring methods now? Blue it is, then."
 
-With the latest version of `colors.js` you can also define **[Custom Themes](https://www.npmjs.com/package/colors#custom-themes)** in `color.js`, which makes our codes more Robust and allows better Encapsulation of data. A nice use case of this maybe:
+With the latest version of `colors.js` you can also define **[Custom Themes](https://www.npmjs.com/package/colors#custom-themes)** in `color.js`, which makes our code more Robust and allows better Encapsulation of data. A nice use case of this maybe:
 
 ```js
 var colors = require('colors');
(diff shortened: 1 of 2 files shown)


XhmikosR (changes requested):
`codes` is wrong IMO

XhmikosR (approved):

locale/en/knowledge/command-line/how-to-get-colors-on-the-command-line.md
```diff
@@ -11,7 +11,7 @@ When working on the command line, it can be both fun and extremely useful to col
 
 First, install it to the directory you'd like to work in.
 
-```js
+```bash
```
Trott: We don't want bash highlighting here.

```suggestion
```console
```

locale/en/knowledge/command-line/how-to-get-colors-on-the-command-line.md
```diff
@@ -11,7 +11,7 @@ When working on the command line, it can be both fun and extremely useful to col
...
 cd mydir
```
Trott: Remove this line, perhaps?

locale/en/knowledge/command-line/how-to-get-colors-on-the-command-line.md
```diff
@@ -77,7 +77,7 @@ console.log(colors.bold.grey.black.blue(stringThree));
 
 **Note**: Unlike the `String.prototype` approach, the chained methods on the `colors` instance are executed left to right i.e., the method closest to the string is finally applied. In the last `console.log` you can think of your terminal saying to itself, "Make this grey. Now, make this black. Now, make this blue. No more coloring methods now? Blue it is, then."
```
Trott:
```suggestion
Unlike the `String.prototype` approach, the chained methods on the `colors` instance are executed left to right i.e., the method closest to the string is finally applied. In the last `console.log` you can think of your terminal saying to itself, "Make this grey. Now, make this black. Now, make this blue. No more coloring methods now? Blue it is, then."
```

Trott:
This article should be removed in my opinion. It is inessential and outdated. 

* There's colorization support in Node.js itself now, and while it's a bit clunkier to use, it should be covered in an article like this.
* This focuses on an external module. But why choose `colors` and not `chalk` or another alternative?
* This is not something that needs to be covered in a knowledge guide. It is inessential. We don't need it, so let's remove it. We have too much stuff.

(Of course, in my opinion, all or nearly all of the knowledge guide should be removed.)

3imed-jaberi (approved):

aymen94 (approved):

PR #2912, opened 2020-01-24
add a missing 'tr' in the CODEOWNERS

ghost:
We now have a new member group 'tr', this is the fixture for it.


Trott:
LGTM I suppose, although none of those CODEOWNERS entries will work because GitHub will not use those entries to notify groups if the groups do not have write access to the repo.

nschonni (approved):

3imed-jaberi (approved):

PR #2913, opened 2020-01-25
simplify main.js

ghost:
Change the "if" statment to two simple statements.


XhmikosR:
Not sure if this is an improvement.

XhmikosR:
@Trott thoughts? I personally don't think this is an improvement.

Trott:
> @Trott thoughts? I personally don't think this is an improvement.

I appreciate the desire to improve the code but I think the existing code is more maintainable. It's easier to understand, at least to me. 

@MaledongGit Is there something in particular about the existing code that makes it problematic that you were hoping to improve?

nickserv (approved):

aymen94 (approved):

nickserv:
This change reduces repetition. I find the current code difficult to read as the duplication makes the intent of the code harder to understand. The variable name also makes it easier to understand.

3imed-jaberi (approved):

osk2 (approved):

static/js/main.js
```diff
@@ -25,12 +25,8 @@
...
   langPickerTogglerElement.addEventListener('click', function () {
     langPickerElement.classList.toggle('hidden')
-
-    if (langPickerTogglerElement.getAttribute('aria-expanded') === 'true') {
-      langPickerTogglerElement.setAttribute('aria-expanded', 'false')
-    } else {
-      langPickerTogglerElement.setAttribute('aria-expanded', 'true')
-    }
+    const isAriaExpanded = langPickerTogglerElement.getAttribute('aria-expanded') === 'true'
+    langPickerTogglerElement.setAttribute('aria-expanded', (!isAriaExpanded).toString())
```
tniessen: I understand @Trott's point, the boolean/string conversions are not pretty. However, if I remember all those questionable DOM / WebIDL conversion specifications correctly, the whole `(...).toString()` is unnecessary:

```suggestion
    langPickerTogglerElement.setAttribute('aria-expanded', !isAriaExpanded)
```

(I luckily haven't had to write frontend JS in a while, so I might be wrong, but it seems to work in FF/IE/Chrome.)

Trott: > I understand @Trott's point,

Clearly I'm in the minority on this. I'm fine with it landing.

XhmikosR:
This whole thing is like premature optimization. I honestly don't see any real benefit from this patch. The code works fine, and we don't even have to get into if `toString()` is needed :P

PR #2909, opened 2020-01-24
fix some typos and enhance 'advanced' knowledge files

ghost:
1. Fixtures for some typos.
2. Add 'console' for some console codes to enhance the formation.
3. Add some backquotes onto some methods.

PR #2923, opened 2020-01-30
Remove console syntax highlighting instances

XhmikosR:
Prism.js doesn't support this and it led to wrong highlighting.

Not sure if prism.js has another language we could use for this, so in the meantime this should fix rendering.

Se my comments in https://github.com/nodejs/nodejs.org/pull/2909#issuecomment-580193103


Trott (approved):
Sorry. `console` was my suggestion. Whoops....

PR #2940, opened 2020-02-06
fix(blog): point v10.19.0 to the correct URL

mmarchini:
URL for v10.19.0 changelog was broken (pointing to unreleased v10.19.1).


sam-github (approved):
Thanks!

PR #2950, opened 2020-02-13
chore: Drop IE 10

nschonni:
Support end January 31st 2020
https://support.microsoft.com/en-us/help/4488955/support-ending-for-internet-explorer-10

### DIFF
--- a/.browserslistrc
+++ b/.browserslistrc
@@ -1,4 +1,4 @@
 # https://github.com/browserslist/browserslist#readme
 
 defaults
-Explorer >= 10
+Explorer 11


XhmikosR:
/CC @Trott 

Personally I wouldn't remove it. The site "works" even on IE 9 (minus the requestAnimationFrame error) and it's good, because people around the world do use such old browsers.

Trott:
@nodejs/website @nodejs/website-redesign

XhmikosR:
We really don't gain anything here, just a few bytes from the CSS. Just let it be for people with old browsers.

Just my 2 cents.

PR #3000, opened 2020-03-03
style(scss): Remove outline border from button

heithemmoumni:


nschonni:
I'm not seeing what this would be used for. Usually you don't want to remove outlines for accessibility reasons, but I don't see a difference in the rendering when trying this with the dev tools in Firefox

heithemmoumni:
cool, ok

XhmikosR:
Yeah, definitely a bad practice to remove the outline unless the effect is replaced with something else.

I'll try to have a look in the next days.

FYI Firefox handles this differently hence why the issue is only visible on Chromium based browsers AFAICT.

XhmikosR:
Alright, I tested this and like I said above, it's bad practice to do this.

PR #2914, opened 2020-01-26
Avoid using "Node" and "Node's"

tniessen:
As per our branding guidelines, both should be avoided, and Node.js should be used instead.

(I think a lot of these documents should be updated / removed anyway...)

PR #3008, opened 2020-03-05
Add official "Introduction to Node.js" as a guide

trevordmiller:
I was trying to find official documentation on core Node.js. The `Introduction to Node.js` at `https://nodejs.dev` is exactly what I was looking for but I had a hard time finding it; I figured there weren't official docs like this until I found the link buried in the `Getting Started Guide` page at `/en/docs/guides/getting-started-guide/`. This pull request makes the official `Introduction to Node.js` more discoverable so that users like me don't have to turn to potentially outdated / biased / incorrect 3rd party information on other sites.

p.s. Thanks for the `Introduction to Node.js`, it is very helpful!


Trott (approved):

XhmikosR:
This needs to be added to the translations too, right?

Trott:
> This needs to be added to the translations too, right?

@XhmikosR What's our process for that? Is it OK to land this and add translations in separate PRs as they come? Should it be added in English in this PR and then people with appropriate language skills translate the ones we can? Something else?

XhmikosR:
Not sure what the process is, personally I'd just add the English change in all translation files here, and then translators can update them later.

nschonni:
No, do not add the English as placeholder content for the other files. Those have been cleaned out because the fallback to english already happens, and you just end up with an out-of-date page when the original English gets changed.
Also, the translations should be a separate PR, as seen by the https://github.com/nodejs/nodejs.org/pull/2914 which got messed up by trying to do it in the same one.

nschonni (approved):

XhmikosR:
That's not what I meant. I meant only update the existent translation files
with the new link as is. Further modifications should happen in separate
PRs, sure.

ghost (approved):
@nodejs/website: This needs translations ;)

nschonni:
@MaledongGit as mentioned above, pushing the translations to this PR is a bad idea. That caused issues and frustration from the PR author last time when they introduced unrelated issues that stalled the PR.

PR #3058, opened 2020-04-02
zh-CN: Fixture of some Chinese translations

ghost:
For "_LTS_ release status is "long-term support", it should be
"长期支持" instead of "长期维护", because we have "Maintance" for "维护"
instead.

@nodejs/nodejs-cn


willin (approved):
LGTM

PR #3042, opened 2020-03-26
Update index.md

sh9temp:

PR #3063, opened 2020-04-02
Fix Japanese typographical errors in the about page

okuryu:
It fixes Japanese typographical errors in the Node.js description page. This is a more appropriate words for Japanese.

The original PR: https://github.com/nodejs/nodejs.org/pull/3042


ghost:
@nodejs/nodejs-ja

watilde (approved):

Issue #2654, opened 2019-10-09
`build:deploy` fails with `EMFILE: too many open files`


XhmikosR:
I'm opening this for ideas about solving the `EMFILE: too many open files` errors that happen when running the full build on CI.

It seems the OS on CI has a maximum limit of 4096 open files which we pass. I tried using graceful-fs, but the error still happens.

* Initial patch to demonstrate the issue: https://github.com/nodejs/nodejs.org/runs/253355870.
* With graceful-fs: https://github.com/nodejs/nodejs.org/runs/253360046

We'll need to run the full build at some point when we have linkinator run for all locales.

XhmikosR:
Perhaps we should use [async queue](https://caolan.github.io/async/v3/docs.html#queue)?

Any help welcome. If we go with the async queue solution #2633 will become obsolete I think.

Trott:
@nodejs/website

phillipj:
You were thinking about using an async queue to not build *all* the locales at one, right? If so that sounds like a good idea 👍

It would ofcourse make the build take a bit longer, but I don't see that as an issue in practise. Sounds much more important and valuable to be able to verify a full build succeeds in CI, than be concerned about the time it takes.

XhmikosR:
Any help welcome. We need to see how much slower it'll be also.

PR #3066, opened 2020-04-03
Run the build in parallel to save time

ghost:
For the issue (https://github.com/nodejs/nodejs.org/issues/2654), because we have a "foreach" run in sequence so that it will take quite a long time to run. Now according to XhmikosR's idea, this is a fixture for that by "Promise.all" to run it in parallel.


build.js
```diff
@@ -301,9 +303,16 @@ function fullBuild (opts) {
       }
       const filteredLocales = locales.filter(file => junk.not(file) && (selectedLocales ? selectedLocales.includes(file) : true))
       const localesData = generateLocalesData(filteredLocales)
+
+      // This is a list of parallel tasks to be run to save time
+      const asyncTasks = []
       filteredLocales.forEach((locale) => {
-        buildLocale(source, locale, { preserveLocale, localesData })
+        asyncTasks.push(buildLocale(source, locale, { preserveLocale, localesData }))
```
LinusU: Instead of using `forEach` + `push` here, you could use the `map` function which does that for you 🙌 

```js
// This is a list of parallel tasks to be run to save time
const asyncTasks = filteredLocales.map((locale) => (
  buildLocale(source, locale, { preserveLocale, localesData })
))
```

XhmikosR:
This should be tested with https://github.com/nodejs/nodejs.org/tree/master-xmr-full-build apart from running it locally to test it.

Also, please try to push your changes to an upstream branch; it makes it easier to test them than having to add new git remotes.

XhmikosR:
Also, please try to push your changes to an upstream branch; it makes it easier to test them than having to add new git remotes.

XhmikosR:
I don't think this is an improvement, at least on my machine. It's actually slower here.

XhmikosR:
Also, tested it on CI https://github.com/nodejs/nodejs.org/runs/557266539#step:8:1 vs https://github.com/nodejs/nodejs.org/runs/557271247#step:8:1

It doesn't really help AFAICT. The problem is with I/O with so many files AFAICT.

build.js
ghost: Thanks! This is fixed.

ghost:
@XhmikosR：Well……yes, it seems slower on the machine, I'll have another try, thanks!

LinusU:
> The problem is with I/O with so many files AFAICT.

Hmm, as far as I understand it, if the problem was that it was much file IO it would indeed be sped up by running in parallel. The non-speed up to me seems to indicate that the problem is CPU bound.

An approach to speeding it up in that case would be to use [worker threads](https://nodejs.org/api/worker_threads.html) 🐎

ghost:
@LinusU: I'll try to compare with diff solutions, thanks!

XhmikosR:
>Hmm, as far as I understand it, if the problem was that it was much file IO it would indeed be sped up by running in parallel. The non-speed up to me seems to indicate that the problem is CPU bound

My CPU usage does not hit 100% at any time with this patch.

ghost:
@LinusU: From the doc, there's one point:

Workers (threads) are useful for performing CPU-intensive JavaScript operations. **They will not help much with I/O-intensive work**. Node.js’s built-in asynchronous I/O operations are more efficient than Workers can be.

Maybe this isn't suitable for I/O work……? Trying....

LinusU:
I think that testing is the only way to be sure 👍 

> Node.js’s built-in asynchronous I/O operations are more efficient than Workers can be.

If the work here is saturated by IO, then surely the proposed solution here wouldn't have made it slower? 🤔

PR #3069, opened 2020-04-03
fix: release cadence post dates are yyyy-mm-dd

codebytere:
Refs https://twitter.com/tomsundone/status/1246209049559027717.

cc @MylesBorins


nschonni (approved):
The rest of the world thanks you for the ISO dates :)

PR #3071, opened 2020-04-06
chore: bump deps and regenerate package-lock.json

nschonni:


XhmikosR:
I already have https://github.com/nodejs/nodejs.org/tree/pck which should pass

XhmikosR:
I'm going to make a PR from my branch because I'm too lazy to push to your fork @nschonni 😛 You should push to the upstream so that's easier.

PR #3086, opened 2020-04-08
Blog: v12.16.2 release post

codebytere:
Refs: https://github.com/nodejs/node/pull/32313


nschonni:
@codebytere is this a non-security release? Just asking because there is a Yarn bump we'll include for docker-node if it is

codebytere:
@nschonni correct, this is not a security release!

PR #3103, opened 2020-04-14
remove logging kb article

### LINKED ISSUE #3101: doc: Error in the "How to log in Node.js" docs - custom logging solution
Regarding the custom logging solution: https://nodejs.org/en/knowledge/intermediate/how-to-log/#custom-logging

The comparison of the `level` and the `debugLevel` is not working like that, because it must be `>=` instead of `<=`. Otherwise if you set the `debugLevel` to something lower than "error" it won't show the error logs. Also I would not recommend declaring a new variable with every call of the logging function. The `levels` should be exported as a property for the logger. My recommendation would be:
```javascript
var logger = exports;

logger.levels = ["debug", "info", "warn", "error"];
logger.debugLevel = "warn";

logger.log = function (level, message) {
    if (logger.levels.indexOf(level) >= logger.levels.indexOf(logger.debugLevel)) {
        if (typeof message !== "string") {
            message = JSON.stringify(message);
        };
        console.log(level + ": " + message);
    }
}
```

And if you want to work with multiple arguments through the rest parameter:
```javascript
logger.log = function (level, ...message) {
    if (logger.levels.indexOf(level) >= logger.levels.indexOf(logger.debugLevel)) {
        console.log(level + ":\t" + message.reduce((prev, curr) => {
            return prev + " " + (typeof curr !== "string" ? JSON.stringify(curr) : curr)
        }, ""));
    }
}
```

Trott:
Closes: https://github.com/nodejs/nodejs.org/issues/3101


ghost (approved):

XhmikosR:
Shouldn't we redirect the removed pages?

Trott:
> Shouldn't we redirect the removed pages?

What should this page be redirected to?

XhmikosR:
> What should this page be redirected to?

Maybe https://nodejs.org/en/docs/ ?

Trott:
> > What should this page be redirected to?
> 
> Maybe https://nodejs.org/en/docs/ ?

On the one hand, good URLs don't break.

On the other hand, if someone is linking to this page, I'd want them to be able to tell from tooling that the link now no longer goes to what they want. And I'm not sure that redirecting to the general docs is a helpful replacement. But I'm not sure it isn't either.

@nodejs/website Thoughts?

phillipj:
I'm not very opinionated, but probably slightly in favour of removing it to tell search robots out there, this page doesn't exist anymore and they better remove traces of it in their index.

For end-users who's following an old URL, sure, it's slightly less convenient. But the docs index page isn't what they asked for anyways, and if that's interesting to them, there's still a Docs-link in the header.

phillipj (approved):

csabapalfi (approved):

aymen94 (approved):

bnb (approved):

nschonni:
> I'm not very opinionated, but probably slightly in favour of removing it to tell search robots out there, this page doesn't exist anymore and they better remove traces of it in their index.

Could also de-index it through metadata if that is the concern

PR #3120, opened 2020-04-19
fix: Enable remark-lint-no-trailing-spaces

nschonni:


locale/uk/about/governance.md
```diff
@@ -81,7 +81,7 @@ TSC може додавати додаткових членів у TSC за ст
 Вони можуть висуватись як і будь–який інший пукт порядку денного
 (дивіться "Зустрічі TSC" нижче).
 
-Не більш як 1/3 від всіх членів TSC можуть бути пов’язані одним роботодавцем.
```
XhmikosR, line 84: This is meant to be a line break.

static/legacy/README.md
```diff
@@ -1,6 +1,7 @@
 # DO NOT REMOVE THIS DIRECTORY
 
-Those assets are still being used by legacy versions of the Node.js docs.  
+Those assets are still being used by legacy versions of the Node.js docs.
```
XhmikosR, line 3: Same here

static/legacy/README.md
```diff
@@ -11,5 +12,6 @@ Those are the first versions with just the API docs and the changelog only:
 * https://nodejs.org/docs/v0.10.27/
 * https://nodejs.org/docs/v0.10.29/
 
-The actual URL rewriting is done in the nodejs.org nginx config:  
-https://raw.githubusercontent.com/nodejs/build/master/setup/www/resources/config/nodejs.org
+The actual URL rewriting is done in the nodejs.org nginx config:
```
XhmikosR, line 15: And here

XhmikosR (changes requested):

nschonni:
@XhmikosR the rule doesn't support double space as line break, so they are converted to explicit line breaks instead. I've opened an upstream issue for support, but I wouldn't block this waiting for it

XhmikosR:
IMHO since this is a perfectly valid Markdown way of doing things and this rule doesn't follow the specs, it shouldn't be merged.

nschonni:
It's the style of the Node formatting rules upstream. It's not that the rule doesn't follow the spec, it just says that style of line break shouldn't be used

Trott:
> It's the style of the Node formatting rules upstream. It's not that the rule doesn't follow the spec, it just says that style of line break shouldn't be used

Maybe add a disable comment in the `locale/uk` file affected and don't worry about the legacy README as long as the resulting format is still intelligible?

nschonni:
@Trott the style isn't being used on the English content, so I don't think it makes sense for the Ukrainian docs only

ghost:
This now is solved by adding https://github.com/nodejs/nodejs.org/blob/cf0873f7b4a8c203530667a0ba7260fb3127b73e/.remarkrc#L11.

nschonni:
@MaledongGit that isn't resolved, that is the current work around that existed when this PR was opened. This was to enable and fix the rule. That's usually the goal of the "temporary" suppression of the rules in `.remarkrc`

PR #3150, opened 2020-04-30
chore: Upgrade remark-preset-lint-node

nschonni:
Disabled the no-literal-urls rule that was added.
That will be done in a separate pass since it should be done as part of
syncing changelogs disables to match upstream.


Trott (approved):

nschonni:
The disabled rules in #2697 might also cover off a bunch of the failures coming from the changelogs

PR #3172, opened 2020-05-13
Feature/git hook i18n translate

JoseJPR:
The main idea of this PR is to provide a new script that controls through the "Husky" npm library with a pre-commit that a contributor cannot upload modified files from the "locale" folder.

A new layer of security to ensure that the contributions in translations are made with Crowdin.

If you modify a file in the "locale" folder you should be notified as you can see in the screenshot.

Captura de pantalla 2020-05-13 a las 18 40 53

The main issue: nodejs/i18n#283

@alexandrtovmach 
@zeke


.git-hooks/locale.js
```diff
@@ -0,0 +1,57 @@
...
+const fs = require('fs')
+const path = require('path')
+
+// Execute Git Diff for get all files with changes.
+const results = execSync('git diff --cached --name-only')
+
+// Create array with all files.
+const arrayResults = results.toString().split('\n')
+// Remove last element with empty line.
+arrayResults.pop()
```
zeke: I think you could do `results.toString().trim().split('\n')` to remove the newline, then you wouldn't have to `pop`

.git-hooks/locale.js
```diff
@@ -0,0 +1,57 @@
...
+
+// Define array files to not change.
+const filesNoChange = []
+
+function gitHook(expresion) {
+
+  // Define RegExp with reg param.
+  const reg = new RegExp(expresion, 'g')
```
zeke: I think this would be more readable as `new RegExp('/locale/', 'g')` rather than making the expression an argument to the function.

.git-hooks/locale.js
```diff
@@ -0,0 +1,57 @@
...
+
+  // Iterate in array for search if RegExp is true or not.
+  arrayResults.forEach(file => {
+    // Control for search if exist or not file.
+    if (fs.existsSync(path.join(__dirname, '..', file)) && reg.test(path.join(__dirname, '..', file))) {
```
zeke: What about first creating `const filepath = path.join(__dirname, '..', file)`  and reusing it in the conditional?

.git-hooks/locale.js
```diff
@@ -0,0 +1,57 @@
...
+      // Include path reference into array for resolve result console.
+      filesNoChange.push({
+        Path: path.join(__dirname, '..', file)
+      })
+    }
+  })
+
+  // Show an error and log table with the list of modified files in the "locale" folder.
+  if (filesNoChange.length !== 0) {
+    console.error('\x1b[31m', '\n ❌ You can not commit changes of the following files:', '\x1b[0m')
```
zeke: What is `x1b[31m`? Can we live without it?

.git-hooks/locale.js
```diff
@@ -0,0 +1,57 @@
...
+    console.table(filesNoChange)
```
zeke: `console.table`!? Wow I didn't know that was a thing. 😎

.git-hooks/locale.js
```diff
@@ -0,0 +1,57 @@
...
+// Execute Git Diff for get all files with changes.
+const results = execSync('git diff --cached --name-only')
+
+// Create array with all files.
+const arrayResults = results.toString().split('\n')
+// Remove last element with empty line.
+arrayResults.pop()
+
+// Define array files to not change.
+const filesNoChange = []
```
zeke:
```suggestion
const changedLocalizedFiles = []
```

zeke (changes requested):
Looks good. I left a few suggestions for refactoring.

alexandrtovmach:
@JoseJPR Could you resolve suggested changes?

package.json
```diff
@@ -21,7 +21,8 @@
     "test:lint:lockfile": "lockfile-lint --allowed-hosts npm github.com --allowed-schemes https: --empty-hostname false --type npm --path package-lock.json",
     "test:lint": "npm-run-all --parallel test:lint:*",
     "test:html": "node scripts/vnu-jar.js",
-    "test:unit": "tape tests/**/*.test.js | faucet"
+    "test:unit": "tape tests/**/*.test.js | faucet",
+    "test:git:locale": "node .git-hooks/locale.js"
```
alexandrtovmach:
```suggestion
    "test:i18n": "node tests/scripts/proctect-i18n.js"
```

package.json
```diff
@@ -83,5 +85,10 @@
     "stylelint-config-twbs-bootstrap": "^2.0.2",
     "tape": "^5.0.0",
     "vnu-jar": "20.3.16"
+  },
+  "husky": {
+    "hooks": {
+      "pre-commit": "npm run test:git:locale"
```
alexandrtovmach:
```suggestion
      "pre-commit": "npm run test:i18n"
```

alexandrtovmach (changes requested):
So far so good. Could you move and rename file:
```
.git-hooks/locale.js -> tests/scripts/protect-i18n.js
```

it looks semantic correctly for me

JoseJPR:
@alexandrtovmach Yes, in the morning we'll have progress.

.git-hooks/locale.js
JoseJPR: I think it's a good idea. Done!

.git-hooks/locale.js
JoseJPR: I think it's a good idea. Done!

.git-hooks/locale.js
JoseJPR: We can live without this code :). I would like to keep it, add the red color to the warning text.

.git-hooks/locale.js
JoseJPR: It seems to me a good option to show the result. Do you think we should remove it?

.git-hooks/locale.js
JoseJPR: I think it's a good idea. Done!

package.json
```diff
@@ -21,7 +21,8 @@
     "test:lint:lockfile": "lockfile-lint --allowed-hosts npm github.com --allowed-schemes https: --empty-hostname false --type npm --path package-lock.json",
     "test:lint": "npm-run-all --parallel test:lint:*",
     "test:html": "node scripts/vnu-jar.js",
-    "test:unit": "tape tests/**/*.test.js | faucet"
+    "test:unit": "tape tests/**/*.test.js | faucet",
+    "test:i18n": "node tests/scripts/proctect-i18n.js"
```
alexandrtovmach: sorry, I suggested typo here

`proctect` -> `protect`

alexandrtovmach (changes requested):

package.json
JoseJPR: Done!

.git-hooks/locale.js
alexandrtovmach: @JoseJPR it looks great, let's keep table for logs

.git-hooks/locale.js
alexandrtovmach: image

For me it looks okay with red cross and normal text, so I think we can live without it.

tests/scripts/protect-i18n.js
```diff
@@ -0,0 +1,56 @@
...
+
+// Create array with all files.
+const arrayResults = results.toString().trim().split('\n')
+
+// Define array files to not change.
+const changedLocalizedFiles = []
+
+function gitHook () {
+  // Define RegExp with reg param.
+  const reg = new RegExp('/locale/', 'g')
```
alexandrtovmach:
```suggestion
  const reg = /locale\/(?!en)/g;
```
We should allow original content changes, so `en` locale should be excluded from protection

tests/scripts/protect-i18n.js
```diff
@@ -0,0 +1,56 @@
...
+      // Include path reference into array for resolve result console.
+      changedLocalizedFiles.push({
+        Path: filepath
+      })
+    }
+  })
+
+  // Show an error and log table with the list of modified files in the "locale" folder.
+  if (changedLocalizedFiles.length !== 0) {
+    console.error('\x1b[31m', '\n ❌ You can not commit changes of the following files:', '\x1b[0m')
```
alexandrtovmach:
```suggestion
    console.error('\n ❌ Translations are handled via Crowdin, and direct changes not allowed. Read more in https://github.com/nodejs/nodejs.org/blob/master/TRANSLATION.md\n    Next files prevented from commit:')
```
We should give descriptive error message with link to guideline.

alexandrtovmach (changes requested):
Few more suggestions 😉

nschonni:
Husky was removed from the nodejs.dev project because of developer friction. I don't think we should add it here

tests/scripts/protect-i18n.js
JoseJPR: Cool!

.git-hooks/locale.js
JoseJPR: Done with commit of @alexandrtovmach

JoseJPR:
Hi @nschonni for me there would be no problem with define a hoot without "husky". @zeke and @alexandrtovmach what do you think?

alexandrtovmach:
@nschonni I believe we're not:
https://github.com/nodejs/nodejs.dev/blob/master/package.json#L82

Could you provide any related info? Husky it's not something bad, and I think it's just one possible way to prevent making translations in wrong way on initial steps by protecting from commit.

Please check related issue here:
https://github.com/nodejs/i18n/issues/283

alexandrtovmach (approved):

nschonni:
https://github.com/nodejs/nodejs.dev/pull/733 is re-removing it, and links to the original removal. It wasn't Husky before, but was removed here https://github.com/nodejs/nodejs.org/pull/2711

alexandrtovmach:
Okay I see.
Gun is not a problem, human with gun in hands - this is problem.
Here I described why husky shouldn't be removed from nodejs.dev https://github.com/nodejs/nodejs.dev/pull/733#pullrequestreview-415413968

The same thing should be for nodejs.org. You removed husky to speed up CI, but it's not a fix. We need just to configure what should run on CI, what shouldn't.

alexandrtovmach:
In case of this PR - husky is only one tool that can help with our goal to prevent wrong translations process as soon as possible

alexandrtovmach:
@zeke Please take a look

tests/scripts/protect-i18n.js
```diff
@@ -0,0 +1,56 @@
...
+ * folder to the repository through a git hook generated
+ * with the Husky library.
+ *
+ * Usage: $ node .git-hook/locale.js
+ *
+ * Happy releasing!
+ */
+
+'use strict'
+
```
zeke, line 15: We still want Actions workflows to be able to change these files, so let's make the Actions environment exempt from this check by adding something like this toward the top of the file:

```js
if (process.env.CI) process.exit()
```

tests/scripts/protect-i18n.js
```diff
@@ -0,0 +1,56 @@
...
+      // Include path reference into array for resolve result console.
+      changedLocalizedFiles.push({
+        Path: filepath
+      })
+    }
+  })
+
+  // Show an error and log table with the list of modified files in the "locale" folder.
+  if (changedLocalizedFiles.length !== 0) {
+    console.error('\n ❌ Translations are handled via Crowdin, and direct changes not allowed. Read more in https://github.com/nodejs/nodejs.org/blob/master/TRANSLATION.md\n    Next files prevented from commit:')
```
zeke:
```suggestion
    console.error('\n ❌ Translations are handled via Crowdin, and direct changes are not allowed. Read more in https://github.com/nodejs/nodejs.org/blob/master/TRANSLATION.md\nPlease revert changes to the following files before committing:')
```

tests/scripts/protect-i18n.js
```diff
@@ -0,0 +1,56 @@
...
+
+function gitHook () {
+  // Define RegExp with reg param.
+  const reg = /locale\/(?!en)/g;
+
+  // Iterate in array for search if RegExp is true or not.
+  arrayResults.forEach(file => {
+    // Control for search if exist or not file.
+    const filepath = path.join(__dirname, '../..', file)
+    console.log(filepath)
```
zeke:
```suggestion
```

zeke (changes requested):
Left some more feedback. Sorry for the delay.

tests/scripts/protect-i18n.js
JoseJPR: In the Github Actions documentation (https://help.github.com/en/actions/configuring-and-managing-workflows/using-environment-variables#default-environment-variables) it refers to some predefined environment variables and there is the one you comment, process.env.CI, so just adding the line you propose would be enough.

JoseJPR: @zeke Done -> https://github.com/nodejs/nodejs.org/pull/3172/files#diff-6d229f474368307be4db86c19d5c3f7cR20

package.json
```diff
@@ -69,6 +70,7 @@
     "cross-env": "^7.0.2",
     "delay-cli": "^1.1.0",
     "faucet": "0.0.1",
+    "husky": "^4.2.5",
```
obensource, line 73: Since `husky` has been regarded as a footgun for developers by some of our contributors – and seeing that it has been added and [removed](https://github.com/nodejs/nodejs.dev/pull/733) from this repo multiple times due to that friction: elevating the conversation about how we want to handle commit-linting across the project (as @MylesBorins mentioned [here](https://github.com/nodejs/nodejs.dev/pull/733#issuecomment-631634322)) seems like the right thing to do – rather than reintroducing it here just to help the i18n initiative protect one repo that consumes its exports from commits made directly to it (which unintentionally green-lights its use for other cases in `/nodejs.org`). Imo, I agree that when there is project-wide consensus on how we want to enforce commit linting (eg. via `husky` or as a CI job), then we can be more confident on how we want 'source of truth' repos like `/i18n` to handle this in order to prevent direct commits elsewhere. 👍

Given that, I'm going to request that we at the very least:
* Hear more opinions here about approving this PR which reintroduces `husky` into this repo, and hopefully get greater consensus on that beyond the folks who're involved with `i18n` and `nodejs.org`.
* Elevate the conversation accordingly to the TSC via [admin](https://github.com/nodejs/admin).

cc @alexandrtovmach @zeke @MylesBorins

obensource (changes requested):
Please see my comment regarding the reintroduction of `husky`. Thanks a million for all your fantastic work @JoseJPR – it's very appreciated! 🙏

package.json
obensource: cc @nodejs/website

zeke: I haven't read all the historical context as it pertains to use of `husky` in the @nodejs org, but if we are going to remove husky, is there some alternative mechanism now being used by the @nodejs project to lint commits?

tests/scripts/protect-i18n.js
```diff
@@ -0,0 +1,61 @@
...
+ */
+if (process.env.CI) process.exit()
+
+// Require main dependences.
+const { execSync } = require('child_process')
+const fs = require('fs')
+const path = require('path')
+
+// Execute Git Diff for get all files with changes.
+const results = execSync('git diff --cached --name-only')
```
Aissaoui-Ahmed, line 28:
```suggestion
const results = execSync('git diff --cached --name-only locale/ :!locale/en')
```

tests/scripts/protect-i18n.js
```diff
@@ -0,0 +1,61 @@
...
+
+// Create array with all files.
+const arrayResults = results.toString().trim().split('\n')
+
+// Define array files to not change.
+const changedLocalizedFiles = []
+
+function gitHook () {
+  // Define RegExp with reg param.
+  const reg = /locale\/(?!en)/g;
```
Aissaoui-Ahmed, line 38:
```suggestion
  const reg = /\//g
```

package.json
```diff
@@ -83,5 +85,10 @@
     "stylelint-config-twbs-bootstrap": "^2.0.2",
     "tape": "^5.0.0",
     "vnu-jar": "20.3.16"
+  },
+  "husky": {
+    "hooks": {
+      "pre-commit": "npm run test:i18n"
+    }
```
Aissaoui-Ahmed, line 92:
```suggestion
  }
```

package.json
Aissaoui-Ahmed, line 73:
```suggestion
```

Aissaoui-Ahmed (changes requested):
Replace with [pre-commit](https://www.npmjs.com/package/pre-commit)

tests/scripts/protect-i18n.js
Aissaoui-Ahmed: get diff of all locale folder and exclude EN
```suggestion
const results = execSync('git diff --cached --name-only locale/ :!locale/en')
```

JoseJPR: Hi @Aissaoui-Ahmed we are controlling that [here](https://github.com/JoseJPR/nodejs.org/blob/feature/git-hook-i18n-translate/tests/scripts/protect-i18n.js#L38). But, do you see any greater benefit in your proposal? :)

tests/scripts/protect-i18n.js
```diff
@@ -0,0 +1,61 @@
...
+
```
Aissaoui-Ahmed:
```suggestion
```

tests/scripts/protect-i18n.js
```diff
@@ -0,0 +1,61 @@
...
+  // Iterate in array for search if RegExp is true or not.
+  arrayResults.forEach(file => {
+    // Control for search if exist or not file.
+    const filepath = path.join(__dirname, '../..', file)
+    if (fs.existsSync(filepath) && reg.test(filepath)) {
```
Aissaoui-Ahmed:
```suggestion
    if (fs.existsSync(filepath) && file !== '') {
```

tests/scripts/protect-i18n.js
Aissaoui-Ahmed: Hi @JoseJPR if you run `git diff --cached --name-only locale/ ':!locale/en'` return only changes on folder `locale` and exclude `locale/en`
you don't need [regex](https://github.com/JoseJPR/nodejs.org/blob/feature/git-hook-i18n-translate/tests/scripts/protect-i18n.js#L38)

Aissaoui-Ahmed (approved):

Aissaoui-Ahmed (approved):

package.json
ghassanmas: @zeke, I think for linting js files, [standard](https://standardjs.com/) is being used. As shown in this snippets, https://github.com/nodejs/nodejs.org/blob/21353d0db79a70b207cdb24b114affaa7b807e12/package.json#L15-L24

If its requires to `npm run test` before each commit, then standard should lint the js file while other modules are used in parallel to lint different type of files  as shown above.

aymen94 (approved):

ghost:
It's been a year and a half since it began, however we don't use "Husky" any more according to some discussions, so this PR will be closed and can also be opened for futher discussion.

PR #3180, opened 2020-05-18
chore: add Catalina native dependencies guide

mmarchini:
With the Catalina macOS upgrade, several users started to face issues when
trying to install native dependencies. There's a guide on `node-gyp` repository
with possible workarounds for the issue, but the guide is hard to find (if you
search for `gyp canalina issues` and variants on Google, you won't get the
guide on first page, and if you search for the error `gyp: No Xcode or CLT
version detected!`, the issue on node-gyp where the problem is discussed and
the guide is linked does not show in the first few results. Adding the guide to
our Guides page might make discoverability better in the future.


Trott:
I don't think this page is a likely place for someone to look for this information, but I don't have a better idea, so not a blocking comment by any means. Just wondering aloud in case someone else has a good idea of where to include this either additionally or instead of this location. Looks good to me as far as the change itself goes.

mmarchini:
@Trott agreed, but at the same time I'm not sure where else we could add this link, and I think it's important for this to be linked on the website since it's quite easy for Catalina users to hit the issue. We could add it in the front page, but that might be a little too much.

Trott:
> @Trott agreed, but at the same time I'm not sure where else we could add this link, and I think it's important for this to be linked on the website since it's quite easy for Catalina users to hit the issue. We could add it in the front page, but that might be a little too much.

Yeah, I don't have a better suggestion and I'm totally fine with this. If a better place is identified later, it can always be moved or added there as well.

Trott (approved):

MylesBorins (approved):
LGTM

PR #3219, opened 2020-06-05
blm: first step towards intl

MylesBorins:
First step in refactoring the BLM page to make it possible to internationalize it.

I've verified locally and there are 0 regressions or changes in visual layout.

While the content is now found in `locale/en/index.md` it is still primarily in HTML due to all the custom styles. This will likely be a blocker in translation. If we actually want to do some translations of this we might want to convert all the content to pure markdown and significantly simplify the custom styles. Before doing this we should like land the remainder of @trott's content updates.


layouts/partials/blm-html-head.hbs
```diff
@@ -0,0 +1,161 @@
...
+  <meta property="og:image:width" content="224">
+  <meta property="og:image:height" content="256">
+  <meta property="og:description" content="#BlackLivesMatter">
+
+  <meta name="twitter:card" content="summary">
+  <meta name="twitter:site" content="@nodejs">
+  <meta name="twitter:title" content="#BlackLivesMatter">
+  <meta name="twitter:image" content="https://nodejs.org/static/images/logo-hexagon-card.png">
+  <meta name="twitter:image:alt" content="The Node.js Hexagon Logo">
+  <style>
```
MylesBorins: this could be broken out into a separate style file... but too much work imho

MylesBorins:
this is ready for review. I've managed to separate style and content, for the most part. Majority of content has now also been refactored to markdown, maintaining some markup for custom styles.

Took some liberties with the layout while refactoring because things were changing anyways  ¯\\\_(ツ)\_/¯. Other than adding titles and layout there were no copy changes. 

Screen shot below

edit: slightly tweak the padding at the top of an h2, updated screen shot

edit: added an hr before content about getting canonical urls, updated screen shot

localhost_8080_en_ (3)

layouts/partials/blm-styles.hbs
```diff
@@ -0,0 +1,122 @@
+<style>
```
MylesBorins, line 1: this could be moved to scss, but honestly it is way easier to integrate as a handlebar partial.

jasnell (approved):

bnb (approved):

Trott:
Not blocking, but I do want to acknowledge that we've received feedback that the BLM content seems US-centric (which is acknowledged on the page itself) and that expanding that content to languages other than English is likely to exacerbate that sense. Again, not blocking, but also don't want to fail to acknowledge that. (Might be good to make a point to get some reviews from non-U.S. residents.)

MylesBorins:
@Trott while this refactoring will make it possible to do intl, it does not mean we 

1. need to do it right away
2. intl implementations need to take the entire page

I was actually imagining that we could include the landing page as a banner for intl sites, potentially include relevant statistics, and use this as a model for what we do on the main page after the 1 week of including the new landing page has passed.

I've been following public discussion closely and seen voices both of support as well as dissent from various international communities. I think we should definitely lock down copy before translating... but I don't think that we should be entirely discouraged by voices of dissent.

MylesBorins:
Landing as it is a rather big layout improvement.. we can keep iterating on copy

PR #3212, opened 2020-06-05
chore: consolidate and clarify Black Lives Matter text

Trott:
Consolidate two related paragraphs into a single paragraph. Use shorter
sentences for more impact. Fix a typo.

PR #3228, opened 2020-06-05
chore: add some (more?) non us cases

### LINKED ISSUE #3226: Anonymous feedback regarding the #BlackLivesMatter site
Hey,

I have received feedback from a third party who wished to stay anonymous regarding the new website being very US-centric in nature.

I am honestly not sure if police brutality towards people of color is a bigger problem in the US (I personally felt a lot more racism in the US walking around with friends who were not white than anywhere else I ever visited).

Would it be possible to add some other non-us examples? Namely I can only speak of my locale [this would probably be the Israeli equivalent](https://en.wikipedia.org/wiki/July_2019_Ethiopian_Jews_protest_in_Israel) - namely one name (Solomon Teka).

benjamingr:
Add some examples from Israel regarding the feedback I've received about US-centrism of the site.

Fixes: https://github.com/nodejs/nodejs.org/issues/3226

### DIFF
--- a/layouts/black-lives-matter.hbs
+++ b/layouts/black-lives-matter.hbs
@@ -201,6 +201,7 @@
       <li>Robert Lawrence White</li>
       <li>Rumain Brisbon</li>
       <li>Ryan Twyman</li>
+      <li>Solomon Teka</li>
       <li>Samuel David Mallard</li>
       <li>Samuel DuBose</li>
       <li>Sandra Bland</li>
@@ -220,6 +221,7 @@
       <li>William Green</li>
       <li>Willie McCoy</li>
       <li>Yassin Mohamed</li>
+      <li>Yehuda Biadga</li>
       <li>... <a href="https://mappingpoliceviolence.org/unarmed">and</a> <a href="https://mappingpoliceviolence.org/unarmed2014">countless</a> <a href="https://apps.npr.org/documents/document.html?id=6933593-NPR-CodeSwitch-Saytheirnameslistv3">more</a> <a href="https://ebwiki.org/">names</a> not listed here whose lives were taken by violence and brutality.</li>
     </ul>
     <p>We've listed the names of the Black folks who have been murdered by police brutality in the U.S. dating back to 2009, both because the U.S. is the origin of the Black Lives Matter movement and currently the heart of the resistance.</p>
--- a/layouts/black-lives-matter.hbs
+++ b/layouts/black-lives-matter.hbs
@@ -224,7 +224,7 @@
       <li>Yehuda Biadga</li>
       <li>... <a href="https://mappingpoliceviolence.org/unarmed">and</a> <a href="https://mappingpoliceviolence.org/unarmed2014">countless</a> <a href="https://apps.npr.org/documents/document.html?id=6933593-NPR-CodeSwitch-Saytheirnameslistv3">more</a> <a href="https://ebwiki.org/">names</a> not listed here whose lives were taken by violence and brutality.</li>
     </ul>
-    <p>We've listed the names of the Black folks who have been murdered by police brutality in the U.S. dating back to 2009, both because the U.S. is the origin of the Black Lives Matter movement and currently the heart of the resistance.</p>
+    <p>We've listed the names of the Black folks who have been murdered by police brutality in the U.S. and other places in the world dating back to 2009, both because the U.S. is the origin of the Black Lives Matter movement and currently the heart of the resistance.</p>
 
     <p>With that said, we fully recognize that white supremacy and police brutality is a global problem. While we have not listed all of Black people who have lost their lives to police murder globally here, their lives mattered.</p>
   </section>


benjamingr:
If anyone wants to learn more: https://en.wikipedia.org/wiki/July_2019_Ethiopian_Jews_protest_in_Israel

jasnell (approved):

benjamingr:
btw: I have no idea how I got a merge commit (I just edited through the GitHub UI I didn't even clone the website repo) - but if that is an issue lmk and I will clone, cherry pick the right commit, rebase and push

addaleax:
The description under the list says that this refers to U.S. cases, you may want to update that as well.

layouts/black-lives-matter.hbs
```diff
@@ -220,9 +221,10 @@
       <li>William Green</li>
       <li>Willie McCoy</li>
       <li>Yassin Mohamed</li>
+      <li>Yehuda Biadga</li>
       <li>... <a href="https://mappingpoliceviolence.org/unarmed">and</a> <a href="https://mappingpoliceviolence.org/unarmed2014">countless</a> <a href="https://apps.npr.org/documents/document.html?id=6933593-NPR-CodeSwitch-Saytheirnameslistv3">more</a> <a href="https://ebwiki.org/">names</a> not listed here whose lives were taken by violence and brutality.</li>
     </ul>
-    <p>We've listed the names of the Black folks who have been murdered by police brutality in the U.S. dating back to 2009, both because the U.S. is the origin of the Black Lives Matter movement and currently the heart of the resistance.</p>
+    <p>We've listed the names of the Black folks who have been murdered by police brutality in the U.S. and other places in the world dating back to 2009, both because the U.S. is the origin of the Black Lives Matter movement and currently the heart of the resistance.</p>
```
Trott: This is going to conflict with https://github.com/nodejs/nodejs.org/pull/3212.  If we land that one first, I'd be happy to suggest a modification. (It probably just means adding something to the affect that instead of being an entirely U.S. list, it is a _predominantly_ U.S. list.

benjamingr: @Trott I don't feel strongly about this PR landing at all (or attribution fwiw), I do think it makes sense to include these names and outline that this is a global issue the project is objecting to.

RichardLitt:
Here's [a database of 461 people killed by police in Canada](https://newsinteractives.cbc.ca/longform-custom/deadly-force). [Regis Korchinski-Paquet](https://www.cbc.ca/news/canada/toronto/regis-korchinski-paquet-toronto-1.5593718) is one of the most recent and not in this database, along with [Chantel Moore](https://www.cbc.ca/news/canada/new-brunswick/edmundston-police-shooting-fatal-1.5597994). 

I was unable to find a simple list for the UK, which also has systemic racism and police brutality. [432 people of aboriginal descent](https://www.theguardian.com/australia-news/2020/jun/01/deaths-in-our-backyard-432-indigenous-australians-have-died-in-custody-since-2008) died in custody in Australia since 1991, which doesn't include fatalities before custody. 

I'm happy to add all of the names. Linking to other countries might be better UX (but, really, who cares? Isn't scrolling the least we can do?).

MylesBorins:
I would like us to get some feedback from more underrepresented groups, especially black americans, before landing this.

I know the intent is 100% good, but want to ensure that it doesn't undermine our message in the eyes of the people we are advocating for.

benjamingr:
@MylesBorins fwiw, I can share more feedback regarding who asked https://github.com/nodejs/nodejs.org/issues/3226 in private. I sent you a message on FB.

In particular the person who asked me about adding Salomon Teka identified themselves as black (but not African-American), resided in the US for a while and was very supportive of the site itself. I suspect they would be _very_ uncomfortable speaking for the whole community or in public (they are a pretty private person).

I suspect I am quite ignorant compared to some parties here on the issue itself and am only trying to forward feedback as I'm getting it.

MylesBorins:
Fwiw myself and others have been getting feedback privately and bringing it
to the repo in cases where people may not be comfortable. Just want to make
sure we don't make these kind of decisions without feedback

Trott:
Rebased to resolve conflicts. I also adjusted the wording at the end. PTAL.

benjamingr:
@Trott this is currently blocked on @MylesBorins - once he gathers the feedback he mentioned in https://github.com/nodejs/nodejs.org/pull/3228#issuecomment-639531609 we can proceed.

Myles is there an ETA on that?

Just to clarify about this particular case: I'm sure it can make certain parties in Israel uncomfortable (like the police) - these two are the most clean cut cases where:

 - A white officer killed a non-white person.
 - That person was not actually a danger to the police officer in question. In Teka's case the police officer was off-duty as well.
 - The police officer did not act according to protocol, in Teka's case they should have shot to the sky before discharging their weapon. In Biadga case as well.
 - If they were white, this would have likely never happened.

I think this boils down to whether this show of solidarity is about "black lives matters" as a cultural problem of racism in the USA or as a global problem. I am pretty sure these particular cases don't diminish the cause - they underscore that the issue has leaked beyond the scope of a US internal affair.

There are other cases as well - I only picked these two because I was asked (by a person affected by it) in order to underscore the globality of the struggle. I am championing this on their behalf but don't actually have particularly strong feelings about whether or not we should do this.

MylesBorins:
I'm still not particularly comfortable with this change as it exists, but have not had the time to engage with black individuals to gather direct feedback. Sadly we do not have many active black contributors in Node.js and we need to rely on outside individuals to give us emotional labor to help answer this for us... and it is genuinely uncomfortable approaching people who do not have a commitment to our project a requesting that labor.

The landing page for Black Lives Matter does call it a global movement, so this seems in line with the mission, although my gut is still conflicted.

So as a thought as to how we could move this forward. If we are going to make this a more international list, it should be broader than these two additional names. On the BLM wikipedia page there are a number of international examples, we should include those names too at a minimum

https://en.wikipedia.org/wiki/Black_Lives_Matter#BLM_international_movement

Lets get this in a place where it is truly representing the global problem, then reach out to folks for review rather than having to consult with them throughout the process.

Thoughts?

benjamingr:
I have gathered feedback from two other parties from the black community in Israel which is impacted by systematic racism who both felt ambivalent about this. They both liked the fact it raises awareness which is important but it "airs dirty laundry in public" which is complicated in the political climate in Israel.

Since neither of the two people affected whom are Node.js users felt strongly about merging this and the one who did isn't in tech - and since some other project members are uncomfortable with this - I'm going to go ahead and close this.

For what it's worth - I did get a lot of messages regarding how "US-Centric" this comes off as.

benjamingr:
(as usual - if someone else feels strongly about it - feel free to reopen)

mmarchini:
Just to share a little more context: The fact that it currently focuses on US police brutality was discussed in the PR where we changed the website, but we decided to keep it mostly unchanged, with an additional note that this problem exists on many other countries. The decision was based on the fact that the Black Lives Matter movement originated in the US, and the critical mass we're seeing in the US is taking unprecedent proportions.

I would need to look further into news of other countries, but at least Brazil was recently protesting police brutality inside the countries with messages of solidarity to the US, but what is happening in the US is not what sparked the protests there (although it might've influenced and also might've brought more people to the streets). I'm not opposed to including more names, but I don't want to blur focus on current events either.

PR #3238, opened 2020-06-07
Clarify handling non-null as an error

veganaiZe:


ghost (commented):
Hi @veganaize：

Thanks for  your great contribution first! Here's my suggestion for your modifications ;)

"if it is" is the shortened expression of "if it's non-null", I don't think it needs to be written in full.
However, if you suggest doing so, maybe you can use something like：

_check if the first parameter is non-null and handle it as an error when non-null._

veganaiZe:
I don't think it's worth a big debate.  I found it confusing.
[ I've provided what I think would be more appropriate.  I think your suggestion sucks, too. ]

locale/en/knowledge/errors/what-are-the-error-conventions.md
```diff
@@ -49,4 +49,4 @@ As you can see from the example, the callback is called with null as its first a
 
 The `callback` function shows the reason for this: it allows a user to easily know whether or not an error occurred. If `null` was not the first argument passed on success, the user would need to check the object being returned and determine themselves whether or not the object constituted an error - a much more complex and less user-friendly approach.
 
-So to wrap it all up, when using callbacks, if an error comes up, then pass it as the first argument. Otherwise, pass `null` first, and then your return arguments. On the receiving end, inside the callback function, check if the first parameter is non-null; if it is, handle it as an error.
+So to wrap it all up, when using callbacks, if an error comes up, then pass it as the first argument. Otherwise, pass `null` first, and then your return arguments. On the receiving end, inside the callback function, check if the first parameter is non-null. If it's non-null then handle it as an error.
```
Trott:
```suggestion
So to wrap it all up, when using callbacks, if an error comes up, then pass it as the first argument. Otherwise, pass `null` first, and then your return arguments. On the receiving end, inside the callback function, check if the first parameter is non-null. If it is non-null, then handle it as an error.
```

Trott:
Most of these knowledge guides have a lot of opportunities for improvement in terms of style and clarity. They are not well-maintained and I would be in favor of getting rid of most or all of them. That said, being explicit about "it" being something that is "non-null" here is an improvement, so I'd be in favor of this change. I also left a suggestion that I think improves it further, but it's optional.

veganaiZe:
Would you also be in favor of burning paper books because they're "no longer maintained"?
I have found the (old) guides to be very concise and helpful.  I just thought I'd help out by correcting one, single ambiguous line.  All anyone had to do was click "OK" but apparently that was too difficult.  This really shouldn't be such a big fucking deal.

Trott (approved):

veganaiZe:
So this is how you guys hijack commits from others?  All from an organization that claims to be white privileged?
Typical.

Trott:
> So this is how you guys hijack commits from others?

Screen Shot 2020-06-23 at 5 08 42 PM

Trott:
> So this is how you guys hijack commits from others? 

I'm not sure if you're frustrated that I re-opened the PR after you closed it, or if you're frustrated that I'm the author on the most recent commit rather than you.

If the former: I was accepting your feedback that it would be simple to accept your PR and there's not a reason to have a whole lot of process or discussion around it. I'm sorry if my actions communicated something less positive. I can certainly understand that it might have been better had I, for example, said something acknowledging the feedback.

If the latter: You are the author of the first commit. Because this repository is set up to default to Rebase rather than Squash, it kept me as the author of the second commit. That wasn't intentional. That was just a result of the way the repo is set up. I can understand how that could be frustrating.

veganaiZe:
You nit picked "it's" versus "it is" and then improperly inserted a comma.  In all honesty I could really care less **_BUT_** this behavior is coming from a group of people who've chosen to take a political stance -- across their homepage.  I'm a little sick of seeing so many white people claim that they're "privileged", yet they continue on with the same undermining behavior.

Rather than virtue signaling I would prefer that the problematic behavior is changed.  Actions speak louder than words.

That aside, thank you for discussing the issues (around the issue) with me.  And thank you for your effort to improve documentation that most others would have mindlessly written off as "obsolete".

Trott:
So it sounds like something I could do differently going forward to make things better for other people would be to nit-pick less within a PR and move those sorts of things to my own PRs or else let them drop entirely.

There is definitely a culture of long discussions about minutiae in the docs. I think the attention to detail is helpful, but there are better ways to achieve that result than attaching it to small changes, especially from first-time contributors. In this case, perhaps I could have approved and landed this PR as-is, and then followed up with my own PR to change "it's" to "it is" here and anywhere else that I think it would be better that way.

veganaiZe:
Thank you again for your patience with me.  In hindsight I do think that your way makes the most sense (with the emphasis on "is").

PR #3249, opened 2020-06-12
Added the minimum criteria for beginners.

suffisme:
As for starting node.js, many gets confused what exactly knowledge and skills required to start this. So if on the first page it is written that a a basic JavaScript and knowledge of callback functions and modules methods is enough to start with node.js, it would give more clearance to the newbies and first time visitor to the websites.


Trott (changes requested):
Welcome, @suffisme, and thanks for the pull request! Let's see what others think, because maybe they'll disagree with me, but I don't think this information belongs here. Or at least not worded like this. Maybe something like "easily" somewhere or "with knowledge of only basic JavaScript!" or something like that? But even that, I'd question. Because it depends what you're trying to do. Many Node.js applications require knowledge of more than basic JavaScript.

suffisme:
Ok, I got you. But all I wanted is “just for beginners“, as many unofficial sources many times pass wrong information. But if the information directly coming from official website that for “starting Nodejs” you should have “atleast” that knowledge in JS.

Thank You for the response.

locale/en/about/index.md
```diff
@@ -7,7 +7,7 @@ trademark: Trademark
 # About Node.js®
 
 As an asynchronous event-driven JavaScript runtime, Node.js is designed to build
-scalable network applications. In the following "hello world" example, many
+scalable network applications, starting for which doesn't require advanced knowledge of JavaScript, you must be familiar with callback functions and basic JavaScript language. In the following "hello world" example, many
```
MaledongGit, line 10: IMO：You don't need to be so wordy to explain it, but just say something like this following:

As an asynchronous event-driven JavaScript runtime, Node.js is designed to build
scalable network applications. It just needs very basic JavaScript knowledge (See PART 1 at https://javascript.info/).

In the following ……

ghost:
IMO：You don't need to be so wordy to explain it, but just say something like this following, this will make others think they don't understand js. However, considering for the very beginners, we can just make tips for them with one simple thing, something like this following (see bold statement):

> As an asynchronous event-driven JavaScript runtime, Node.js is designed to build
> scalable network applications. **It's based on the very basic JavaScript (See PART 1 at https://javascript.info/).**
> 
> ...

ghost:
Another good idea is that we can add a link directly onto "JavaScript", others can see if they want to know the basic knowledge：

As an asynchronous event-driven [JavaScript ](https://javascript.info/js)runtime, ....

What do you think of this? @Trott 

PS：@nodejs/website

willin (approved):

Trott:
This has been inactive for long enough that it seems like we can close it. Please comment or re-open if that's not right.

PR #3268, opened 2020-06-20
zh-CN: improve the Chinese translation of about/index.md

dead-horse:


dead-horse:
I'll invite some native to check the translate. cc @joyeecheung  @yorkie

ghost (dismissed):

locale/zh-cn/about/index.md
```diff
@@ -25,23 +25,20 @@ server.listen(port, hostname, () => {
 });
 ```
 
-这与今天使用 OS 线程的更常见并发模型形成了对比。基于线程的网络效率相对低下，使用起来非常困难。此外，Node.js 的用户不必担心死锁过程，因为没有锁。Node 中几乎没有函数直接执行 I/O 操作，因此进程从不阻塞。由于没有任何阻塞，可伸缩系统在 Node 中开发是非常合理的。
+这与当今比较常见的采用操作系统线程的并发模型形成了鲜明对比。基于线程的网络效率相对较低且更难以使用。此外，由于没有锁，Node.js 的用户不用担心进程死锁的问题。Node.js 中几乎没有函数直接执行 I/O，所以进程从不阻塞，因为没有任何阻塞，所以用 Node.js 来开发可扩展系统是非常合理的。
```
fengmk2: `所以进程从不阻塞` 中的 `所以` 可以去掉。否则与后面的因为所以不搭配。

fengmk2 (approved):
Others LGTM.

locale/zh-cn/about/index.md
dead-horse: Almost no function in Node.js directly performs I/O, so the process never blocks. Because nothing blocks, scalable systems are very reasonable to develop in Node.js.

Node.js 中几乎没有函数直接执行 I/O，进程就不会被阻塞，因为没有任何阻塞，所以用 Node.js 来开发可扩展系统是非常合理的。

ghost (commented):
No other comments but just three suggestions：

1）"so the process never blocks" ：In the origin tranlation, "never" isn't translated into Chinese.

2）Almost no function in Node.js directly performs I/O, so the process never blocks.  Because nothing blocks, scalable systems are very reasonable to develop in Node.js.

“**so the process never blocks.  Because nothing blocks**"：This is a typical English expression, however it's a bit too wordy in Chinese expression. So if we suggest translating it, just use ”既然如此".

3）“perform I/O” can be translated into “执行 I/O 操作” instead of “执行 I/O”， the first is easy to understand.

All in all, my suggestion is：

Node.js 中几乎没有函数直接执行 I/O 操作，其进程从不会被阻塞；既然如此，用 Node.js 来开发可扩展系统是非常合理的。

yorkie:
@dead-horse May I ask you to have small updates on https://github.com/nodejs/nodejs.org/blob/master/locale/zh-cn/docs/guides/event-loop-timers-and-nexttick.md? There is "事件轮询" words in this article.

locale/zh-cn/about/index.md
```diff
@@ -25,23 +25,20 @@ server.listen(port, hostname, () => {
 });
 ```
 
-这与今天使用 OS 线程的更常见并发模型形成了对比。基于线程的网络效率相对低下，使用起来非常困难。此外，Node.js 的用户不必担心死锁过程，因为没有锁。Node 中几乎没有函数直接执行 I/O 操作，因此进程从不阻塞。由于没有任何阻塞，可伸缩系统在 Node 中开发是非常合理的。
+这与当今比较常见的采用操作系统线程的并发模型形成了鲜明对比。基于线程的网络效率相对较低且更难以使用。此外，由于没有锁，Node.js 的用户不用担心进程死锁的问题。Node.js 中几乎没有函数直接执行 I/O 操作，其进程从不会被阻塞，因此用 Node.js 来开发可扩展系统是非常合理的。
```
yorkie, line 28: > 此外，由于没有锁，Node.js 的用户不用担心进程死锁的问题

IMO we need to emphasize that this is in user-land to avoid other ambiguities.

willin (approved):
Appreciate

locale/zh-cn/about/index.md
MaledongGit: I strongly suggest you submitting it into the "Issues" of the doc, this is a discussion instead of suggestions for the translation. Because if you do this, the English version will be first changed.

I think you can directly do changes to the English version and let tech members have a review first, and then we can translate them into Chinese, and call others to related languages as well ;)

ghost: > > 此外，由于没有锁，Node.js 的用户不用担心进程死锁的问题
> 
> IMO we need to emphasize that this is in user-land to avoid other ambiguities.

I think you can directly do changes to the English version and let tech members have a review first, and then we can translate them into Chinese, and call others to related languages as well ;)

yorkie: > Furthermore, users of Node.js are free from worries of dead-locking the process, since there are no locks.

Hmm, the English version is OK because it mentioned the user at the former sentence, however, the Chinese description in this PR doesn't guarantee this.

yorkie:
> May I ask you to have small updates on https://github.com/nodejs/nodejs.org/blob/master/locale/zh-cn/docs/guides/event-loop-timers-and-nexttick.md? There is "事件轮询" words in this article.

And IMO this inconsistency in the official documentation is a little bit confusing to developers.

Issue #3263, opened 2020-06-19
Broken link on BLM page


Trott:
* **URL**: https://nodejs.org/en/black-lives-matter/
* **Browser version**: Chrome 83
* **Operating system**: macOS

The word _more_ links to https://apps.npr.org/documents/document.html?id=6933593-NPR-CodeSwitch-Saytheirnameslistv3 but that page isn't loading. Console is reporting a 404 error for https://www.documentcloud.org/documents/6933593-NPR-CodeSwitch-Saytheirnameslistv3.js. This was reported to me by someone else and I am having the same expreience. Might be a temporary outage for that site, or might be a link that needs fixing, or...?

<img width="1236" alt="Screen Shot 2020-06-18 at 5 05 00 PM" src="https://user-images.githubusercontent.com/718899/85083208-e8a33c00-b185-11ea-8629-bb50fc1ad640.png">

h/t @stopfstedt

jonchurch:
Here's an archive.org link that could replace it https://web.archive.org/web/20200607130311/https://apps.npr.org/documents/document.html?id=6933593-NPR-CodeSwitch-Saytheirnameslistv3

Or alternatively, a wikipedia page that lists lists of people killed by police in the US https://en.wikipedia.org/wiki/Lists_of_killings_by_law_enforcement_officers_in_the_United_States#Lists_of_killings

nel417:
I swapped out the link in [more] for the wiki link but I'm having some failing tests. I would love to contribute to this, this is super cool. Anything I can do to get the tests passing? here's my PR https://github.com/nodejs/nodejs.org/pull/3277

Thanks!


**Edit after posting this**
I think my failed tests were from my linter and code not matching the style guide, Just resubmitted with passing tests.

MylesBorins:
Closing as I believe this is fixed

PR #3277, opened 2020-06-24
changed "more" link on BLM page

nel417:
from issue #3263 the "more" link was broken (not broken, just the NPR link wasn't loading) and a community member listed two possible links that could take its place. I have replaced the link with the Wikipedia article.


nel417 (commented):
Changed link for [more] to one of the alternate links listed in issue

PR #3278, opened 2020-06-24
Update black-lives-matter.md

nel417:
#3263 Updated the black lives matter page's [more] link with a link from a list someone for the community posted would be eligible. The npr link would not load


Trott (approved):
I think we should eventually swap out this link for one specifically documenting black lives. But this is an improvement over a link that does not work at all, so I'm 👍 on it.

Trott:
Welcome, @nel417, and thanks for the pull request!

pierreneter (approved):

pierreneter:
👍 Totally agree to merge this PR

PR #3298, opened 2020-07-03
Recommend node-glob for file search

steelywing:
Recommend node-glob for file search in detail info with examples.


MaledongGit:
@steelywing : Maybe in the English version first?

MaledongGit:
@steelywing : Maybe you can submit for the English version first for us to review instead of for the 'fr' version....

ghost (changes requested):
@steelywing : Thanks for your contribution to extend the 'node-glob search'. 

However, I found you're editing the "fr" version, maybe you can submit for the English version first for us to review instead of for the 'fr' version....This will be blocked until you switch to the English version.

PR #3300, opened 2020-07-06
Add n to package manager page

shadowspawn:
Adding https://github.com/tj/n to the list of options for "Installing Node.js via package manager" as one of the popular options for Mac and Linux.

I included a brief overview, link to home page, and quick start instructions for the easiest scenario.


nschonni (approved):

locale/en/download/package-manager.md
```diff
@@ -164,6 +165,19 @@ Or build manually from pkgsrc:
 cd pkgsrc/lang/nodejs && bmake install
 ```
 
+## n
+
+`n` is a simple to use node version manager for Mac and Linux. Specify the target version to install using a rich syntax, or select from a menu of previously downloaded versions.
```
ljharb: it might be worth calling out here that `n` changes the system-wide version of node, as opposed to a per-shell version manager

Trott:
```suggestion
`n` is a simple to use Node.js version manager for Mac and Linux. Specify the target version to install using a rich syntax, or select from a menu of previously downloaded versions. 
```

We try to use _Node.js_ (and joke that "the js is silent") in text and reserve `node` for the executable.

ljharb: hopefully the dot is silent too

shadowspawn: Hmm, I will be reviewing my READMEs. I had missed that pattern. 😄

shadowspawn: I am not sure if it is worth getting into the detail, but `n` does follow a different pattern to many other managers. Would this be helpful?

> `n` can install Node.js system-wide or user-local, and limited support for running a version for a single command.

ljharb: That sounds good - by "user-local" tho, that's still user-wide, right? like you can't have two tabs open as the same user with different node versions available at the same time?

(that's the difference i think is important to call out)

shadowspawn: Yes, I was referring to user-wide. 

I'm taking care with the wording, as `n` is more flexible than it once was. You can have two tabs open as the same user with different node versions available at the same time. There is support for modifying the `PATH` to run a command from a vanilla install of particular Node.js. If the command is a shell, this is somewhat similar to the `nave` use pattern.

```sh
bash-3.2$ node --version
v12.20.2
bash-3.2$ n exec lts zsh
~ % node --version
v14.15.5
```

I'll think some more...

shadowspawn: Take 2:

> `n` is a simple to use node version manager for Mac and Linux. Specify the target version to install using a rich syntax, or select from a menu of previously downloaded versions. The versions are installed system-wide or user-wide, and for more targeted use you can run a version directly from the cached downloads.

ljharb: Sounds great!

ljharb:
LGTM

locale/en/download/package-manager.md
shadowspawn: (Thanks @Trott for suggestion.)

Trott (approved):

PR #3340, opened 2020-08-16
fix: ensure primary content areas wrapped in main landmark element

jendowns:
Description

Hey everyone! ✨ 
 
I noticed that the pages on the site are not wrapped in a [`<main>` sectioning element](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/main), so the goal of this PR is to ensure that each page includes this semantic markup. Including a `<main>` landmark on a page improves accessibility by [helping assistive tech quickly identify the primary content of the page](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/main#Accessibility_concerns).

Additional resources:
- [W3C Recommendation - The `main` element](https://www.w3.org/TR/html51/grouping-content.html#the-main-element)
- [WCAG 2.1 checkpoint: 4.1.2 - Name, Role, Value](https://www.w3.org/WAI/WCAG21/Understanding/name-role-value.html)

Please let me know if you have any questions or concerns about this change! Or if I missed a page. 😅


nschonni (approved):

XhmikosR:
This should be good to go; I don't see any CSS selectors to break after a quick look.

ghost (approved):

ghost:
Thanks for your contribution @jendowns!

PR #3341, opened 2020-08-16
fix: use nav elements with descriptive aria-labels

jendowns:
## Description

This PR sets out to make 2 related improvements to the navigation of the site. Please let me know what you think -- thank you!

### 1. Ensure that all navigation components use a `<nav>` element
I noticed that docs page has a secondary navigation wrapped in an `<aside>`.

In this case I believe it would be more helpful to use `<nav>` to more directly signal the purpose of the markup, as it includes navigation links (reference: [`<nav>`: The Navigation Section element via MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/nav)). By comparison, an `<aside>` element is associated with complementary content "indirectly related to the document's main content" (reference: [`<aside>`: The Aside element via MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/aside)).

### 2. Ensure that all `<nav>` elements have descriptive `aria-label` attributes
There are instances where multiple `<nav>`s appear on a page, as is the case with the newly-revised Docs page and the current News page.

Whenever there are multiple `<nav>`, it helps to either point to an existing label (via `aria-labelledby`) or create a label (via `aria-label`). For this PR, I opted to use `aria-label` because I did not see any existing headings that would be ideal to point to with `aria-labelledby`. Related resource: [WCAG 2.1 -- Example using `aria-label` to distinguish between navigation landmarks](https://www.w3.org/WAI/WCAG21/Techniques/aria/ARIA6#example-1-distinguishing-navigation-landmarks)

### DIFF
--- a/layouts/blog-index.hbs
+++ b/layouts/blog-index.hbs
@@ -25,7 +25,7 @@
       </ul>
 
       {{#if pagination}}
-      <nav class="pagination">
+      <nav aria-label="pagination" class="pagination">
         {{#if pagination.prev.path}}
         <a href="/{{ site.locale }}/{{ pagination.prev.path }}/">&lt; Newer</a>
         {{/if}}
--- a/layouts/css/styles.scss
+++ b/layouts/css/styles.scss
@@ -50,7 +50,7 @@ article a {
 }
 
 .has-side-nav {
-  aside {
+  nav {
     width: 200px;
     float: left;
     margin-top: 1.5em;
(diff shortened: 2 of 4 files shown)


DerekNonGeneric (approved):
This looks great &mdash; thanks so much for doing this!

/cc @Trott

Trott (approved):

PR #3347, opened 2020-08-24
refactor(release-post.js): use array destructuring instead of indexes

umairx97:
**What change does this PR introduces?**
This PR refactors `scripts/release-post.js` to use array destructuring instead of using indexes

**Does this PR contain a breaking change?** 
No


willin (approved):

willin (approved):

umairx97:
thanks for approving this again @willin

XhmikosR:
@Trott while this looks good, I don't think it's enforced by standard. So, it's your call :)

Trott (approved):

PR #3364, opened 2020-09-09
chore: add link to OpenJSF Slack channel

mmarchini:


nschonni (approved):

ahtee:
Should the line about the **node-slackers** slack group be moved into this sub-list since it is real-time chat related? Referring to the last point in this list:
```markdown
* [Node Slackers](https://www.nodeslackers.com/) is a Node.js-focused Slack community.
```

mmarchini:
I'm happy either way.

locale/en/get-involved/index.md
```diff
@@ -8,7 +8,9 @@ layout: contribute.hbs
 ## Community Discussion
 
 * The [GitHub issues list](https://github.com/nodejs/node/issues) is the place for discussion of Node.js core features.
-* For real-time chat about Node.js development go to `irc.freenode.net` in the `#node.js` channel with an [IRC client](https://en.wikipedia.org/wiki/Comparison_of_Internet_Relay_Chat_clients) or connect in your web browser to the channel using [freenode's WebChat](https://webchat.freenode.net/#node.js).
+* For real-time chat about Node.js development use the OpenJS Foundation Slack or IRC
+  * For Slack, [join the OpenJSF workspace](https://slack-invite.openjsf.org/) and then join the [`#nodejs`](https://openjs-foundation.slack.com/archives/CK9Q4MB53) channel
```
MylesBorins, line 12: Super nitty... but I feel like IRC should be first on the list since it is the official channel for development. thoughts?

mmarchini: I'm trying to avoid the "official" discussion, but the OpenJSF Slack channel is just as official in this case (since it's a Foundation channel), isn't it?

MylesBorins: That's totally fair, an apologies for muddying the water. Let me rephrase. #node-dev is the primary channel for node.js core development. The IRC predates the slack. IRC is alphabetically sorted before Slack.

Those are all reasons I would say to put IRC first.

Now with that said, slack is way easier to use and more approachable for most folks, so that could be a reason to list it first. Also TBH, I misread and thought this was referring to node slackers not the openjs slack, so apologies for that.

I don't have strong feelings here, just had an initial reaction to share, which honestly is maybe even less relevant with the reminder that we are talking about the OpenJS slack

mmarchini: > Now with that said, slack is way easier to use and more approachable for most folks, so that could be a reason to list it first

That's the main reason I listed it first. Not a hill I'm willing to die on though, so if folks want IRC first I can change that.

> Also TBH, I misread and thought this was referring to node slackers not the openjs slack, so apologies for that

Node Slackers is already listed below :) I kept it there because it's not an "official" community per say, so didn't want to mix with the other real-time chat alternatives. But as @ahtee suggested, we can move it to the real time chat list if folks prefer that way.

ahtee: I didn't realize nose-slackers was unofficial I just wondered if it was somehow misplaced 🙃.  Could also be helpful to note that node-slackers in unofficial versus going to the official channels (core development 👉 IRC, other development or questions 👉 Slack, community channels 👉 node-slackers) if that's how it's set up. If it were up to me I like alphabetical lists if there are no icons on the page to bring your eye to the resources you are looking for without reading 🙃.

mmarchini: We didn't add any notes wrt officialness as the discussion can get pretty long and heated :)

Issue #3372, opened 2020-09-14
Change language-picker icon so it's not the Google Translate icon


missmatsuko:
I saw that a language picker was added in #2294, that's great!

However, I noticed that the language picker icon uses the Google Translate icon:
<img width="52" alt="Google Translate icon" src="https://user-images.githubusercontent.com/17884198/93033583-b02bc680-f5eb-11ea-82ad-456f07f6f9ee.png">

I think this makes it look like the site will be automatically translated with Google Translate, rather than actual human translations.

PR #3382, opened 2020-09-16
Update language-picker icon

RamonRomeroQro:
#3372 
Update header.hbs: Change Google Translate Icon to a generic Translate icon.
Material.io icon under Apache license version 2.0.

Screen Shot 2020-09-16 at 3 50 57


XhmikosR:
I'm not sure this icon is clear as to what it does. We could use a flag icon or a globe icon maybe?

PR #3404, opened 2020-09-27
Update test.yml

XhmikosR:
* specify `CI:true`
* move Node.js version to an environment variable
* remove info which is now printed by actions/setup-node


Trott (approved):

.github/workflows/test.yml
```diff
@@ -1,11 +1,16 @@
...
+
 on:
   push:
     branches-ignore:
       - "translations"
       - "dependabot/**"
   pull_request:
 
+env:
+  CI: true
```
nschonni: Already set by default https://docs.github.com/en/free-pro-team@latest/actions/reference/environment-variables#default-environment-variables

nschonni (commented):
👍 I like the removal of those `--version` steps.
Don't really mind the node version variable, but that data is really already in there too in the setup-node step as you mention in the discription
```
/opt/hostedtoolcache/node/12.18.4/x64/bin/node --version
v12.18.4
/opt/hostedtoolcache/node/12.18.4/x64/bin/npm --version
6.14.6
```
Is it work using https://github.com/actions/setup-java if we care about the particular Java version?

.github/workflows/test.yml
XhmikosR: Oh, good, I guess they finally added it, because it wasn't the case in the beginning.

XhmikosR:
The default Java version should be fine :)

PR #3408, opened 2020-09-27
Correct example code for "How to serve static files"

sethbattin:
The suggested code did not work as written.  I was looking for a terse script to display some local html content, so i googled "node server static".  Although this article is only the 5th result, it should be suitable for a copy-paste, turnkey solution.  That is the spirit of the article.

The article could use some other modernization as well, but I'll leave that out to avoid muddying the main point.

I didn't verify any of the other claims of the article, i.e. mime correctness, caching, or filesystem traversal protection.


Trott:
The knowledge base is definitely out of date in many places. And it is a shame that it recommends specific packages in a lot of places. But yeah, if this makes it correct (and judging from the node-static README, it does), then 👍  and thanks for the udpate.

Trott (approved):

PR #3431, opened 2020-10-03
Added NodeJS Installation 

Xurde-glitch:


nschonni:
This appears to just be info copy/pasted from https://github.com/nodejs/nodejs.org/blob/master/locale/en/download/package-manager.md

PR #3457, opened 2020-10-16
Fix spelling of "release" in Ukrainian translation

1valdis:
First column name on Releases page was spelt incorrectly in Ukrainian.


rodion-arr (approved):
Good catch 👍

nschonni:
Google translate (😄 ) says the old one said "resize" and the new one says "release"

PR #3475, opened 2020-10-27
CI: switch to Node.js 14.x.

XhmikosR:
Switch to Node.js 14.x.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -8,7 +8,7 @@ on:
   pull_request:
 
 env:
-  NODE: 12.x
+  NODE: 14.x
 
 jobs:
   test:


Trott (approved):

Trott:
Do we want to consider having a matrix that includes all supported versions of Node.js so that we can make sure the website tests pass no matter what (supported) version of Node.js the developer is running? Something like the `strategy` and `matrix` stuff at  https://docs.github.com/en/free-pro-team@latest/actions/guides/building-and-testing-nodejs#starting-with-the-nodejs-workflow-template?

XhmikosR:
I don't mind adding that but I'd like to do it in another PR where I will split tests so that it doesn't slow us down.

PR #3476, opened 2020-10-27
Switch to `@metalsmith/permalinks`

XhmikosR:
metalsmith-permalinks seems to be deprecated.

@Trott although the repo (https://github.com/segmentio/metalsmith-permalinks) is not archived nor can I find any references, the npm package is deprecated https://www.npmjs.com/package/metalsmith-permalinks


XhmikosR:
OK, I guess they have deprecated the npm package but they haven't released a new npm version with the new package name...

PR #3479, opened 2020-10-27
Update about-release-schedule.hbs

XhmikosR:
Fix the image dimensions to be what we end up with instead of the original image ones.


Trott (approved):
Rubber-stamp LGTM

PR #3482, opened 2020-10-27
feat: update release schedule

bmuenzenmeyer:
The release table at https://nodejs.org/en/about/releases/ is missing the Fermium codename cell

image

This PR resolves that by running `npm run load-schedule`

image


richardlau:
https://nodejs.org/en/about/releases/ includes "Fermium" for me, perhaps you were getting an old cached version from CloudFlare? 
image

bmuenzenmeyer:
I see that now too 🤷 that must have been it. Curious why the `master` source of this data is so outdated - perhaps it's unused. I don't know enough of the codebase yet but I may revisit someday.

Cheers to the Node 14 release!

richardlau:
> 
> 
> I see that now too 🤷 that must have been it. Curious why the `master` source of this data is so outdated - perhaps it's unused. I don't know enough of the codebase yet but I may revisit someday.
> 

I'm not overly familiar with how the website gets built, but I think the file in `master` gets overwritten by https://github.com/nodejs/nodejs.org/blob/20b7838d5485772aa4b399667e1e28a9dcf2737b/package.json#L13 
which is called by the deploy script https://github.com/nodejs/nodejs.org/blob/20b7838d5485772aa4b399667e1e28a9dcf2737b/package.json#L11 so the file probably is unused and just a placeholder?

bmuenzenmeyer:
yeah it sounds like an artifact that likely doesn't need to be tracked in git then

PR #3503, opened 2020-11-09
Updated the Language selector icon

### LINKED ISSUE #3372: Change language-picker icon so it's not the Google Translate icon

DantrazTrev:
Fixes #3372 

I've changed the language selector icon to a better one, that is easily identifiable for all users 
image

As mentioned in the issue
The current icons makes it look like the site will be automatically translated with Google Translate, rather than actual human translations.


XhmikosR:
@nschonni @Trott thoughts?

layouts/partials/header.hbs
```diff
@@ -24,9 +24,9 @@
...
+      <svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 512 512" width="20" height="20" aria-hidden="true" focusable="false" fill="currentColor">
```
XhmikosR, line 27: We should try and keep the previous dimensions

Trott:
> @nschonni @Trott thoughts?

I don't have an opinion on this one and am happy to defer to you, @nschonni, and anyone else in the website group who might be paying more attention to this particular interface element.

XhmikosR:
I'm not sure I like this new icon TBH. But the previous one is wrong too, because it's from Google translate.

nschonni:
Where did the new icon come from? I'm not sure if the current one was properly sourced either 😄

XhmikosR:
That I don't know either. Better put this on hold then?

ghost:
What about this article about the language icon? Use for free.

http://www.languageicon.org/

DerekNonGeneric:
> What about this article about the language icon? Use for free.
> 
> http://www.languageicon.org/

It looks like we are going to be going w/ that suggestion. Thanks for the link, I was able to convert that to an optimized SVG for whoever wants to do the honors. @DantrazTrev, you may be able to simply modify your PR to use the following SVG.

https://codepen.io/DerekNonGeneric/pen/bGroexG

Also, please take dark mode into account if you do end up going for it as this icon would need to be made into a lighter color to contrast the dark background.

DerekNonGeneric (changes requested):
> What about this article about the language icon? Use for free.
> 
> http://www.languageicon.org/

It looks like we are going to be going w/ that suggestion. Thanks for the link, I was able to convert that to an optimized SVG for whoever wants to do the honors. @DantrazTrev, you may be able to simply modify your PR to use the following SVG.

https://codepen.io/DerekNonGeneric/pen/bGroexG

Also, please take dark mode into account if you do end up going for it as this icon would need to be made into a lighter color to contrast the dark background.

PR #3517, opened 2020-11-12
Break out Snaps in the package manager page

richardlau:
Creation of Node.js snaps on the Snap store has been [moved into
the nodejs org](https://github.com/nodejs/admin/issues/567). Move it into its own section and reference as an
alternative method for CentOS/Fedora/RHEL/Debian/Ubuntu.

Refs: https://github.com/nodejs/admin/issues/567
Refs: https://github.com/nodejs/nodejs.dev/pull/1049#issuecomment-726161819

cc @mhdawson @rvagg

### DIFF
--- a/locale/en/download/package-manager.md
+++ b/locale/en/download/package-manager.md
@@ -12,7 +12,7 @@ title: Installing Node.js via package manager
 * [Android](#android)
 * [Arch Linux](#arch-linux)
 * [CentOS, Fedora and Red Hat Enterprise Linux](#centos-fedora-and-red-hat-enterprise-linux)
-* [Debian and Ubuntu based Linux distributions, Enterprise Linux/Fedora and Snap packages](#debian-and-ubuntu-based-linux-distributions-enterprise-linux-fedora-and-snap-packages)
+* [Debian and Ubuntu based Linux distributions](#debian-and-ubuntu-based-linux-distributions)
 * [FreeBSD](#freebsd)
 * [Gentoo](#gentoo)
 * [IBM i](#ibm-i)
@@ -23,6 +23,7 @@ title: Installing Node.js via package manager
 * [OpenBSD](#openbsd)
 * [openSUSE and SLE](#opensuse-and-sle)
 * [SmartOS and illumos](#smartos-and-illumos)
+* [Snap](#snap)
 * [Solus](#solus)
 * [Void Linux](#void-linux)
 * [Windows](#windows)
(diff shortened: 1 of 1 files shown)


Trott (approved):

locale/en/download/package-manager.md
```diff
@@ -72,10 +73,20 @@ dnf module install nodejs:12
 
 For CentOS/RHEL 7 Node.js is available via [Software Collections](https://www.softwarecollections.org/en/scls/?search=NodeJS).
 
-## Debian and Ubuntu based Linux distributions, Enterprise Linux/Fedora and Snap packages
+### Alternatives
+
+Packages compatible with CentOS, Fedora and RHEL are available via
+* [Node.js snaps](#snap).
+* [NodeSource Node.js binary distributions](#debian-and-ubuntu-based-linux-distributions).
```
Trott: We're not great about doing this, but as a rule, we should [avoid lists that complete a sentence fragment because these can be difficult to translate](https://docs.microsoft.com/en-us/style-guide/scannable-content/lists#introductory-text).

```suggestion
These resources provide packages compatible with CentOS, Fedora, and RHEL.
* [Node.js snaps](#snap)
* [NodeSource Node.js binary distributions](#debian-and-ubuntu-based-linux-distributions)
```

locale/en/download/package-manager.md
```diff
@@ -72,10 +73,20 @@ dnf module install nodejs:12
 
 For CentOS/RHEL 7 Node.js is available via [Software Collections](https://www.softwarecollections.org/en/scls/?search=NodeJS).
 
-## Debian and Ubuntu based Linux distributions, Enterprise Linux/Fedora and Snap packages
+### Alternatives
+
+These resources provide packages compatible with CentOS, Fedora, and RHEL.
```
mhdawson, line 78: I think this should be "These resources provide packages compatible with CentOS, Fedora, and RHEL. if you don't need support"

mhdawson: or something along those lines so it's clear that you won't get support if you use those versus the ones in software collections.

richardlau: Isn't that information covered by the paragraph at the top of the page?

> ***Note:*** The packages on this page are maintained and supported by their respective packagers, **not** the Node.js core team. Please report any issues you encounter to the package maintainer. If it turns out your issue is a bug in Node.js itself, the maintainer will report the issue upstream.

mhdawson: What I'm trying to get across is that Red Hat only supports the packages which are installed via software collections.  So while the alternatives are worth mentioning, \you won't get support from the "package maintainer. " as stated in that paragraph.

richardlau: I'll try to think of some appropriate words (although if anyone has suggestions please suggest them). For the alternatives you do get support from the package maintainer but the maintainer is different (for example the package maintainer of the NodeSource packages is NodeSource). 

Maybe keep the sentence and explicitly call out the package maintainer in the bullets?
e.g., 
> These resources provide packages compatible with CentOS, Fedora, and RHEL.
> * [Node.js snaps](#snap) maintained by https://github.com/nodejs/snap
> * [Node.js binary distributions](#debian-and-ubuntu-based-linux-distributions) maintained by [NodeSource](https://github.com/nodesource/distributions)

rvagg (approved):

locale/en/download/package-manager.md
richardlau: @mhdawson I've updated the wording of the bullet points to explicitly state where the alternatives are maintained. PTAL.

mhdawson (approved):
LGTM

PR #3520, opened 2020-11-15
docs: make note formatting more consistent

GeorgeSapkin:
The note formatting seems to differ among various parts of the docs, so I've tried to normalize it a bit. I've only touched the English bits for now. I've picked the style that seems to be the most popular:

> ~~**Note:**~~ Details here.


Trott:
Looks fine to me, although I wonder if the thing to do is remove `Note:` entirely.

GeorgeSapkin:
What are you suggesting? Just leave the note text without the word `Note:`?

> ~~**Note:**~~ Details here.

I don't have a strong opinion. I'm updating the event loop guide and this irks me somewhat.

Trott:
> What are you suggesting? Just leave the note text without the word `Note:`?
> 
> > ~**Note:**~ Details here.
> 
> I don't have a strong opinion. I'm updating the event loop guide and this irks me somewhat.

Yes, that's what I'm suggesting, but like you, I don't have a particularly strong opinion. The change here is 👍 by me, as would getting rid of `Note:` entirely. Maybe let's see what @nodejs/website folks thing.

phillipj (approved):
LGTM either way, with or without *Note:*.

If I'd have to choose, I'd lean slightly towards getting rid of *Note:*, primarily because the text that follows it has now been updated to be markdown quotes, which I'm assuming will highlight it a bit compared to other non-quote parts of these guides.

GeorgeSapkin:
I removed `Note:` and rebased.

sonicdoe:
To support removing “Note:” completely, we’ve made the same change two years ago in https://github.com/nodejs/node/pull/18592.

PR #3551, opened 2020-11-30
Add banner and link to Node.js cert and training discounts

brianwarner:
Signed-off-by: Brian Warner <brian@bdwarner.com>

### DIFF
--- a/build.js
+++ b/build.js
@@ -278,8 +278,8 @@ function getSource (callback) {
         },
         banner: {
           visible: true,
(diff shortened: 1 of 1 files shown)


mhdawson (approved):
LGTM. 

I think the banner for the security release was up long enough so no concern about removing that.

Adding a banner for the sale was discussed with TSC and CommComm and no issues were raised.

joesepi (approved):
I agree with Michael.

LGTM

PR #3674, opened 2021-02-10
fix: Missing 14.15.5 Windows 64 install link

nschonni:
@BethGriggs in case this was held back on purpose


BethGriggs (approved):
Ah thanks for catching. It's a little odd that this happened as the promotion script confirmed all assets were present 🤔

nschonni:
Yeah, it doesn't look like it was even the last asset to upload either https://nodejs.org/dist/v14.15.5/

PR #3685, opened 2021-02-12
package.json: require Node.js >= 12

ghost:
'Fibers' ONLY supports Node.js's version >= 12.X, otherwise there'll be
an error when running the local project. So the minimal version of
Nodejs should be 12.X.

See: https://www.npmjs.com/package/fibers#supported-platforms. This is the related screenshot:
image


XhmikosR:
Well, CI seems to work fine regardless of that. That being said, Node.js 10 is close to EOL.

Not sure how to proceed.

/CC @Trott

Trott:
> Well, CI seems to work fine regardless of that. That being said, Node.js 10 is close to EOL.
> 
> Not sure how to proceed.
> 
> /CC @Trott

I think it's OK to go with 12.0.0 and above for purposes of building this site, just to be cautious.

Trott (approved):

PR #3689, opened 2021-02-13
zh-CN: fixture of error translation

ghost:
"If a consensus cannot be reached that has no objections"  doesn't mean "如果实在无法达成共识但是又没有异议".  This is the fixture for it.


ghost:
@nodejs/nodejs-cn: Please have a review, thanks!

laosb:
I doubt whether we can simply leave *that has no objections* part out. To me reaching a consensus doesn't necessarily mean *no objections*. Is there any reference on why *no objections* is added in the original?

ghost:
>To me reaching a consensus doesn't necessarily mean _no objections_. 

You mean we have diff ideas but we can still have the same idea? Should this be translated as "如果无法达成无反对意见……" better to mean what you say.

laosb:
I tend to preserve the no objection part. Rephrase it a little and 如果無法達成無異議的共識 is better IMO.

MaledongGit:
> I tend to preserve the no objection part. Rephrase it a little and 如果無法達成無異議的共識 is better IMO.

求同存异……明白了。Thanks!

ghost:
求同存异……明白了。Thanks!
@laosb

PR #3695, opened 2021-02-14
v15.1.0.md: fix wrong syntax highlighting

XhmikosR:
So, this is a tricky one.

1. the build errors aren't real errors; the build continued and resulted in wrong highlighting, see for example https://github.com/nodejs/nodejs.org/runs/1896479847#step:6:31. AFAICT this is by design in the metalsmith-prism package: https://github.com/Availity/metalsmith-prism/blob/8db23a433d600b3fd13e33779cbd264e053d7caf/lib/index.js#L41-L47
2. maybe this should be implemented in the changelog maker or something?

Before:

image

After:

image

Any of these should do the job: `shell-session`, `sh-session`, `shellsession`

### DIFF
--- a/locale/en/blog/release/v15.1.0.md
+++ b/locale/en/blog/release/v15.1.0.md
@@ -104,7 +104,7 @@ When generating snapshots, garbage collection may be triggered and bring the hea
 
 Generating V8 snapshots takes time and memory (both memory managed by the V8 heap and native memory outside the V8 heap). The bigger the heap is, the more resources it needs. Node.js will adjust the V8 heap to accommondate the additional V8 heap memory overhead, and try its best to avoid using up all the memory avialable to the process.
 
-```console
+```shell-session
 $ node --max-old-space-size=100 --heapsnapshot-near-heap-limit=3 index.js
 Wrote snapshot to Heap.20200430.100036.49580.0.001.heapsnapshot
 Wrote snapshot to Heap.20200430.100037.49580.0.002.heapsnapshot


targos:
LGTM, thanks!

Is the rule to always use `shell-session` instead of `console` ? The issue I see is that GitHub doesn't highlight it at all: https://github.com/nodejs/nodejs.org/blob/22ced65f915b3899683f0309154d58470ad8cc0b/locale/en/blog/release/v15.1.0.md#take-a-v8-heap-snapshot-just-before-running-out-of-memory-experimental
It's okay for `nodejs.org`, but the same markdown source is used for the changelog at https://github.com/nodejs/node/blob/master/doc/changelogs/CHANGELOG_V15.md#take-a-v8-heap-snapshot-just-before-running-out-of-memory-experimental and https://github.com/nodejs/node/releases/tag/v15.1.0

XhmikosR:
> Is the rule to always use `shell-session` instead of `console` ?

Any of `shell-session`, `sh-session`, `shellsession` will work with prismjs, but yeah, this isn't what GitHub's syntax highlighting works with.

Hence why I think there should be some kind of enforcement for this specific case which we have hit more than once (#2923)

XhmikosR:
The other options would be

1. ask upstream in prismjs to add an alias for this
2. see if we can do this manually
3. switch to another solution altogether

targos:
Maybe we can do something in the [release post script](https://github.com/nodejs/nodejs.org/blob/master/scripts/release-post.js) to transform occurences of ```` ```console```` to ```` ```shell-session```` ?

XhmikosR:
Yup, that could work, but I'm not sure I have the time to propose a patch :)

That being said, a fix there along with a test case should be a pretty good fix.

Trott (approved):

PR #3540, opened 2020-11-20
chore: Convert Standard to ESLint plugin and Prettier plugin

nschonni:

PR #3697, opened 2021-02-14
fixture of warnings in standard check of js

ghost:
This is the fixture for main.js to remove all the warnings of "using let or const instead of var".
Here's what we see before the fixture:

捕获

Here's what we see after this fixture:

image


XhmikosR:
These are just warnings and IE is supposed to be supported last time I checked.

-1 from me unless the supported browsers have changed, in which case this should be applied in .browserslistrc.

ghost:
OK, I'll let this file be ignored bacause of this..., wait a  monent and I'll submit another PR here. Thanks!

ghost:
@XhmikosR : Consider there's only one file disabled, I just ignore this file.

XhmikosR:
I wouldn't ignore the whole file, rather the specific rules.

But wait for others to chime in because things might have changed.

MaledongGit:
Great! I didn't think of that. Now ONLY "var" check rule is disabled.

ghost:
@XhmikosR : Thank you for your suggestion! I didn't think of that. Now ONLY "var" check rule is disabled.

static/js/main.js
```diff
@@ -1,3 +1,6 @@
+/* eslint-disable no-var */
+// Notice: IE is still supported, so disable eslint for the file
```
Trott: Probably worth being specific about what version we support:

```suggestion
// Notice: IE 11 is still supported, so disable eslint for the file
```

nschonni:
This is covered in https://github.com/nodejs/nodejs.org/pull/3540, and I disagree supporting below IE11 is a goal. Any OS that can't run IE11 can't run any version of NodeJS released in many years

static/js/main.js
nschonni: `const` is supported in IE11, as mentioned in https://github.com/nodejs/nodejs.org/pull/3540, it's only <IE11 that even has an issue with this

ghost:
> This is covered in #3540, and I disagree supporting below IE11 is a goal. Any OS that can't run IE11 can't run any version of NodeJS released in many years

Let's Wait more comments here to decide which version will be kept for diff ideas on it.

Whether IE 11 below will be supported or not?  Can we need to support all the browsers to see the website of Nodejs (I mean for view only)?

Ping @nodejs/tsc  for final decision

Trott:
Microsoft [stopped supporting IE 10 over a year ago](https://docs.microsoft.com/en-us/lifecycle/announcements/internet-explorer-10-end-of-support). I don't think we should worry about making special accommodations in our JavaScript for browser versions that are no longer supported by the browser vendor. Anyone browsing the web with IE 10 is no doubt used to many websites not working.

That said, I'd be supportive of making sure that the site is still functional without JavaScript. Optional features like dark mode don't have to work and the site does not have to be aesthetically pleasing with JavaScript not running. But a user coming to nodejs.org should be able to read documentation and download Node.js, even if they are using an old and unsupported browser that can't run our JavaScript code.

MaledongGit:
> But a user coming to nodejs.org should be able to read documentation and download Node.js, even if they are using an old and unsupported browser that can't run our JavaScript code.

@Trott：Maybe My PR can be approved first and when Nodejs doesn't support browsers below IE 11 totally，@nschoinni's work can be merged  and as a backup solution now?

ghost:
According to Trott's suggestion, it seems we can make the file ignore syntax checks (related to `var`) so as to still allow others to view, download...ect, this is a fixture for it to avoid warnings and when Node.js group has claimed that it DOESN'T support browsers < IE11 an more, maybe nschonni's suggestion will be taken. So I'll merge mine first.

PR #3698, opened 2021-02-14
fixture of the error when rendering with Prism

ghost:
"Metal-prism" has referred Prism, but it doesn't support coding language of "console", so "batch" is used here as a fixture instead.

See more at: https://prismjs.com/#supported-languages

Before Fixture:
image

After Fixture:
image


XhmikosR:
This is invalid.; The proper fix is in #3695

ghost:
OK, it seems you've submitted a fixture :)

PR #3699, opened 2021-02-14
fixture of rendering error of "console" in Prism

ghost:
"Metal-Prism" needs a range of coding languages to be rendered with, however the script will generate some languages that don't match what it really needs. This is a fixture for it.

For more, please see: https://github.com/nodejs/nodejs.org/pull/3695.


XhmikosR:
You could probably add a new CHANGELOG.md file with a few lines instead of 1.6K lines :)

EDIT: or even just add a new section in the existent changelog file? Not sure what's the procedure.

MaledongGit:
> You could probably add a new CHANGELOG.md file with a few lines instead of 1.6K lines :)
> 
> EDIT: or even just add a new section in the existent changelog file? Not sure what's the procedure.

IMO：I prefer using the raw version, just like what others did in the previous test by referring the real generated md file to make a real test;)

Wait for others' comments.

ghost:
> You could probably add a new CHANGELOG.md file with a few lines instead of 1.6K lines :)
> 
> EDIT: or even just add a new section in the existent changelog file? Not sure what's the procedure.

Good idea, I just keep the content to be mocked for test.

aymen94 (approved):

PR #3704, opened 2021-02-15
add 'RO' into CODEOWNERS

ghost:
Add a missing member group "RO" into CODEWINERS.


ghost:
PS: @nodejs/nodejs-ro：You don't need to add all the files into your folder but just keep files that are translated into your own language ;)

nschonni (approved):
Doesn't hurt, but these CODEOWNERS assignments don't currently work unless the groups have write access to the repo (which we don't have setup for the i18n groups)

ghost:
@nschonni : Thanks for telling me about that, and I'll join in the group if possible.

PR #3707, opened 2021-02-16
changed 'web server' to 'web application server'

Snickdx:
The term application server is more appropriate here. Web servers are programs which serve static files such as Apache, IIS, Nginx. While node can do this, it is my understanding is that node is not a replacement for these technologies.


rodion-arr (changes requested):
As for me - `web server` is perfectly fine here. Web server is not only about serving static files - it's about processing HTTP requests in general.

Trott:
Welcome @Snickdx and thanks for the pull request. I see what you're saying, but I'm not sure this makes the text easier to understand. Also, the demo code is, in fact, a web server. It does not need Apache or nginx to be served over HTTP. Node.js handles the HTTP processing itself. (We don't recommend writing code like that for production. But it's useful to show the batteries-included aspect of Node.js.)

Snickdx:
Hi thanks for the quick feedback and I get your points. I'm curious as to how the term "Application Server" is barely used in modern-day jargon. Do you think it still has a place in today's terminology? Maybe it's more appropriate when describing frameworks like express? IE node is a pure runtime but when you are using express you are building an App server?

XhmikosR:
I believe this is fine as is too. @Trott should we close this?

Trott:
> I believe this is fine as is too. @Trott should we close this?

Sure. It can always be re-opened by OP or any collaborator who has a different opinion.

Thanks, @Snickdx! Hope this doesn't discourage future contributions!

PR #3774, opened 2021-03-31
fixture of Contributor Loading

ghost:
Now our contributor loading page will show a template something like "Thanks for username's contribution". It seems this won't be shown until all the variables from the remote GitHub loaded successfully. So first hide the contributor card and then show it when ready.

### DIFF
--- a/layouts/partials/contributor-card.hbs
+++ b/layouts/partials/contributor-card.hbs
@@ -1,4 +1,4 @@
-<div class="contributor-card">
+<div class="contributor-card" style="display: none;">
   <a href="#" rel="nofollow noopener noreferrer">
     <div class="spinner-border" role="status">
       <span class="sr-only">Loading...</span>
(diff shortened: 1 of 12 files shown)


layouts/partials/contributor-card.hbs
```diff
@@ -1,4 +1,4 @@
-<div class="contributor-card">
+<div class="contributor-card" style="display: none;">
```
Trott, line 1: Should this be `visibility: hidden` instead of `display: none`?

ghost: I understand what you mean: No matter whether the data is loaded, the space should be left.
But it looks so ugly, so are you sure you want this? Something like:


IMO: If it fails to load, the space shouldn't be shown and take place until all the data is loaded successfully....

XhmikosR:
This was intentional though, so that the page jump isn't that big.

layouts/partials/contributor-card.hbs
Trott: Ah, yeah, I can see the merit of `display:none` in that case.

PR #3784, opened 2021-04-05
creation of a new folder

yurii-valeriiovych:


nschonni:
Thanks @yurii-valeriiovych, looks like you're trying to translate a bunch of content, but also have included a bunch of English-only copies of files (which we've previously removed). Check out https://github.com/nodejs/nodejs.org/blob/master/TRANSLATION.md and look at adding translations on Crowdin, or submit PRs of individual translated pages here

PR #3821, opened 2021-04-26
Add WSL install bullet option to downloads page

mattwojo:


mattwojo:
This was update was previously proposed: https://github.com/nodejs/nodejs.org/pull/2692#pullrequestreview-304772901
...and denied because at that time, running a Windows Insider build was required for using WSL 2. WSL 2 is currently the ideal NodeJS install location and dev experience on a Windows machine, so it would be nice to have representation as an option on this downloads page.
I resubmitted as the install doc URL has changed.
@Trott 
*Also had some more in-depth discussion with @MylesBorins and will be following up regarding additional suggestions for WSL representation across the NodeJS sites.

layouts/partials/download-list.hbs
```diff
@@ -1,5 +1,6 @@
 <section>
   <ul>
+    <li><a href="/{{site.locale}}/{{site.download.install-on-wsl.link}}/">{{site.download.install-on-wsl.text}}</a></li>
```
MylesBorins:
```suggestion
    <li><a href="https://docs.microsoft.com/windows/dev-environment/javascript/nodejs-on-wsl">{{site.download.install-on-wsl.text}}</a></li>
```

You don't need to use the `site.locale` for the URL since it is an external link.

I also think this should likely be at the bottom of the list (I believe we have tended to add things in chronological order)

locale/en/site.json
```diff
@@ -82,7 +82,11 @@
       "verify-text": "How to verify"
     },
     "install-on-linux": {
-      "text": "Installing Node.js via binary archive"
+      "text": "Installing Node.js via binary archive on Linux"
```
MylesBorins:
```suggestion
      "text": "Installing Node.js via binary archive"
```

I don't think there is a need to make this copy change (is it perhaps left over from the earlier version?

locale/en/site.json
```diff
@@ -82,7 +82,11 @@
...
+    },
+    "install-on-wsl": {
+      "link": "https://docs.microsoft.com/windows/dev-environment/javascript/nodejs-on-wsl",
+      "text": "Install on Windows Subsystem for Linux (WSL)",
```
MylesBorins:
```suggestion
      "text": "Install on Windows Subsystem for Linux (WSL)"
```

No need to include the link, we can just embed it (unless we are planning to share a localized link

MylesBorins (commented):
Some small nits but generally in support of this

layouts/partials/download-list.hbs
nschonni: Why is this going on the top, rather than just as part of the rest of the package manager links?

nschonni (commented):
Not going to block it now that WSL2 is GA 🎆 
Just not sure why it's getting treated separately from the other package manger instructions we have

locale/en/site.json
mattwojo: Yes - looks like a leftover. :)

locale/en/site.json
mattwojo: No need to localize the link. The linked doc auto-sniffs the users loc and updates the language accordingly.

layouts/partials/download-list.hbs
mattwojo: Myles just filed an update to move it to the bottom. Wasn't aware of your protocol for adding new items. Now moved to the bottom of the list.

Not the purpose of this PR, but arguably it seems that a reordering for relevance would make this list more user-friendly to meet the most common use-case scenarios... 1) Installing Node.js via package manager, 2) Previous releases, 3) Install via WSL (?), etc.  
It can feel like a bit of a disconnect / point of confusion to look at that list and see "Signed SHASUMS.." as the first bullet. Anyone new to NodeJS or unfamiliar with SHASUMS may just stop reading as that seems very advanced / in-depth, whereas having "Installing via package manager" seems like it should be front and center.

mattwojo: In regard to separate treatment from other package managers, this isn't just a call to install via package manager, it's an entirely different install location. The primary purpose is to highlight that NodeJS is a better overall dev experience on WSL using a Windows machine for most scenarios so want to raise awareness of it as an option.

layouts/partials/download-list.hbs
```diff
@@ -1,5 +1,6 @@
 <section>
   <ul>
+    <li><a href="https://docs.microsoft.com/windows/dev-environment/javascript/nodejs-on-wsl">{{site.download.install-on-wsl.text}}</a></li>
```
MylesBorins: looks like this is still top of the list and needs to be moved to the bottom.

Regarding the order, all the other links are to resources provided directly from the Node.js project. Considering that this is an off site resource I think it only make sense to include it as the last link

to @nschonni's question regarding why this isn't in the other package manager instructions I do see this as distinct as it is about setting up on a platform, not a specific way to install Node.js, you still need to manage it on WSL, in which case nvm may be the right choice. It would likely be good to do a follow up to improve some other documentation related to this, but I think that have a clear pointer to high quliaty maintained instructions on using Node.js with WSL is a net positive

MylesBorins (approved):
LGTM

MylesBorins:
@nodejs/website can we get another approval on here please. As I work for a subsidiary of MS I would like another approval before landing this.

XhmikosR:
While the linked page gives some good info, I don't think the WSL2 installation is a good one. There are still outstanding issues with WSL2 and 3rd-party software.

Honestly, I don't see why WSL is treated as a different system? I mean, it's the same steps as one would do if they used a linux distro, isn't it?

mattwojo:
> There are still outstanding issues with WSL2 and 3rd-party software.

@XhmikosR  - Could you identify what the outstanding issues are that you feel make installing NodeJS with WSL2 a poor idea? We at Microsoft have been promoting WSL2 as a preferred path for NodeJS devs as it often seems to have closer alignment with many of their deployment environment and preferred Bash tooling. I'm not aware of any outstanding issues that would be considered a blocker on this recommendation. 

> Honestly, I don't see why WSL is treated as a different system? I mean, it's the same steps as one would do if they used a linux distro, isn't it?

WSL is a different system. It's a Linux kernel running in alignment with Windows and requires a special installation. Once installed, then I agree that it runs primarily the same as any other Linux system, but for anyone with a machine running Windows 10, this requires special steps and is a uniquely new experience that I've heard from many NodeJS devs has improved their workflow and solved issues that had blocked them in the past.

XhmikosR:
> @XhmikosR - Could you identify what the outstanding issues are that you feel make installing NodeJS with WSL2 a poor idea? We at Microsoft have been promoting WSL2 as a preferred path for NodeJS devs as it often seems to have closer alignment with many of their deployment environment and preferred Bash tooling. I'm not aware of any outstanding issues that would be considered a blocker on this recommendation.

The various incompatibilities with other programs and WSL2 is a real deal breaker to me. Last time I tried WSL2, VirtualBox couldn't work with it. I saw they (VB) recently released a new version which might fix the issues, but personally I'm too scared to try. That being said, VirtualBox is only an example, and I'm pretty sure other issues exist or existed.

VB issue: https://github.com/MicrosoftDocs/WSL/issues/798

Oh, and the slowness which I believe still exists due to the way WSL2 is implemented, https://github.com/microsoft/WSL/issues/4197

My point is that one does not need WSL2 to just set up Ubuntu or their favorite distro and develop Node.js apps. Please allow me to choose to not get into the marketing stuff, as I don't like it. 🙂

PS. I'm a Windows user myself, mostly out of habit at this point, though. The WSL introduction was a pretty nice thing, and I'm sure that after WSL2 matures, WSL2 will be fun to work with too.

mattwojo:
Thanks for this feedback @XhmikosR, I would appreciate if you could directly address any "deal breakers".  

VirtualBox is compatible with WSL 2 as of VB version 6.1.20.  The "slowness" issue that you link is only in regard to working on Linux files stored on the Windows OS and vice-versa, which is not recommended unless there is a specific reason for doing so. Speed and performance are excellent when working with WSL 2 for any of the common NodeJS development scenarios based on the data that I've seen.

It's true that no one _needs_ to use WSL, but many developers like the integration. Listing WSL as an option (the preferred option for most devs) doesn't take away from the option to develop directly on a Linux OS.

Unless you have a direct reason supporting that WSL 2 is "immature", I'm wondering why you would block it from being listed as an option for other developers to try and make and determination for themselves?

aymen94 (approved):

MylesBorins:
@XhmikosR are you blocking this from landing or simply -0?

XhmikosR:
Nah, you can land it although personally I really don't see the need for WSL **2**, but I guess if the issues are solved, everyone will eventually move to WSL **2**. I'm still too scared to try WSL **2**, but I will when I have some time.

ghost:
@nodejs/translators: Plz translate into your own language for it.

Issue #3802, opened 2021-04-16
"404 Page Not Found" Cannot work properly when we don't have a language


ghost:
Now for the Nodejs page, if we have a language that doesn't exist, a 404 page will be shown.
E.g: https://nodejs.org/cn

However, if you see the page, everything doesn't seem work properly (Language Icon cannot be clicked to choose a language, "Contributor" is still running....), ect.

![image](https://user-images.githubusercontent.com/52018749/115014627-29bec680-9ee5-11eb-8db5-757d615d29f6.png)

IMO: I tend to say that this is a bug, and we should redirect the current user to the default (English) page, something like "
https://nodejs.org/en/404", and this will make everything work properly. It'd be better if we can keep the wrong URL but still make them work properly (or something we can use like https://nodejs.org/en/404?from=https://nodejs.org/cn)?

Not sure whether we have a Ngnix group to fix it by redirecting the user to the correct 404 page with related language...For I later noticed that we have multi-language 404 page...?

PS: If we have the language but with a wrong part back, something like (https://nodejs.org/zh-cn/wrongLink), this is fine：
![image](https://user-images.githubusercontent.com/52018749/115016607-dd28ba80-9ee7-11eb-9142-67ccb9dc1ab1.png)

What do you think of this?

MaledongGit:
Not sure whether we have a Ngnix group to fix it by redirecting the user to the correct 404 page with related language...For I later noticed that we have multi-language 404 page...?

PS: If we have the language but with a wrong part back, something like (https://nodejs.org/zh-cn/wrongLink), this is fine.

nschonni:
Might be a duplicate of https://github.com/nodejs/nodejs.org/issues/2984, but there is more details here

ghost:
Thank you @nschonni to notify me on another issue related to this，sorry for the late response because of busy work.

Ping @nodejs/build ，maybe two senarios should be included after reading the issues：

1. https://nodejs.org/wronglang
2. https://nodejs.org/rightlang/wronglocation

MrJithil:
@MaledongGit For a part of the statements, "language icon will not work properly", I added a fix for that. #3823 

Could you please elaborate specifically what are the other areas which not working while accessing a invalid locale. 

I think, we can work together to rectify all these issues.

ghost:
> @MaledongGit For a part of the statements, "language icon will not work properly", I added a fix for that. #3823
> 
[...]

PR #3823, opened 2021-04-27
fix null check for invalid lang

MrJithil:
`null` check is missing before accessing the child props of `currentLangElement`.

Error Screenshot:
image

Impact:
There will be no impact on the current application behavior. Instead, it will fix an issue caused due to missing null checks.

### DIFF
--- a/static/js/main.js
+++ b/static/js/main.js
@@ -20,11 +20,13 @@
     }
   })
 
-  langPickerTogglerElement.setAttribute('title', currentLangElement.textContent)
+  if (currentLangElement) {
+    langPickerTogglerElement.setAttribute('title', currentLangElement.textContent)
 
-  // Remove the current selected language item, because we don't need to choose it
-  // any more unless we want to switch to a new language
-  langPickerElement.removeChild(currentLangElement.parentNode)
+    // Remove the current selected language item, because we don't need to choose it
+    // any more unless we want to switch to a new language
+    langPickerElement.removeChild(currentLangElement.parentNode)
+  }
 
   langPickerTogglerElement.addEventListener('click', function () {
     langPickerElement.classList.toggle('hidden')


MrJithil:
@MylesBorins Could you please help me to run this test pipeline?

ghost:
Thank you @MrJithil  for your 1st contribution!
However I don't think it a good solution, because it just seems like a "workaround" to hide the problem instead of solving it. If you see my https://github.com/nodejs/nodejs.org/issues/3802, you will easily find that the problem is there's something wrong with the route itself:

Take this as an example:
```
https://nodejs.org/wronglocation  => This will return a bad 404 page
https://nodejs.org/correctlocation/invalid locations  => This will return a correct 404 page
```

So the problem now is: How to make the senario 1 fit for senario 2 to make them work both properly? Maybe we also need @nodejs/build to modify the route or to make a better solution ;)

MrJithil:
> Thank you @MrJithil for your 1st contribution!
> However I don't think it a good solution, because it just seems like a "workaround" to hide the problem instead of solving it. If you see my #3802, you will easily find that the problem is there's something wrong with the route itself:
> 
> Take this as an example:
> 
> ```
> https://nodejs.org/wronglocation  => This will return a bad 404 page
> https://nodejs.org/correctlocation/invalid locations  => This will return a correct 404 page
> ```
> 
> So the problem now is: How to make the scenario 1 fit for scenario 2 to make them work both properly? Maybe we also need @nodejs/build to modify the route or to make a better solution ;)

Hi @MaledongGit , Here I am trying to address the language change icon not working while the wrong route. The missing null check breaks the rest of the execution, and it leads to an unresponsive language change icon.

The router issue is a second part, I am definitely will look into it. I tagged issue #3802 here, because it states the unresponsive language change icon.

Anyway, since this is a missing null check, in my opinion, this code needed.

ghost:
> The missing null check breaks the rest of the execution, and it leads to an unresponsive language change icon.

When does it happen? Does it only happen when route is bad? If for the bad route, maybe this shoud be fixed by route; however if it happens in a normal route, this is another bug/issue.

MrJithil:
> > The missing null check breaks the rest of the execution, and it leads to an unresponsive language change icon.
> 
> When does it happen? Does it only happen when route is bad? If for the bad route, maybe this shoud be fixed by route; however if it happens in a normal route, this is another bug/issue.

This is happening only for bad language selection. 

NJS Change Lang not working

In the above recording, we can see that while users in the wrong route, they can't use the change language button. All other links are working.

ghost:
> > > The missing null check breaks the rest of the execution, and it leads to an unresponsive language change icon.
> > 
> > 
> > When does it happen? Does it only happen when route is bad? If for the bad route, maybe this shoud be fixed by route; however if it happens in a normal route, this is another bug/issue.
> 
> This is happening only for bad language selection.
> 
> NJS Change Lang not working
> 
> In the above recording, we can see that while users in the wrong route, they can't use the change language button. All other links are working.

Ah yes, so let's go to the start point：Maybe this isn't the code error, what we should do is to make the wrong URL to be directed to a correct 404 page by Nginx instead of coding directly....

What do you think of this?

MrJithil:
> > > > The missing null check breaks the rest of the execution, and it leads to an unresponsive language change icon.
> > > 
> > > 
> > > When does it happen? Does it only happen when route is bad? If for the bad route, maybe this shoud be fixed by route; however if it happens in a normal route, this is another bug/issue.
> > 
> > 
> > This is happening only for bad language selection.
> > NJS Change Lang not working
> > In the above recording, we can see that while users in the wrong route, they can't use the change language button. All other links are working.
> 
> Ah yes, so let's go to the start point：Maybe this isn't the code error, what we should do is to make the wrong URL to be directed to a correct 404 page by Nginx instead of coding directly....
> 
> What do you think of this?

Yeah. We can proceed with that. Let me check it.

MrJithil:
@MaledongGit I traversed through the build as well. The issue is because of the missing null check. Because of the missing null check, the scripts are broken. The code after the error (I pasted it on the first screen) is not running and it causes the non-working of the functionalities.

If you still wonder, this is not the right fix, could you please run this branch in your local and send me the screen recording?

MrJithil:
@Trott , @targos @jasnell  Could you please review this once you are free?

jasnell (approved):

MrJithil:
@jasnell Sorry to tag you again and again. Seems I like you are a more frequent visitor during my timezone. Could you please merge this PR on my behalf?

jasnell:
Someone will definitely look at it. Just in general, there's no need to ping specific individuals to have them merge prs. It may take a few days but there are enough people that keep an eye on the prs and will get to merging them when they are ready

ghost:
@MrJithil: I'll take this as a temp solution to this problem, better than nothing to be solved. Thank you!

PR #3824, opened 2021-04-28
add husky pre-commit hook

MrJithil:
A developer must initiate the commands to run unit tests and syntax checks before submitting the changes.

If someone forgot to run the commands to do so may lead to failures in the pipeline. So, the same could be achieved using husky `pre-commit` implementation.

Each time a developer commits their changes, the pre-commit hooks will get triggered and stop that commit if any unit test fails or lint errors. If all the tests and checks are passed, the commit will be a success.

The developer need not worry about the hook implementations since hook integration is automated by adding an npm `prepare` script.

The details have been added to the `README.md` file.


MylesBorins (changes requested):
We ended up removing husky commit hooks in the past because it caused significant friction while developing. We have extensive CI testing which I think is sufficient

MrJithil:
Okay. Closing the MR.

PR #3836, opened 2021-04-30
downloads: add macOS ARM64 link for current version.

### LINKED ISSUE #3811: Missing link for Apple Silicon (arm64) on download page
Starting with v16.0.0, we have official Apple Silicon tarballs.
It should be available on https://nodejs.org/en/download/current/

XhmikosR:
image

Fixes #3811

---

I used `versionTypeCurrent` so this will only show up for Node.js 16.x.

### DIFF
--- a/layouts/partials/primary-download-matrix.hbs
+++ b/layouts/partials/primary-download-matrix.hbs
@@ -59,11 +59,12 @@
         <th>{{downloads.MacOSInstaller}} (.pkg)</th>
         <td colspan="2"><a href="https://nodejs.org/dist/{{version.node}}/node-{{version.node}}.pkg">64-bit</a></td>
       </tr>
-
       <tr>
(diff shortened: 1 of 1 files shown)


richardlau:
Might it look more compact to have a single "macOS Binary (.tar.gz)" row with two columns ("x64" and "ARM64" (or even "Apple Silicon" as that is what Apple call it))?

XhmikosR:
Yeah, I'm not sure about naming so let me know what you prefer.

As for using a row with two columns, it doesn't sound bad. I could switch
to that if everyone agrees.

MylesBorins (approved):
LGTM

targos:
I'd prefer two columns

XhmikosR:
I switched to a two-column row.

Let me know about the text. I went with ARM64 but I'm not very familiar with macOS myself.

richardlau (approved):
I fine with either "ARM64" or "Apple Silicon".

PR #3838, opened 2021-05-01
docs: change built in debugger url doc to the current version

MrJithil:
the Built-In Debugger url redirects to an old version of the document.

Updating it to a current LTS version.


Trott:
How about we use `https://nodejs.org/docs/latest/api/debugger.html`? Then we never ever have to update the URL in these docs again.

MrJithil:
> How about we use `https://nodejs.org/docs/latest/api/debugger.html`? Then we never ever have to update the URL in these docs again.

@Trott It's a good suggestion.

But, in my opinion, we have to point only to the LTS version. Else, we definitely go with the suggestion.

MrJithil:
@Trott Are we good to go with this? Or do I need to think different approach?

If yes, please check here #3857

PR #3841, opened 2021-05-03
dependabot: switch to a weekly interval

XhmikosR:
Daily seems a bit too much IMHO


MylesBorins (approved):
LGTM

MylesBorins:
At one point I think we had this turned on only for security dependencies because of the rate of noise. That could also be a solution

nschonni (approved):

XhmikosR:
We could even switch it to a monthly interval later, whatever most people prefer.

PR #3843, opened 2021-05-03
vnu-jar.js: switch to `execFile`

XhmikosR:


XhmikosR:
BTW I'm still not sure if this script is as optimal as it could be (I don't even remember why I had to use `spawn` for example). Any help to improve it is welcome.

nschonni:
Is there a benifit to using the script vs something like https://github.com/w3c/aria-practices/blob/ea4ade330500ac65cf1e4e3334de892c2a42cec7/package.json#L22

XhmikosR:
Yeah. I believe only Windows 32-bit support. Ideally the needed workaround should be upstream, but time...

PR #3845, opened 2021-05-03
main.js: drop support for `innerText`

XhmikosR:
`textContent` is supported on IE >= 9.

### DIFF
--- a/static/js/main.js
+++ b/static/js/main.js
@@ -186,7 +186,6 @@
     n.cpuClass === 'x64'
     ? 'x64'
     : 'x86'
-  var text = 'textContent' in d ? 'textContent' : 'innerText'
   var buttons = d.querySelectorAll('.home-downloadbutton')
   var downloadHead = d.getElementById('home-downloadhead')
   var dlLocal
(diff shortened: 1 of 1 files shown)


static/js/main.js
```diff
@@ -214,15 +213,15 @@
     switch (os) {
       case 'Mac':
         versionIntoHref(buttons, 'node-%version%.pkg')
-        downloadHead[text] = dlLocal + ' macOS (x64)'
+        downloadHead.textContent = dlLocal + ' macOS (x64)'
```
XhmikosR, line 216: @targos should this button show ARM64 too for Node.js >= 16? If so, we'll need to adapt it later.

Trott (approved):

static/js/main.js
targos: Where is this?

XhmikosR: This is the text above the buttons show in the homepage.

targos: I see. That case is a bit ambiguous, because the text applies to both buttons (LTS and Current) but ARM64 is only available for Current.

PR #3848, opened 2021-05-03
de: add missing strings

XhmikosR:
There are probably more, I just noticed these, so I went with the English strings. I believe it's better than having missing text...


nschonni (approved):
Wasn't sure if these would just "fallback", but probably better to include for the translators to flag that they're missing

XhmikosR:
The only fallback we currently have is if a whole file is missing. So, in all other cases, strings will not be in sync.

Anyway, if anyone spots more like these, we should fix them.

PR #3849, opened 2021-05-03
Update to the latest Google Fonts code

XhmikosR:


layouts/partials/html-head.hbs
```diff
@@ -2,11 +2,10 @@
   <meta charset="utf-8">
   <title>{{#if title}}{{ title }} | {{/if}}{{ site.title }}</title>
 
-  <link rel="dns-prefetch" href="https://fonts.googleapis.com">
-  <link rel="dns-prefetch" href="https://fonts.gstatic.com">
   <link rel="dns-prefetch" href="https://api.github.com">
+  <link rel="preconnect" href="https://fonts.gstatic.com">
```
XhmikosR, line 6: I'm unsure about switching to preconnect since we are using `fallback`. The default `display` value is `swap`.

MrJithil (approved):

PR #3854, opened 2021-05-03
README.md: minor tweaks

XhmikosR:


README.md
```diff
@@ -74,8 +74,8 @@ Before submitting, you must pass all the unit tests and syntax checks by running
 
 ## Notice
 
-* Multiple locales can be built by using comma separated values in the `DEFAULT_LOCALE` variable. i.e: `DEFAULT_LOCALE=en,es,it`.
-* For other 'Serve/Build Options' and 'Test Options', please see `package.json` in details.
+* Multiple locales can be built by using comma-separated values in the `DEFAULT_LOCALE` variable. i.e: `DEFAULT_LOCALE=en,es,it`.
```
Trott: Optional nit: `e.g.` is what is really meant here, rather than `i.e.`. But since it's obviously an example, it can just be dropped entirely:

```suggestion
* Multiple locales can be built by using comma-separated values in the `DEFAULT_LOCALE` variable: `DEFAULT_LOCALE=en,es,it`.
```

README.md
```diff
@@ -74,8 +74,8 @@ Before submitting, you must pass all the unit tests and syntax checks by running
...
+* For other "Serve/Build Options" and "Test Options", please see `package.json` in details.
```
Trott: It's not clear to me why these are capitalized and in quotations. I'm also not sure what "in details" means at the end of this sentence. Maybe this?

```suggestion
* For other options, see `package.json`.
```

Trott (approved):
Looks good to me with or without my suggestions.

XhmikosR:
Yeah, the README could use more love later for sure. :) I mean, I don't even know why we have the examples with `{{FOO}}`, but I thought I'd leave these for someone else.

PR #3857, opened 2021-05-04
feat: add metalsmith-one-replace plugin for dynamic contents

MrJithil:
We have added a new metalsmith plugin `metalsmith-one-replace`.

The major intention I added here to replace the `Built-in Debugger` URL was too old and needs to be pointed to the `LTS` version.
Now, we don't want to change the URLs each time we released a new `LTS`. The plugin will dynamically generate the URLs.

Evidence:
NodeJsLatestDebuggerUrl


Trott:
I think we're better off pointing at https://nodejs.org/docs/latest/api/debugger.html. I don't know why we'd have to point at LTS. Either way, the docs are going to be wrong when 16.x becomes LTS in October because 16.x does not have `node debug script.js` as described in that document. It needs to be updated to use `node inspect script.js` instead.

MrJithil:
> I think we're better off pointing at https://nodejs.org/docs/latest/api/debugger.html. I don't know why we'd have to point at LTS. Either way, the docs are going to be wrong when 16.x becomes LTS in October because 16.x does not have `node debug script.js` as described in that document. It needs to be updated to use `node inspect script.js` instead.

Since the LTS is recommended for most users, who are more new to the NodeJS sometimes, in my opinion, we should point to LTS rather than the current version.

Anyway, this plugin can dynamically replace the versions. So, in that case, whenever we change an LTS version, it will reflect after a build and deployment. 

For eg: if we want to switch to the current version, we can change the code as below:
```
varValues: {
    currentVersion: `latest-${latestVersionInfo.current.nodeMajor}`
}
```

So, we can merge this pull request, and open a discussion if we need to finalize the version later. In the future, it will be a single-line change. 
CC: @Trott

Trott:
> Since the LTS is recommended for most users, who are more new to the NodeJS sometimes, in my opinion, we should point to LTS rather than the current version.

Sure, I'm good with that.

MrJithil:
> > Since the LTS is recommended for most users, who are more new to the NodeJS sometimes, in my opinion, we should point to LTS rather than the current version.
> 
> Sure, I'm good with that.

Thanks @Trott

MrJithil:
> > Since the LTS is recommended for most users, who are more new to the NodeJS sometimes, in my opinion, we should point to LTS rather than the current version.
> 
> Sure, I'm good with that.

@Trott  Could you please review these changes?

Trott:
@XhmikosR Is this something you're comfortable reviewing?

Trott (approved):
LGTM

XhmikosR:
> 
> 
> @XhmikosR Is this something you're comfortable reviewing?

Not thoroughly, but perhaps we don't need the plugin at all. We already parse and modify the content in the other custom plugins and this seems to be redundant, let alone it might slow things down (again, I haven't tested the patch).

I'd do this with a custom plugin like we do for anchor links, github links, and so on. Alternatively, I'd look into what else from the custom code could be done with the plugin so that things are more streamlined.

MrJithil:
> > @XhmikosR Is this something you're comfortable reviewing?
> 
> Not thoroughly, but perhaps we don't need the plugin at all. We already parse and modify the content in the other custom plugins and this seems to be redundant, let alone it might slow things down (again, I haven't tested the patch).
> 
> I'd do this with a custom plugin like we do for anchor links, github links, and so on. Alternatively, I'd look into what else from the custom code could be done with the plugin so that things are more streamlined.

It's a built time plugin. So it won't slow down things as it's a one-time activity. Could you please provide the sample where we use the  `anchor links` replacements? So that I can refer and use it here too.

MrJithil:
Can I close this pr? @Trott

Trott:
@nodejs/website This looks OK to me, but I'd really like to get a more informed set of eyes on it. I don't pay that close attention to the website build process. Anyone able to look at this and give it a reasonably confident 👍  or 👎 ?

aymen94 (approved):

Trott:
Thanks for your patience, @MrJithil!

MrJithil:
Thanks @Trott

PR #3866, opened 2021-05-11
Blog: v14.17.0 release post

danielleadams:
Refs: https://github.com/nodejs/node/pull/38507


XhmikosR:
It seems the homepage isn't updated, but the news page is. Not sure what's going wrong, but someone should have a look if the issue persists.

XhmikosR:
NVM, now it's solved after merging another PR. IMHO deployment should be moved to Actions ASAP to make things simpler and more robust.

PR #3894, opened 2021-05-24
docs : add missing Strings

leejiwoo2021:
While reading the article, I found that the function syntax was missing.


leejiwoo2021:
I have confirmed that there is a PR for the same content
https://github.com/nodejs/nodejs.org/pull/3718

PR #3950, opened 2021-07-03
fix:  Pop-up Menu cannot close when clicking outside it

Medialo:
Fix the bug when click outside the menu, it doesn't close.
Ref: https://github.com/nodejs/nodejs.org/issues/3907

PR #3952, opened 2021-07-04
doc: Add syntax commands in README.md

ghost:
Add related syntax commands in README.md to notify the users that we can use them to avoid syntax errors.
Related to https://github.com/nodejs/nodejs.org/pull/3950


ghost:
@XhmikosR & @Trott ：Considering some new comers don't know how to use them and make mistakes (like: https://github.com/nodejs/nodejs.org/pull/3950). So that's the reason why I added this PR, based on a new comer to make things more clear and easy.

If any of you is free, plz have a review ;)

XhmikosR:
There's no need for new npm scripts... You can just pass `npm run test:lint:stylelint -- --fix` etc

ghost:
> There's no need for new npm scripts... You can just pass `npm run test:lint:stylelint -- --fix` etc

Fixed it directly in README.md as the guide for the new comers ;)

PR #3959, opened 2021-07-06
fix: Only Nodejs's logo visible part can be clicked

ghost:
On the main page of Nodejs, it seems the anchor link's width is too large, this results in a problem from the record that even if you click outside the logo, it also jumps. It would be better only Nodejs's logo (visible part) can be clicked.

Screen Record:
动画

Solution:
Just use another div to wrap and limit the width and height of img to make sure ONLY the visible part can be clicked and jump.


ghost:
@XhmikosR and @Trott : Any though on this?

Trott:
@nodejs/website Anyone able to review this? Seems OK to me, but I don't trust myself with CSS. (Would `rem` be better than `px`, for example? I don't know but I bet a whole bunch of you folks do.)

XhmikosR:
Nah, the proper fix is to just change `display: block` to `inline-block` for the logo here:

https://github.com/nodejs/nodejs.org/blob/b6b02407d5f3174513152bcedb11ab3ee4108a2e/layouts/css/page-modules/_header.scss#L149

ghost:
> Nah, the proper fix is to just change `display: block` to `inline-block` for the logo here:
> 
> https://github.com/nodejs/nodejs.org/blob/b6b02407d5f3174513152bcedb11ab3ee4108a2e/layouts/css/page-modules/_header.scss#L149

It seems you are very familiar with CSS, I'll have a try and see you, thanks!

PR #3963, opened 2021-07-09
zh-CN: Add '环境' to clearify the meaning of Nodejs

AkarinLiu:
For English, we can say "运行时" directly, however to be more clear, it would be better to say "运行时环境" in Chinese.


ghost (approved):
@LiuDarkShaoYe : Thanks for your contribution!

PR #3965, opened 2021-07-09
refactor: moving banner to ext json file

lancemccluskey:
[Relevant issue](https://github.com/nodejs/nodejs.dev/issues/1502)
[Relevant comment in issue](https://github.com/nodejs/nodejs.dev/issues/1502#issuecomment-872201961)

### Changes proposed:
* Pull `banner` out of `build.js` into external json file.

This is the first item based on the relevant comment above. I'll put the GH action in a separate PR once this is approved.


ghost:
@lancemccluskey: Thanks for your contribution, However, please use this to fix your syntax first:
image

ghost:
Ping @benhalverson and @Trott

benhalverson:
One question for the nodejs.org team is how long are these banners usually posted for? Instead of a boolean value we could have a defined date and time for when its posted and when it expires

{
  "title": "New security releases to be made available July 1st, 2021"
  "url": "https://nodejs.org/en/blog/vulnerability/july-2021-security-releases/"
  "url_type": "vulnerability"
  "start": "2021-06-29"
  "end": "2021-07-31"
}

I'm not sure how that will break the current implementation of nodejs.org but it's an idea.

Either way works for me 👍🏻

lancemccluskey:
@MaledongGit Would yall be interested in using Husky to lint when committing? I use it at work at its helpful to avoid situations like this

nschonni:
> @MaledongGit Would yall be interested in using Husky to lint when committing? I use it at work at its helpful to avoid situations like this

No, it was removed because it caused more friction than help

nschonni:
Maybe https://github.com/nodejs/nodejs.org/blob/master/locale/en/site.json would be a better place for this, as it would potentially also support i18n (which we currently don't do for the banners)

lancemccluskey:
> Maybe [https://github.com/nodejs/nodejs.org/blob/master/locale/en/site.json](https://github.com/nodejs/nodejs.org/blob/master/locale/en/site.json?rgh-link-date=2021-07-10T01%3A48%3A44Z) would be a better place for this, as it would potentially also support i18n (which we currently don't do for the banners)

Yeah thats a great idea ill move it. I just saw other json files in src which is why I went for it originally

nschonni:
👍 your current approach was also my first though too 😄

ghost:
@lancemccluskey: Any preparations for i18n if you put it into site.json? I remember that it was hard-coded before....

BTW: You can run `npm run test` to do all the checks and do submit ;)

build.js
```diff
@@ -37,6 +37,8 @@ const anchorMarkdownHeadings = require('./scripts/plugins/anchor-markdown-headin
 const loadVersions = require('./scripts/load-versions')
 const latestVersion = require('./scripts/helpers/latestversion')
 
+const banner = require('./locale/en/site.json').banner
```
nschonni: There is the `i18nJSON` function that returns the value or the default

lancemccluskey:
> @lancemccluskey: Any preparations for i18n if you put it into site.json? I remember that it was hard-coded before....
> 

uhh I have none. Im assuming prep for that would be done later?

build.js
lancemccluskey: oof, maybe i should read first lol

build.js
```diff
@@ -293,11 +293,7 @@ function getSource (callback) {
           text: '#BlackLivesMatter',
           link: '/en/black-lives-matter/'
         },
-        banner: {
-          visible: true,
-          text: 'New security releases now available for 16.x, 14.x, and 12.x release lines',
-          link: '/en/blog/vulnerability/july-2021-security-releases/'
-        }
+        banner: i18nJSON().banner
```
nschonni: You might be able to drop this entirely and do it directly in the template https://github.com/nodejs/nodejs.org/blob/b6b02407d5f3174513152bcedb11ab3ee4108a2e/layouts/index.hbs#L21

lancemccluskey: Good call, you can def use it there. Since the `blacklivesmatter` object has a boolean property also, should I just pull that into the `site.json` and do the same thing? @nschonni

nschonni (approved):
Think this is a good first pass as-is. The other parts that could either be done here or later:
- Do the same for the BLM banner config
- Add the concept of the "timed" banners, but might be better to do separately, since it would require a new cron/ci job

build.js
nschonni: Yup, I think doing the same would make sense if you want to do that here too

lancemccluskey:
> Think this is a good first pass as-is. The other parts that could either be done here or later:
> 
> * Do the same for the BLM banner config
> * Add the concept of the "timed" banners, but might be better to do separately, since it would require a new cron/ci job

ill add in the BLM banner here too since its relevant and is a quick change. I agree about doing the cron job later. I was thinking GH action?

nschonni:
> I was thinking GH action?

Yeah, and potentially some sort of "workflow_run" on the nodejs.dev side to pick up changes, or just an equivalent cron

nschonni:
Looks like the build is failing because of a double slash at the end of the link https://github.com/nodejs/nodejs.org/pull/3965/checks?check_run_id=3036710327#step:9:20
not sure if removing the trailing slash in the link would make it happy or if it's something else

lancemccluskey:
> Looks like the build is failing because of a double slash at the end of the link [#3965 (checks)](https://github.com/nodejs/nodejs.org/pull/3965/checks?check_run_id=3036710327#step:9:20)
> not sure if removing the trailing slash in the link would make it happy or if it's something else

Yeah i have no idea, ill try that though

ghost:
Hi all:

Sorry to bother you, but don't you think it's something duplicated after your code is merged?
Maybe "NEW RELEASE...." info can be removed from the menu but just to "keep Black Lives matter"?

image

Trott:
> Maybe "NEW RELEASE...." info can be removed from the menu but just to "keep Black Lives matter"?

I may be misunderstanding you, but the BLM banner was removed weeks ago.

EDIT: Oh, I see now! The nav has extra stuff.

nschonni:
Yeah, this change looks like it tripped up with a greedy loop that's pulling in everything in the site.json to build in the header.hbs
Trying to see how to filter it

ghost:
> > Maybe "NEW RELEASE...." info can be removed from the menu but just to "keep Black Lives matter"?
> 
> I may be misunderstanding you, but the BLM banner was removed weeks ago.

Yes, that's my first question after this submit (why "Black Lives Matter" is still here but in site.json it's been "false"....? It confuses me a lot).

However, "NEW Security Release..." shouldn't be in the menu as it's something duplicated?

nschonni:
It's a "bug" in the looping on the nav.main helper stuff. https://github.com/nodejs/nodejs.org/pull/3968 is a monkey patch that no longer pulls it into the nav

nschonni:
Monkey patch worked for now, and I landed it since it was an "issue" on the live site. Have a PR to add the dates, but I'll leave that open for a proper review.

PR #3969, opened 2021-07-11
chore: Add date range for banner expiry

nschonni:
Didn't add any GitHub Action to support this, as the `.org` would need a cron on the build server since the deployment is done by webhook there right now

### DIFF
--- a/layouts/index.hbs
+++ b/layouts/index.hbs
@@ -18,7 +18,7 @@
           </div>
         {{/if}}
 
-        {{#if site.banners.index.visible}}
+        {{#if ( betweenDate site.banners.index.startDate site.banners.index.endDate ) }}
           <p class="home-version home-version-banner">
             <a href="{{ site.banners.index.link }}">{{ site.banners.index.text }}</a>
           </p>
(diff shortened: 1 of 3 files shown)


Trott (approved):

Trott:
> Didn't add any GitHub Action to support this, as the `.org` would need a cron on the build server since the deployment is done by webhook there right now

It will at least remove the banner when something else lands, rather than requiring a specific PR to remove it.

Might be a bit kludgy...er, ambitious, but I suppose we could add a scheduled GitHub Action that opens a PR when the banner is expired, and maybe even lands it too.

nschonni:
Think there is an issue around moving from the webhook to GitHub Actions for the deploy, but not sure how much work is involved in changing that. If that happens, the cron GitHub Action should be doable

ghost (commented):
Feel great

ghost (approved):

lancemccluskey (approved):
Thanks for adding this in! I know my approvals dont count, but here ya go lol

ghost:
@nodejs/translators: Here comes a new improvement——We can now localize our own banner.

PR #3970, opened 2021-07-11
zh-CN: Localize the Security Info and add DateTime limit

ghost:
Since the banner has been moved to site.json with the datetime limitation, it's time to localize it.
Ref: https://github.com/nodejs/nodejs.org/pull/3969


MaledongGit:
Notice: This will be merged after https://github.com/nodejs/nodejs.org/pull/3969.

PR #3971, opened 2021-07-11
improve: add "EditOnGitHub" on every necessary page

ghost:
Here's the 1st version (Deprecated now, just keep for history):

mainIndex

downloadLTS

downloadCur~~

### DIFF
--- a/scripts/plugins/githubLinks.js
+++ b/scripts/plugins/githubLinks.js
@@ -1,24 +1,25 @@
 'use strict'
 
+const sep = require('path').sep
+// add suffix (".html" or sep for windows test) to each part of regex
+// to ignore possible occurrences in titles (e.g. blog posts)
+const isEditable = `(security|index).html|(about|download|docs|foundation|get-involved|knowledge)\\${sep}`
+const isEditableReg = new RegExp(isEditable)
+
 // This middleware adds "Edit on GitHub" links to every editable page
 function githubLinks (options) {
   return (files, m, next) => {
-    // add suffix (".html" or "/") to each part of regex
-    // to ignore possible occurrences in titles (e.g. blog posts)
-    const isEditable = /security\.html|about\/|docs\/|foundation\/|get-involved\/|knowledge\//
-
     Object.keys(files).forEach((path) => {
-      if (!isEditable.test(path)) {
+      if (!isEditableReg.test(path)) {
         return
       }
 
       const file = files[path]
-      const url = `https://github.com/nodejs/nodejs.org/edit/master/locale/${options.locale}/${path.replace('.html', '.md')}`
-      const editText = options.site.editOnGithub || 'Edit on GitHub'
+      path = path.replace('.html', '.md').replace(/\\/g, '/')
+      const url = `https://github.com/nodejs/nodejs.org/edit/master/locale/${options.locale}/${path}`
 
-      const contents = file.contents.toString().replace(/<h1(.*?)>(.*?)<\/h1>/, (match, $1, $2) => {
-        return `<a class="edit-link" href="${url}">${editText}</a> <h1${$1}>${$2}</h1>`
-      })
+      const contents = file.contents.toString() +
+      ` <input type = "hidden" id = "editOnGitHubUrl" value="${url}"/> `
 
       file.contents = Buffer.from(contents)
     })
--- a/.remarkrc
+++ b/.remarkrc
@@ -10,6 +10,7 @@
     ["remark-lint-no-literal-urls", false],
     ["remark-lint-no-trailing-spaces", false],
     ["remark-lint-no-undefined-references", false],
-    ["remark-lint-prohibited-strings", false]
+    ["remark-lint-prohibited-strings", false],
+    ["remark-preset-lint-node/remark-lint-nodejs-links", false]
   ]
 }
--- a/layouts/download-current.hbs
+++ b/layouts/download-current.hbs
@@ -11,7 +11,7 @@
         <div class="download-header">
           <h1>{{downloads.headline}}</h1>
         </div>
-
+        {{{contents}}}
         {{> primary-download-matrix version=project.latestVersions.current versionTypeCurrent="true"}}
         {{> secondary-download-matrix version=project.latestVersions.current versionTypeCurrent="true"}}
 
--- a/layouts/download.hbs
+++ b/layouts/download.hbs
@@ -11,7 +11,7 @@
         <div class="download-header">
           <h1>{{downloads.headline}}</h1>
         </div>
-
+        {{{contents}}}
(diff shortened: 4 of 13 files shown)


ghost:
@Trott and @XhmikosR : What do you think of this change? This change will let you show what you change in both windows and linux, as well as add more "Edit On GitHub" on possible pages.

XhmikosR:
The separator isn't even needed because those are permalinks so they use forward slashes.

As for the change itself, I can't really comment on it. I like simple things and this is something I personally don't like.

MaledongGit:
> The separator isn't even needed because those are permalinks so they use forward slashes.

This 'sep' here is just for Windows platform and Linux to be compatable, if I don't do this, we cannot see locally on windows.

> As for the change itself, I can't really comment on it. I like simple things and this is something I personally don't like.

Yes, but according to the codes of adding links of GitHub, in order not to make bad effects by breaking changes on the original md and other files, maybe mine is a better way.

But still a Big Thank you for your tips.

@nodejs/website : Any other ideas on it?

ghost:
Now after my deep thoughts in both UI and codes, I've changed by moving a GitHub icon onto the header, just like what others' website by GitHub. It looks more beautiful and practical:


MainEng

For diff language, it has the default tooltip:
MainChina

Here's the missing download part:
Download

Codes look simple but useful, easy to understand ;)

ghost:
Ping @nodejs/website : Any other ideas on it? I'll leave it for some days for your suggestions.

nickserv:
Maybe this could go in a footer or sidebar.

ghost:
> Maybe this could go in a footer or sidebar.

Thanks for your suggestion, many website pages on GitHub has a link on its head, and to be more clear, GitHub's icon is better:)

Take React as an example:
image

> or sidebar.

Can you point out where to put by circuling in red with a screenshot here ;)

scripts/plugins/githubLinks.js
```diff
@@ -1,24 +1,27 @@
 'use strict'
 
+const sep = require('path').sep
+// add suffix (".html" or sep for windows test) to each part of regex
+// to ignore possible occurrences in titles (e.g. blog posts)
+const isEditable = `(security|index).html|(about|download|docs|foundation|get-involved|knowledge)\\${sep}`
```
XhmikosR, line 6: I still don't see **why** this is needed. These are permalinks, so they will **always** use `/` regardless of the OS.

nickserv:
I guess there isn't a sidebar on most pages, but the header and footer could work. React is a good example, and we already have a Google Translate button in the header anyway.

scripts/plugins/githubLinks.js
ghost: Yes, for Linux and other systems (exclude windows), the original codes work perfectly. However as you see, if for windows (I'm now working on Win, and because it uses '\' instead of '/' as the path sep, so I cannot see the result right or not according to '/' in the gitHubLink.js (because nothing would be found and replaced).

So in order to be compatable with all the platforms, sep is a better choice IMO :)

Or maybe any good other ideas?

ghost:
> I guess there isn't a sidebar on most pages, but the header and footer could work. React is a good example, and we already have a Google Translate button in the header anyway.

Yes, that's also what I think now by just putting it beside the Mul-language icon :)

Trott:
Nothing against GitHub, but I'm not excited about prominently featuring a company's logo on all webpages like that. I just removed the GitHub mark from docs a few days ago. https://github.com/nodejs/node/pull/39251

scripts/plugins/githubLinks.js
Trott: I'm not sure I understand. Are you viewing it with a `file:` URL or something like that?

ghost:
> Nothing against GitHub, but I'm not excited about prominently featuring a company's logo on all webpages like that. I just removed the GitHub mark from docs a few days ago. [nodejs/node#39251](https://github.com/nodejs/node/pull/39251)

Sorry I didn't notice it....I'll change it to "Edit on GitHub" as a link and put it there.

scripts/plugins/githubLinks.js
ghost: @Trott and @XhmikosR : No. what I mean is: if we do tests on Windows, because the "files" in "scripts\plugins\githubLinks.js" will have the spe = "\", however the original filtering condition is something like "downloads/", this means for windows platform, we cannot get the result (we cannot see "Edit on GitHub" on the locally running test page).

ghost:
https://user-images.githubusercontent.com/52018749/125402445-ffe34600-e3e6-11eb-8a47-4924d8b1b563.mp4

Trott:
Sounds good to me. Anyone looking at this more closely want to review it?

scripts/plugins/githubLinks.js
XhmikosR: I explained it to you... these are links. They will always use `/`. This is not needed.

silverwind:
I don't think that button should be in the header. I'd put it into into the footer instead, besides `Report Node.js issue`, without an icon.

MaledongGit:
> I don't think that button should be in the header. I'd put it into into the footer instead, besides `Report Node.js issue`, without an icon.

@silverwind :
What about putting it beside "Get Help" without an icon? I mean once you cannot "get help" immediately, you can DIY by using Edit on GitHub (they belong to the same part) . Something like this:

image

ghost:
> I don't think that button should be in the header. I'd put it into into the footer instead, besides `Report Node.js issue`, without an icon.

Since @silverwind and @nickmccurdy have ideas to put it in the footer.... I've changed to this following:

1. For a normal English version:
English

2. For a non-English version:
Jap

3. For some page (such as blog), you will never see the result:
dontallowedit

4. For those who don't translate it into English, it will use English in default:
Ariab

scripts/plugins/githubLinks.js
ghost: @XhmikosR : I see you mean for the output files, yes, they will always use "/", but on windows platform in gitHhubLinks.js, you'll never get any files to be coped with, because windows is using "\", this means if you using "npm run serve", you will never see the result of "Edit on GitHub" on the page, and you don't know what you did is right or not. 

So considering those who are using windows, I have to use spe instead of hard code here. You can have a try on windows, if you hard coded, then you know why I changed to this :)

XhmikosR: No, these are links so they will always use `/` regardless of OS. That is why it is working all this time just fine.

MaledongGit:
@silverwind：Maybe a rubber stamp Approve？

XhmikosR:
@MaledongGit I still don't get why you use path.sep. We are 100% sure there will not be a `\` present, there's no reason to complicate things...

Secondly, why do we even need the new JS at all? We have already access to the page path in metalsmith,

This shouldn't have landed IMHO.

PR #3977, opened 2021-07-14
fix: fix code display in release blog

### LINKED ISSUE #3976: single quotes are being converted to `&apos` and displaying incorrectly in code snippet for new release blog post
Anybody know why the single quotes are being converted to `&apos` and displaying incorrectly on the website?

image

_Originally posted by @Trott in https://github.com/nodejs/nodejs.org/pull/3974#r669481512_

Trott:
Fixes: https://github.com/nodejs/nodejs.org/issues/3976

### DIFF
--- a/locale/en/blog/release/v16.5.0.md
+++ b/locale/en/blog/release/v16.5.0.md
@@ -18,7 +18,7 @@ Node.js now exposes an experimental implementation of the
 While it is experimental, the API is not exposed on the global object and is only
 accessible using the new `stream/web` core module:
 
-```mjs
+```js
 import { ReadableStream, WritableStream } from 'stream/web';
 // Or from 'node:stream/web'
 ```


targos:
The problem will come back next time we put an mjs code block in release notes. Node's linter forbids ESM syntax in js blocks

Trott:
> The problem will come back next time we put an mjs code block in release notes. Node's linter forbids ESM syntax in js blocks

Best option is probably to add `mjs` and `cjs` options for code highlighting in this repo. What do you think, @aduh95?

Refs: https://github.com/nodejs/node/pull/37162

We probably don't need anything as elaborate as what's in the docs (with a toggle switch and all that). We probably just need to map `mjs` and `cjs` to `js` in the highlighter somehow, or else add something in our build process that converts `mjs` and `cjs` code fences to `js` code fences.

richardlau:
If it's just the release posts we already remap `console` in https://github.com/nodejs/nodejs.org/blob/a654c6b40e892f064752a6bd4b99117eba745f02/scripts/release-post.js#L107-L113

aduh95:
I think we need to do the same thing I did in https://github.com/nodejs/node/pull/37311.

Trott:
> I think we need to do the same thing I did in [nodejs/node#37311](https://github.com/nodejs/node/pull/37311).

If I understand correctly, that would fix lint errors but not the display error that we saw here.

aduh95:
Hum indeed, I misread the issue. It looks like a bug in `Prism`:

```
Failed to load prism syntax: mjs
Error: Cannot find module 'prismjs/components/prism-mjs.js'
```

I'm able to workaround the issue by adding these lines at the top of `build.js`:

```js
const { languages } = require('prismjs');
languages.mjs = languages.js;
```

nschonni:
I wonder if maybe the "preLoad" option is also applicable https://www.npmjs.com/package/metalsmith-prism#preload-optional

PR #3987, opened 2021-07-19
add v14 into schedule.json

ghost:
Since v14 of Nodejs has been published, we've missed the info, this adds it into json.

### DIFF
--- a/source/schedule.json
+++ b/source/schedule.json
@@ -60,6 +60,13 @@
     "lts": "2019-10-22",
     "maintenance": "2021-04-01",
     "end": "2022-04-01",
-    "codename": ""
+    "codename": "Erbium"
+  },
+  "v14": {
+    "start": "2020-04-21",
+    "lts": "2020-10-27",
+    "maintenance": "2021-10-19",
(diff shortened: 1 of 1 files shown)


richardlau:
The entire contents of this file gets replaced during `deploy`. https://github.com/nodejs/nodejs.org/blob/27cafec87ce5d4635d0a3740c1ec2977262f12b5/package.json#L11-L13

nschonni:
Yeah, there was a PR to just remove this file, but I don't recall why it didn't land. Maybe just left as a rough fallback, but always wiped out as part of the normal build

Trott:
> Yeah, there was a PR to just remove this file, but I don't recall why it didn't land. Maybe just left as a rough fallback, but always wiped out as part of the normal build

If it's left as a fallback, I suppose it might be worth updating from time to time.

ghost:
Thanks @richardlau to tell me, and I had a try, yes the file is actually updated.
What about excluding this file from Git and remove it? Because we can run it manually, no need to update it again and again, just for deploy only....

PR #3988, opened 2021-07-19
Update simple-profiling.md

1110sillabo:
Fix missing negation before users[username] to check if there is no user with the inserted username. This aligns the code snippet with the one that below which implements the same check (line 62).

### DIFF
--- a/locale/en/docs/guides/simple-profiling.md
+++ b/locale/en/docs/guides/simple-profiling.md
@@ -37,7 +37,7 @@ app.get('/newUser', (req, res) => {
 
   username = username.replace(/[!@#$%^&*]/g, '');
 
-  if (!username || !password || users[username]) {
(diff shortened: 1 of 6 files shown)


nschonni (approved):

ghost (approved):
Some files are missing the "!", add up it to them.

ghost:
@1110sillabo & @nschonni: 

Agree with your ideas on this fixture, so I merged and changed all the other related files first.

However this snippet of codes make me feel ambigious: I mean this action is "addNewUser", so according to the original codes, if we use:

```js
if (!username || !password || users[username]) {...}
```

This also can be explained as "We cannot add a new user that has existed in the 'users'", so we must check whether the "users" have the user already or not by its "username"...

What do you think of it? Which is better to understand? I think it has two diff meanings.....

PR #3992, opened 2021-07-20
Update index.md

s-mrb:
doc: fix slight error

Line 35:

> Almost no function in Node.js directly performs I/O, so the process never blocks. Because nothing blocks, scalable systems are very reasonable to develop in Node.js.

`process never blocks` could be confusing for lots of learners, explicitly add **when** the `process never blocks` i.e. in case of async APIs


ghost (approved):
@s-mrb: Thanks for you 1st contribution! It's fine with me. Waiting for others' review.
Ping @nodejs/documentation.

aymen94 (approved):

PR #4022, opened 2021-08-03
feat: change from CommonJS to ModuleJs

ghost:
Considering there're more and more Nodejs modules using ModuleJs instead of CommonJS (such as 'junk'), and we've now made 'moduleJS' module in a stable function, so this will change the extension from 'js' to 'mjs' as a try. 'mjs' can be also checked with eslint as js files.

Ref:
1. https://nodejs.org/dist/latest-v14.x/docs/api/esm.html#esm_import_specifiers
2. https://github.com/nodejs/nodejs.org/pull/4020


ghost:
Maybe this is the first time to have a play with msj with js (mixed as an experinment), I meet some problems during code changes, but now everything seems better for me.

Hope to listen to your ideas @nodejs/website

XhmikosR:
Note that there are plenty of other files that are still using CJS. My suggestion would be to either do it for every file consistently and enforce it through linting, or just leave it as is.

ghost:
@XhmikosR : Thanks! After several trials, I found some npm packages' inner codes don't support es6 very well (such as metal-smith...,ect). So I closed it.

PR #4047, opened 2021-08-15
zh-CN: fix some translations

ghost:
1. Current => "最新尝鲜版", because "当前版" is a bit ambigious in
   understanding, but "最新尝鲜版" means this is the latest version with
   some experimental functions (Maybe not safe in production). Something like https://github.com/nodejs/node/blob/master/doc/changelogs/CHANGELOG_V16.md#notable-changes-3

2. Remove useless tranlations nodes.


willin (approved):
LGTM

PR #4050, opened 2021-08-15
improve: refer the Nodejs history links in one file

ghost:
Due to this sample (https://github.com/nodejs/nodejs.org/pull/4049) and many other related samples, it seems Node.js' history links are very easily forgotten by us and we have to maintain files one by one. This fixture will let all the files refer the same 'nodeVersionLinks.md', which means what we do now is just to modify this file instead of modifying translated files one by one in case of missing.

### DIFF
--- a/locale/ca/docs/index.md
+++ b/locale/ca/docs/index.md
@@ -25,20 +25,7 @@ També descriu els mòduls inclosos que proporciona Node.js, mes no documenta el
 
 ### Buscant la referència de l'API per a una versió anterior?
 
-* [Node.js 15.x](https://nodejs.org/docs/latest-v15.x/api/)
-* [Node.js 13.x](https://nodejs.org/docs/latest-v13.x/api/)
-* [Node.js 12.x](https://nodejs.org/docs/latest-v12.x/api/)
-* [Node.js 11.x](https://nodejs.org/docs/latest-v11.x/api/)
-* [Node.js 10.x](https://nodejs.org/docs/latest-v10.x/api/)
-* [Node.js 9.x](https://nodejs.org/docs/latest-v9.x/api/)
-* [Node.js 8.x](https://nodejs.org/docs/latest-v8.x/api/)
-* [Node.js 7.x](https://nodejs.org/docs/latest-v7.x/api/)
-* [Node.js 6.x](https://nodejs.org/docs/latest-v6.x/api/)
-* [Node.js 5.x](https://nodejs.org/docs/latest-v5.x/api/)
-* [Node.js 4.x](https://nodejs.org/docs/latest-v4.x/api/)
-* [Node.js 0.12.x](https://nodejs.org/docs/latest-v0.12.x/api/)
-* [Node.js 0.10.x](https://nodejs.org/docs/latest-v0.10.x/api/)
-* [Totes les versions](https://nodejs.org/docs/)
+{#var nodeVersionLinks}* [Totes les versions](https://nodejs.org/docs/)
 
 </div>
 
--- a/locale/en/docs/index.md
+++ b/locale/en/docs/index.md
@@ -23,20 +23,7 @@ This documentation describes the built-in modules provided by Node.js. It does n
 
 ### Looking for API docs of previous releases?
 
-* [Node.js 15.x](https://nodejs.org/docs/latest-v15.x/api/)
-* [Node.js 13.x](https://nodejs.org/docs/latest-v13.x/api/)
-* [Node.js 12.x](https://nodejs.org/docs/latest-v12.x/api/)
-* [Node.js 11.x](https://nodejs.org/docs/latest-v11.x/api/)
-* [Node.js 10.x](https://nodejs.org/docs/latest-v10.x/api/)
-* [Node.js 9.x](https://nodejs.org/docs/latest-v9.x/api/)
-* [Node.js 8.x](https://nodejs.org/docs/latest-v8.x/api/)
-* [Node.js 7.x](https://nodejs.org/docs/latest-v7.x/api/)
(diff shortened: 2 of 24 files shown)


locale/en/docs/index.md
```diff
@@ -23,20 +23,7 @@ This documentation describes the built-in modules provided by Node.js. It does n
...
-* [Node.js 9.x](https://nodejs.org/docs/latest-v9.x/api/)
-* [Node.js 8.x](https://nodejs.org/docs/latest-v8.x/api/)
-* [Node.js 7.x](https://nodejs.org/docs/latest-v7.x/api/)
-* [Node.js 6.x](https://nodejs.org/docs/latest-v6.x/api/)
-* [Node.js 5.x](https://nodejs.org/docs/latest-v5.x/api/)
-* [Node.js 4.x](https://nodejs.org/docs/latest-v4.x/api/)
-* [Node.js 0.12.x](https://nodejs.org/docs/latest-v0.12.x/api/)
-* [Node.js 0.10.x](https://nodejs.org/docs/latest-v0.10.x/api/)
-* [All versions](https://nodejs.org/docs/)
+{#var nodeVersionLinks}* [All versions](https://nodejs.org/docs/)
```
ghost, line 26: This is purposely written like this, because we need to make a whole ```<ul>....</ul>```. If you begin with a new line for [All Versions], it will generate ```<li><p>...<p></li>```, this isn't what it used to be.

PS: I suspect this is a Bug for 'metalsmith-one-replace'

nodejsVersions.md
```diff
@@ -0,0 +1,18 @@
+<!--
```
nschonni: The root doesn't seem like the right place for this. Maybe in the layouts or https://github.com/nodejs/nodejs.org/tree/master/layouts/partials

nschonni (dismissed):

nodejsVersions.md
MaledongGit: @nschonni: Thanks for the idea, maybe this is a mistake by you. What I aim is to just notify the user the usage of the md file instead of saying about layouts...What's more: I cannot find a suitable place to put it in the doc.hbs? Or do you think where to put? Or remove it as you think it's useless?

ghost: I'll remove it.

nschonni: Sorry, my comment wasn't about the comments at the top of the file, it was that this whole file doesn't belong in the root of the project and should be in the layouts

MaledongGit: What about putting into the 'source'? I notice that there're many data from 'source' folder, such as 'schedule.json', 'versions.json'?

ghost: Oh thanks! You reguard it as a template file, that's fine with me ;)

MaledongGit: Oh thanks! You reguard it as a template file, that's fine with me ;)

ghost: Oh thanks! You reguard it as a template file, that's fine with me ;)

locale/ca/docs/index.md
```diff
@@ -25,20 +25,7 @@ També descriu els mòduls inclosos que proporciona Node.js, mes no documenta el
...
-* [Node.js 9.x](https://nodejs.org/docs/latest-v9.x/api/)
-* [Node.js 8.x](https://nodejs.org/docs/latest-v8.x/api/)
-* [Node.js 7.x](https://nodejs.org/docs/latest-v7.x/api/)
-* [Node.js 6.x](https://nodejs.org/docs/latest-v6.x/api/)
-* [Node.js 5.x](https://nodejs.org/docs/latest-v5.x/api/)
-* [Node.js 4.x](https://nodejs.org/docs/latest-v4.x/api/)
-* [Node.js 0.12.x](https://nodejs.org/docs/latest-v0.12.x/api/)
-* [Node.js 0.10.x](https://nodejs.org/docs/latest-v0.10.x/api/)
-* [Totes les versions](https://nodejs.org/docs/)
+{#var nodeVersionLinks}* [Totes les versions](https://nodejs.org/docs/)
```
nschonni, line 28: If you rename the template to `.hbs`, I think you can use include style like `{{> node-versions }}`. and move this one translatable string of the last bullet into the regular `site.json`

ghost: I thought it before, but maybe not suitable. Files such as 'locale\en\docs\index.md' will be rendered as a whole {{content}} in the 'doc.hbs' , and this is only a part of md file. So to use variable-replacer is better IMO:

无标题

So if I make it as 'hbs' file, how can I refer it into the {{content}}? I cannot say something like '{>nodejsVersions}', it will make a big change on the original page after rendering.....

MaledongGit:
@nschonni: Maybe I cannot understand what you said: I once tried to rename the md to hbs, and it cannot be referred by md directly, something like this is wrong:

image

image

However, if I do something like this:

image

This doesn't seem properly as what it used to be....

Can you guide me? Thanks anyway!

locale/ca/docs/index.md
ghost: @nschonni: I've removed this md file as a data (just like versions.json) into "source" folder, and refer it into md to be rendered with.

nschonni: Sorry, I've been trying to figure it out locally too. Not sure if it's a limitation of the Metalsmith, or if I'm also not getting the syntax right

ghost: @nschonni: Nothing, I thought you'd something I didn't know about XD, so I ask you again and again...

nschonni: Maybe https://github.com/fidian/metalsmith-handlebars-contents would work

PR #4054, opened 2021-08-17
de: typo fix

ghost:
In "https://nodejs.org/de/download/current/", this isn't right for 'LTS' version but 'Aktuellste' only.

[Before fixing]
wrong

[After fixing]
image


ghost:
Ping @nodejs/nodejs-de

tniessen (approved):

PR #4055, opened 2021-08-17
Correct adding new users handle

minhtaile2712:
Update that if user has already existed, response error 400


ghost (approved):

MaledongGit:
@minhtaile2712: If possible, would you please replace all the codes in the localized files of the other languages? Thanks for your first contribution!

PR #4061, opened 2021-08-21
feat: make previous versions of Nodejs pagable and filterable

ghost:
Since we've now got a large table with latest versions of Nodejs, we don't need previous links but table instead.

What's more, in order to make people download the version quickly, the new feature is to offer you a quick but simple search for Nodejs's version, as well as npm version.

Here's the demo:

https://user-images.githubusercontent.com/52018749/130311405-49831bcf-ec86-4fcb-9396-801d2b4fb7c4.mp4

Notice: We've only supported IE >= 10, see https://github.com/nodejs/nodejs.org/pull/2602


static/js/jquery.min.js
```diff
@@ -0,0 +1,5 @@
+/*! jQuery v1.12.4 | (c) jQuery Foundation | jquery.org/license */
```
nschonni: This is a super old version

MaledongGit: Yes, but considering we should allow some old Explorers to download，I have to use it——Just like we have to use “var” in the "main.js".

ghost: @nschonni: Sorry, I remember we'd discussed about the version of browsers for ages...but now, it seems we've only suppoorted IE >=10... https://github.com/nodejs/nodejs.org/pull/2602. It's been fixed.

PR #4062, opened 2021-08-22
chore: update remark tools to latest versions

Trott:


.remarkrc
```diff
@@ -10,6 +10,6 @@
     ["remark-lint-no-literal-urls", false],
     ["remark-lint-no-undefined-references", false],
     ["remark-lint-prohibited-strings", false],
-    ["remark-preset-lint-node/remark-lint-nodejs-links", false]
+    ["remark-preset-lint-node/remark-lint-nodejs-links.js", false]
```
nschonni, line 13: Shouldn't this work without the extension change? Or is there something we missed upstream?

Trott: Yes, this change is unnecessary. I've removed it.

Trott: Er, nope, looks like it totally matters. Reverting.

Trott: So it works without the `.js` (and doesn't seem to work with the `.js`--is that because including the extension is just wrong and it loads nothing or is that a bug?). Additionally, when it's set to work, it flags absolute paths as broken links even though they work fine on GitHub and/or on the website. I guess technically they shouldn't be absolute links because we don't really know where these files are being hosted and if they're in a subdirectory or not. (GitHub helpfully maps `/` to the repo rather than to GitHub's root.)

Because there are a lot of unordered refs to fix too, I think I'll also make it without `.js` and make the paths relative paths. That's going to be hundreds of changed lines though.

Trott: I wonder if we should configure the linter to ignore absolute URLs? There's too much uncertainty around them and I'm finding lots of situations where I don't have much of a choice other than to disable the lint rule with a comment. (Examples include links to /ar/docs/foo when foo isn't there but the web server than automatically serves /en/docs/foo instead, and links to stuff in /static which one level above everything in /locale in the repo but at the same level on the website.)

Trott: I've pushed the change with the rule enabled so we can see all the things it flags.

Trott:
Moving this comment to the main area so it doesn't get lost:

I wonder if we should configure the linter to ignore absolute URLs? There's too much uncertainty around them and I'm finding lots of situations where I don't have much of a choice other than to disable the lint rule with a comment. (Examples include links to /ar/docs/foo when foo isn't there but the web server than automatically serves /en/docs/foo instead, and links to stuff in /static which one level above everything in /locale in the repo but at the same level on the website. In both of those cases, using a relative path will still be flagged as broken because the relative path that will work on the website is different than the relative path that works according to the linter.)

.remarkrc
Trott: Argh! No, it's supposed to turn the rule off. Never mind, I was right the first time: Leaving off the `.js` fails to turn the rule off. I'm going to put it back. This is ready to land and we can investigate the `.js` thing separately. (I don't think it's a bug. See https://nodejs.org/api/esm.html#esm_mandatory_file_extensions)

Trott:
Argh! No, it's supposed to turn the rule off. Never mind, I was right the first time: Leaving off the `.js` fails to turn the rule off. I'm going to put it back. This is ready to land and we can investigate the `.js` thing separately. (I don't think it's a bug. See https://nodejs.org/api/esm.html#esm_mandatory_file_extensions)

Trott:
> Argh! No, it's supposed to turn the rule off. Never mind, I was right the first time: Leaving off the `.js` fails to turn the rule off. I'm going to put it back. This is ready to land and we can investigate the `.js` thing separately. (I don't think it's a bug. See https://nodejs.org/api/esm.html#esm_mandatory_file_extensions)

If we want it to work without the file extension, we need to add `"exports"` to the `package.json` in `remark-preset-lint-node` but either way, it will work *with* the extension too.

.remarkrc
Trott: If we want it to work without the file extension, we need to add "exports" to the package.json in remark-preset-lint-node but either way, it will work with the extension too.

nschonni (approved):

PR #4086, opened 2021-09-01
[Arabic] aug 31 2021 security release post announcement

Aissaoui-Ahmed:
Ref: #4082

### DIFF
--- a/locale/ar/site.json
+++ b/locale/ar/site.json
@@ -163,10 +163,7 @@
   },
   "banners": {
     "index": {
-      "startDate": "2021-08-25T16:00:00.000Z",
-      "endDate": "2021-09-08T16:00:00.000Z",
-      "text": "إصدارات الأمانة الجديدة متوفرة الآن لإصدارات 14.x و 12.x.",
-      "link": "blog/vulnerability/aug-2021-security-releases2/"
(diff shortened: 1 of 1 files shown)


locale/ar/site.json
```diff
@@ -165,7 +165,7 @@
     "index": {
```
ghost, line 165: Nothing but just a trick.

Maybe 'startDate', 'endDate' and 'link' can be removed, because it will be automatically in sync with the English version, what you can do is just to keep your translations. Otherwises you have to update datatimes, as well as links.

For more, you can refer my change similar with yours here, without 'startDate', 'endDate'……ect：https://github.com/nodejs/nodejs.org/blob/56c4f1d7c0ad86232c4c2ccaaf014f75473fa476/locale/zh-cn/site.json#L163-L166

BTW: It's still OK with me if you keep them, but I just think it's a bit too bothering to change them from time to time. What I think is to just change the translation instead of datetimes, links...ect.

Hope it helps :)

Aissaoui-Ahmed: Thanks @MaledongGit

aymen94 (approved):

ghost (approved):
Fine with your modifications ;)

PR #4101, opened 2021-09-07
fix: correct "Coming soon" links in release posts

targos:
Replace "Coming soon" with link if available or delete the line when
there is no corresponding file on the server.


ghost (approved):
@targos : Fine with me, waiting for others' resp.

BTW: I'll submit another PR to remove "Coming Soon" / change when there's nothing to record....Is it OK?

https://github.com/nodejs/nodejs.org/blob/ceb3379cdd10e7068c90354b9fe363339416884a/scripts/release-post.js#L156

PR #4115, opened 2021-09-18
fix typo in simple-profiling.md

meixg:
Check if the username exists, "400" will be back.

### DIFF
--- a/locale/en/docs/guides/simple-profiling.md
+++ b/locale/en/docs/guides/simple-profiling.md
@@ -59,7 +59,7 @@ app.get('/auth', (req, res) => {
 
   username = username.replace(/[!@#$%^&*]/g, '');
 
-  if (!username || !password || users[username]) {
+  if (!username || !password || !users[username]) {
     return res.sendStatus(400);
   }
 


ghost (approved):

MaledongGit:
@meixg: Thanks for your 1st great contribution!

ghost:
@meixg: Sorry by mistake, I found the real logic is: If the user name Exists, a 400 error will be back. So this cannot be as a commit and I've reverted it back to what it used to be.

meixg:
> @meixg: Sorry by mistake, I found the real logic is: If the user name Exists, a 400 error will be back. So this cannot be as a commit and I've reverted it back to what it used to be.

there are two api, newUser and auth.

newUser should be 'If the user name Exists, a 400 error will be back.'

auth should be 'if the user name **not** exists, a 400 error will be back.'

ghost:
@meixg: Ah....Thanks!

And it seems you've missed the other files of translations...I've fixed them totally.

ghost:
@meixg : Thanks for your good catch, and here's my fix for it, you can approve it if free to have a review.

PR #4118, opened 2021-09-20
fix for simple-profiling.md of other translations

ghost:
Just for the other translations' missing typos.
Ref: https://github.com/nodejs/nodejs.org/pull/4115.


nschonni:
You reverted the linked PR

ghost:
@nschonni: Sorry, after a discussion with @meixg, It seems we have two interfaces but for the 2nd one, it seems the logic should be "When the user send a request to check whether the user exists or not, if not, an error will be returned....".

So I re-opened it to have you a review, maybe this is my fault...

PR #4127, opened 2021-09-22
Flatten package-lock.json

XhmikosR:
Not sure again why this keep happening... Anyway, aprox. -500 packages.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -19,20 +19,20 @@
       "dev": true
     },
     "@babel/core": {
-      "version": "7.15.0",
-      "resolved": "https://registry.npmjs.org/@babel/core/-/core-7.15.0.tgz",
-      "integrity": "sha512-tXtmTminrze5HEUPn/a0JtOzzfp0nk+UEXQ/tqIJo3WDGypl/2OFQEMll/zSFU8f/lfmfLXvTaORHF3cfXIQMw==",
+      "version": "7.15.5",
+      "resolved": "https://registry.npmjs.org/@babel/core/-/core-7.15.5.tgz",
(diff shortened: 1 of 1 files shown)


XhmikosR:
@Trott it seems this was introduced mostly in https://github.com/nodejs/nodejs.org/commit/6f5fa269d98afb43655bc2efec9478efea2ea6af#diff-053150b640a7ce75eff69d1a22cae7f0f94ad64ce9a855db544dda0929316519

Could this be an issue with npm 6.x? If so, I don't suppose we could get this improved?

Trott:
I tend to use npm@7 but will switch to npm@6 if I get the warning message about updating the package-lock.json file version so I can avoid churn, but obviously I am not avoiding churn here. 🤔

Trott (approved):

Trott:
> Not sure again why this keep happening... Anyway, aprox. -500 packages.

@nodejs/npm Would switching to npm@7 / package-lock version 2 avoid this kind of thing?

XhmikosR:
Generally, it's not a new issue with package-lock.json getting bloated. It usually happens with Dependabot. Unsure if switching to npm 7.x will fix this, but I know for a fact the issue happens since I've been hitting this for years on many projects.

For example: https://github.com/nodejs/nodejs.org/commit/5bbfb3b5f8337df887f82fff0b5919c0c5a89529#diff-053150b640a7ce75eff69d1a22cae7f0f94ad64ce9a855db544dda0929316519

Trott:
npm 7 seems to produce much larger package-lock.json files, or at least that's what I'm seeing.

Are there actual problems we've run into with a large package-lock file? It seems like there's a lot of repetition of information, but not anything wrong in the file. I try to treat package-lock.json as an implementation detail that I don't need to worry about. (I don't mind PRs to slim it down. I just want to make sure there aren't problems and that I shouldn't care more about the size than I do.)

XhmikosR:
Yeah, that's a different thing you are describing.

Here we are talking about suboptimal dependencies tree. Plenty of deps are
not flattened for some reason.

PR #4136, opened 2021-09-29
fix: address "coming soon" 404 cache hit problem

### LINKED ISSUE #2751: scripts/release-post.js always fills "Source Code" link with "Coming soon"
See with version 16.8.0:

image

It's been happening for a while now. I didn't open an issue before because I thought it was my fault somehow, but I just saw that it also happened to @danielleadams last week: https://github.com/nodejs/nodejs.org/commit/1cfee5b6ff123962216c3a5030ff0f5cf248151a#diff-f51dba60b01bf6ae650269a3e87fdbefd528776c0c11b35f8a2396408aeaf5d4R150

rvagg:
Fixes: https://github.com/nodejs/build/issues/2751

This is entirely untested (and I'm editing this in my browser too) but this might be a way of addressing the problem. I'm not super happy about exposing this address but it exists and shouldn't suffer from the hard-cache problem that seems to be causing https://github.com/nodejs/build/issues/2751; it'll give accurate results wrt what's actually available on the server, not just what the edge cache has.

### DIFF
--- a/scripts/release-post.js
+++ b/scripts/release-post.js
@@ -151,7 +151,8 @@ function findAuthorLogin (version, section) {
 }
 
 function urlOrComingSoon (binary) {
-  return sendRequest({ url: binary.url, method: 'HEAD' }).then(
+  const url = binary.url.replace('nodejs.org', 'direct.nodejs.org')
+  return sendRequest({ url: url, method: 'HEAD' }).then(
     () => `${binary.title}: ${binary.url}`,
     () => {
       console.log(`\x1B[32m "${binary.title}" is Coming soon...\x1B[39m`)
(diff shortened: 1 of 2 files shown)


targos:
I think the URL needs to be changed in the tests too:
https://github.com/nodejs/nodejs.org/blob/17e27ee17e80574898d07f316787b9d6a06506be/tests/scripts/release-post.test.js#L48
https://github.com/nodejs/nodejs.org/blob/17e27ee17e80574898d07f316787b9d6a06506be/tests/scripts/release-post.test.js#L63

rvagg:
would a releaser or web maintainer mind taking this over?

nschonni (approved):

MaledongGit:
Ping @nodejs/releasers , please have a review so as to let it merged.

PR #4156, opened 2021-10-08
Update anatomy-of-an-http-transaction.md

sumitkt:
at line no 85,   'at'  was present after Getting. So deleted it.


nschonni:
Sorry, this is already correct English

Trott:
> Sorry, this is already correct English

It's as intended, but it is also somewhat colloquial and I wonder if removing `at` wouldn't make it better. (I don't actually know.) Or maybe "Getting access to the body data" instead of "Getting at the body data"?

PR #4139, opened 2021-09-30
Add dark theme style

NriotHrreion:
issue #3507 

I've added the dark theme style.
It will look more comfortable for the dark theme user (like me).

PR #4159, opened 2021-10-09
improve Dark Mode

ghost:
Ref: #4139.

1. Change the header's background color to make it more light so as to
   be enhanced according to @Trott's idea.
2. Fix the blog's "code" part and its font color to make it more
   visible.

****
FixMe

Blog

### DIFF
--- a/layouts/css/_variables.scss
+++ b/layouts/css/_variables.scss
@@ -17,11 +17,12 @@ $home-secondary-links-color: #51744e;
 
 // for dark-theme
 $dark-black: #090c15;
-$dark-black2: #141a2b;
+$dark-black2: #233056;
 $dark-gray: #242424;
 $dark-green: #2c372a;
 $dark-green2: #84ba64;
 $dark-green3: #64de64;
+$dark-code-background: #2c3437;
 
 @mixin padded-link($padding) {
   padding-right: $padding;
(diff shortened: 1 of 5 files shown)


ghost:
@NriotHrreion, @Trott. Plz have a review, thanks!

NriotHrreion (changes requested):
The switcher's position is not suitable for the mobile client. 
It's better to move it to the center.

image

Trott:
I've never noticed it until this PR, but the highlighting in the docs vs. the rest of the website is very different.

About page:

image

Docs:

image

Trott:
Doesn't necessarily need to be fixed in this PR because it's already a problem on the main branch so could be fixed separately, but at certain viewport widths, the dark mode/light mode toggle sits on top of the News nav item:

image

Trott (approved):

ghost:
@Trott: Here's my fix for the background color of code...ect (The same color as what we in in the doc).

About
Blog

@NriotHrreion: The problem you've mentioned above doesn't only appear in the dark mode, but in the light mode (Not only for your icon, but the language picker as well). I'll create another Bug issue to re-produce it in detail. So please cancel your reject request here because it doesn't belong to this fix ;)

ghost:
Related Issue: https://github.com/nodejs/nodejs.org/issues/4166

NriotHrreion (approved):
That's OK, LGTM

NriotHrreion:
@MaledongGit: The switcher is still have a problem. I fixed that. See: #4174

PR #4173, opened 2021-10-11
Address issue in file system security knowledge base article (directory traversing)

zefir-git:
The [fs security intro article](https://nodejs.org/en/knowledge/file-system/security/introduction/) suggests preventing directory traversal by checking whether the requested directory starts with the pre-defined root directory. However, the code suggested only checks if the pre-defined root directory is part of the path (could be anywhere in the path).

Old code:
```js
var rootDirectory = '/var/www/';
var filename = path.join(rootDirectory, userSuppliedFilename);
if (filename.indexOf(rootDirectory) !== 0) {
  return respond('trying to sneak out of the web root?');
}
```
This allows traversing to any path as long as `/var/www` is present anywhere. This complicates the attack but does not prevent it.

This PR suggests using `if (!filename.startsWith(rootDirectory))` to ensure only files inside `/var/www` are accessed, disallowing paths like `/etc/traversed/directory/var/www/secret_file`. Secondarily, this PR proposes the use of ES6 syntax.

### DIFF
--- a/locale/en/knowledge/file-system/security/introduction.md
+++ b/locale/en/knowledge/file-system/security/introduction.md
@@ -39,14 +39,14 @@ Directory traversal means that an attacker tries to access files outside of the 
 This example assumes that you already checked the `userSuppliedFilename` variable as described in the "Poison Null Bytes" section above.
 
 ```javascript
-var rootDirectory = '/var/www/';
+const rootDirectory = '/var/www/';
 ```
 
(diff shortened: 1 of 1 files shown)


gumbol:
The code would check if root directory is at the start (index 0) first. an /var/www/ along the path after that wont affect it because its within the root path

zefir-git:
> The code would check if root directory is at the start (index 0) first. an /var/www/ along the path after that wont affect it because its within the root path

My bad; totally missed it's `!== 0`. For some reason I assumed `> -1` which is incorrect.

Issue #4167, opened 2021-10-11
Flash of Incorrect Theme when Dark Theme is chosen


U-C-S:
When reloading a page or Navigating across the site with Dark Theme on, There's a slight Flash of Light Theme. 

Probably, due to HTML and CSS being loaded first and later, JavaScript code that changes the theme. To Fix it, Might want to make sure the JS code is loaded before CSS.

U-C-S:
It can already be observed in this GIF from the Dark Mode PR: https://github.com/nodejs/nodejs.org/pull/4139#pullrequestreview-769533552

NriotHrreion:
I've fixed this problem and committed.

NriotHrreion:
> I've fixed this problem and committed.

@U-C-S: The PR has been merged.

PR #4174, opened 2021-10-12
Fix dark mode switcher flash bug

NriotHrreion:
issue: #4167

### DIFF
--- a/layouts/partials/html-head.hbs
+++ b/layouts/partials/html-head.hbs
@@ -34,4 +34,6 @@
   {{#each site.feeds}}
   <link rel="alternate" href="/{{ ../site.locale }}/{{ link }}" title="{{ text }}" type="application/rss+xml">
   {{/each}}
+
+  <script src="/static/js/themeSwitcher.js"></script>
 </head>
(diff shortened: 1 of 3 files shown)


layouts/partials/html-head.hbs
```diff
@@ -34,4 +34,6 @@
   {{#each site.feeds}}
   <link rel="alternate" href="/{{ ../site.locale }}/{{ link }}" title="{{ text }}" type="application/rss+xml">
   {{/each}}
+
+  <script src="/static/js/themeSwitcher.js"></script>
```
U-C-S, line 38: will it be a good idea to just embed the JavaScript code into HTML file, instead of a new js file??

NriotHrreion: I think it's better to be a new js file instead embedding into the html.

MaledongGit:
@U-C-S: Is it fine with you? Can I help to merge it?

XhmikosR:
This results in duplicate code in main.js and the new JS file. Why can't we have everything in a HEAD script?

PR #4181, opened 2021-10-15
Fix ES6 and Guides links to go Turkish pages.

enverarslan:


nschonni (approved):
Makes sense, that file has been translated https://github.com/nodejs/nodejs.org/blob/main/locale/tr/docs/es6.md and isn't just a placeholder

PR #4183, opened 2021-10-16
Added browser detection notification

NriotHrreion:
issue: #3960

I think it's very important for the users who are using old browser. So I added and committed it.

### DIFF
--- a/layouts/index.hbs
+++ b/layouts/index.hbs
@@ -1,4 +1,12 @@
 <!DOCTYPE html>
+
+<!--[if let IE 9]>
+  <h2>{{ site.browserOutdatedTitle }}</h2>
+  <p>{{ site.browserOutdatedText }}</p>
+<![endif]-->
+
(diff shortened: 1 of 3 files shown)


NriotHrreion:
@MaledongGit @Trott Can you have a review?

Trott:
I'll defer to others, but I'm not sure this is useful in 2021. My impression is that if someone is browsing the internet with IE9, then they are used to half the internet being completely broken for them and they know they are running an old browser.

NriotHrreion:
> I'll defer to others, but I'm not sure this is useful in 2021. My impression is that if someone is browsing the internet with IE9, then they are used to half the internet being completely broken for them and they know they are running an old browser.

I'm sure that there must be some people still using IE9. Not all are using Chromium. It's necessary to add this.

XhmikosR:
I don't think it's worth adding this, either. It's totally useless nowadays.

NriotHrreion:
well.

PR #4192, opened 2021-10-19
change security banner end date

Trott:
With 17.0.0 about to drop and 16.x going LTS, remove two-week old security banner. (We should set those banners to be there for two weeks and not a month. Otherwise there will always be a security banner and people will stop seeing it.)

### DIFF
--- a/locale/en/site.json
+++ b/locale/en/site.json
@@ -164,7 +164,7 @@
   "banners": {
     "index": {
       "startDate": "2021-10-05T16:00:00.000Z",
-      "endDate": "2021-11-05T16:00:00.000Z",
+      "endDate": "2021-10-19T16:00:00.000Z",
       "text": "New security releases now available October 12th, 2021",
       "link": "blog/vulnerability/oct-2021-security-releases/"
     },


richardlau:
The start/end dates were a reasonable idea but the way things currently work they don't really gain anything over the previous `visible` boolean as in order to actually change the state or contents of the banner the website needs to be rebuilt (i.e. just because we tick past the end date the site won't update to remove the banner until the next time it is rebuilt).

Trott:
> The start/end dates were a reasonable idea but the way things currently work they don't really gain anything over the previous `visible` boolean as in order to actually change the state or contents of the banner the website needs to be rebuilt (i.e. just because we tick past the end date the site won't update to remove the banner until the next time it is rebuilt).

I think it will have benefit if we can have the discipline to not put up a banner for a month at a time. A week in general is plenty, and two weeks in unusual cases. At least that's my opinion.

BethGriggs (approved):

Trott:
> (i.e. just because we tick past the end date the site won't update to remove the banner until the next time it is rebuilt).

That's true, but (and please correct me if I'm wrong!) it will rebuild when we do things like land dependabot changes and whatnot. We can also add a GitHub Action scheduled rebuild once a day or something if we don't see commits landing with sufficient frequency. (But I don't think we need that.)

richardlau:
> > (i.e. just because we tick past the end date the site won't update to remove the banner until the next time it is rebuilt).
> 
> That's true, but (and please correct me if I'm wrong!) it will rebuild when we do things like land dependabot changes and whatnot. We can also add a GitHub Action scheduled rebuild once a day or something if we don't see commits landing with sufficient frequency. (But I don't think we need that.)

It will rebuild every time the GitHub webhook fires when something is merged into `main` (GitHub Actions is not involved).

PR #4194, opened 2021-10-19
chore: remove remark-gfm

Trott:
remark-gfm is included in remark-preset-lint-node now so it no longer
needs to be included separately.

This also moves us to package-lock.json format version 2, hence the large diff there.

### DIFF
--- a/.remarkrc
+++ b/.remarkrc
@@ -1,6 +1,5 @@
 {
   "plugins": [
-    "remark-gfm",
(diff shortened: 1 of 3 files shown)


Trott:
This needs a review. @nschonni @XhmikosR @aduh95 

I'm happy to switch back to package-lock v1 if that makes a difference in reviewer enthusiasm.

aduh95 (approved):

PR #4195, opened 2021-10-19
chore: update CONTRIBUTING.md

Trott:
Update CONTRIBUTING.md with some cosmetic (remark-suggested) changes,
one or two comma-splice fixes, and some deletion of unnecessary detail.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -2,7 +2,7 @@
 
 ## Code of Conduct
 
-The Code of Conduct explains the *bare minimum* behavior
+The Code of Conduct explains the _bare minimum_ behavior
 expectations the Node Foundation requires of its contributors.
 [Please read it before participating](https://github.com/nodejs/admin/blob/HEAD/CODE_OF_CONDUCT.md).
 
@@ -13,8 +13,8 @@ expectations the Node Foundation requires of its contributors.
 
 ## Logging Issues
 
-Log an issue for any question or problem you might have. When in doubt, log an issue,
-any additional policies about what to include will be provided in the responses. The only
+Log an issue for any question or problem you might have. When in doubt, log an issue.
+Any additional policies about what to include will be provided in the responses. The only
 exception is security disclosures which should be sent privately.
 
 Collaborators may direct you to another repository, ask for additional clarifications, and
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -25,9 +25,7 @@ project's Code of Conduct.
 
 ## Contributions
 
-Any change to resources in this repository must be through pull requests. This applies to all changes
-to documentation, code, binary files, etc. Even long term collaborators must use
-pull requests.
+Any change to resources in this repository must be through pull requests.
```
nschonni, line 28: Is this enforced through branch policies right now?

nschonni (approved):

CONTRIBUTING.md
Trott: No, I don't believe it is. It looks to me like we'd need to check a box to make it enforced via tooling. I don't *think* this would impact any of the tooling in this repo, but others are in a better position to know.

image

Especially given the unusually high number of people with write access to this repo (another thing worth addressing, I think), protecting the main branch from direct push seems like a good idea.

nschonni: yeah, I think the "allow admins to bypass" is a separate option, but that's a much smaller group AFAIK. An admin could always toggle it on/off even if that is enforced, and that action just shows up in the org audit log

Trott: I've gone ahead and checked the box. We can always uncheck it if it breaks something.

mhdawson (approved):
LGTM

CONTRIBUTING.md
richardlau: > I've gone ahead and checked the box. We can always uncheck it if it breaks something.

We now appear to require a review on PRs which is unexpected given that we haven't required reviews for the release posts before https://github.com/nodejs/node/blob/master/doc/guides/releases.md#17-create-a-blog-post
> Always use pull-requests on the [nodejs.org](https://github.com/nodejs/nodejs.org) repository. Be respectful of the website team, but you do not have to wait for PR sign-off.

image

Trott: OK, I've unchecked the box again because not increasing the burden on releasers is way more important than enforcing this policy through tooling. We can re-visit to sort things out, but would need to involve someone to represent the release process.

PR #4200, opened 2021-10-20
Re-style release schedule table in both light & dark mode

### LINKED ISSUE #4180: Release Schedule table is very unpleasant and hard to read
* **URL**: https://nodejs.org/en/about/releases/
* **Browser version**: Firefox 93
* **Operating system**: Ubuntu 20.04

Hi, I visited the Node.js website and found out some redesigns have taken place and the dark mode is much better, though when I visited [Release Schedule](https://nodejs.org/en/about/releases/) page I saw that the table is very unpleasant and hard to read the content, looks like it uses some white color depending odd/even rules, which is a great thing to do to improve table readability, but the colors are not fit to be together, I think we can do better in terms of colors and make it better, here is a screenshot of the table:

image

> PS: I will be more than happy to help fix this issue, please let me know if you have any preference regarding the table design or appearance as there is a `redesign` taking place, otherwise I can work on it on my own.

akamfoad:
Fixes #4180


Trott (approved):
This fixes the color contrast accessibility issue.

Trott:
Before:

image


After:

image

NriotHrreion (approved):
It's better than before. I like it 👍.

ekalkutin (approved):

PR #4209, opened 2021-10-26
Update Docker example to use most recent LTS

rauno56:
Update node LTS version for the guide for docerizing node applications.

I happened to read the guide at the moment when the new LTS was swapped over to 16. My first intention was to actually suggest changing the output of `docker ps` which included `npm start` in the command column, which will not be the output if this guide is followed, but I don't think it's much of an issue. Or?

Related: https://github.com/nodejs/nodejs.org/pull/2983, https://github.com/nodejs/nodejs.org/pull/2953, https://github.com/nodejs/nodejs.org/issues/2988


DerekNonGeneric (approved):
Thanks!

Issue #4210, opened 2021-10-27
When viewing the release page in dark mode, the text is unreadable.


mochi-sann:
<img width="1195" alt="screenshot 2021-10-27 14 51 12" src="https://user-images.githubusercontent.com/44772513/139007742-eee04622-b26c-494e-b9c1-9609f7ebac99.png">

When I open https://nodejs.org/en/download/releases/ and look at it in dark mode, I can't read the version name and other information in white, as shown in the image.

* **URL**: https://nodejs.org/en/download/releases/
* **Browser version**:  Google Chrome | 95.0.4638.54 (Official Build) （x86_64）
* **Operating system**: Mac OS

DerekNonGeneric:
I can confirm that it is indeed an unintentional bug and an accessibility concern at that. It was introduced in https://github.com/nodejs/nodejs.org/pull/4200, which was merged only a few hours ago. Unless someone beats me to it, I think this should be pretty simple to handle and should have a patch for the fix shortly.

Thanks for the report!

PR #4211, opened 2021-10-27
Revert "Re-style release schedule table in both light & dark mode"

DerekNonGeneric:
Reverts nodejs/nodejs.org#4200

I have decided to revert since we need a quick fix to address the a11y problem, which is more pressing than the cosmetic feature implementation that somehow was not tested on dark mode, which several people (like myself) have set as the default.

### DIFF
--- a/layouts/css/_variables.scss
+++ b/layouts/css/_variables.scss
@@ -2,7 +2,7 @@ $body-max-width: 980px;
 
 $black: #000;
 $white: #fff;
-$white-alpha-b3: #ffffffb3;
+
 $node-green: #43853d;
 //$bright-green: #028a00;
 $light-green: #eaf5e9;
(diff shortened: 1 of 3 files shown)


Trott:
> cosmetic feature

It was not cosmetic. That change fixed color contrast issues elsewhere.

> somehow was not tested on dark mode

It was indeed tested in dark mode. I don't know why you are saying this.

DerekNonGeneric:
@Trott, I was just doing some research on this because of the issue in #4210 that I was able to reproduce. I have now figured out why this could have appeared to have been looking fine in dark mode to one person, but not to another.

From what I can tell so far, there currently at least two ways for a user of this site to enable dark mode:
- use the toggler button (likely what you and the PR author have been using)
- communicating your preference as a setting in your browser (likely what the OP of the issue and myself use)

Probably not the popular opinion and I will put it back (sorry), but the screenshot from https://github.com/nodejs/nodejs.org/issues/4210 is how it also looked for me. The problem was not going to be solved as simply as I had hoped originally as I noticed that there are several ways to enable dark mode here&hellip;

Trott:
I definitely see the color contrast problem reported in #4210. But the revert re-introduced *other* color contrast problems. Nothing you or I or someone else can't figure out and fix (separately or together). I just didn't think the comments about it not being tested etc. were accurate and could be interpreted as disrespectful to the folks who did review and test.  (Would be great to get automated color-contrast testing on all pages, similar to broken-link testing, so that we can stop playing whack-a-mole with color contrast issues!)

PR #4215, opened 2021-10-28
fix: make highlight-box text visible in dark mode

jonchurch:
Visit https://nodejs.org/en/blog/npm/peer-dependencies/ in darkmode, scroll to bottom and see the highlight-box appears empty.

In light mode the text color is the same as what's used for the article body in that mode, so for the fix I used the same color that the dark mode article body uses.

## Before 📷:
before
-------

## After 📷:
after

### DIFF
--- a/layouts/css/layout/_dark-theme.scss
+++ b/layouts/css/layout/_dark-theme.scss
@@ -82,6 +82,7 @@
     }
 
     .highlight-box {
+      color: $light-gray3;
       background-color: $dark-code-background;
     }
   }


Trott:
Nice! If you're feeling it, there's also *very* low contrast on the byline that could be fixed. You can barely see Domenic's name or the fact that it was posted way back in 2013....

image

Trott (approved):

DerekNonGeneric (approved):
Wow, had no idea about this one. Thanks for fixing it!

PR #4217, opened 2021-10-31
fix: add commits section to Gallium release post

DerekNonGeneric:
Adds the &ldquo;Commits&rdquo; section that was missing from the Gallium release post.

/cc @richardlau @BethGriggs @targos @danielleadams


richardlau (changes requested):
We don’t include the release commit in the list of commits. Since there were no other commits, the commits section was left out (see the first post in https://github.com/nodejs/node/pull/40536).

DerekNonGeneric:
Okay, so you are saying that the rationale behind the commit link (and entire section) not being present is on purpose due to there only being a single commit, but if we need to make some other single-commit release (not a simple bit-flipping LTS one), I think I would want it to show up there&hellip;

Thanks for letting me know; I will close this one out.

PR #4236, opened 2021-11-09
dependency bump: marked from 3.0.8 to 4.0.1

humble-barnacle001:
Fixes the bump issue of dependabot in #4233 caused by bumping the major version of `marked`


nschonni (approved):

.gitignore
```diff
@@ -12,3 +12,6 @@ npm-debug.log
 
 # Netlify
 .netlify
+
+#VSCode settings; makes life easier for some, tougher for most
+.vscode/
```
Trott: If you run VS Code or whatever creates this directory, you should probably put this in a .gitignore file in your home directory. I'd prefer we don't try to accommodate every IDE separately in our repo .gitignore.

Trott (approved):
LGTM. I'd prefer we pull out the `.gitignore` change, but I'm OK if we don't.

.gitignore
humble-barnacle001: Removed in 956b1ac

humble-barnacle001:
Bump to fix for #4249

humble-barnacle001:
> LGTM. I'd prefer we pull out the `.gitignore` change, but I'm OK if we don't.

@Trott  Done

Issue #4235, opened 2021-11-09
The <input> inside the releases table overflowed.


18202409203:
* **URL**: https://nodejs.org/en/download/releases/
* **Browser version**: Chrome / Firefox
* **Operating system**: Ubuntu 16

The `<input>` inside the table overflowed.

![chrome](https://user-images.githubusercontent.com/30512639/140892218-afdfdb8b-37dd-49af-a268-323d4b3bbe9b.png)

I think it should be `box-sizing: border-box;`

tymnim:
The issue seems to be in `static/js/fancyTable.min.js`. I am not sure if there's a non-minified file of this.

if `("<input>",{placeholder:r.inputPlaceholder,style:"width:100%;"+r.inputStyle})` replaced by `("<input>",{placeholder:r.inputPlaceholder,style:"width:100%;box-sizing:border-box;"+r.inputStyle})`, it should work. I hope someone can apply this change.

thomscoder:
I have worked on it and pull requested. Hope it fixes the issue

Trott:
> The issue seems to be in `static/js/fancyTable.min.js`. I am not sure if there's a non-minified file of this.

It looks like that file was added by @MaledongGit in 6eba6d1b912ca902aa70b02f51bf6cd2abc7fdce without any approval by other contributors. Hopefully they can indicate where the unminified code can be found, etc.

Trott:
> It looks like that file was added by @MaledongGit in [6eba6d1](https://github.com/nodejs/nodejs.org/commit/6eba6d1b912ca902aa70b02f51bf6cd2abc7fdce) without any approval by other contributors. Hopefully they can indicate where the unminified code can be found, etc.

(The search feature is a nice improvement, I think, so I'm surprised someone didn't 👍  that PR at some point!)

Trott:
> It looks like that file was added by @MaledongGit in [6eba6d1](https://github.com/nodejs/nodejs.org/commit/6eba6d1b912ca902aa70b02f51bf6cd2abc7fdce) without any approval by other contributors. Hopefully they can indicate where the unminified code can be found, etc.

Looks like the unminified source would be in https://github.com/myspace-nu/jquery.fancyTable.

thomscoder:
I have worked on this. This issue is currently in PR

nschonni:
Looks like the latest fancy tables version has fixed it 
[...]

PR #4243, opened 2021-11-10
fix overflowed input

thomscoder:


thomscoder:
There was an overflowed input in the releases tables. I think it was caused by some extra padding. This should fix the issue

Trott:
Seems like the package-lock.json change is unrelated? Can we remove it?

Trott:
I'm concerned that if/when we update the code from source, the bug will be re-introduced. I'm not sure what the best path here is. 

* Float this patch anyway?
* Put the unminified source in the repo here and minify it as part of our pipeline?
* Remove the dependency and replace it with something else or write it ourselves?
* Submit a fix upstream?
* Fix outside of the minified file via CSS or something else?

Part of the issue is...uh...I haven't looked closely at the source of the problem yet.

package-lock.json
```diff
@@ -4,7 +4,6 @@
   "requires": true,
   "packages": {
     "": {
-      "name": "nodejs.org",
```
nschonni, line 7: This change doesn't seem related

thomscoder:
> Seems like the package-lock.json change is unrelated? Can we remove it?

Yes. My bad!

package-lock.json
thomscoder: Yes, my bad. Should've noticed the double push

humble-barnacle001:
> There was an overflowed input in the releases tables. I think it was caused by some extra padding. This should fix the issue

Could please post a link and/or SS of the page so that I could look closely into the source?

thomscoder:
> > There was an overflowed input in the releases tables. I think it was caused by some extra padding. This should fix the issue
> 
> Could please post a link and/or SS of the page so that I could look closely into the source?

Issue link: https://github.com/nodejs/nodejs.org/issues/4235
Bug's URL link: https://nodejs.org/en/download/releases/ (input in the table is overflowing)

humble-barnacle001:
> I'm concerned that if/when we update the code from source, the bug will be re-introduced. I'm not sure what the best path here is.
> 
>     * Float this patch anyway?
> 
>     * Put the unminified source in the repo here and minify it as part of our pipeline?
> 
>     * Remove the dependency and replace it with something else or write it ourselves?
> 
>     * Submit a fix upstream?
> 
>     * Fix outside of the minified file via CSS or something else?
> 
> 
> Part of the issue is...uh...I haven't looked closely at the source of the problem yet.


I think what we could probably do is:
- Submit a fix to the upstream; If the person is unresponsive, we could then copy it. Till then I could probably patch the page with the `inputStyle` option of the dependency
- Use `jquery.fancytable` from npm instead of copying the file directly so that dependabot takes care of the update

What say @thomscoder ?

humble-barnacle001:
Changing `static/js/previousVersion.js` to this (adding: `inputStyle: 'box-sizing: border-box;'`)
Screenshot from 2021-11-13 20-28-51

Fixes the issue without changing the source-code for the js because I think this is due to the CSS of the website rather than a bug.
I would those suggest to at least include the original source https://github.com/myspace-nu/jquery.fancyTable in the `static/js/fancyTable.min.js` to give credit as well as provide an uplink for reference. Also we could use the CDN link instead of copying it seperately.
This should probably be a new PR @thomscoder

humble-barnacle001:
Alternatively, adding this CSS option (inspired by Bootstrap since the theme closely follows its pattern) in the file `layouts/css/_base.scss` is a worldwide fix for any such issue but needs to investigated in other pages to see possible disruptions in other pages

```Css
*,
*::before,
*::after {
  box-sizing: border-box;
}
```

nschonni:
Opened up https://github.com/nodejs/nodejs.org/pull/4250 that might conflict with this, unless you get the change accepted upstream

thomscoder:
> https://github.com/myspace-nu/jquery.fancyTable

Alright! Should I proceed with the fix? and we treat the conflict (@nschonni mentioned) that may arise separately?

nschonni:
Looks like it's now been reported there too https://github.com/myspace-nu/jquery.fancyTable/issues/21

humble-barnacle001:
> > https://github.com/myspace-nu/jquery.fancyTable
> 
> Alright! Should I proceed with the fix? and we treat the conflict (@nschonni mentioned) that may arise separately?

I think you could proceed with the fix in https://github.com/nodejs/nodejs.org/pull/4243#issuecomment-968082507 in a new PR after #4250 is patched through

humble-barnacle001:
Fixed in https://github.com/myspace-nu/jquery.fancyTable/issues/21#issuecomment-968996194 version 1.0.25

nschonni:
I've updated the #4250 with the new patched version

nschonni:
Closing this one since I landed the updated version from NPM

PR #4246, opened 2021-11-11
chore: Use Node 16 in CI

nschonni:
Now that it is the LTS branch

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -7,9 +7,6 @@ on:
       - "dependabot/**"
   pull_request:
 
-env:
-  NODE: 14.x
-
 jobs:
   run:
     name: Node on ${{ matrix.os }}
@@ -27,7 +24,7 @@ jobs:
       - name: Set up Node.js
         uses: actions/setup-node@v2
         with:
-          node-version: "${{ env.NODE }}"
+          node-version: 16
           cache: npm
 
       - run: java -version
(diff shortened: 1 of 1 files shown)


trivikr (approved):

.github/workflows/test.yml
```diff
@@ -27,7 +24,7 @@ jobs:
       - name: Set up Node.js
         uses: actions/setup-node@v2
         with:
-          node-version: "${{ env.NODE }}"
+          node-version: 16
```
Trott: For what it's worth, I believe this should always install the latest LTS release:

```suggestion
          node-version: lts/*
```

Trott (approved):
LGTM with or without my suggestion (since I only *think* my suggestion will work but I'm not sure)

humble-barnacle001:
Dependency installation with node 16 gives the depreciation warnings:
```
npm WARN deprecated sprintf@0.1.5: The sprintf package is deprecated in favor of sprintf-js.
npm WARN deprecated fs-tools@0.5.0: Use shelljs, glob, mkdirp, rimraf, or other alternatives
npm WARN deprecated coffee-script@1.12.7: CoffeeScript on NPM has moved to "coffeescript" (no hyphen)
```

DerekNonGeneric (approved):
According to the docs, that does appear to be a valid version range, which allows the nvm LTS syntax: `lts/erbium`, `lts/fermium`, `lts/*`. Guess that would be the Bash version of the command (they do not specify).

Refs: https://github.com/actions/setup-node#usage

PR #4256, opened 2021-11-16
meta: use logo-hexagon instead of card

VoltrexKeyva:
Remove the `logo-hexagon-card` PNG only used in meta properties and replace it with the `logo-hexagon` PNG for a better fitting logo with a transparent background to blend in with the whatever it is used on; instead of a logo with a white background that does not look good on all platforms that embed the meta properties, such as Discord for example (The one shown is the `logo-hexagon-card` PNG):

Screenshot_20211116-093639_Discord

This would also be consistent with many projects using transparent backgrounds for their meta logos.


VoltrexKeyva:
cc @Trott @MaledongGit @nschonni

Trott:
Pinging folks who have modified the logo files: @XhmikosR @MylesBorins 

I'll rubber-stamp it if no one else weighs in, but I don't know a lot about "Should we use SVG instead?" and questions like that.

Trott (dismissed):
Rubber-stamp LGTM

Trott:
@nodejs/website Anyone want to review this one?

XhmikosR:
IMHO this shouldn't land. The two images are not identical. They have different padding.

The proper way would be to create a new logo-hexagon-card-white.png PNG file, optimize it and use it without removing the old one.

VoltrexKeyva:
> IMHO this shouldn't land. The two images are not identical. They have different padding.
> 
> The proper way would be to create a new logo-hexagon-card-white.png PNG file, optimize it and use it without removing the old one.

I don't think it's a good idea to create another PNG file that is alternatively identical to the original logo-hexagon-card PNG file since it wouldn't really change much and would probably look the same way as the original does, which would seperate how the meta properties are shown from the other big projects that use it in the correct/consistent ways.

The padding shouldn't matter much here as it is only used for meta properties which doesn't affect anything (or at least nothing much), this is just for consistency and for better icon for URL embedders in various platforms.

Most URL embedders nowadays already do a great job at modifying the metadata of the given media files such as these PNG files to set a correct padding or resize them in the best ways fit for the platform they're used on, which most of the platforms already use one of those great URL embedders or make their own. 😄

It would also be a waste to keep the old one as it'll be unused entirely.

XhmikosR:
It's **not** identical. Transparent vs white background. Also, the images might be used by other Node.js projects as we've seen in the past.

Either way, my take is that a new image should be created from the vector files with the same padding and dimensions, then properly optimized and committed. **If** we are 100% sure the current PNG is not used elsewhere, sure, we can remove it, otherwise it doesn't hurt anyone leaving the old image there.

VoltrexKeyva:
@XhmikosR alright, so I've went ahead to create another PNG file with the exact same padding and dimensions but the only difference being that the background is transparent, but as I've already mentioned, it does not look as appealing and consistent just like the other projects' meta property logos.

The current one (using the already existing `logo-hexagon` PNG file):

Screenshot_20211126-123142

The new one as you've suggested (using the new `logo-hexagon-card-white` PNG file):

Screenshot_20211126-123505

Although they don't look that different, the new one might seem off, but that might be just up to the preference, but the main goal here is to be consistent; so, what do you think?

silverwind (approved):
I wouldn't care about padding. Having fewer different files seems better.

aymen94 (approved):

aymen94:
LGTM

silverwind:
Regarding links to the old files, maybe a redirect can be added for `logo-hexagon-card.png` to `logo-hexagon.png`. I think it needs to be done in the nginx configuration which is in a different repo somewhere.

silverwind:
For the redirect, I think it needs to be added to https://github.com/nodejs/build/blob/master/ansible/www-standalone/resources/config/nodejs.org.

jonymusky (approved):
LGTM

osk2 (approved):
LGTM

shannonenz (approved):

bnb (changes requested):
Could you check with https://www.opengraph.xyz/ before landing? I want to make sure this looks good on socials in addition to chat apps.

VoltrexKeyva:
> Could you check with https://www.opengraph.xyz/ before landing? I want to make sure this looks good on socials in addition to chat apps.

It generally looks almost the same but more refined as the transparency makes it blend in, although it kinda looks a bit zoomed in.

Current:



New:



Current:



New:

bnb:
Yeah, the current one (I think I'm the one who made it) intentionally had padding so it wouldn't be zoomed in. Could you add the same (or similar) amount of padding so it'll show correctly? ❤️

ovflowd:
@VoltrexKeyva are you still interested in this PR?

VoltrexKeyva:
> @VoltrexKeyva are you still interested in this PR?

I completely forgot about this PR, though there has been many changes as to how the metadata is shown and some of the changed files were deleted, I'm going to close this PR to see if I can open another with the appropriate changes.

PR #4277, opened 2021-12-03
docs: add links for external references

mapag-g:
Hey y'all! 👋 

Noticed links were missing for `backpressuring-in-streams` article.

Added links for standard docs and some useful resources like the one explaining backpressuring concept.


nschonni:
Thanks it looks like you've edited one of the stub copies in the Spanish folder. The original is in https://github.com/nodejs/nodejs.org/blob/main/locale/en/docs/guides/backpressuring-in-streams.md.
It looks like Crowdin mangled the anchor links in that article when it made the copies

PR #4280, opened 2021-12-06
fix: remove outdated security policy, add link

Trott:
Remove outdated security policy page. Replace all links to it with a
link to the equivalent information in the core repository (where it was
recently updated) so that we aren't duplicating the information in
multiple places, inevitably resulting in contradictory information when
only one place is updated.

Refs: https://github.com/nodejs/security-wg/issues/775


Trott:
@nodejs/security-triage

Trott:
(We'll probably want to wait on landing this until we can get a redirect in place over in the build repo.)

mcollina (approved):
lgtm

rcowsill:
Will this break external links to https://nodejs.org/en/security/, or would they be redirected to https://github.com/nodejs/node/blob/HEAD/SECURITY.md#security automatically?

Also, this appears to drop translations for the security page. All locales point at the same security policy in English.

Trott:
> Will this break external links to https://nodejs.org/en/security/, or would they be redirected to https://github.com/nodejs/node/blob/HEAD/SECURITY.md#security automatically?

Redirect is in https://github.com/nodejs/build/pull/2819. As alluded to in https://github.com/nodejs/nodejs.org/pull/4280#issuecomment-986363343, this PR should probably wait until that one lands.

> Also, this appears to drop translations for the security page. All locales point at the same security policy in English.

Correct. This is the case for all of our links to the GitHub repository. It would be great to find a way to keep translations, but I'd rather link to correct security-reporting information in English only than link to incorrect information in many languages. The reporting process will need to be done in English anyway (which is something else that it would be great to fix somehow).

danielleadams (approved):

mhdawson:
I'm a bit torn, we currently have a number of translations.

While I get the lets have one copy I'm not 100% that should trump supporting people for who English is not their first language.

A compromise might be a page which is internationalized that has less specifics and then points to the English page but not sure that really is any better.

I do think that we should delete translations when a page is updated to ensure that they are in sync with the english translations.

Trott:
> While I get the lets have one copy I'm not 100% that should trump supporting people for who English is not their first language.

Many of those translations don't mention HackerOne and instruct people to email security@nodejs.org.

Ideally, I'd like to provide translations. But if given the choice between translations and correct/up-to-date information, that choice is easy when it comes to security reporting.

I think we should land this and then open a separate issue (maybe at the TSC level--there are a lot of things I want to talk about concerning our websites) about how to get a robust i18n program going.

PR #4285, opened 2021-12-06
chore: remove standard

nschonni:
We migrated from the CLI to the ESLint plug-ins, so it isn't needed


DerekNonGeneric (approved):
RSLGTM

DerekNonGeneric:
As followup to https://github.com/nodejs/nodejs.org/pull/3540#discussion_r748567497, seems like the more-desirable config would be [`eslint-config-semistandard`](https://www.npmjs.com/package/eslint-config-semistandard).

PR #4286, opened 2021-12-06
chore: switch to semistandard

nschonni:
From https://github.com/nodejs/nodejs.org/pull/3540#discussion_r748567497
Don't feel strongly either way, but I guess this is probably more in line with the main node repo


DerekNonGeneric (approved):
Thanks, you're the best!

PR #4287, opened 2021-12-06
chore: Turn on ESLint for markdown

nschonni:
Turn on linting/fixing for JS blocks in markdown and autofix many formatting issues automatically. Ignored a few more that come up because it parses the document as a whole scope


Trott (approved):
I'm OK with this as it is, but I do think I'd prefer changing the quotes rule to require single quotes as that would probably reduce churn here and also align it with Node.js core practices. (Perhaps it's time to create eslint-plugin-node-core similar to remark-plugin-lint-node.)

Trott (approved):

.eslintrc
```diff
@@ -9,5 +9,26 @@
   ],
   "rules": {
     "prettier/prettier": "error"
-  }
+  },
+  "overrides": [
+    {
+        "files": ["**/*.md"],
```
XhmikosR: Indentation seems off in this file

.eslintignore
```diff
@@ -2,3 +2,5 @@
 static/legacy/
 external/
 build/
+# Top level await isn't supported till ESLint 8
+locale/en/blog/release/v17.0.0.md
```
XhmikosR, line 6: Are we blocked because standard is still using eslint 7.x?

I think I've expressed my opinion in the past, standard does not catch many of real issues and personally I never use it. prettier doesn't catch real issues either.

Just thinking out loud for the future :)

XhmikosR:
I can't review every file, but I suppose these are autofixable mostly, right?

From a quick look, the diff looks good, I'll try to have another look tomorrow.

.eslintignore
nschonni: Yeah, I believe the ESLint upgrade is blocked on prettier, but eslint-plugin-promise just released a compatible version, so that might have been the last one remaining

.eslintrc
nschonni: Yeah, reformatted. Might make sense to just convert this to the JS version later

locale/zh-cn/docs/guides/domain-postmortem.md
```diff
@@ -127,16 +139,15 @@ let uid = 0;
 
 // Setting up temporary resources
 const buf = Buffer.alloc(FILESIZE);
-for (let i = 0; i < buf.length; i++)
-  buf[i] = ((Math.random() * 1e3) % 78) + 48; // Basic ASCII
+for (let i = 0; i < buf.length; i++) buf[i] = ((Math.random() * 1e3) % 78) + 48; // Basic ASCII
```
XhmikosR, line 142: Just wondering, don't we enforce brackets in these cases?

nschonni: I think it's the `curly` rule coming from Standard

XhmikosR (approved):

.eslintignore
XhmikosR: All good, as long as someone remembers to revert this later :)

XhmikosR:
We should look into enabling ESLint caching later.

PR #4304, opened 2021-12-18
Update devDependencies

XhmikosR:
* linkinator                        ^2.16.2  →   ^3.0.1
* marked                             ^4.0.7  →   ^4.0.8
* remark-cli                        ^10.0.0  →  ^10.0.1
* stylelint                        ^13.13.1  →  ^14.1.0
* stylelint-config-twbs-bootstrap    ^2.2.4  →   ^3.0.0
* switch to @metalsmith/markdown 1.4.0

Also, specify `--report-needless-disables` for stylelint while at it

918 -> 839 packages


package.json
```diff
@@ -17,7 +17,7 @@
     "test:lint": "npm-run-all --parallel --aggregate-output test:lint:*",
     "test:lint:js": "eslint .",
     "test:lint:md": "remark -qf .",
-    "test:lint:stylelint": "stylelint \"layouts/css/**/*.{css,scss}\"",
+    "test:lint:stylelint": "stylelint \"layouts/css/**/*.{css,scss}\" --rd",
```
nschonni: What do the new flags do?

XhmikosR: Reports needless disables. Too lazy to split it, sorry 👼

package.json
```diff
@@ -30,6 +30,7 @@
     "node": ">=12"
   },
   "dependencies": {
+    "@metalsmith/markdown": "^1.4.0",
```
XhmikosR, line 33: Might be worth someone going through the metalsmith packages we use and see if there are new scoped ones available

package.json
nschonni: No worries, maybe the full version would be better `--report-descriptionless-disables`

nschonni (approved):

package.json
XhmikosR: Feel free to change it, I don't mind. I too prefer to long ones but npm scripts become easily bloated hence why I went with the short one here.

PR #4306, opened 2021-12-18
CI: restrict pushes event and add `workflow_dispatch`

XhmikosR:
This will allow us to trigger the workflow manually

Might be worth restricting on push now that we can manually trigger the workflow. I'll leave that for someone else to tackle later.


Trott (approved):

bnb:
Is there a specific reason we'd need to manually trigger that you had in mind?

Not opposed at all (I advocated for this feature to be shipped 😅) , just want to make sure adding it actually solves a (potential) problem.

XhmikosR:
The main idea is to restrict triggering the actions on all pushes (we should change it only to main since we always use PRs) and so this feature makes it possible to test a branch before opening a PR or without having a PR.

Right now, we have duplicate runs for pushes and PRs.

XhmikosR:
I updated the PR with the proposed changes, please have another look. :)

Trott (approved):

aymen94 (approved):

PR #4307, opened 2021-12-18
package.json: add `--continue-on-error` flag

XhmikosR:
This should allow the `test:lint` script to finish and not just exit immediately if one of the subscripts fails. The script will still fail, just not immediately without the other scripts finishing first.


aymen94:
LGTM 😁

aymen94 (approved):

PR #4309, opened 2021-12-19
Move linkinator to Actions

XhmikosR:
852 -> 744 packages

Opening this for feedback.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -7,6 +7,9 @@ on:
   pull_request:
   workflow_dispatch:
 
+env:
+  FORCE_COLOR: 2
+
 jobs:
   run:
     name: Node on ${{ matrix.os }}
@@ -44,4 +47,10 @@ jobs:
         run: npm run test:html
 
       - name: Run linkinator
-        run: npm run test:linkinator
+        uses: JustinBeckwith/linkinator-action@v1
+        with:
+          linksToSkip: "^(?!http://localhost)"
+          paths: en
+          recurse: true
+          serverRoot: build
+          verbosity: error
--- a/package-lock.json
+++ b/package-lock.json
@@ -47,7 +47,6 @@
         "eslint-plugin-node": "^11.1.0",
         "eslint-plugin-prettier": "^4.0.0",
         "eslint-plugin-promise": "^6.0.0",
-        "linkinator": "^3.0.2",
         "lockfile-lint": "^4.6.2",
         "nock": "^13.2.1",
         "node-fetch": "^2.6.6",
@@ -60,8 +59,7 @@
         "stylelint": "^14.1.0",
         "stylelint-config-twbs-bootstrap": "^3.0.0",
         "tape": "^5.3.2",
-        "vnu-jar": "21.10.12",
-        "wait-on": "^6.0.0"
+        "vnu-jar": "21.10.12"
       },
       "engines": {
         "node": ">=12"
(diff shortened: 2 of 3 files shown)


.github/workflows/test.yml
```diff
@@ -7,6 +7,9 @@ on:
       - "dependabot/**"
   pull_request:
 
+env:
+  FORCE_COLOR: 2
```
XhmikosR, line 11: This doesn't gain us a lot and arguably doesn't belong in this PR, so I can either split it or drop it. Only the `npm ci` step currently benefits.

XhmikosR:
@Trott you are probably aware, but it seems that PRs don't require approvals, only statuses to pass. IMHO this should change.

Trott:
> @Trott you are probably aware, but it seems that PRs don't require approvals, only statuses to pass. IMHO this should change.

@nodejs/releasers Would requiring a review approval for this repository increase friction for releases? I'm thinking it would because y'all probably immediately merge new-release blog posts after opening them, right?

If that's the case, one workaround might be to set the releasers team as admins for this repository and allow administrators to bypass the requirement.

Another possibility would be to enable auto-merge and releasers can use that. (Auto-merge will land the PR once all requirements are met, which in this case would be a passing CI and however many approvals we decide should be required.)

Trott (approved):

XhmikosR:
To clarify, I don't want to increase the work or friction for releases. My main concern is that anyone with push rights on this repo can deploy something on the main server without someone else having a second look.

Also, I'm unsure if 2FA is enforced for everyone. It's just that IMHO it looks bad from a security perspective is all.

Trott:
> Also, I'm unsure if 2FA is enforced for everyone.

2FA is required and enforced. (GitHub allows us to set up that requirement for the entire org, so that's where it's enforced. If you're in the nodejs GitHub org, you must have 2FA.)

It might not apply to outside collaborators, but this repo doesn't have any of those.

XhmikosR:
Your call, as long as you think the current approach is safe enough :)

Trott:
> Your call, as long as you think the current approach is safe enough :)

I agree with you that there should be an approval requirement/process. I just want to make sure I don't trip up releasers in the process. 

I also think we should pare down the people who can commit to this repo. Like, *way* down.

XhmikosR:
> I also think we should pare down the people who can commit to this repo. Like, _way_ down.

Totally agree on this. Unsure how to organize this, though. Unless there are people who are available to review more actively.

Anyway, sorry for getting sidetracked, I just wanted to bring the issue to your attention because something doesn't feel right to me.

aymen94 (approved):

targos:
> > @Trott you are probably aware, but it seems that PRs don't require approvals, only statuses to pass. IMHO this should change.
> 
> @nodejs/releasers Would requiring a review approval for this repository increase friction for releases? I'm thinking it would because y'all probably immediately merge new-release blog posts after opening them, right?

Probably. It's rare that I get a review before merging the blog post PR.

> If that's the case, one workaround might be to set the releasers team as admins for this repository and allow administrators to bypass the requirement.

Another workaround can be to allow the releasers team to bypass pull request requirements:

image

> Another possibility would be to enable auto-merge and releasers can use that. (Auto-merge will land the PR once all requirements are met, which in this case would be a passing CI and however many approvals we decide should be required.)

I'd be +1 on enabling auto-merge right now. It's annoying to have to wait for CI to pass. Releases have already other long manual steps.

Trott:
Pull requests now require one approval before landing. The releasers team is the one team configured to be able to bypass the requirement.

Trott:
Auto-merging is also enabled.

PR #4310, opened 2021-12-19
Remove unused Contributor Card code

XhmikosR:


static/js/themeSwitcher.js
```diff
@@ -1,6 +1,6 @@
+/* eslint-disable no-var */
```
nschonni: Why?

XhmikosR: Because we only use `var`s. And this file wrongfully used const. Either we do it consistently everywhere and drop old IE support later, or we just stick to what we did.

Trott (approved):
I'd prefer we didn't disable the ESLint rule but won't block on that.

Trott:
> I'd prefer we didn't disable the ESLint rule but won't block on that.

Oh, guess I should wait until it's ready-to-review before approving....

XhmikosR:
> I'd prefer we didn't disable the ESLint rule but won't block on that.

Yeah, I think we should do that in a separate PR then and be consistent with it. Now it's a warning, it should be an error and applied everywhere.

> Oh, guess I should wait until it's ready-to-review before approving....

I need to test this a bit more, hopefully tomorrow.

static/js/main.js
```diff
@@ -68,14 +68,15 @@
   });
 
   function setTheme(theme) {
-    document.querySelector('html').setAttribute(themeAttr, theme);
+    document.documentElement.setAttribute('data-theme', theme);
```
XhmikosR: This doesn't belong to this PR but I was too lazy to split it.

PR #4311, opened 2021-12-19
Update npm scripts

XhmikosR:
* enable caching for eslint and stylelint
* double quote globs

Let me know about the cache location.

Unfortunately, remark does not seem to have a cache implementation and it seems it's the slowest of our lint scripts.


nschonni:
One approach i've seen is just to used the default location at the root, the `action/cache` can be used in CI to restore it

XhmikosR:
Do we want to use the cache on CI too? I usually don't use it on CI TBH, but happy to move it to the root folder.

Our main issue is still remark, but I have no idea what we can do since it doesn't have caching.

Trott:
> Our main issue is still remark, but I have no idea what we can do since it doesn't have caching.

It's not perfect, but over in `nodejs/node`, we have a timestamp file and our `make` task only lints what is newer than the timestamp file. That implementation uses some `Makefile` magic, but maybe we could do something similar.

Ref: https://github.com/nodejs/node/blob/23637e9a3b208892449268290143dad4391fee45/Makefile#L1257-L1276

XhmikosR:
Also, https://github.com/unifiedjs/unified-engine/issues/22 which is what remark is using, in case anyone wants to work on it.

XhmikosR:
How about landing this as is and @nschonni or someone else improves it further later for CI and perhaps remark?

It should speed local development a lot, so that's a win.

Trott (approved):

PR #4314, opened 2021-12-20
Lossy compress PNG files

XhmikosR:
Before: 1,55 MB (1.627.160 bytes)
After:   720 KB (737.908 bytes)


nschonni:
I'm not sure if these are updated during the deploy or not from https://github.com/nodejs/nodejs.org/blob/9239c9eb30ed78b58722ef111179f55eb687d366/package.json#L10

XhmikosR:
The external ones are probably not updated indeed. I plan to drop those from this patch, I just opened the PR with all the changes first.

bnb:
might be worth also documenting how you did this so we can reproduce it in the future :)

bnb (approved):
LGTM, would love to see the approach documented in a reproducible way though.

XhmikosR:
I use plenty of tools and custom scripts, but here it is:

1. https://kraken.io/web-interface lossy
2. `optipng -o7`
3. pngout with various blocksizes
3. `zopflipng -lossy_transparent -y --iterations=500`

PR #4320, opened 2021-12-21
main.js: consistency changes

XhmikosR:
Later, we should enforce const/let properly. This patch just streamlines the code with what we are currently using.

### DIFF
--- a/static/js/main.js
+++ b/static/js/main.js
@@ -34,9 +34,9 @@
     langPickerElement.removeChild(currentLangElement.parentNode);
   }
 
-  const toggleFunction = function () {
+  var toggleFunction = function () {
     langPickerElement.classList.toggle('hidden');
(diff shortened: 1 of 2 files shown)


KenG-pro:
Thanks hope things work out

aymen94 (approved):

Trott:
@XhmikosR Do you want to rebase so this can be landed?

XhmikosR:
I'm not sure this PR makes sense anymore... I mean, I see pretty modern features being used like `??` which contradicts with the browsers in .browserslistrc.

aymen94:
@XhmikosR I think not make sense.

XhmikosR:
Either way, .browserslistrc should be updated and `var` should be removed.

aymen94:
@XhmikosR yes, we can update it by making a new PR.

PR #4322, opened 2021-12-23
Update devDependencies and downgrade eslint-plugin-promise

XhmikosR:
eslint-config-semistandard requires eslint-plugin-promise@5

FYI @Trott @nschonni so that you don't land the update in the future.

Unsure why things worked but if I remove the lock file locally I get the errors as expected:

```
C:\Users\xmr\Desktop\nodejs.org>npm i
npm ERR! code ERESOLVE
npm ERR! ERESOLVE unable to resolve dependency tree
npm ERR!
npm ERR! While resolving: nodejs.org@undefined
npm ERR! Found: eslint-plugin-promise@6.0.0
npm ERR! node_modules/eslint-plugin-promise
npm ERR!   dev eslint-plugin-promise@"^6.0.0" from the root project
npm ERR!   peer eslint-plugin-promise@">=4.2.1" from eslint-config-semistandard@16.0.0
npm ERR!   node_modules/eslint-config-semistandard
npm ERR!     dev eslint-config-semistandard@"^16.0.0" from the root project
npm ERR!
npm ERR! Could not resolve dependency:
npm ERR! peer eslint-plugin-promise@"^4.2.1 || ^5.0.0" from eslint-config-standard@16.0.3
npm ERR! node_modules/eslint-config-standard
npm ERR!   peer eslint-config-standard@">=16.0.3" from eslint-config-semistandard@16.0.0
npm ERR!   node_modules/eslint-config-semistandard
npm ERR!     dev eslint-config-semistandard@"^16.0.0" from the root project
npm ERR!
npm ERR! Fix the upstream dependency conflict, or retry
npm ERR! this command with --force, or --legacy-peer-deps
npm ERR! to accept an incorrect (and potentially broken) dependency resolution.
npm ERR!
npm ERR! See C:\Users\xmr\AppData\Local\npm-cache\eresolve-report.txt for a full report.

npm ERR! A complete log of this run can be found in:
npm ERR!     C:\Users\xmr\AppData\Local\npm-cache\_logs\2021-12-23T07_23_56_061Z-debug.log
```

PR #4323, opened 2021-12-23
package.json: add missing properties

XhmikosR:
`npm init` (using npm 8.1.3) is using the explicit repository and bugs properties but I guess it was intentional to skip them? If not, let me know so that I add them:

```json
  "repository": {
    "type": "git",
    "url": "git+https://github.com/nodejs/nodejs.org.git"
  },
  "bugs": {
    "url": "https://github.com/nodejs/nodejs.org/issues"
  },
```


nschonni:
Kinda -0, since this is an unpublished package, so I don't think these are really needed

XhmikosR:
I agree, but I personally keep the defaults. I noticed it recently here: https://github.com/nodejs/nodejs.org/pull/4322#issue-1087451709

```
npm ERR! While resolving: nodejs.org@undefined
```

Either way, I'm fine with dropping the patch, I just happened to notice this.

Trott:
Totally neutral on this one. Happy if it lands, happy if we close it. Whichever.

Trott (approved):

Issue #4063, opened 2021-08-23
Default-to-English-when-translation-is-absent gets weird when language is RTL


Trott:
* **URL**: https://nodejs.org/ar/docs/guides/blocking-vs-non-blocking/
* **Browser version**: n/a
* **Operating system**: n/a

If you go to (for example) the Arabic URL for a page that hasn't yet been translated to Arabic, the English language version is displayed, which is certainly better than displaying nothing and seems like desirable behavior. However, in RTL languages like Arabic, the English text displays RTL, which is probably not what we want. (Although if someone who is a native speaker of a RTL language tells me that this rendering makes it easier to understand, then that's good enough for me.)

Notice that the text is right justified and that terminating punctuation appears at the start of each sentence rather than at the end.

![image](https://user-images.githubusercontent.com/718899/130389384-cd15843d-c800-40eb-8e67-a11f0d4234f1.png)

thomscoder:
I worked on it on #4336

PR #4336, opened 2021-12-29
auto detect ltr and rtl text

thomscoder:
This small bug fix is related to #4063 
The bug presents when manually inserting the url https://nodejs.org/ar/docs/guides/blocking-vs-non-blocking/ or when in an rtl language clicking href that lead to partially translated pages.

The fix is simple enough yet tricky. The fix applies a dir attribute that sets the direction according to the first strong character found in the text block: https://www.w3.org/International/tests/repo/results/the-dir-attribute-auto

I had no problem on Edge as well.

Other solution might be the <bdi> tag or the unicode-bidi rule:
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/bdi

I fixed some warnings of non-closed tags in external/survey-2018/index.html while I was at it.


Trott:
@nodejs/website

aymen94 (approved):

Issue #4166, opened 2021-10-10
Icons misPositioning and background color doesn't match what we see in doc


ghost:
1. In the light mode for Node.js org, the background color is still dark. But in the doc, the color seems light-gray more:

【About of Node.js org】
![About](https://user-images.githubusercontent.com/52018749/136678356-34e457f5-548b-44df-933e-ec0869446d76.JPG)

【In the doc】
![Doc](https://user-images.githubusercontent.com/52018749/136678379-e025d6ad-a1a4-4c5b-8616-4cdc23bbfb8b.JPG)

2. When the window is smaller, the language icon and the dark/light switcher will override the menu:

https://user-images.githubusercontent.com/52018749/136678389-7d32af57-f3f3-409c-994a-4a49add56094.mp4

Trott:
Hi, @MaledongGit. What you wrote here about the background color is correct, of course. In addition, I was talking about the highlight colors, which are also different. In the about section, `const` is a different color than in the docs, for example. Same for `require`. And so on.

ghost:
@Trott: That's because in Node.js org, we use something like:

```
```javascript
```
However in Node, we use:

```
```cjs
```

Maybe this will be wholly replaced.

ghost:
There're still some problems needing discussions:

1. The background color is somehow using "cjs", however it seems we CANNOT directly use it in the Node.js org (there'll be an error thrown out because prismjs doesn't support it for now). I don't know where to find "prism-cjs.js"?

2. When the browser's window width's getting smaller and smaller but the nav menu doesn't change to a new line, how to cope with it? Put it directly under the menu or ……?

![image](https://user-images.githubusercontent.com/52018749/136686666-df51ec4d-ce03-43ab-a027-ea6064bf34d3.png)

NriotHrreion:
> There're still some problems needing discussions:
> 
> 1. The background color is somehow using "cjs", however it seems we CANNOT directly use it in the Node.js org (there'll be an error thrown out because prismjs doesn't support it for now). I don't know where to find "prism-cjs.js"?
> 2. When the browser's window width's getting smaller and smaller but the nav menu doesn't change to a new line, how to cope with it? Put it directly under the menu or ……?
> 
> ![image](https://user-images.githubusercontent.com/52018749/136686666-df51ec4d-ce03-43ab-a027-ea6064bf34d3.png)

This is what I think:

[...]

PR #4338, opened 2021-12-30
fix position of language picker and dark/light mode switch buttons

thomscoder:
Fix of buttons over text. The issue was also mentioned in #4166 
Issue:
Schermata 2021-12-30 alle 12 30 26

Issue was cause by some position absolute not relative

Fix:
Schermata 2021-12-30 alle 12 29 08

fix consists in wrapping element inside a div to have a bit more control over the pair and some media queries.


thomscoder:
Just noticed I opened the branch from a non merged one #4336 . I don't know if it's better to close this and open a new PR or checking both at same time.

PR #4340, opened 2021-12-31
Azure Cli

MarcusSaviour:
az vm create \
    --resource-group myResourceGroup \
    --name myVM \
    --image Win2019Datacenter \
    --public-ip-sku Standard \
    --admin-username azureuser


Trott:
I've blocked this user from the org.

PR #4344, opened 2022-01-04
doc: jan-2022 sec release pre-announcement

bengl:


mhdawson (approved):
LGTM

locale/en/blog/vulnerability/jan-2022-security-releases.md
```diff
@@ -0,0 +1,35 @@
...
+
+The 12.x release line of Node.js is vulnerable to three medium severity issues.
+
+## Release timing
+
+Releases will be available on, or shortly after, Monday, January 10th, 2021.
+
+## Contact and future updates
+
+The current Node.js security policy can be found at https://nodejs.org/en/security/. Please follow the process outlined in https://github.com/nodejs/node/blob/master/SECURITY.md if you wish to report a vulnerability in Node.js.
```
mhdawson:
```suggestion
The current Node.js security policy can be found at https://github.com/nodejs/node/blob/HEAD/SECURITY.md#security. 
Please follow the process outlined in 
```

mhdawson:
Pushed a few commits trying to fixup the link in the standard boilerplate that was removed since the last security release

nschonni:
I think the change to the linkanator action in https://github.com/nodejs/nodejs.org/pull/4309 is causing the build failure rather than an actual issue

mhdawson:
@nschonni what would you suggest we do ?

locale/en/site.json
```diff
@@ -163,10 +163,10 @@
   },
   "banners": {
     "index": {
-      "startDate": "2021-11-30T00:00:00.000Z",
-      "endDate": "2021-12-07T00:00:00.000Z",
-      "text": "Node.js training and exams discounted through Dec. 6th",
-      "link": "https://openjsf.org/blog/2021/11/29/test-your-skills-how-good-are-you-with-node-js/"
+      "startDate": "2022-01-04T16:00:00.000Z",
+      "endDate": "2022-02-04T16:00:00.000Z",
```
Trott:
```suggestion
      "endDate": "2022-01-25T16:00:00.000Z",
```

locale/en/site.json
```diff
@@ -163,10 +163,10 @@
...
+      "text": "New security releases to be made available January 10th, 2022",
+      "link": "blog/vulnerability/jan-2022-security-releases/"
```
Trott:
```suggestion
      "link": "en/blog/vulnerability/jan-2022-security-releases/"
```

tniessen (approved):

Trott:
> @nschonni what would you suggest we do ?

https://github.com/nodejs/nodejs.org/pull/4344/files#r778403894 (adding the `en/` locale prefix to the link) should fix it, I think. I'll commit that now and we'll see.

mhdawson:
@Trott thanks for trying that. Since I don't think the link was necessarily broken before I was not sure that updating the text would not break something else.

mhdawson:
And I see it's not obvious how to actually fix it...

nschonni:
Looking at https://github.com/nodejs/nodejs.org/commit/a800a29f4fbb288175aa45f510b8b2b53b2c1f75#diff-25e01f5bf6e807a24898f24cc295bb5a83a84540f8709b82187987b09a0e62c1 I don't think the locale is usually included

mhdawson:
@nschonni  +1, I'd validated that the PR was using the same format as before.

Trott:
> @Trott thanks for trying that. Since I don't think the link was necessarily broken before I was not sure that updating the text would not break something else.

It was broken when I built the site from this PR and the change I made fixed it, but maybe it fixed it in one place and broke it in another? I'll keep looking....

Trott:
OK, I think I see what's going on. On the live site. nodejs.org redirects to nodejs.org/en and the link works. However in the test site, that redirect doesn't happen, so without the /en, it's broken. Linkinator is probably testing both pages, so no matter which you do, you get a broken link. That's my guess, anyway.

locale/en/site.json
```diff
@@ -163,10 +163,10 @@
...
   "banners": {
     "index": {
-      "startDate": "2021-11-30T00:00:00.000Z",
-      "endDate": "2021-12-07T00:00:00.000Z",
-      "text": "Node.js training and exams discounted through Dec. 6th",
-      "link": "https://openjsf.org/blog/2021/11/29/test-your-skills-how-good-are-you-with-node-js/"
+      "startDate": "2022-01-04T16:00:00.000Z",
+      "endDate": "2022-01-25T16:00:00.000Z",
+      "text": "New security releases to be made available January 10th, 2022",
+      "link": "en/blog/vulnerability/jan-2022-security-releases/"
```
Trott:
```suggestion
      "link": "blog/vulnerability/jan-2022-security-releases/"
```

mhdawson:
Going to try the full link. May not be the right answer but want to see if it works

Trott:
> Going to try the full link. May not be the right answer but want to see if it works

I'd guess that will almost certainly fail CI because you'll be running on a weird port or something like that.

locale/en/site.json
```diff
@@ -163,10 +163,10 @@
...
   "banners": {
     "index": {
-      "startDate": "2021-11-30T00:00:00.000Z",
-      "endDate": "2021-12-07T00:00:00.000Z",
-      "text": "Node.js training and exams discounted through Dec. 6th",
-      "link": "https://openjsf.org/blog/2021/11/29/test-your-skills-how-good-are-you-with-node-js/"
+      "startDate": "2022-01-04T16:00:00.000Z",
+      "endDate": "2022-01-25T16:00:00.000Z",
+      "text": "New security releases to be made available January 10th, 2022",
+      "link": "blog/vulnerability/jan-2022-security-releases/"
```
mhdawson:
```suggestion
      "link": "https://nodejs.org/en/blog/vulnerability/jan-2022-security-releases/"
```

Trott:
I think the most expedient workaround for now might be to disable linkinator, which I'll do in a commit here and then open an issue.

Trott:
Wait! I'm totally wrong. It's finding a genuinely broken link.

mhdawson:
The full link seems to pass the previous failing job. It may not be the right answer for supporting other languages, but I'm going to suggest we land this way and open an issue to fixup the linkinator separately.

Trott:
It's finding ` <link rel="canonical" href="https://nodejs.org/en/blog/vulnerability/jan-2022-security-releases/">` in the rendered HTML for the blog post, but that link doesn't yet exist. It will as soon as this lands. So we'll need to find a linkinator workaround for such things. In the meantime, disabling linkinator....

Trott:
And we're not the first to have this issue: https://github.com/JustinBeckwith/linkinator/issues/175

mhdawson:
@Trott putting the full link also seemed to work so we can either do that or disable the linkinator, whichever you think it better.

Trott:
> @Trott putting the full link also seemed to work so we can either do that or disable the linkinator, whichever you think it better.

Well now I'm completely mystified.

mhdawson:
@Trott I think what is supposed to happen is that site.json has a partial link and then the build process fills in the rest of the link including the locale.  However, full links are also supported so the full link should still work when the site is published for the English locale at least and the linkinator does not flag it.

Trott:
> and the linkinator does not flag it.

But it's the same link linkinator is flagging otherwise. But whatever, it worked, let's land this thing. I can figure out the gap in my understanding some other time.

PR #4400, opened 2022-02-01
locale/sr-initial commit

lxmarinkovic:
Hi @Trott , here is a pull request with the initial commit. The rest of docs will fix over the weekend


Trott:
> Hi @Trott , here is a pull request with the initial commit. The rest of docs will fix over the weekend

Please stop tagging me on everything. There are lots of other people who review and maintain stuff in this repo. Thanks!

Trott (changes requested):
The package-lock update seems unrelated. Also, please don't copy in English language templates. If the template isn't there, then the English one will be used anyway. So don't copy them in. Just let them not be there. If you copy it in, then updates in the English template won't show up in the translation results.

@nodejs/website Do we need to do anything special to make the new language appear in the options for the internationalization menu? Or does it happen automagically?

PR #4428, opened 2022-02-18
Add automatic light/dark theme switching based on prefers-color-scheme if the user has not explicitly asked for a static theme

nobelharvards:
Currently, the website's initial theme is set based on the user's `prefers-color-scheme`.

This value is fixed when returning to the website. It does not bother to check `prefers-color-scheme` again to see if the user might have changed their OS/browser level theme setting unless the cookies/cache are wiped.

So if a user has their OS/browser theme set to dark, the website will match their setting with a dark theme on their first visit, but if they were to switch their OS/browser theme to light (either on the spot or on a return visit), the website would remain dark.

This may be desirable if the user has explicitly asked for a static theme by clicking on the toggle theme button, but it otherwise feels like a missed opportunity to not have the website dynamically follow the OS/browser level theme via `prefers-color-scheme`.

This change allows for automatic light/dark theme based on the user's `prefers-color-scheme` if the user has not explicitly asked for a static theme by clicking on the toggle theme button. If the user asks for a static theme, the theme will no longer change when `prefers-color-scheme` changes.

This will not work when JavaScript is disabled. That will require CSS media queries.


Trott:
@nodejs/website

DerekNonGeneric (approved):

osk2 (approved):

static/js/themeSwitcher.js
```diff
@@ -12,7 +12,6 @@
 
   function setTheme(theme) {
     document.querySelector('html').setAttribute(themeAttr, theme);
-    window.localStorage.setItem('theme', theme);
```
Trott, line 15: I would expect with this line removed that the last selected theme color would no longer be remembered by the browser, but a quick test (assuming I did it correctly) shows that it is in fact being remembered. Is localstorage being set somewhere else and this line is a duplicate or something?

nobelharvards: There is another `window.localStorage.setItem('theme', theme);` in setTheme in main.js.

`setTheme` and `getTheme` in `themeSwitcher.js` are called when the page is first loaded.

There are separate setTheme and getTheme functions in main.js that are called when the theme toggle is clicked.

You can verify this behaviour by putting logging statements in the set and get theme functions in both files. setTheme in themeSwitcher.js is called once on loading, `setTheme` in main.js is called when the theme toggle is clicked.

So if someone visits for the first time or clears cookies/cache, the initial theme will match their OS/browser theme, but with this change, that preference won't be stored. The event listener will also change the theme without having to refresh the page.

Once someone clicks the theme toggle button, however, the theme preference will be saved and the theme will become static.

Before this change, the website would set its initial theme based on `prefers-color-scheme`, but it stores that preference and does not check `prefers-color-scheme` again, resulting in a static theme without the user explicitly asking for it.

This was likely written by someone who has their OS/browser set to always light or always dark. This change will accomodate people who schedule their OS/browser to switch light/dark theme based on time of day or just switch system theme regularly for other reasons.

Trott (approved):

Trott:
Thanks for the contribution! 🎉

PR #4451, opened 2022-03-01
Fixed typo (bug in code example)

ackinc:


Trott:
@nodejs/diagnostics

RafaelGSS (changes requested):
The current code is correct. It aims to return `400` if the user already exists, otherwise, it creates.

nschonni:
I think this has come up and reverted back and forth a few times. It might be worth having a code comment in the sample to clear up the confusion

ackinc:
> The current code is correct. It aims to return `400` if the user already exists, otherwise, it creates.

Not sure I agree, @RafaelGSS . Why would the handler of the `/auth` route, whose goal is to check whether a supplied password belongs to the specified user or not, return a 400 if the user exists?

In fact, just a few lines below my change, the handler refers to `user[username]`. This is unreachable code if the handler returns 400 if `user[username]` already exists.

Perhaps you're thinking of the `/newUser` route, which is defined elsewhere in the same file.

RafaelGSS (approved):
LGTM

RafaelGSS:
> In fact, just a few lines below my change, the handler refers to user[username]. This is unreachable code if the handler returns 400 if user[username] already exists.

You right. I was seeing another handler.

benjamingr (dismissed):

Trott (approved):

Issue #4361, opened 2022-01-11
Fix/re-add Run HTML validator


mhdawson:
The HTML validator check was removed in https://github.com/nodejs/nodejs.org/commit/1a9b04876fd03c6e9d0cccb8b5df717c3c7ffb7c.

The problem appears to be that  the website cannot handle multiple consecutive posts that contain the exact same headings in the post. See https://github.com/nodejs/nodejs.org/pull/4356#issuecomment-1009480699

mhdawson:
@nodejs/website 

I'll note that having the same headings in multiple posts is a regular thing for security releases so maybe there was some additional checking added recently?

mhdawson:
@nschonni FYI since you likely have the most insight into the recent changes to the vnu-jar.js which is what I think is used for that check.

tniessen:
I think the check itself is not the problem. The generated HTML should not contain the same IDs.

nschonni:
Problem code
```html
      <ul class="blog-index">
          <li>
            <time datetime="2022-01-10T19:27:39-0500">10 Jan</time>
            <a href="/en/blog/release/v16.13.2/">Node v16.13.2 (LTS)</a>

              <div class="summary">
                <h4 id="header-improper-handling-of-uri-subject-alternative-names-medium-cve-2021-44531">Improper handling of URI Subject Alternative Names (Medium)(CVE-2021-44531)<a id="improper-handling-of-uri-subject-alternative-names-medium-cve-2021-44531" class="anchor" href="#improper-handling-of-uri-subject-alternative-names-medium-cve-2021-44531" aria-labelledby="header-improper-handling-of-uri-subject-alternative-names-medium-cve-2021-44531"></a></h4><p><a href="/en/blog/release/v16.13.2/">Read more...</a></p>
              </div>
          </li>
          <li>
            <time datetime="2022-01-10T19:02:52-0500">10 Jan</time>
            <a href="/en/blog/release/v14.18.3/">Node v14.18.3 (LTS)</a>

              <div class="summary">
[...]

PR #4469, opened 2022-03-06
remove summaries

Trott:
Because our blog posts are mostly only release announcements trying
to have summaries probably isn't that useful. The summaries also
sometimes get misrendered in small ways (as is the case for the
latest-as-of-this-writing summary for the 17.6.0 release).


Trott:
Before: 
image

After: 
image

Trott:
I'm on the fence with this one. I like that it simplifies our build pipeline a bit. I think the pages *look* better with the summaries, but I'm not sure they are any better functionally. Maybe they are.

DerekNonGeneric:
Please leave the commit hashes.

Trott:
> Please leave the commit hashes.

Why?

DerekNonGeneric:
They look better imo, but do as you wish.

It seems nicer for people who want to see what went on, but if that's also in the individual blog posts themselves, then I don't really see a problem.

nschonni (approved):
This will also be able to fix https://github.com/nodejs/nodejs.org/issues/4361

DerekNonGeneric:
> This will also be able to fix https://github.com/nodejs/nodejs.org/issues/4361

Welp, maybe @Trott wants to add [fixes syntax](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue) to the commit message? Or is that bug more complicated?

Trott:
> Welp, maybe @Trott wants to add [fixes syntax](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue) to the commit message? Or is that bug more complicated?

This won't quite fix it. Once this lands, it can be fixed with a subsequent PR that re-enables link-checking.

DerekNonGeneric:
> Once this lands, it can be fixed with a subsequent PR that re-enables link-checking.

I must have missed when link-checking was disabled/removed (guess that was being handled by [`vnu-jar`](https://www.npmjs.com/package/vnu-jar)), but would imagine that that would not be too difficult to add as a follow-on PR and would gladly make such a PR to help move this along.

<details>

As you likely know, nowadays, there are so many ways people do link-checking, but not sure what would be ideal for this particular repository. I have seen some really good ways of doing it, and the way that I liked the most would probably not be appropriate for this repository since it involves using [Perl](https://www.perl.org/) for pre-deployment checking or the links [w3c/link-checker](https://github.com/w3c/link-checker) &mdash; feel free to read up more about it at on [their documentation page](https://dev.w3.org/perl/modules/W3C/LinkChecker/docs/checklink).

I have not done extensive research into its usage, but maybe its online service version may be a potentially good solution (it seems to not be functional at the moment).

A cursory look at other issues in this repo regarding link-checking only turned up #2865 &mdash; I also found a [Metalsmith Plugin](https://www.npmjs.com/package/@fidian/metalsmith-broken-link-checker) recently published to npm that could be a viable solution.

</details>

I wonder how it should be done here? Are there better ideas that have been discussed in the past that would help move this PR along? I have not found any others.

Trott:
> I wonder how it should be done here? Are there better ideas that have been discussed in the past that would help move this PR along? I have not found any others.

I believe we used linkinator in a GitHub Action workflow.

PR #4483, opened 2022-03-13
Remove call to deprecated function util.inherits

louwers:
### Affected URL(s)

https://nodejs.org/en/docs/guides/event-loop-timers-and-nexttick

### Description of the problem

On the bottom of this help page, `util.inherits` is used.

This function is [deprecated](https://nodejs.org/docs/latest/api/util.html#utilinheritsconstructor-superconstructor). I think the docs should set an example and not use deprecated functionality! The example should be updated with `class` based syntax.


XhmikosR (approved):

ghost (approved):

ghost:
@nodejs/translators : This file has been updated, please do translations when possible :)

PR #4497, opened 2022-03-17
doc: post-release announcement Mar 2022 OpenSSL Updates

joesepi:
~DO NOT MERGE UNTIL RELEASES ARE OUT~

Discussed in private with @richardlau and @mhdawson in preparation for the releases.

### DIFF
--- a/locale/en/blog/vulnerability/mar-2022-security-releases.md
+++ b/locale/en/blog/vulnerability/mar-2022-security-releases.md
@@ -19,7 +19,7 @@ More details are available at https://www.openssl.org/news/secadv/20220315.txt
 Impacts:
 * All versions of the 17.x, 16.x, 14.x, and 12.x releases lines.
 
-# Downloads and release details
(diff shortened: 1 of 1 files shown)


locale/en/blog/vulnerability/mar-2022-security-releases.md
```diff
@@ -1,19 +1,40 @@
 ---
-date: 2022-03-16T23:22:00.000Z
+date: 2022-03-18T09:00:00.000Z
```
richardlau: I can't remember if we ran into issues if the date/time was in the future, so we might want to check this value before merging.

joesepi: Yup. Was just about to update it to be more accurate. Thanks!

richardlau (approved):

richardlau:
I believe all the release blog posts are live on the website now. So we should be good to go with this.

PR #4517, opened 2022-03-26
improved grammar

chriskorosu:
Changed comma to parentheses to make it clear that what previously followed the comma is a side-note, not a separate statement with equal weight

Changed "includes" to "also applies to" to remove grammatical ambiguity which allowed for the following interpretation: "strings can contain object properties"

### DIFF
--- a/locale/en/knowledge/javascript-conventions/what-is-json.md
+++ b/locale/en/knowledge/javascript-conventions/what-is-json.md
@@ -65,7 +65,7 @@ There are a few rules to remember when dealing with data in JSON
 format. There are several gotchas that can produce invalid JSON as well.
 
 * Empty objects and arrays are okay
-* Strings can contain any unicode character, this includes object properties
+* Strings can contain any unicode character (this also applies to object properties)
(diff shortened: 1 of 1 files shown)


locale/en/knowledge/javascript-conventions/what-is-json.md
```diff
@@ -65,7 +65,7 @@ There are a few rules to remember when dealing with data in JSON
 format. There are several gotchas that can produce invalid JSON as well.
 
 * Empty objects and arrays are okay
-* Strings can contain any unicode character, this includes object properties
+* Strings can contain any unicode character (this also applies to object properties)
```
Trott, line 68:
```suggestion
* Strings can contain any unicode character. (This also applies to object properties.)
```

Trott:
I suspect we should remove this article entirely and replace it with a link to https://developer.mozilla.org/en-US/docs/Learn/JavaScript/Objects/JSON. We really don't need to maintain a JSON tutorial. That's outside of the scope of the website.

Trott (approved):

PR #4529, opened 2022-04-06
Allow inexact filtering and filtering by LTS

tchetwin:
Right now it is not possible to filter by, e.g. `16`, or `Gallium`.
This would be useful to take a quick glance at just a single major release.
The `"auto"` setting allows for exact matches to be performed by quoting the input, e.g. `"Node.js 16.13.2"`.

### DIFF
--- a/static/js/previousVersion.js
+++ b/static/js/previousVersion.js
@@ -5,7 +5,7 @@ $(function () {
     pagination: true,
     perPage: 10,
     globalSearch: true,
-    exactMatch: "auto",
+    exactMatch: 'auto',
     globalSearchExcludeColumns: [3, 4, 6, 7],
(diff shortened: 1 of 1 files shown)


static/js/previousVersion.js
```diff
@@ -5,8 +5,8 @@ $(function () {
     pagination: true,
     perPage: 10,
     globalSearch: true,
-    exactMatch: true,
-    globalSearchExcludeColumns: [2, 3, 4, 6, 7],
+    exactMatch: "auto",
```
Trott:
```suggestion
    exactMatch: 'auto',
```

tchetwin: 🙇‍♂️ whoops!

Trott (approved):

Trott:
Thanks for the fix!

PR #4543, opened 2022-04-11
Update trademark notice

brianwarner:
Cleaning up an older version of the trademark policy to reflect the transfer of the Node.js marks to the foundation.

Signed-off-by: Brian Warner <brian@bdwarner.com>


Trott (approved):
Should there be a comma before “don’t hesitate”?

PR #4545, opened 2022-04-16
fix: add Node.js 16.x and 17.x to generated version link list

austinkelleher:
`source/nodejsVersions.md` is used to generate a list of version links
on each locale page. Node.js 16.x and 17.x were missing.

**OLD**: https://nodejs.org/en/docs/#looking-for-api-docs-of-previous-releases

**NEW**:

Screen Shot 2022-04-16 at 3 04 38 PM


Trott:
16.x and 17.x are missing because those are the current versions already linked on the left nav and everywhere else.

austinkelleher:
@Trott Thanks for the feedback. This makes sense to me. How about I close this PR and open a separate PR that adds a comment in `nodejsVersions.md` which will call out the detail that you highlighted?

austinkelleher:
@Trott Fwiw, I think it would be nice if the current versions showed up in the list alongside the other versions instead of having the distinction in the sidebar. It may not be the best phrasing currently though: `"Looking for API docs of previous releases?`"...

I was going to recommend updating the phrasing to something that is less specific to other versions, but then the other locale docs would need to be updated 😅.

Trott:
@austinkelleher I'm fine with either of the options you suggest. I was just explaining why those two versions don't currently appear in there. Let's do whatever would be the best approach for end users.

source/nodejsVersions.md
```diff
@@ -3,6 +3,8 @@
  'locale' folder at 'docs/index.md' to replace the variable
  in order to avoid missing adding Node.js links.
 -->
+* [Node.js 17.x](https://nodejs.org/docs/latest-v17.x/api/)
+* [Node.js 16.x](https://nodejs.org/docs/latest-v16.x/api/)
```
nschonni: Now that 18 is out, we need the 17 link, but the 16 is still in the side bar
```suggestion
```

austinkelleher:
Thanks for the update @nschonni!

nschonni (approved):

nschonni:
Thanks @austinkelleher if you want to take more of a generated list approach/helper template, that might be another option to not need to bump this in the future

austinkelleher:
@nschonni Good idea and I'm happy to work on this! 🔨 

I think we should be able to leverage the information that is fetched from `node-version-data` and generate the major version info that, today, we are storing in `nodejsVersions.md`. https://github.com/nodejs/nodejs.org/blob/main/scripts/load-versions.js#L6

How does that sound to you?

nschonni:
Sure, I think it might be a helper to put in https://github.com/nodejs/nodejs.org/tree/main/scripts/helpers and update the locale files too

PR #4553, opened 2022-04-19
Blog: Node.js 18 release announcement and post

BethGriggs:


richardlau (approved):

mhdawson (approved):
LGTM

pratikpc:
> Prebuilt binaries for 32-bit Windows will initially not be available due to issues building the V8 dependency in Node.js. We hope to restore 32-bit Windows binaries for Node.js 18 with a future V8 update.

If this is the case, would removing the x86 download links be a good idea from the website temporarily?

https://nodejs.org/en/download/current/ does not show this warning

PR #4560, opened 2022-04-24
Replacing a broken link

neo-clj:
Hi.
The link that I replaced was broken; because mxr.mozilla.org doesn't exist anymore.
Here are some more general pages that lead to the link I added to this document:
https://wiki.mozilla.org/CA/Included_Certificates
https://wiki.mozilla.org/CA/FAQ
https://wiki.mozilla.org/CA/Included_CAs


Trott (commented):
I don't think using a Salesforce page for a list Mozilla keeps is the way to go, although I won't block this or anything because it's better than having a broken link. But I think the entire paragraph and much other material should be removed or replaced with a link. We should not be trying to offer a tutorial on what a TLS connection is. Omit it or else refer to another resource like https://developer.mozilla.org/en-US/docs/Web/Security/Transport_Layer_Security or https://en.wikipedia.org/wiki/Transport_Layer_Security.

Trott:
> I don't think using a Salesforce page for a list Mozilla keeps is the way to go,

Although as you point out, Mozilla itself links there, so I guess it's OK. Still, I think the material probably should be removed entirely.

Trott:
@nodejs/crypto Thoughts on removing this document entirely or pointing it somewhere else? I get the impression that these things aren't kept very up to date in general.

Trott (approved):

Issue #4392, opened 2022-01-31
doc command for Fedora installation with different version seem to be wrong


option-greek:
- **URL**: https://nodejs.org/en/download/package-manager/#centos-fedora-and-red-hat-enterprise-linux
- **Browser version**: Firefox
- **Operating system**: Fedora

To install nodejs version 12 on fedora, the page says to run:

`dnf module install nodejs:12`

If I try the same with versions 14 or 16, dnf  throws an error saying:

`No default profiles for module nodejs:14. Available profiles: default, development, minimal`

The correct command that worked as referenced on this website (https://linuxconfig.org/how-to-install-node-js-on-redhat-8-linux)  is:

`dnf module install nodejs:16/default`

richardlau:
There’s an open PR, https://github.com/nodejs/nodejs.org/pull/4279, although there appears to be an inconsistency on Fedora when compared to CentOS/RHEL.

option-greek:
I guess its because some one stopped marking the default profiles for  fedora starting with release 14 of NodeJS. I'm not sure that particular lack of default profile marking is on NodeJS side or Fedora side and where to correct it. 

But as it stands, installing NodeJS > 12 is broken on Fedora currently with the default instructions provided on nodejs.org. May be this can be left as a note  on the website to prevent needless googling ? (it works from 12-16 so there is no loss of function when explicitely specifying the profile name). 

Since the behavior is different in CentOS/RHEL and looks like CentOS is dead/dying (with stream being the replacement), splitting Fedora's documentation in nodejs.org website will be nice.

kasicka:
@option-greek Thanks for bringing this up, we'll try to fix this in Fedora and then I'll review the current documentation.

kasicka:
Update: defaults were fixed/changed to `common`, I'll get back to this issue when the changes show up in repos for F34 and F35.

mhdawson:
Looks like some of the updates are in F35

```shell 
[root@2d21e2dd9e4a /]# dnf module list nodejs
Fedora 35 - x86_64                                                                                              16 MB/s |  79 MB     00:05    
Fedora 35 openh264 (From Cisco) - x86_64                                                                       3.4 kB/s | 2.5 kB     00:00    
Fedora Modular 35 - x86_64                                                                                     3.6 MB/s | 3.3 MB     00:00    
Fedora 35 - x86_64 - Updates                                                                                   9.8 MB/s |  28 MB     00:02    
[...]

PR #4566, opened 2022-04-25
add a note for RHEL when install nodejs > 12

sakurayang:
https://bugzilla.redhat.com/show_bug.cgi?id=1728643

---
1728643 – NodeJS stream does not have a default profile set


ghost (commented):
Is this problem related to Node.js instead of Linux system? If true, maybe it should be sent to Node Bug feed?

sakurayang:
i think it's a problem of software distributor.
according to [this page](https://docs.fedoraproject.org/en-US/modularity/glossary/#_default_module_stream)

> Default Module Stream is a Module Stream pre-selected by the software distributor

i dont really know who distributes it, so i dont know if i should report it as a bug.

Churru14 (commented):
6091dbd989c849ed8bfbc6933272426f31e43a1f

locale/en/download/package-manager.md
```diff
@@ -92,6 +92,8 @@ dnf module install nodejs:12
 
 For CentOS/RHEL 7 Node.js is available via [Software Collections](https://www.softwarecollections.org/en/scls/?search=NodeJS).
 
+For Node.js > 12 is not available by this way because [this](https://bugzilla.redhat.com/show_bug.cgi?id=1728643) bug.
```
richardlau, line 95: Default profiles (named `common`) were added to Node.js 14 and 16: https://github.com/nodejs/nodejs.org/issues/4392

Is that not working for you?

If you are referring to Software Collections then the referenced bugzilla is irrelevant (https://bugzilla.redhat.com/show_bug.cgi?id=1728643 refers to modules streams which are are RHEL 8 feature). My understanding is Node.js versions >12 are not available on Software Collections because they build Node.js dynamically linked to OpenSSL and the version of OpenSSL in RHEL 7 is too old (OpenSSL 1.0.2).

mhdawson: +1 to what @richardlau says you can install Node.js 12, 14, and 16 on Fedora/CentOS/RHEL on RHEL 8 and 12 and 14 on RHEL 7.

Can you share the sequence of command lines that are not working for you and which version of the OS you are using.

sakurayang: it can install but must supply profile named `default`.
so maybe we should change the command instead of add a note ?
pastebin [here](https://pastebin.com/iJsB9w0a)
my os is Fedora 32.

mhdawson: @sakurayang I think the issue is that you are using Fedora 32. Based on an earlier issue I believe we have fixed it in later versions (ie with later versions, not sure but possibly 34 or later) you won't need to specify default.

sakurayang: okay, I test again on both Fedora 32 and 33 with the same problem. But it solved in 35. 34 did not have default profile for nodejs16.
The reason I do this PR is that I found the problem unresolved when I followed #4392     .
Sorry for taking up your time.

mhdawson: @sakurayang no problem hopefully as time moves forward all of the Fedora versions in people reach for will have the fix. Going to close the issue.

PR #4575, opened 2022-04-27
fix: broken security link

sanamhub:


nschonni:
These links aren't broken, https://github.com/nodejs/node/blob/HEAD/SECURITY.md#security works

sanamhub:
> These links aren't broken, https://github.com/nodejs/node/blob/HEAD/SECURITY.md#security works

Did you fixed in repo or what just happened? That link wasn't working 5 minute before.

nschonni:
Maybe you hit GitHub's Unicorn error page or something, but `HEAD` alias's to the default branch and there was no change

PR #4578, opened 2022-05-01
 chore: update links and ignore some json files

ghost:
1. 'shedule.json' and 'versions.json' links are out of date, this
is the fixture for it.

2. ignore these files, because they are only for the current building.

### DIFF
--- a/layouts/about-release-schedule.hbs
+++ b/layouts/about-release-schedule.hbs
@@ -14,7 +14,7 @@
         {{{ contents }}}
 
         <p>
-          <img src="https://github.com/nodejs/Release/blob/main/schedule.svg" alt="{{ title }}" width="760" height="396">
+          <img src="https://raw.githubusercontent.com/nodejs/Release/master/schedule.svg?sanitize=true" alt="{{ title }}" width="760" height="396">
         </p>
 
         {{{ release-schedule }}}
--- a/layouts/about-release-schedule.hbs
+++ b/layouts/about-release-schedule.hbs
@@ -14,7 +14,7 @@
         {{{ contents }}}
 
         <p>
-          <img src="https://raw.githubusercontent.com/nodejs/Release/master/schedule.svg?sanitize=true" alt="{{ title }}" width="760" height="396">
+          <img src="https://github.com/nodejs/Release/blob/main/schedule.svg" alt="{{ title }}" width="760" height="396">
         </p>
 
         {{{ release-schedule }}}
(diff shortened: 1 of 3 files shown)


ghost:
@nodejs/releasers : Please have a review and do we really need to keep these files here?

nschonni:
AFAIK, they're kept so an offline build will still work, but I don't think these updates are needed

richardlau:
> 1. 'schedule.json' is out of date, generated using https://github.com/nodejs/lts-schedule.

This is not correct. `schedule.json` is a copy of https://github.com/nodejs/Release/blob/main/schedule.json: https://github.com/nodejs/nodejs.org/blob/3cf2aeb4f66ab6f972d69c7fd585396871276a03/package.json#L14
The URL in `package.json` could be updated to use the `main` branch of the release repository.

FWIW https://github.com/nodejs/lts-schedule is used to generate the SVG file, https://github.com/nodejs/Release/blob/main/schedule.svg, referenced in 
https://github.com/nodejs/nodejs.org/blob/f2285a6beea2b9808fa22efce8bab8938eea5d9a/layouts/about-release-schedule.hbs#L17
Again the URL could be updated to use `main`.




> @nodejs/releasers : Please have a review and do we really need to keep these files here?

That's completely up to the website team to decide. The canonical schedule for 1. in the https://github.com/nodejs/Release repository and the versions table for 2. is https://nodejs.org/download/release/index.json. As @nschonni points out having copies in this repository could enable offline builds at the expense of currency although I don't know if the website build would not just error out if it cannot fetch the current versions of those files from the web.

MaledongGit:
@richardlau: I'll re-updated and fix these issues, thank you for your guidance.

layouts/about-release-schedule.hbs
```diff
@@ -14,7 +14,7 @@
         {{{ contents }}}
 
         <p>
-          <img src="https://raw.githubusercontent.com/nodejs/Release/master/schedule.svg?sanitize=true" alt="{{ title }}" width="760" height="396">
+          <img src="https://github.com/nodejs/Release/blob/main/schedule.svg" alt="{{ title }}" width="760" height="396">
```
nschonni: I don't think you can just reference the URL if you want it to display an image. I think it needs to use the githubusercontent URL. I the comment was about changing the branch part

ghost: Oops……fixed, sorry.

PR #4611, opened 2022-05-23
Add list of courses about Node.js

skillcombo:
Helpful resource for community


Trott:
Thanks, but I think it would be far better to keep this list very, very short.

PR #4635, opened 2022-06-04
Update README.md

saintgabriel0:
Corrections of misspelled words

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,9 +6,9 @@
 
 ## What is this repo?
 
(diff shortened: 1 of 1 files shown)


nschonni:
Sorry, but most of this are incorrect changes that would break command line statements or don't improve the existing text

Trott:
It probably is far enough in the past that all the text about the io.js merger and things like that can be removed.

Issue #4688, opened 2022-06-28
Changelog links point to the old branch (master) instead of main, causing a redirect and blue banner


Schweinepriester:
- **URL**: https://nodejs.org/
- **Browser version**: Any
- **Operating system**: Any

![image](https://user-images.githubusercontent.com/2644614/176305535-d50736bf-01c5-46bf-a592-c817fb316ef0.png)

Both those links

- https://github.com/nodejs/node/blob/master/doc/changelogs/CHANGELOG_V16.md#16.15.1
- https://github.com/nodejs/node/blob/master/doc/changelogs/CHANGELOG_V18.md#18.4.0

then require a redirect and trigger the distracting blue banner, for example:

![image](https://user-images.githubusercontent.com/2644614/176305777-827a5fe7-f7ce-4333-9074-103eeef4026c.png)

Looking at https://github.com/nodejs/nodejs.org/blob/main/layouts/index.hbs the props seem to be

- `{{changeloglink project.latestVersions.lts.node}}`
- `{{changeloglink project.latestVersions.current.node}}`

but I haven't found their source, otherwise I probably would have sent a PR 0:-)

nschonni:
Probably coming from https://github.com/nodejs/nodejs.org/blob/c1703267c30d2bfc19da5708afcb7169ef1a31c6/scripts/helpers/changeloglink.js#L9

nschonni:
I didn't realize that they had changed the default branch, so it's probably worth doing a larger cleanup

ghost:
A large update for these links.....

ghost:
@Schweinepriester: This is fixed at #4692 , I'll close it and for other problems, you can re-open it and feedback.

PR #4692, opened 2022-07-02
fix: update the useless links for all the related files

ghost:
Since "master" has been renamed to "main", all the pervious links should
be also renamed.

Refs: https://github.com/nodejs/nodejs.org/issues/4688.


ghost:
@nodejs/documentation : I'm not sure whether the docs under "blog" folder has a script or not to generate links with "master", but no matter what it is, I've helped to change them all for these old files. Please pay attention to the newly-generated links of blog files with such links.

Trott (approved):
Rubber-stamp LGTM. I forget if we have a link checker, but it would be good to run one.

richardlau:
> Rubber-stamp LGTM. I forget if we have a link checker, but it would be good to run one.

We had one -- I think we turned it off during a previous security release as it was blocking the announcement.

PR #4693, opened 2022-07-04
fix: Html validator checks re-added

ghost:
In order to avoid some HTML links, tags' missing, Html validator checks should be re-added here.

Refs: https://github.com/nodejs/nodejs.org/pull/4358. https://github.com/nodejs/nodejs.org/issues/4361.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -63,6 +63,9 @@ jobs:
       - name: Run unit tests
         run: npm run test:unit
 
+      - name: Run HTML validator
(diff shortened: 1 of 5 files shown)


ghost:
@mhdawson, @richardlau and @tniessen: You can have a review of this PR, for it has re-added the Html links without any other problems when checking.....It seems this is already fixed we can revert it.

willin (approved):
👍🏻

mhdawson (approved):
LGTM

PR #4719, opened 2022-07-14
Jump to the related 404 page file for the local server when running

ghost:
Now for the local server.js when it's running and you've inputted an invalid route, it will ONLY return you "Not Found" as a plain text instead of a whole HTML page, we should make it act as what we see from the official site.

Here's the fix in demo:
Demo


MaledongGit:
PS：In order to see what's wrong in https://github.com/nodejs/nodejs.org/issues/3430, the current issue should be fixed so that we can decide what to do for the next (whether it's the problem of metal-smith or Ngnix), I'll deeply see the issue....

aymen94 (approved):

PR #4752, opened 2022-07-30
fix: remove the recursive link of Node.js API docs

ghost:
Since there's a recursive link of Node.js API docs when clicking "New Issue" on GitHub of Node.js org in GitHub, we should refer the link directly to the Node.js official site instead of the website itself.

【Issue Reproduced】
微信图片_20220730175217

### DIFF
--- a/.github/ISSUE_TEMPLATE/config.yml
+++ b/.github/ISSUE_TEMPLATE/config.yml
@@ -4,5 +4,5 @@ contact_links:
     url: https://github.com/nodejs/help/issues/
     about: File an issue in our help repo.
   - name: 📗 Node.js API Docs
-    url: https://github.com/nodejs/nodejs.org/issues/new/choose
-    about: Please open an issue in the main Node.js repo, prefixed with "doc".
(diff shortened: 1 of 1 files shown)


.github/ISSUE_TEMPLATE/config.yml
```diff
@@ -3,6 +3,3 @@ contact_links:
   - name: ⁉️ Need help with Node.js?
     url: https://github.com/nodejs/help/issues/
     about: File an issue in our help repo.
-  - name: 📗 Node.js API Docs
-    url: https://github.com/nodejs/nodejs.org/issues/new/choose
```
nschonni, line 7: This should be pointed to nodejs/node repo not this one, since that is where the docs actually live.

ghost: Thank you, and I've fixed that by offering a direct link to the API Docs of Nodejs itself.

nschonni (approved):

nschonni:
I think the nodejs.dev has the same issue if you want to do the same fix there

ghost:
> I think the nodejs.dev has the same issue if you want to do the same fix there

Thank you for tipping me :)

PR #4773, opened 2022-08-16
Run linkinator against the live site.

Theo13111:
update


Trott:
User blocked based on this and their trail of similar nuisance activity on GitHub.

PR #4779, opened 2022-08-19
Website guides: highlight potential security risk with above code

lirantal:
## Reasons for the proposed change
I'm raising this pull request to better highlight potential risks with the above code. New developers who may scour through the website guides might be more prone to a copy-and-paste method of testing things out. Nothing's wrong with doing that for learning purposes, but some might be too quick to do that and skip reading through the paragraph.

I'm also happy to move that part at the beginning fo the paragraph to immediately draw attention and call it out, and then continue with the explanation.


Trott:
This change looks good to me, but honestly, I'd be OK with removing the content entirely. We don't need to have a separate knowledge base doc on how to write a web server. People generally shouldn't be rolling their own HTTP servers in Node.js to serve static files. Having this document suggests that maybe people should.

lirantal:
I'm good with either way. I presumed this was mostly about appealing to new developers to the platform and showing them how easy it is to get started.

Trott (approved):

aymen94 (approved):

PR #4792, opened 2022-08-23
feat: Use native test instead of 'tape'

ghost:
Since Node has released its own 'Test Runner', here's the upgrade for nodejs.org by using node natvie test runner as a feature experiment. This has shortened the time in the process of unit tests.

Here's the detailled info for the new "Test Runner" as a result:


ghost:
Reopen it because Nodejs version is latest.

tniessen:
Note that there are no stability guarantees for `node:test` since it is an experimental feature and not recommended for production. Unless the Node.js version used in workflows is fixed to a specific version, workflows might break when changes are made as part of new Node.js versions.

ghost:
@tniessen ：I see……I've converted it into a draft version until the released version of Nodejs is OK... Thanks.

PS：Everything seems nice now even in the experimental mode.

tniessen:
> I see……I've converted it into a draft version until the released version of Nodejs is OK... Thanks.
>
> PS：Everything seems nice now even in the experimental mode.

To clarify, I am _not_ saying that this PR should not be merged sooner. It will likely take a while until the feature is considered stable, and I myself am not opposed to large or even breaking changes to the test runner if they are considered to be an improvement. In such cases, it would be up to the maintainers of repositories such as this one to keep up with those changes. If someone can commit to doing that, I don't see a problem here.

On the plus side, projects belonging to Node.js (such as this repository) could actually be good places to test the test runner.

ghost:
OK，I'll keep it until the final release of the Test Runner comes.

tests/scripts/release-post.test.js
```diff
@@ -1,32 +1,31 @@
...
     releasePost.explicitVersion('4.1.1').then((version) => {
-      t.equal(version, '4.1.1');
-      t.end();
+      assert.equal(version, '4.1.1');
+      done();
     });
   });
 
-  t.test('rejects when given an falsy argument', (t) => {
+  await t.test('rejects when given an falsy argument', (_, done) => {
```
benhalverson: I would stick with one style of async programming 
async / await, .then, or callbacks. Not all 3 at once.

ghost: @benhalverson: Thank you and it's been fixed.

PR #4804, opened 2022-08-31
zh-CN: improve Translation

LawssssCat:
Some sentences are not easy to read in Chinese, so I slightly modified them without modifying the meaning of the sentences.

.


willin (approved):
LGTM

PR #4811, opened 2022-09-05
AK

AKIN2022:


Trott:
Users only other contribution on GitHub at this time is opening an issue in the Brave browser repo that is a blank template. I've blocked them.

PR #4820, opened 2022-09-13
remove "International" list from Get Involved

Trott:
* We don't vet any of these groups and have little idea of their quality or values.
* Calling it "International" is a very U.S./English-language-centered approach.
* Many of these groups appear to have not been active for some time.
* We can't possibly list all groups, but listing some invites everyone else to want their group listed.
* The Node.js website is simply not the place for this content.

Much of the remainder of the page needs to be removed or edited too, but I'm starting at the bottom of the page.


ovflowd (approved):
LGTM and agree on that.

PR #4821, opened 2022-09-14
refer people to release repo for release schedule

Trott:
Do not duplicate the release schedule and information on the site. Refer people to the release repo so they can see the most up-to-date and canonical information.


Trott:
This is one where I could certainly be convinced that we should get a redirect in place before landing it. There likely are people that have the releases page bookmarked solely for the release schedule. I would want to redirect them seamlessly to the schedule information on GitHub instead.

ovflowd (approved):
LGTM! Agree having this statically defined within the website isn't a good idea.

Great that on nodejs.dev we directly retrieve the release schedule from an API.

Trott:
Whoops, forgot about my note about including a redirect before landing this. Probably should have had a `blocked` label applied or something. I'll see about getting the nginx updated for a redirect.

PR #4857, opened 2022-10-03
chore: specify that Node.js is open source and cross-platform

Trott:
The fact that Node.js is open source and cross-platform is far more important to call out than that it runs on V8 so let's do that.


ovflowd (approved):
LGTM! (And agreed!)

richardlau (approved):

PR #4315, opened 2021-12-20
Add CodeQL Action

XhmikosR:
- [ ] See if everyone's OK with the cron job
- [ ] Tackle the errors

About this error: https://github.com/nodejs/nodejs.org/security/code-scanning/1?query=ref%3Arefs%2Fpull%2F4315%2Fmerge, I still do not get why we need this client side. I'm pretty sure I expressed my objection in the relevant PR, but due to lack of time, I couldn't spend more time then.

Does anybody recall why we need this client side and why we don't generate the Edit on GitHub links on build time?

EDIT: I see now it was done #3971. I still don't quite get what the issue was and why we can't fix it on build time...

PR #4867, opened 2022-10-08
remove useless js snippets of codes for "Edit On GitHub"

ghost:
We don't need to implement it through the client side but serve side only by generating them together.

Refs: https://github.com/nodejs/nodejs.org/pull/4315.

### DIFF
--- a/scripts/plugins/githubLinks.js
+++ b/scripts/plugins/githubLinks.js
@@ -17,10 +17,18 @@ function githubLinks(options) {
       const file = files[path];
       path = path.replace('.html', '.md').replace(/\\/g, '/');
       const url = `https://github.com/nodejs/nodejs.org/edit/main/locale/${options.locale}/${path}`;
-
-      const contents =
-        file.contents.toString() +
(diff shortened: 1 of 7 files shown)


Trott:
@nodejs/website PTAL

ovflowd (approved):

scripts/plugins/githubLinks.js
```diff
@@ -17,10 +17,17 @@ function githubLinks(options) {
...
-      const contents =
-        file.contents.toString() +
-        ` <input type = "hidden" id = "editOnGitHubUrl" value="${url}"/> `;
+      const editOnGitHubTrans = options.site.editOnGithub || 'Edit on GitHub';
+      const contents = file.contents
+        .toString()
+        .replace(
+          /<div class="openjsfoundation-footer-edit"><\/div>/,
+          (match, $1, $2) => {
+            return `<div class="openjsfoundation-footer-edit">
```
ovflowd: Could we put the callback somewhere else? Like separate `const`?

ghost: OK，it's done.

Issue #4874, opened 2022-10-11
Remove code+learn 


AugustinMauroy:
### Enter your suggestions in details:

Remove code+ learn page
This work is in progress on nodejs.dev

nschonni:
Not sure what this is tracking, I don't think this repo ever had this content to remove

Trott:
> Not sure what this is tracking, I don't think this repo ever had this content to remove

This: https://github.com/nodejs/nodejs.org/blob/bbe4dddb0926a62356aaab2b0b5ca7dd77fd776b/locale/en/get-involved/code-and-learn.md

nschonni:
Thanks, I wasn't seeing staring at the site nav 😆 , I see it's here https://nodejs.org/en/get-involved/code-and-learn/

HassanBahati:
@AugustinMauroy @Trott @nschonni does this mean we need to remove this file [https://github.com/nodejs/nodejs.org/blob/bbe4dddb0926a62356aaab2b0b5ca7dd77fd776b/locale/en/get-involved/code-and-learn.md](https://github.com/nodejs/nodejs.org/blob/bbe4dddb0926a62356aaab2b0b5ca7dd77fd776b/locale/en/get-involved/code-and-learn.md) that renders as [https://nodejs.org/en/get-involved/code-and-learn/](https://nodejs.org/en/get-involved/code-and-learn/) ?

nschonni:
Yes, plus the translated files, and check for any links to the page

HassanBahati:
> Yes, plus the translated files, and check for any links to the page

okay. let me get started with working on this

HassanBahati:
> Yes, plus the translated files, and check for any links to the page

@nschonni please find the submitted pull request at #4883

PR #4883, opened 2022-10-16
Remove code+learn

HassanBahati:
This is in reference to #4874 . @nschonni @Trott kindly review.

- [x] Remove all links to [https://nodejs.org/en/get-involved/code-and-learn/ ](https://nodejs.org/en/get-involved/code-and-learn/)
- [x] Remove `code-and-learn.md` file from all translations


nschonni (approved):
LGTM, I'm pretty sure that the side bar is generated off those JSON files, so this should do it

Trott:
This looks good to me. Thinking about this more, though, I wonder if the thing to do is to remove it from the navigation (the `.json` files here) but leave the `.md` file so that people reaching it from elsewhere still get the content? Otherwise, we'd probably want to figure out a redirect.

nschonni:
Some of the other stuff I think needed a redirect, but since this appears to be a list of old conference/meetup events that hasn't been updated since 2019 I think it's OK. If someone really wants to see it, they're probably OK to check it out on Archive.org

HassanBahati:
> 

@nschonni @Trott does this mean we should put back the `.md` files and only maintain removal from the `.json` files ?

AugustinMauroy (approved):
LGTM

Issue #4884, opened 2022-10-17
“Edit on GitHub” cannot point to an md file that isn't translated yet


ghost:
### URL:

https://nodejs.org/ja/about/governance/

### Browser Name:

N/A

### Browser Version:

N/A

### Operation System:

N/A

### How to reproduce the issue:

1. Open your browser and switch to the URL I've mentioned above.
2. Click "Edit on GitHub" ("GitHub 上で編集" in Japanese).
3. Expected: Directed to the English md file.
4. Actually it goes to the Page 404.

**Reason:**  Any md file that isn't translated yet will be replaced by the English version page instead.

**Solution:** Check the file path really exists or not, if not, just change the "local" to "en" instead of the current one.

Related issues about "Edit on GitHub":  https://github.com/nodejs/nodejs.org/pull/4880.

MaledongGit:
I'm fixing the problem and I'll submit a new fix here linking to the issue.

ghost:
Before fixing it I want to confirm：

1. If the file path doesn't exist in the local language, direct us to the 'en' as the default page for editing.
2. If the file path doesn't exist in the local language, "Edit on GitHub" shouldn't be seen（This means the current language page hasn't been translated yet, so there should be no URL for the file）

I wanna ask all：Which is better？

nschonni:
I would lean more towards 2, with maybe a "Help translate this page" pointing to some instructions instead. Reasoning that, a copy of the old file would need to be create, so even a link to GitHub "Add File" wouldn't be useful, and 1 would encourage people spending time to translate the EN page, which would be rejected as a PR

ghost:
> I would lean more towards 2, with maybe a "Help translate this page" pointing to some instructions instead. Reasoning that, a copy of the old file would need to be create, so even a link to GitHub "Add File" wouldn't be useful, and 1 would encourage people spending time to translate the EN page, which would be rejected as a PR

From your nice suggestions tips me：When the file doesn't exist, I'll do this：

A、Add a new file, with the name of the missing one.
B、There's a tip in the newly-created file in memory on github to guid the user there, something like this following（Let's just say there's something missing in zh-cn, the generated link may be the following, you can click to have a review:）
C、Anyone who can cancel the PR at any time by clicking the "Cancel" buttong or submitting it as a new PR.
D、So the link name is still "Edit on GitHub" with two diff ways of processing (Either add  a new file with the name, or edit the current file if it exists).

```
[...]

PR #4885, opened 2022-10-17
fix：Wrong URL to a missing translation page

ghost:
There're some files that aren't translated into the locale, which causes the "404 Page Not Found" Error when clicking "Edit on GitHub".

Now the current solution is: If the file doesn't exist, it will create a new blank file, with a link of translation guide there to guide what to do next. The user can either continue translating or abandoning it by clicking "Cancel" or "Submit".

Demo is here: https://github.com/nodejs/nodejs.org/new/main/locale/zh-cn/locale?filename=aaa.md&value=create%20your%20translation%20here,for%20more%20you%20can%20see:%20https://github.com/nodejs/nodejs.org/blob/main/TRANSLATION.md

Refs：https://github.com/nodejs/nodejs.org/issues/4884


DominusKelvin:
Great fix @MaledongGit

aymen94 (approved):

PR #4895, opened 2022-10-26
fix: remove obsolete links to Node.js Collection and CommComm

Trott:
Node.js Collection and the Community Committee are not active entities.


joesepi (approved):
LGTM. Thanks

PR #4898, opened 2022-10-28
doc: add API link for Node.js 16.x

ghost:
Since Node.js 16.x is now out of active mode, add this link into 'Node.js versions'.

### DIFF
--- a/source/nodejsVersions.md
+++ b/source/nodejsVersions.md
@@ -4,6 +4,7 @@
  in order to avoid missing adding Node.js links.
 -->
 * [Node.js 17.x](https://nodejs.org/docs/latest-v17.x/api/)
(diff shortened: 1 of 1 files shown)


ghost:
PS：I find it's getting longer and longer with the Node.js versions created.... So should we remove many useless (no longer supported versions) here to keep those useful only?

IMO：Just ref https://github.com/nodejs/release#release-schedule, so those whose versions are lower than 14.x will be removed, just to keep the latest versions of Nodejs, they are frequently used and viewed. For other versions, you can click "All Versions" to find what you want.



So the yellow-blocked versions of node.js will be removed, they are out of supoort.

Trott (approved):

PR #4900, opened 2022-10-28
doc: fix title november -> November

RafaelGSS:
as titled

### DIFF
--- a/locale/en/blog/vulnerability/openssl-november-2022.md
+++ b/locale/en/blog/vulnerability/openssl-november-2022.md
@@ -1,7 +1,7 @@
 ---
 date: 2022-10-28T19:00:01.316Z
 category: vulnerability
-title: OpenSSL november security release
+title: OpenSSL November Security Release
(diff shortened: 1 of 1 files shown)


richardlau (approved):

locale/en/blog/vulnerability/openssl-november-2022.md
```diff
@@ -1,7 +1,7 @@
 ---
 date: 2022-10-28T19:00:01.316Z
 category: vulnerability
-title: OpenSSL november security release
+title: OpenSSL November Security Release
```
Trott, line 4: Probably should be `OpenSSL November security release` [per the Microsoft Style Guide](https://learn.microsoft.com/en-us/style-guide/capitalization#sentence-style-capitalization-in-titles-and-headings) (which we've adopted so that we don't have to debate what goes into our own style guide), but also probably not worth "fixing" it. Just FYI for future reference.

PR #4909, opened 2022-11-02
Suggested update for the openssl-november-2022.md

pyotam:
Added that Node.js v17.x is also affected by the vulnerabilities. I think it is important to mention even though it is no longer supported to avoid creating a false sense of security for folks still using 17.x. This is especially true since otherwise, users won't know they have OpenSSL 3.x in their environment since it is compiled into node and hence won't be flagged by most scanners/SCA tools.

### DIFF
--- a/locale/en/blog/vulnerability/openssl-november-2022.md
+++ b/locale/en/blog/vulnerability/openssl-november-2022.md
@@ -23,8 +23,8 @@ the _highest_ severity issue is CRITICAL.
 
 [security policy](https://www.openssl.org/policies/secpolicy.html).
 
-Node.js v18.x and v19.x use OpenSSL v3.
-Therefore these release lines are impacted by this update.
+Node.js v17.x, v18.x and v19.x use OpenSSL v3.
+While Node.js v17.x is no longer supported, the v18.x and v19.x release lines are impacted by this update.
 
 Node.js 14.x and v16.x are not affected by this OpenSSL update.
 
--- a/locale/en/blog/vulnerability/openssl-november-2022.md
+++ b/locale/en/blog/vulnerability/openssl-november-2022.md
@@ -23,8 +23,11 @@ the _highest_ severity issue is CRITICAL.
 
 [security policy](https://www.openssl.org/policies/secpolicy.html).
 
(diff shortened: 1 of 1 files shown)


locale/en/blog/vulnerability/openssl-november-2022.md
```diff
@@ -23,8 +23,8 @@ the _highest_ severity issue is CRITICAL.
 
 [security policy](https://www.openssl.org/policies/secpolicy.html).
 
-Node.js v18.x and v19.x use OpenSSL v3.
-Therefore these release lines are impacted by this update.
+Node.js v17.x, v18.x and v19.x use OpenSSL v3.
+While Node.js v17.x is no longer supported, the v18.x and v19.x release lines are impacted by this update.
```
Trott:
```suggestion
Node.js v17.x, v18.x, and v19.x use OpenSSL v3.

Node.js v18.x and v19.x will be updated to address this issue.

Support for Node.js v17.x ended in June 2022. It will not be updated. Please migrate to a supported version of Node.js.
```

Trott:
@RafaelGSS What do you think?

RafaelGSS:
I'm -1 on that.

> I think it is important to mention even though it is no longer supported to avoid creating a false sense of security for folks still using 17.x.

Just the fact of being EOL should be enough to avoid using those versions, regardless of security status. If we start to assess EOL lines, people will start thinking they were _still_ maintained somehow.

Trott:
> Just the fact of being EOL should be enough to avoid using those versions, regardless of security status. If we start to assess EOL lines, people will start thinking they were _still_ maintained somehow.

Perhaps the text in here that says it's unsupported and won't get an update and to please migrate to a supported line will have the desired effect of encouraging people to move off of 17.x if they're still using it. 

I wouldn't want to do this for all release lines, but it might not be obvious to a lot of users that just because 16.x and 18.x are supported, that doesn't mean that 17.x is also supported.

That said, I don't feel strongly about this. I'm in favor, but not something I'm going to spend a lot of time trying to persuade others if they disagree.

RafaelGSS:
I'm good if you think that's a good call.

Trott (approved):

RafaelGSS (approved):

PR #4916, opened 2022-11-04
doc: nov-2022 sec release pos-announcement

juanarbol:


RafaelGSS (approved):

richardlau:
Are we updating the banner as well?

mhdawson:
@richardlau we should, I missed suggesting that in the internal review.

RafaelGSS:
I think @juanarbol is away. Since I don't have access to his branch, I've created a fix in another PR https://github.com/nodejs/nodejs.org/pull/4917

PR #4931, opened 2022-11-12
fix for codeQL

ghost:
Use 'fs.existSync' and 'fs.writeFileSync' together instead of a callback functions, because it will make the chance for file changes.

Refs: https://github.com/nodejs/nodejs.org/pull/4930.

### DIFF
--- a/scripts/release-post.js
+++ b/scripts/release-post.js
@@ -211,23 +211,21 @@ function writeToFile(results) {
   );
 
   return new Promise((resolve, reject) => {
-    fs.access(filepath, fs.F_OK, (err) => {
-      if (!err && process.argv[3] !== '--force') {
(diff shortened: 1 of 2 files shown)


XhmikosR:
I wouldn't touch external/survey-2018/full.js; instead we should just ignore the issue.

EDIT: I ignored the issue. The other fix is probably valid, but I'll leave this for someone else to review.

ghost:
> I wouldn't touch external/survey-2018/full.js; instead we should just ignore the issue.
> 
> EDIT: I ignored the issue. The other fix is probably valid, but I'll leave this for someone else to review.

OK, ignore it, too :)

ovflowd (approved):

PR #4957, opened 2022-11-30
feat: add slider for cyber week

Trott:
Given our avoidance of even basic analytics tracking (e.g., Google Analytics) on privacy grounds, I'm going to guess that this will give people pause. But maybe not because there's a GDPR-compliant cookie tracker config thingy? I don't know. Let's find out! @nodejs/website


nschonni:
Yeah, not a fan of trackers. My simple suggestion if the foundation wants some click through tracking, just added a basic query string to the banner link like `?campaign=nodejs.org`

PR #4963, opened 2022-12-09
Added flags : 'a' option to fs.createWriteStream

VatoQv:
Without flags : 'a' option, nothing is saved in output.txt file.


AugustinMauroy (commented):
I think if you want to add the Flag you have to explain what it does.

VatoQv:
Default 'flags' option for fs.createWriteStream is 'w', which opens file for writing, the file is created (if it does not exist) or truncated (if it exists). 

Since most browsers send more than one request to a server at a time, each new request, and therefore each new createWriteStream call deletes old data from 'output' file. So, if the request with the POST data from form element is not the last request (which in my experience it never is), output file is left blank, which is the not desired result.

'a' flag opens file for appending data, and nothing is being truncated. This way piped data is safe.

If this doesn't make sense or is not a solution to the problem, then I'll just close pull request.

AugustinMauroy:
Yeah, I mean add explication in the document.

AugustinMauroy:
No you don't have to close it, you can add the explanation and it will improve the tutorial. But if you are not happy with your work, you are free to close it.

AugustinMauroy (approved):
🔥 LGTM 🔥

PR #5039, opened 2023-02-02
chore: add link for nodejs Mastodon verification

Trott:
We should find the right place for this on the page but right now, I'm putting it after the footer with no link text so that we can get the official Mastodon account verified as the owner of the site.

### DIFF
--- a/layouts/partials/footer.hbs
+++ b/layouts/partials/footer.hbs
@@ -21,6 +21,9 @@
 
 </footer>
 
+<!-- Verification of official Mastodon profile. -->
+<a rel="me" href="https://social.lfx.dev/@nodejs"></a>
+
 <script src="/static/js/main.js" async defer></script>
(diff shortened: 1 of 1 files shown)


AugustinMauroy (approved):
LGTM! Should we open an issue in nodejs.dev

joesepi (approved):
LGTM!

Trott:
> LGTM! Should we open an issue in nodejs.dev

No and yes. No need to replicate what I'm doing here, which is putting a link with empty text into the site. That's only needed on the website that the account links to. In this case, that will be (or already is) nodejs.org.

However, if we want to provide a link to the Mastodon account, probably adding it wherever we already might have a link to the Twitter account, then yeah, go ahead and add that to nodejs.dev. (And if there's a place on nodejs.org, hey, we can add it there too.)

PR #5079, opened 2023-03-08
Chore(Style): introduce branding from nodejs Website Redesign

AugustinMauroy:
# Proposal
This is the beginning of the style/layout migration.

For that I created this little piece of code to support both styles at the same time.

The idea is just to go ahead. For the user it doesn't bring anything it's only for our contributor.

We could have put it in a PR with modifications of style and layout. But it seemed cleaner to do it in two shots.

> The difference in syntax between the two variables allows them to co-exist.

# Checklist

* [X] Test if the "old" design work good.
* [X] Test if the new var work good.


ovflowd (approved):
LGTM! This is the first step for the styles of nodejs.dev. Nicely done!

AugustinMauroy:
> This is the first step for the styles of nodejs.dev. Nicely done!

I'm so Hype !!!!

PR #5087, opened 2023-03-09
fix broken links

imba-tjd:
https://nodejs.org/en/download/package-manager/#debian-and-ubuntu-based-linux-distributions

图片

`https://github.com/nodesource/distributions/blob/master/README` doesn't exist. But the source md is correct (with .md suffix).


ovflowd (approved):
Looks good to me! :shipit: 

Thank you for the first-time contribution 🎉

Trott (approved):

PR #5098, opened 2023-03-10
fix: scroll to top button on small viewports

### LINKED ISSUE #5090: "Scroll to top" button broken on mobile
### URL:

https://nodejs.org/en/blog/release/v18.15.0

### Browser Name:

Chrome

### Browser Version:

110

### Operation System:

macOS

1. Go to https://nodejs.org/en/blog/release/v18.15.0
2. Scroll down until the "Scroll to top" button appears
3. Reduce the width of the window until the button is replaced with a small green line

Working:
CleanShot 2023-03-09 at 13 38 45

Not working:
CleanShot 2023-03-09 at 13 39 16

mikeesto:
Resolves https://github.com/nodejs/nodejs.org/issues/5090

Looking at this CSS rule, perhaps the original intention was to hide the button on mobile. I don't think there's any harm to show it for now. This rule does not hide it completely in any event, so I think overall this PR is an improvement.

### DIFF
--- a/styles/page-modules/_scrollToTop.scss
+++ b/styles/page-modules/_scrollToTop.scss
@@ -41,10 +41,4 @@ html {
       opacity: 1;
     }
   }
-
-  @media (max-width: 600px) {
-    span {
-      display: none;
-    }
-  }
 }


ovflowd (approved):
Sounds a good plan. I also believe we can hide this on mobile.

LGTM!

Thank you for your contributions @mikeesto!

SpeedyCodeYT:
I think that the scroll to top button has no use on mobile when you can just use the back button on androids. iPhone users should use their fingers instead,

SpeedyCodeYT (approved):
Looks good to me!
thanks for your contribution!!
cheers

PR #5100, opened 2023-03-10
fix: AnchorLinks don't work properly

ghost:
This fixture mainly focuses on two bugs:

1. For Heading \<h1\>, it doesn't render as what we see for "h2" , "h3"..,and the "id" is always null. So the solution is to render the the children (here it must be a string) and return it in default.

【Before】
WrongForH1

【After】
CorrectForH1

3. For some very very long English titles or non-English titles in md, the generated anchor name in default will be very hard to read/write in md files if we want to refer. What we used to do is to give it an alian name in short, Take Chinese generated HTML files will be a good example to explain.

【Before】
WrongForChineseAnchorLink

【After】
CorrectForChineseAnchorLink


components/AnchoredHeading.tsx
```diff
@@ -7,11 +7,42 @@ type AnchorHeadingProps = PropsWithChildren<{
...
 
-  // This regex replacement is used for <!-- --> comments within the heading
-  const sanitizedId = id?.replace(/---(.+)---(.*)/, (_, f) => f);
+  let sanitizedId =
+    id ?? children?.toLocaleString().toLocaleLowerCase().replace(/\x20/g, '-');
+
+  if (sanitizedId) {
+    const foundAnchorAndTitle = COMMENT_FOR_HEADANCHOR.exec(sanitizedId);
+    if (foundAnchorAndTitle) {
+      sanitizedId = foundAnchorAndTitle[1];
```
ovflowd: Can we make only one assignment here? Like `sanitizedId = foundAnchorAndTitle[1].replace....`

components/AnchoredHeading.tsx
```diff
@@ -7,11 +7,42 @@ type AnchorHeadingProps = PropsWithChildren<{
...
 const AnchoredHeading = ({ children, level, id }: AnchorHeadingProps) => {
   const HeadingLevelTag = `h${level}` as any;
 
-  // This regex replacement is used for <!-- --> comments within the heading
-  const sanitizedId = id?.replace(/---(.+)---(.*)/, (_, f) => f);
+  let sanitizedId =
+    id ?? children?.toLocaleString().toLocaleLowerCase().replace(/\x20/g, '-');
+
+  if (sanitizedId) {
+    const foundAnchorAndTitle = COMMENT_FOR_HEADANCHOR.exec(sanitizedId);
```
ovflowd, line 35: Can we add a white space here?

ovflowd (approved):
Overall looks good to me! I added just two minor suggestions. Thank you!

components/AnchoredHeading.tsx
ghost: Do you mean the white space shouldn't be converted to "-"?

I intend to do that, however I find it will be converted in default....So this is only a sync conversion :)

components/AnchoredHeading.tsx
ghost: Fix it :)

components/AnchoredHeading.tsx
ovflowd: Oh, I meant an empty line :) to separate this line to the next

ghost: > Oh, I meant an empty line :) to separate this line to the next

Do you mean writing like this following:

```js
  let sanitizedId =
    id ?? 
    children?.toLocaleString().toLocaleLowerCase().replace(/\x20/g, '-');
```

When running "lint:js -- --fix", it automatically fix it like it....

If I get you wrong, please point it out :)

ovflowd: No no, I literally meant adding a blank line between 35 and 36 😅

ghost: Got it and fixed it, thanks for your patience XD

PR #5105, opened 2023-03-12
fix: download links cannot be refreshed through js

ghost:
【Problem】

https://github.com/nodejs/nodejs.org/issues/5102.

【Reason】 

As far as I see, when using 'querySelectorAll', it will first search for the whole page and return you the result directly, sometimes the link name of the element is changed but NOT saved in the storage of querySelector DOM itself. 

So A better solution is to search for the element after the DOM frame is loaded without loading the DOM contents...ect.

For more：
1. https://developer.mozilla.org/en-US/docs/Web/HTML/Element/script#defer
2. https://developer.mozilla.org/en-US/docs/Web/API/Document_object_model/Locating_DOM_elements_using_selectors

### DIFF
--- a/pages/_document.tsx
+++ b/pages/_document.tsx
@@ -1,5 +1,4 @@
 import React from 'react';
-import Script from 'next/script';
 import { Html, Head, Main, NextScript } from 'next/document';
 import Link from 'next/link';
 
@@ -13,7 +12,7 @@ export default function Document() {
       <body>
         <Main />
         <NextScript />
-        <Script strategy="beforeInteractive" src="/static/js/legacyMain.js" />
+        <script src="/static/js/legacyMain.js" defer async />
         <Link rel="me" href="https://social.lfx.dev/@nodejs" />
       </body>
     </Html>
--- a/.github/workflows/github-pages.yml
+++ b/.github/workflows/github-pages.yml
@@ -35,8 +35,6 @@ jobs:
 
       - name: Setup GitHub Pages
         uses: actions/configure-pages@v3
-        with:
-          static_site_generator: next
 
       - name: Restore Next.js cache
         uses: actions/cache/restore@v3
--- a/layouts/DownloadLayout.tsx
+++ b/layouts/DownloadLayout.tsx
@@ -6,11 +6,11 @@ import SecondaryDownloadMatrix from '../components/Downloads/SecondaryDownloadMa
 import { useNextraContext } from '../hooks/useNextraContext';
 import { useNodeData } from '../hooks/useNodeData';
 
-import type { LegacyDownloadsFrontMatter, NodeVersionData } from '../types';
+import type { LegacyDownloadsFrontMatter } from '../types';
 
 const DownloadLayout = ({ children }: PropsWithChildren) => {
   const nextraContext = useNextraContext();
-  const { currentLtsVersion = {} as NodeVersionData } = useNodeData();
+  const { currentLtsVersion } = useNodeData();
 
   const { downloads } = nextraContext.frontMatter as LegacyDownloadsFrontMatter;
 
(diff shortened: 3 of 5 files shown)


ghost:
PS：Another way is to change the querySelector to getElementById(), which will find the real element again and again, no matter whether the page is stored or not (Live in searching). But now it seems to change the loading way of js gets me also fine.

pages/_document.tsx
```diff
@@ -13,7 +12,7 @@ export default function Document() {
       <body>
         <Main />
         <NextScript />
-        <Script strategy="beforeInteractive" src="/static/js/legacyMain.js" />
+        <script src="/static/js/legacyMain.js" defer async />
```
ovflowd, line 15: Why removing the `Script` tag from `nextjs`? Which kind o tests you made that resulted in this conclusion?

ovflowd:
With all due respect, I don't think your PR is fixing anything.

pages/_document.tsx
MaledongGit: This reason is hard to explain....But after a period of experienment, I found if I changed to this, it seems the bug was fixed... You can have a try after applying it on your local code. And That's the reason why I thought of it.

Any other better suggestions is welcomed.

ovflowd:
@MaledongGit, my concerns here are the following:

- You seem to be doing many unrelated changes without explaining why.
- The description of the PR has nothing to do with the changes nor explains why these changes
- You should know that changing the script to `async defer` doesn't solve any issue; you're deferring the script's load to be the last JavaScript module to be loaded. This can cause many complications, as we're invoking the script "start" function on the Component mount of' Theme'. By deferring its load, the script may not be available when React attempts to call its start method. The `next/script` with `beforeInitiative` type ensures this is the first script to load immediately after Next.js's Framework is loaded (before the Application gets loaded).
- The bug mentioned in #5102 has nothing to do with the changes of this PR; please read #5025 for context. 
- The screenshot attached to #5102 has an "error" on the console. That is not an error. `next/link` tries to prefetch routes, but we should not use `next/link` for external routes.

Please, keep these points in consideration. I'm inclined to close this PR, as it doesn't solve any issues, but I want to hear you out, with your points, on why you changed the props of an unrelated component and the `label-for` part of the Component.

ghost:
> why you changed the props of an unrelated component and the label-for part of the Component.

Do you mean "labels["download-for"] ? Ah yes, this is a typo error and it should be this instead of "labels["download"] in i18n. I'm now inverstigating in it and found the problem, just fix it together.

ghost:
> The screenshot attached to https://github.com/nodejs/nodejs.org/issues/5102 has an "error" on the console. That is not an error. next/link tries to prefetch routes, but we should not use next/link for external routes.

Ah yes, I also found it later, can I help to change them?

I found there're many external routes in "\<Link\>", however some of them work properly with console errors, should I change them or just fix this?

ghost:
In the end, I found it seems we're now having another fix for bugs like this, so I'd close mine.
Thanks!

pages/_document.tsx
ghost: Maybe the customized event handlers' problem for listening after all the other js files of next.js are loaded successfully for querySelectoAll()……I also wonder if it's a better solution？

ovflowd: Not really, it's more like the following:

1. The next/link component takes an href argument, which is pretty much what our JavaScript legacy script does, replace the URL
2. But the thing is that `next/link` also sets up "click" events and other kinds of events. These events are managed by a React component, meaning if there's a re-render or whatever the internal logic is, it will ignore our "href" attribute.
3. Not to mention, "click" event has precedence over "href". So basically, us replacing the "href" probably has no effect at that point in time. Note that I am just making a wild guess here.

But the actual solution is for now to not use next/link, we probably also should not use it for external links, as it was primarily made for internal links.

ovflowd:
> Do you mean "labels["download-for"] ? Ah yes, this is a typo error and it should be this instead of "labels["download"] in i18n. I'm now inverstigating in it and found the problem, just fix it together.

Yeah, I wonder why you made the change? 🤔

ovflowd:
> Ah yes, I also found it later, can I help to change them?

I found there're many external routes in "<Link>", however some of them work properly with console errors, should I change them or just fix this?

Sure, but I made a PR already hours hours ago hehe

PR #5109, opened 2023-03-13
fix: update the translation label for the home page

ghost:
For the home page of nodejs.org, it seems the download info should be "labels['download-for']" as an i18n message instead of "labels['download']", because it fetches the info from "data-dl-local", whose meaning is the same as "labels['download-for']":

【Details Info from the Home page code behind】
https://github.com/nodejs/nodejs.org/blob/6f88680c53e195042099f003ea41b5b35e202706/layouts/IndexLayout.tsx#L25

【Details Info from the legacyMain.js】
https://github.com/nodejs/nodejs.org/blob/6f88680c53e195042099f003ea41b5b35e202706/public/static/js/legacyMain.js#L99-L113

Thus in the Index.md file, the "download" i18n is useless (no-where to refer), I've also removed them out of here.

### DIFF
--- a/layouts/IndexLayout.tsx
+++ b/layouts/IndexLayout.tsx
@@ -23,7 +23,7 @@ const IndexLayout = ({ children }: PropsWithChildren) => {
           <Banner />
 
           <h2 id="home-downloadhead" data-dl-local={labels['download-for']}>
-            {labels['download']}
+            {labels['download-for']}
           </h2>
 
           <HomeDownloadButton {...currentLtsVersion!} />
--- a/pages/ar/index.md
+++ b/pages/ar/index.md
@@ -2,7 +2,6 @@
 layout: index.hbs
 labels:
   current-version: الإصدار الحالي
-  download: تحميل
   download-for: تحميل النسخة الخاصة بـ
   other-downloads: تحميل نسخ أخرى
   current: النسخة الحالية
--- a/pages/ca/index.md
+++ b/pages/ca/index.md
@@ -2,7 +2,6 @@
 layout: index.hbs
 labels:
   current-version: Versió actual
-  download: Descarregar
   download-for: Descarregar per
   other-downloads: Altres Descàrregues
   current: Actual
--- a/pages/de/index.md
+++ b/pages/de/index.md
@@ -2,7 +2,6 @@
 layout: index.hbs
 labels:
   current-version: Aktuelle Version
-  download: Download
   download-for: Herunterladen für
   other-downloads: Andere Downloads
   current: Aktuell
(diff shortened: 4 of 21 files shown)


ghost:
PS：According to the real code, the text contents of \<h2\> should be removed（I mean we shouldn't keep "labels['download-for']" in the tag）

https://github.com/nodejs/nodejs.org/blob/6f88680c53e195042099f003ea41b5b35e202706/layouts/IndexLayout.tsx#L26

Because it's been totally replaced through js, see above.

However since "legacyMain.js" will be removed later, so I still keep it here as a template so that we can (may be replacing the template value) in the future....

So which is better? I just keep it here for future design because I don't know what'll happen here if "legacyMain.js" is removed...

ovflowd (changes requested):
We should keep the current behaviour. The current `download` label is working as expected. If you notice, this is the default message. Our `legacyMain` tries to identify the current running OS and then replaces the label `textContent` with the `label-for` + OS.

But you can see that the switch doesn't have a default statement. If it doesn't identify which OS you're running, we should still use the regular `download` label.

So in the end, this PR is not fixing anything. And if we remove the "download" label, if we don't identify the OS, it will show like "Download for", but for what?

So yes, default behaviour is indeed correct :)

ghost:
> We should keep the current behaviour. The current `download` label is working as expected. If you notice, this is the default message. Our `legacyMain` tries to identify the current running OS and then replaces the label `textContent` with the `label-for` + OS.
> 
> But you can see that the switch doesn't have a default statement. If it doesn't identify which OS you're running, we should still use the regular `download` label.
> 
> So in the end, this PR is not fixing anything. And if we remove the "download" label, if we don't identify the OS, it will show like "Download for", but for what?
> 
> So yes, default behaviour is indeed correct :)

To be honest, when I submit the PR after a period of thought.

Could it be possible to run Nodejs out of Mac,Linux(any other platform, maybe based on unix/linux that can be checked rightly) and Windows? Does Nodejs support the platform as well when to download Nodejs on an unknown platform....Just feeling a bit strange :(

ovflowd:
> Does Nodejs support the platform as well when to download Nodejs on an unknown platform....Just feeling a bit strange :(

It is not an unknown platform. Just not Linux, Windows or macOS. We support a huge amount of platforms. But for the website, we only offer "customised" downloads for Linux, macOS and Windows.

ghost:
> It is not an unknown platform. Just not Linux, Windows or macOS. We support a huge amount of platforms. But for the website, we only offer "customised" downloads for Linux, macOS and Windows.

If so, I'll close this PR, due to diff understanding of the support of Nodejs.

PR #5125, opened 2023-03-14
fix: windows 11 cannot detect CPU processor successfully

ghost:
Refs:
1. https://github.com/nodejs/nodejs.org/issues/5028.
2. https://github.com/nodejs/nodejs.org/pull/5025.

Due to the author doesn't have free time to fix this, I'd like to help.


ghost:
@styfle or @ovflowd：Could any of you have free time to review it?

public/static/js/legacyMain.js
```diff
@@ -80,14 +80,40 @@ const listenScrollToTopButton = () => {
...
+
   const userAgent = navigator.userAgent;
   const osMatch = userAgent.match(/(Win|Mac|Linux)/);
-
   const os = (osMatch && osMatch[1]) || '';
 
   const arch =
-    userAgent.match(/x86_64|Win64|WOW64/) || navigator.cpuClass === 'x64'
+    userAgent.match(/x86_64|Win64|WOW64/) ||
+    userAgent.cpuClass === 'x64' ||
```
ovflowd: There is no such thing as `userAgent.cpuClass`, in the past there existed a `navigator.oscpu` and `navigator.cpuClass` but they were long time removed from all modern browsers. IE6-11 might actually still use it. But then keep it as `navigator.cpuClass`, please, and add a note that this is an IE11 specific attribute.

public/static/js/legacyMain.js
```diff
@@ -80,14 +80,40 @@ const listenScrollToTopButton = () => {
   });
 };
 
-const detectEnviromentAndSetDownloadOptions = () => {
+const detectEnviromentAndSetDownloadOptions = async () => {
```
ovflowd, line 83: Let's not make this asynchronous.

public/static/js/legacyMain.js
```diff
@@ -80,14 +80,40 @@ const listenScrollToTopButton = () => {
...
+   * @see
+   * [MDN](https://developer.mozilla.org/en-US/docs/Web/API/NavigatorUAData/getHighEntropyValues)
+   * and
+   * [MSFT](https://learn.microsoft.com/en-us/microsoft-edge/web-platform/how-to-detect-win11)
+   *
+   * @returns
+   * 32 or 64, depending on the CPU bitness of the user's device.
+   * If not supported, it returns undefined.
+   */
+  async function getCPUBitnessForWin11() {
```
ovflowd: I think instead of making a complex asynchronous function we can simply:

```js
const userAgent = navigator.userAgent;
const userAgentData = navigator.userAgentData;
const osMatch = userAgent.match(/(Win|Mac|Linux)/);

const os = (osMatch && osMatch[1]) || '';

// detects the architecture through regular ways on user agents that 
// expose the architecture and platform information
// @note this is not available on Windows11 anymore
// @see https://learn.microsoft.com/en-us/microsoft-edge/web-platform/how-to-detect-win11
let arch === userAgent.match(/x86_64|Win64|WOW64/) ? 'x64' : 'x86';

// detects the platform through a legacy property on navigator
// available only on firefox
// @see https://developer.mozilla.org/en-US/docs/Web/API/Navigator/oscpu
if (navigator.oscpu && navigator.oscpu.length) {
  arch = navigator.oscpu.match(/x86_64|Win64|WOW64/) ? 'x64' : 'x86';
}

// detects the platform through a legacy property on navigator
// only available on internet explorer
if (navigator.cpuClass && navigator.cpuClass.length) {
  arch = navigator.cpuClass === 'x64' ? 'x64' : 'x86';
}

// detects the architecture and other platform data on the navigator
// available only on Chromium-based browsers
// @see https://developer.mozilla.org/en-US/docs/Web/API/NavigatorUAData/getHighEntropyValues
if (userAgentData && userAgentData.getHighEntropyValues) {
  // note that getHighEntropyValues returns an object
  const platform = await userAgentData.getHighEntropyValues(['bitness']);

  if (platform && platform.bitness) {
    // note that platform.bitness returns a string
    arch = platform.bitness === "64" ? 'x64' : 'x86';
  }
}

...
```

ovflowd: In my snippet above you have a full "safe" and easy-to-read and understand approach. I'd appreciate if you update your PR with this snippet.

ovflowd (changes requested):
The intent is great, but the code is messy and kinda hard to read, I added a comment with a better example.

public/static/js/legacyMain.js
ghost: Hi @ovflowd：

There're two things to confirm:

1. Do you mean to remove the **async** function and then paste what you've written directly inside "detectEnviromentAndSetDownloadOptions"? 

If so, the function "userAgentData.getHighEntropyValues" is a Promise-based (awaitable one). It must be inside an async method. So "detectEnviromentAndSetDownloadOptions" should still be an async one.

2. If according to your flattern code, because each browser has its own checking way, if mixed together, some checking ways matches while the others not, you'll get an error result in the end. We don't need to go through each "if" and assign to "arch".

All in all, I tend to write it like this following by seperating the complicated conditions into several functions with comments and call them and chain them by "||"：

```js
const detectEnviromentAndSetDownloadOptions = async () => {

const userAgent = navigator.userAgent;

// detects the architecture through regular ways on user agents that 
// expose the architecture and platform information
// @note this is not available on Windows11 anymore
// @see https://learn.microsoft.com/en-us/microsoft-edge/web-platform/how-to-detect-win11
function getCPUProcessBit() {
 if (!userAgent) return undefined;
 return userAgent.match(/x86_64|Win64|WOW64/) ? 'x64' : 'x86';
}

// detects the platform through a legacy property on navigator
// only available on internet explorer
function getCPUProcessBitForIE() {
if (navigator.cpuClass && navigator.cpuClass.length) {
  return navigator.cpuClass === 'x64' ? 'x64' : 'x86';
 }
return undefined;
}

// detects the platform through a legacy property on navigator
// available only on firefox
// @see https://developer.mozilla.org/en-US/docs/Web/API/Navigator/oscpu
function getCPUProcessBitForFireFox() {
 if (navigator.oscpu && navigator.oscpu.length) {
  return navigator.oscpu.match(/x86_64|Win64|WOW64/) ? 'x64' : 'x86';
 }
 return undefined;
}

// detects the architecture and other platform data on the navigator
// available only on Chromium-based browsers
// @see https://developer.mozilla.org/en-US/docs/Web/API/NavigatorUAData/getHighEntropyValue
async function getCPUProcessBitForChrome() {
 const userAgentData = navigator.userAgentData;
 if (userAgentData && userAgentData.getHighEntropyValues) {
  // note that getHighEntropyValues returns an object
  const platform = await userAgentData.getHighEntropyValues(['bitness']);

  if (platform && platform.bitness) {
    // note that platform.bitness returns a string
    return platform.bitness === "64" ? 'x64' : 'x86';
  }
 }
 return undefined;
}

const osMatch = userAgent.match(/(Win|Mac|Linux)/);
const os = (osMatch && osMatch[1]) || '';

let arch = getCPUProcessBit() ||
           (await getCPUProcessBitForChrome()) ||
           getCPUProcessBitForFireFox() ||
           getCPUProcessBitForIE();

if (!arch) 
  arch = 'x86';
}

// Your other codes here.....
```

ghost: > In my snippet above you have a full "safe" and easy-to-read and understand approach. I'd appreciate if you update your PR with this snippet.

Sorry but what do you mean by "safe" here? Do you think this method, if an async one, will be unsafe to be called? Though the function will be executed in an async way, however it isn't referred by any other functions and it may be safe.

Correct me if I take your meaning wrong :)

ghost:
@ovflowd : Thanks for your patience to add up a lot of things missing in the rough codes, however there're still some things to discuss :)

public/static/js/legacyMain.js
ghost: If not, the awaitable method cannot be called....., and the compiler will report an error, because "userAgentData.getHighEntropyValues" is an async method, which must be called by "await" in an async method.

public/static/js/legacyMain.js
```diff
@@ -80,14 +80,40 @@ const listenScrollToTopButton = () => {
...
+    if (
+      navigator.userAgentData &&
+      navigator.userAgentData.getHighEntropyValues
+    ) {
+      return await navigator.userAgentData.getHighEntropyValues(['bitness']);
+    }
+    return undefined;
+  }
+
+  const getWin11CPUProcessorBitness = getCPUBitnessForWin11();
```
styfle: I don't think Win11 is relevant here. This works for any OS and any browser that supports the `getHighEntropyValues()` API 

I tested on Edge and Chrome on Windows and macOS to confirm.

public/static/js/legacyMain.js
ovflowd: I don't want to argue here but look at your snippet's complexity versus mine's simplicity.

ovflowd: > Do you mean to remove the async function and then paste what you've written directly inside "detectEnviromentAndSetDownloadOptions"?

Nope.

> If so, the function "userAgentData.getHighEntropyValues" is a Promise-based (awaitable one). It must be inside an async method. So "detectEnviromentAndSetDownloadOptions" should still be an async one.

My original comment about not making it asynchronous was based if we only used promises. But then I made a snippet with a nice clean code using async/await. So, I apologise for the confusion.

> If according to your flattern code, because each browser has its own checking way, if mixed together, some checking ways matches while the others not, you'll get an error result in the end. We don't need to go through each "if" and assign to "arch".

Ehh, not really. If you look carefully, you'll notice that:

1. the `oscpu` check only runs on Firefox
2. the `cpuClass` will only run on IE11 and prior
3. The userAgentData will only run on Chromium

Meaning, no condition is being superseded. You can test this manually to reach your own conclusions.

public/static/js/legacyMain.js
ghost: > I don't think Win11 is relevant here. This works for any OS and any browser that supports the `getHighEntropyValues()` API

Any browser? Are you sure? From MDN document, it seems FF and Safari won't be supported.
image

What's more: Maybe we still consider IE6 - IE11? They won't support such functions.....
Correct me if I take you wrongly.... :D

public/static/js/legacyMain.js
ghost: Got it, thanks I'll fix mine.

public/static/js/legacyMain.js
styfle: I think you misunderstood. I said "any browser that supports".

More importantly, it works on every OS – not just Win11.

ovflowd: Yeah (agree with @styfle here)

ovflowd (approved):
Hey @MaledongGit thanks for the changes, and I also appreciate the patience with dealing with me 🙈🫠 I know my communication at times can sound off or harsh, but I swear, have the best of intentions here ❤️

Let's get this merged!

styfle (approved):

ovflowd:
Btw: @MaledongGit don't forget to test this on Firefox and Chrome on Windows if possible 🙃

ghost:
> Btw: @MaledongGit don't forget to test this on Firefox and Chrome on Windows if possible

No problem. I have Windows 10 with latest FF, Chrome and IE.

@styfle : If you have Windows 11, can you help to test it?

ghost:
@ovflowd & @styfle：

Firstly, FF and Chrome go fine with me now. I think everything goes fine with me now on Win 10.

Secondly, notice that IE 11 doesn't support new js functions or properties...ect, when debugging locally, but it doens't matter when keeping these codes, because [IE11 is out of service by Microsoft for Edge.](https://learn.microsoft.com/en-us/lifecycle/announcements/internet-explorer-11-end-of-support) And now if someone tends to use IE11 on Windows 10, it will directly guide you to the Edge download page, which means you cannot use IE11 in normal or smoothly. Have u noticed that on the latest version of IE11 and windows 10?

image

In the end, just waiting for @styfle 's tests on Win 11 for FF/Chrome.

ghost:
@ovflowd：Should we alert a message or something similar to notify those using IE browser to change to Edge/Chrome/FF? Because they won't work properly for Node.js loading and they are totally out of service?

Another way is to put something like polyfill.js to be compatable with IE, but is it possible to do so? IE's is out of its way.

If you think possible, maybe this will be done later.

ghost:
> Hey @MaledongGit thanks for the changes, and I also appreciate the patience with dealing with me 🙈🫠 I know my communication at times can sound off or harsh, but I swear, have the best of intentions here ❤️
> 
> Let's get this merged!

Nothing special, that's THE REAL TALK sincerely, problems and ideas may occur here.
Glad to share this time with you!!!! XD XD

ovflowd:
> Secondly, notice that IE 11 doesn't support new js functions or properties...ect, when debugging locally, but it doens't matter when keeping these codes, because [IE11 is out of service by Microsoft for Edge.](https://learn.microsoft.com/en-us/lifecycle/announcements/internet-explorer-11-end-of-support) And now if someone tends to use IE11 on Windows 10, it will directly guide you to the Edge download page, which means you cannot use IE11 in normal or smoothly. Have u noticed that on the latest version of IE11 and windows 10?

We can add support to IE11 through https://nextjs.org/docs/basic-features/supported-browsers-features#browserslist, but I wonder @Trott should we actually still support IE11? I can imagine people downloading Node.js through Windows Server or older machines, so I assume, yes?

@MaledongGit if you want, feel free to make a PR with a Browserlist config that includes IE11 :)

styfle:
I use a VM to test windows. Is there a vercel preview deployment url we can use to test? I'm not seeing it on this PR?

ovflowd:
> I use a VM to test windows. Is there a vercel preview deployment url we can use to test? I'm not seeing it on this PR?

We don't use Vercel yet.

styfle:
> We don't use Vercel yet.

I see. Let me know if you need any help setting it up 👍 

In the meantime, I deployed this branch here https://nodejs-2jdiq6e4o-uncurated-tests.vercel.app/en

I also confirmed it works on Windows 11 64 bit ARM on Edge, Chrome, and Firefox ✅

ghost:
Thanks for both, I first merged this bug fix and I'll submit another new PR to support IE when free :)

PR #5136, opened 2023-03-15
fix: sed command line in "Diagonostics"

ghost:
Refs: https://github.com/nodejs/nodejs.org/pull/4636.

This is a fix for the other translated languages in union. Some bashes are missing to be updated.


ovflowd (approved):

ovflowd:
@MaledongGit if possible, please do not update translations through here. Otherwise Crowdin gets confused.

PR #5133, opened 2023-03-15
Update 2 links in Don't Block the Event Loop

ghost:

PR #5137, opened 2023-03-15
fix: correct the missing two links

ghost:
Refs: https://github.com/nodejs/nodejs.org/pull/5133.

Here's the fixture for the missing links in the other translations.


ovflowd (approved):

ovflowd:
@MaledongGit if possible, please do not update translations through here. Otherwise Crowdin gets confused.

PR #5159, opened 2023-03-18
fix: fix the navigation menu's style item when selected

ghost:
When select a menu item in the navigation menu in the pages such as 'Doc' or 'About', the style should be changed as when it's hovered.

E.g: When switching to 'About' and click 'About' or "Governance', the menu item style should be changed to 'light-green', with white fonts.

### DIFF
--- a/styles/styles.scss
+++ b/styles/styles.scss
@@ -41,7 +41,7 @@ article a {
     font-size: inherit;
     margin-bottom: 15px;
 
-    + p {
+    +p {
       margin-top: 0;
     }
   }
@@ -66,6 +66,11 @@ article a {
 
     li {
       padding: 0.1em 0.5em;
+
+      a:focus {
+        background-color: $active-green;
+        color: $white;
+      }
     }
 
     a {
@@ -77,8 +82,9 @@ article a {
     }
 
     .active {
-      > a,
-      > a:hover {
+
+      >a,
+      >a:hover {
         background-color: $active-green;
         color: $white;
       }
--- a/styles/layout/_dark-theme.scss
+++ b/styles/layout/_dark-theme.scss
@@ -42,6 +42,16 @@ body.dark {
       }
     }
 
+    // For Navigation Menu in dark mode
+    .container.has-side-nav {
+      nav > ul > li{
+        a:focus {
+        background-color: $dark-green2;
+        color: $white;
+        }
+      }
+    }
+
     .active > a {
       background-color: $active-green;
       color: $white;


ovflowd:
Could you please attach screenshots, and explain why this is needed? Why this wasn't a thing before?

styles/styles.scss
```diff
@@ -83,6 +83,16 @@ article a {
         color: $white;
       }
     }
+
+    // For Navigation Menu in light mode
+    .container.has-side-nav {
```
nschonni: This is already inside the `.has-side-nav` selector, so the SCSS nesting looks incorrect

ghost:
> Could you please attach screenshots, and explain why this is needed?
image

As you can see above, when your mouse is over the sub navigator menu items, its style will be changed, so what I did now is to make the "selected" style as the same as mouse hover.

> Why this wasn't a thing before?

I remember it used to be so but now it isn't....? Or I had a bad memory?
And what's more: this will make you clear to see what sub navigation item you're in now :)

ghost:
@ovflowd : It seems you're angery with me. Sorry for the inconvenience.

As a Nodejs lover, I just want to make the page more beautiful and let the user experience better. Maybe we have many diff in understanding how to design or fix some bugs, however notice I'm not making trouble with you. 

So next time if possible, Maybe I'll start a discussion first to decide something to be fixed/changed or not?

styles/styles.scss
ghost: @nschonni : Thanks for pointing it out, and a quick fix for it :)

ovflowd:
Oh lol, not angery at all. Just want to know why we're doing this supposed fix now, why wasn't this done at any point before 🙃

Anyhow, could you (maybe) show on the screenshot what changed? Having a hard time to notice the change 🙈

ovflowd:
Also a note about "making the website more beautiful" and "design improvements". These are very subjective, that's why usually we talk about design changes and feature requests before doing them on discussions/issues, but it's alright also to discuss directly on the PR.

But again, what you might find better, not necessarily is what someone else will find better 👀

(Not that I am saying that's the case!! 😅)

Anyhow, I would very soon rather have us focusing on porting the nodejs.dev design over here. Check out the discussions tab 👀

ghost:
> Anyhow, could you (maybe) show on the screenshot what changed? Having a hard time to notice the change 🙈

Sorry for the late reply because of my hard work....
The change is: when you switch to a certain item in the left navigator menu by clicking it, the style would be the same as when your mouse is over it.

【Before】
Click "Governance", the "Governance" loses its style, so you don't know which has been choosen.



【After】
Click "Governance", the "Governance" still keeps its style as your mouse is over it.

ghost:
The reason why I think is a bug: If you review your website in the mobile size or chorme's mobile size, it just plays as what I showed you here. However in a PC it doesn't.

You can reproduce it directly in the debugging mode of mobile at Chrome :)

Now here's the mobile style for your current Nodejs website:

ovflowd:
I'm sorry, but I don't understand what you're talking about if you're saying that the bug is that when in one location, the side navigation is not being highlighted, this is not a CSS bug.

It is a bug, but not because of CSS. It was introduced when we removed the `trailingSlash` config option from Next.js. [This](https://github.com/nodejs/nodejs.org/pull/5162) PR is the proper fix to the issue.

Please let me know if that PR fixes the behaviour you're talking about because I needed clarification on the bug you are referring to.

ghost:
> I'm sorry, but I don't understand what you're talking about if you're saying that the bug is that when in one location, the side navigation is not being highlighted, this is not a CSS bug.
> 
> It is a bug, but not because of CSS. It was introduced when we removed the `trailingSlash` config option from Next.js. [This](https://github.com/nodejs/nodejs.org/pull/5162) PR is the proper fix to the issue.
> 
> Please let me know if that PR fixes the behaviour you're talking about because I needed clarification on the bug you are referring to.

Good catch and really the class "active" is missing, and I have a bad idea to fix that.
Thank you!

PR #5171, opened 2023-03-21
fix: "Home" menu item misses the arrow when selected

ghost:
Refs:
1. https://github.com/nodejs/nodejs.org/pull/5162.
2. https://github.com/nodejs/nodejs.org/issues/5160.

Reason:  For home, the localised route will end with a slash, but the asPath will not. So we need to remove the slash from the localised route to make it the same as what we can get from useRoute().

Before Fix: 

网页捕获_21-3-2023_101454_nodejs org

After Fix:

Right


hooks/useLocale.ts
```diff
@@ -17,12 +17,19 @@ export const useLocale = () => {
...
     isCurrentLocaleRoute: (route: string, allowSubPath?: boolean) => {
-      const localisedRoute = localisedPath(route);
+      let localizedRoute = localisedPath(route);
       const asPathJustPath = asPath.replace(/[#|?].*$/, '');
 
+      // For home, the localised route will end with a slash, but the asPath will not.
+      // So we need to remove the slash from the localised route to make it the same
+      // as what we can get from useRoute().
+      if (localizedRoute.endsWith('/')) {
+        localizedRoute = localizedRoute.slice(0, localizedRoute.length - 1);
```
ovflowd: I'd say it's better to update the `linkWithLocale` util with this logic! Instead of editing this here 👀

ovflowd (commented):
Good stuff, but I still believe this doesn't address the root cause, and it is a workaround.

The "issue" is still on `navigation.json` and not here. You can update the "home" link (https://github.com/nodejs/nodejs.org/blob/main/navigation.json#L3) by removing the "/".

As the "localized" routes come from `navigation.json` and are premade by us, it might be easier to update the entry on `navigation.json`.

I say that because not in all cases do we want to remove the trailing slash of the localized route. What if it is an external link?

So I don't think adding an extra logic with unknown side-effects without adding unit tests (which we don't have yet) would be good here. (To be fair, I didn't make any unit test for any of the existing components, with the hypothesis that a lot of the content here will eventually just be purged).

Could you give it a try to update `navigation.json`?

ghost:
> Good stuff, but I still believe this doesn't address the root cause, and it is a workaround.
> 
> The "issue" is still on `navigation.json` and not here. You can update the "home" link (https://github.com/nodejs/nodejs.org/blob/main/navigation.json#L3) by removing the "/".
> 
> As the "localized" routes come from `navigation.json` and are premade by us, it might be easier to update the entry on `navigation.json`.
> 
> I say that because not in all cases do we want to remove the trailing slash of the localized route. What if it is an external link?
> 
> So I don't think adding an extra logic with unknown side-effects without adding unit tests (which we don't have yet) would be good here. (To be fair, I didn't make any unit test for any of the existing components, with the hypothesis that a lot of the content here will eventually just be purged).
> 
> Could you give it a try to update `navigation.json`?

Yes, I had a try first but it failed. Because if I  removed "/" for home, and when you click Home, it has an arrow, and when you click any of the other item, you'll still get an arrow (At least two arrows).

So that's the reason why I had to change it, I know this is a work around and I'm trying to find out (maybe a better way in navitation.json)??

ovflowd:
> Yes, I had a try first but it failed. Because if I removed "/" for home, and when you click Home, it has an arrow, and when you click any of the other item, you'll still get an arrow (At least two arrows).

Well, it is good to debug what each variable has as value and find another way of doing so.

> So that's the reason why I had to change it, I know this is a work around and I'm trying to find out (maybe a better way in navitation.json)??

Maybe we can use the remove last slash workaround, but I still believe the logic should reside somewhere else and ensure it is only applied to internal links. And please, test all possible usages of this (ensure everything is working).

ghost:
> > Yes, I had a try first but it failed. Because if I removed "/" for home, and when you click Home, it has an arrow, and when you click any of the other item, you'll still get an arrow (At least two arrows).
> 
> Well, it is good to debug what each variable has as value and find another way of doing so.
> 
> > So that's the reason why I had to change it, I know this is a work around and I'm trying to find out (maybe a better way in navitation.json)??
> 
> Maybe we can use the remove last slash workaround, but I still believe the logic should reside somewhere else and ensure it is only applied to internal links. And please, test all possible usages of this (ensure everything is working).

image

That's the result if we don't do so, wired!

ghost:
@ovflowd : Now I combined your ideas with mine——That's to say: I'll check whether the last character is '/' or not, and if for yes, just ignore it, while the other is to keep what it is.

util/linkWithLocale.ts
```diff
@@ -1,4 +1,9 @@
 import { UrlObject } from 'url';
 
-export const linkWithLocale = (locale: string) => (path: string | UrlObject) =>
-  `/${locale}${path.toString()}`;
+export const linkWithLocale = (locale: string) => {
+  return (path: string | UrlObject) => {
+    path = path.toString();
+    if (path === '/') return `/${locale}`;
```
ovflowd: This is a really cool solution! Could we convert this in a ternary operator?

ovflowd (approved):
Really cool! That's what I call a sleek solution! 🎉

PR #5233, opened 2023-04-05
📎 chore(migration): migrate banner component

manishprivet:
## Changes

Migrates Banner Component From https://github.com/nodejs/nodejs.dev/tree/main/src/components/CommonComponents/Banner

Addresses #5193


AugustinMauroy (approved):
LGTM

site.json
```diff
@@ -36,7 +36,8 @@
       "startDate": "2022-12-16T17:00:00.000Z",
       "endDate": "2022-12-30T16:00:00.000Z",
       "text": "Node.js assessment of OpenSSL 3.0.7 security advisory",
-      "link": "https://nodejs.org/en/blog/vulnerability/openssl-fixes-in-regular-releases-dec2022/"
+      "link": "https://nodejs.org/en/blog/vulnerability/openssl-fixes-in-regular-releases-dec2022/",
+      "bannerBtnText": "Read more"
```
nschonni: "Read more" links are usually an accessibility smell. Why is this needed vs the old link-only version?

manishprivet: The original component in `nodejs.dev` had two versions.

One with a Banner Text + a button, which was generally used for announcements, which looks like this.

image

And one like the classic one, where one can directly give an html, e.g., `<p>Banner Text</p>` and a link, and it will look like this.
image

Both versions are here. However the addition of `bannerBtnText` in `site.json` will not affect anything on the website, as the added component is not being used anywhere as of now. I added it here just for reference.

For the new banner component, the input interface looks something like this:

```ts
{
  text: string;
  link: string;
  bannerBtnText: string;
} | {
  html: string;
  link: string;
}
```

components/CommonComponents/Banner/index.tsx
```diff
@@ -0,0 +1,70 @@
...
+  bannersIndex: WebsiteBanner;
+}
+
+const useTextContent = ({ text, link, bannerBtnText }: WebsiteBanner) =>
+  useMemo(() => {
+    if (text) {
+      return (
+        <p>
+          <a href={link} target="_blank" rel="noopener noreferrer">
+            <button type="button">{bannerBtnText || 'Read More'}</button>
```
nschonni: This is invalid HTML as `<button>` can't be nested in a `<a>`

components/CommonComponents/Banner/Banner.stories.tsx
HinataKah0, line 1: Nit: move files from `components/CommonComponents/Banner` to `components/Common/Banner` as described in the issue description

.storybook/main.js
```diff
@@ -5,6 +5,7 @@ const config = {
     '@storybook/addon-links',
     '@storybook/addon-essentials',
     '@storybook/addon-interactions',
+    'storybook-addon-sass-postcss',
```
ovflowd: Another PR is already adding this, seems like we actually need this hmm.

site.json
ovflowd: I don't think we should add `bannerBtnText` to `site.json`. It should probably be on `i18n/locales/en.json` as "components.common.banner.button.text"

manishprivet: The data for banner is present in the `site.json` only, and as mentioned above, the buttonText was previously configurable according to the text in banner. For example, there can be an announcement for a youtube video, so the button should say something like `Watch Now`, at-least that was the original purpose of this configuration IMO, I could be wrong though.

I think it makes sense to keep all the configurable data for banner at the same place, but happy to change it if we are fixed on keeping the button text as `Read More` 🤔 

Btw we should also make the banner data localised, but I am guessing that we can tackle that in a separate issue/pr

.storybook/main.js
manishprivet: Yup

components/CommonComponents/Banner/index.tsx
manishprivet: This is fixed

components/CommonComponents/Banner/Banner.stories.tsx
manishprivet: Done

site.json
ovflowd: Sorry, but I have to disagree. You should 'at least' if you're adding this new feature to make `bannerBtnText` a i18n string that fallbacks to the content of `bannerBtnText` if there's no i18n key for that banner button :)

ovflowd: > Btw we should also make the banner data localised, but I am guessing that we can tackle that in a separate issue/pr

I would rather not delegate such a trivial and important change to a separate PR.

components/Common/Banner/index.tsx
```diff
@@ -0,0 +1,75 @@
...
+import styles from './index.module.scss';
+import { WebsiteBanner } from '../../../types';
+
+export interface BannerProps {
+  bannersIndex: WebsiteBanner;
+}
+
+const useTextContent = ({ text, link, bannerBtnText }: WebsiteBanner) =>
+  useMemo(() => {
+    if (text) {
```
nschonni, line 15: Might be another disconnect with the Storybook vs the actual usage, but these banner displays are based off the start/end dates in the JSON that dictate how long the banner is shown for a campaign

.storybook/main.js
shanpriyan: @manishprivet Were you able to process sass files for story without this addon?

components/Common/Banner/index.tsx
manishprivet: @nschonni That part is actually being checked here https://github.com/nodejs/nodejs.org/pull/5233/files#diff-ce3ccfab991d0d5b74fb505155f7ae195213df71c5ccd974fef7e7153a2e5681R52

site.json
manishprivet: @ovflowd The thing is that these tasks have been dealing with migrating the components. The migrated component here takes the bannerData as props, and will be used in the implementation part of the component. What I was thinking that maybe I can remove the bannerBtnText from site.json altogether, and when we will start implementing the components in the pages, then we can think of a way to implement new format of website banner. As other components that are being migrated from nodejs.dev also will require changes in the original configuration when we will try to use them

ovflowd: If the task is just about migrating the component, you shouldn't add `bannerBtnText` as this deviates from the migration. If you're not open to considering other angles here, then please remove this change, and open another PR after this one gets merged so we discuss that change.

I won't approve of a change that intentionally introduces a point of failure that "will" someday be addressed.

ovflowd (changes requested):
Per my comment, please either remove `bannerBtnText` or address its i18n issues.

site.json
manishprivet: Agreed, my bad as well. The task should have been focused on migrating the component only. I'll remove the entry in site.json in a few hours and then we can merge it. 

Should I create a new interface as well for the new Banner component, or adding the btnText in old interface will work?

ovflowd: I guess the appropriate solution for now is a fixed banner button text "Read more" that is a i18n translation key. So no need for a bannerBtnText on the interface or something.

I know your concern is about allowing the user to change the text of that button on a per-banner basis, we can address that in a future PR 🙃

manishprivet: Yeah that can work 👍🏼

components/Common/Banner/Banner.stories.tsx
```diff
@@ -0,0 +1,36 @@
+import type { Meta, StoryObj } from '@storybook/react';
+import Banner from './';
+
+const meta: Meta<typeof Banner> = {
```
ovflowd: Since this meta is used only on the default export, can't you make it an inline export? (I doubt you need to make a `typeof Banner` here).

components/Common/Banner/Banner.stories.tsx
```diff
@@ -0,0 +1,36 @@
...
+  title: 'Banner',
+  component: Banner,
+};
+
+export default meta;
+type Story = StoryObj<typeof Banner>;
+
+const startDate = new Date();
```
ovflowd, line 14: I would extract this to a tiny function in the same file, but just to not have it coexisting with the root-level code, as it could be confusing

i18n/locales/ca.json
```diff
@@ -1,39 +1,40 @@
...
-    "components.downloadReleasesTable.releases": "Releases",
-    "components.downloadReleasesTable.docs": "Documentació",
-    "components.header.buttons.toggleLanguage": "Toggle Language",
-    "components.header.buttons.toggleDarkMode": "Toggle dark/light mode",
-    "components.pagination.next": "Següent | ",
-    "components.pagination.previous": "Anterior",
-    "layouts.blogPost.author.byLine": "{author, select, null {} other {By {author}, }}",
-    "layouts.blogIndex.currentYear": "News from {year}"
-  }
+  "components.footer.scrollToTop.button": "Torna al començament",
```
ovflowd, line 2: Any reason why git thinks the whole file got changed?

i18n/locales/en.json
```diff
@@ -34,6 +34,7 @@
   "components.header.buttons.toggleDarkMode": "Toggle dark/light mode",
   "components.pagination.next": "Newer | ",
   "components.pagination.previous": "Older",
+  "components.common.banner.button.text": "Read More",
```
ovflowd: **Important:** New translation keys should be added always at the end of the locale file.

ovflowd (commented):
I left two nitpicks, but there's one major thing to address yet, can you remove the new translation key from all files except `en.json`? The key should be added to the translated languages through Crowdin, not through here :)

i18n/locales/ca.json
manishprivet: There was an indentation issue here

manishprivet:
> I left two nitpicks, but there's one major thing to address yet, can you remove the new translation key from all files except en.json? The key should be added to the translated languages through Crowdin, not through here :)

@ovflowd Regarding this, I myself didn't want to copy it in each file, but then I saw these two lines present in each file in English language, so I thought maybe a key was needed for Crowdin

```json
  "components.header.buttons.toggleLanguage": "Toggle Language",
 ```

**Comment by ovflowd** [comment]
> @ovflowd Regarding this, I myself didn't want to copy it in each file, but then I saw these two lines present in each file in English language, so I thought maybe a key was needed for Crowdin

That's fine. Just means that the key was not translated yet.

File: i18n/locales/ca.json [hunk 9]
```diff
@@ -1,39 +1,40 @@
 {
-    "components.footer.scrollToTop.button": "Torna al començament",
```
ovflowd, line 2: I see this file is still updated. Can you revert this file?

ovflowd (approved):
LGTM! Thank you for the considerable effort here. I left one last comment that must be addressed before we merge this PR, but otherwise, fine!

i18n/locales/ca.json
manishprivet: This file had wrong indentation, should we revert it?

ovflowd: Oh, no let's keep the change then. But just don't add the translation key

manishprivet: Great, I removed the key

rodion-arr (approved):

components/Common/Banner/index.tsx
```diff
@@ -0,0 +1,82 @@
...
+  useMemo(() => {
+    if (html) {
+      const sanitizedHtml = sanitize(html);
+
+      return (
+        <a
+          href={link}
+          target="_blank"
+          rel="noopener noreferrer"
+          dangerouslySetInnerHTML={{ __html: sanitizedHtml }}
```
shanpriyan: nits 
```suggestion
 
      return (
        <a
          href={link}
          target="_blank"
          rel="noopener noreferrer"
          dangerouslySetInnerHTML={{ __html: sanitize(html) }}
```

shanpriyan (approved):
LGTM! 🚀

components/Common/Banner/index.tsx
```diff
@@ -0,0 +1,81 @@
...
+      );
+    }
+
+    return null;
+  }, [text, link, bannerBtnText]);
+
+const useHtmlContent = ({ html, link }: WebsiteBanner) =>
+  useMemo(() => {
+    if (html) {
+
```
mikeesto:
```suggestion
```

To make the linter happy

PR #5234, opened 2023-04-05
(website redesign) Feat(shellbox): migration

AugustinMauroy:
* Introduce shellBox in commonComponents

---
Related to #5193


.storybook/main.js
```diff
@@ -5,6 +5,7 @@ const config = {
     '@storybook/addon-links',
     '@storybook/addon-essentials',
     '@storybook/addon-interactions',
+    'storybook-addon-sass-postcss',
```
manishprivet: #5233 will add this, so maybe should keep a eye on that if that gets merged first 👀

components/CommonComponents/ShellBox/index.stories.tsx
```diff
@@ -0,0 +1,12 @@
+import ShellBox from '.';
+
+export default {
+  title: 'CommonComponents/ShellBox',
+  component: ShellBox,
+};
+
+export const Default = () => (
+  <ShellBox textToCopy="echo 'Hello World'">
+    echo &rsquo;Hello World&rsquo;
```
manishprivet: Since the content will come from a md file, can we use a template string variable to mimic that through props instead of directly writing in the story. Also makes easier to try component in the storybook

components/CommonComponents/ShellBox/index.module.scss
```diff
@@ -0,0 +1,75 @@
+.shell-box {
```
manishprivet: I believe this should be `shellBox`, as it is used here

https://github.com/nodejs/nodejs.org/pull/5234/files#diff-a315b5aa06fa6054f4a1a831400da67bd944d83eda8da6aa7484783b0b9be4dbR19

```suggestion
.shellBox {
```

i18n/locales/es.json
```diff
@@ -35,5 +35,6 @@
   "components.pagination.next": "Siguiente | ",
   "components.pagination.previous": "Anterior",
   "layouts.blogPost.author.byLine": "{author, select, null {} other {By {author}, }}",
-  "layouts.blogIndex.currentYear": "News from {year}"
+  "layouts.blogIndex.currentYear": "News from {year}",
+  "components.codeBox.copy": "{copied, select, true {copied}other {copy}}"
```
ovflowd: Translation key should probably be `components.common.shellBox.copy"

components/CommonComponents/ShellBox/index.tsx
```diff
@@ -0,0 +1,31 @@
...
+import styles from './index.module.scss';
+
+interface Props {
+  textToCopy: string;
+}
+
+const ShellBox = ({ children, textToCopy }: React.PropsWithChildren<Props>) => {
+  const [copied, copyText] = useCopyToClipboard();
+
+  const handleCopyCode = async (event: React.MouseEvent<HTMLButtonElement>) => {
```
ovflowd: Shouldn't the children be the text to copy? Afaik you can simple convert the `children.toString()` no?

hooks/useCopyToClipboard.ts
```diff
@@ -0,0 +1,31 @@
+import { useEffect, useState } from 'react';
+
+const copyToClipboard = (value: string) => {
+  if (typeof window === 'undefined') {
```
ovflowd:
```suggestion
  if (typeof navigator === 'undefined') {
```

As we're using `navigator.` here

hooks/useCopyToClipboard.ts
```diff
@@ -0,0 +1,31 @@
...
+};
+
+// eslint-disable-next-line no-unused-vars
+export const useCopyToClipboard = (): [boolean, (text: string) => void] => {
+  const [copied, setCopied] = useState(false);
+
+  const copyText = (text: string) => copyToClipboard(text).then(setCopied);
+
+  useEffect(() => {
+    if (!copied) {
```
ovflowd: What about:

```js
if (copy) {
  const timerId = setTimeout(() => setCopied(false), 3000);

  return () => clearTimeout(timerId);
}
```

ovflowd:
BTW as mentioned in other PRs the folder should be `components/Common`

tniessen:
Is there some context on this somewhere? I don't see any helpful comments or documentation, nor a reference to previous related work. Is this effectively a fancy `<pre>` for shell code?

components/CommonComponents/ShellBox/__tests__/__snapshots__/index.test.tsx.snap
```diff
@@ -0,0 +1,25 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
```
tniessen: Is Jest already widely used in this repo? Is that conventional/necessary with Next.js? IIRC there was significant pushback against jest from some node core collaborators in other contexts, but I don't have much insight into this.

components/CommonComponents/ShellBox/index.stories.tsx
AugustinMauroy: No it is used to make the shellbox on the index page. It has nothing to do with markdown.

i18n/locales/es.json
AugustinMauroy: It is the same as for codebox so I propose to merge it into one key

components/CommonComponents/ShellBox/index.tsx
AugustinMauroy: I will investigate to see exactly how it is used

hooks/useCopyToClipboard.ts
AugustinMauroy: Oh ok. I'm currently ready the difference.

components/CommonComponents/ShellBox/__tests__/__snapshots__/index.test.tsx.snap
AugustinMauroy: Jest has been setup. There is no test on the codebase yet. And I don't think we'll take another test package. Because Jest is very good for react (IMHO)

components/CommonComponents/ShellBox/index.module.scss
AugustinMauroy: I don't understand how it's work for nodejs.dev 🤔

components/CommonComponents/ShellBox/__tests__/__snapshots__/index.test.tsx.snap
ovflowd: > Is Jest already widely used in this repo? Is that conventional/necessary with Next.js? IIRC there was significant pushback against jest from some node core collaborators in other contexts, but I don't have much insight into this.

@tniessen I believe it's mostly because we've been using it on nodejs.dev. But I'm all in for finding a substitute if that's the will of the core collaborators. I'm all fine with Jasmine/Mocha. WDYT?

ovflowd: I'd also be fine with native unit testing provided by Node.js, just not sure if it works well with DOM testing yet (haven't checked that yet)

i18n/locales/es.json
ovflowd: Note that the key should contain the name of the component. shellBox ain't codeBox.

components/CommonComponents/ShellBox/index.module.scss
ovflowd: CSS modules natively translate hyphen classes to camelCase. I think it is done by Next already https://github.com/vercel/next.js/discussions/11267

AugustinMauroy:
@tniessen It's an component form nodejs.dev use on index page.
So if you have any question say it.

tniessen:
> So if you have any question say it.

@AugustinMauroy I did:

> Is this effectively a fancy `<pre>` for shell code?

components/CommonComponents/ShellBox/__tests__/__snapshots__/index.test.tsx.snap
tniessen: I don't have an opinion on this; I don't know what share of node core collaborators intend to work on this new repository structure at all, and I don't have an opinion on frontend testing frameworks myself. If the folks who do maintain the new website are happy with this, then that's great.

AugustinMauroy:
> Is this effectively a fancy `<pre>` for shell code?

yes and no. it's use to do this.

Capture d’écran 2023-04-06 à 17 16 42

So it's never use to do this

```md
  ```js
  const a = 42
  '``
```

hooks/useCopyToClipboard.ts
AugustinMauroy: Done. And good idea my dear

components/CommonComponents/ShellBox/index.tsx
AugustinMauroy: it's use like that.
```tsx
<ShellBox textToCopy="choco install nvs">
      <span className={styles.installTextNoSelect}>$</span>
      <span className={styles.installTextCommand}>choco </span>install nvs
    </ShellBox>
```

So we don't touch to that .

i18n/locales/es.json
AugustinMauroy: But then if we follow my proposal for CodeBox the key will be `components.shellBox.copy`

manishprivet:
> Is there some context on this somewhere? I don't see any helpful comments or documentation, nor a reference to previous related work.

@tniessen The reference issue for this PR is here at #5193

components/CommonComponents/ShellBox/__tests__/__snapshots__/index.test.tsx.snap
ovflowd: I'd say let's keep at Jest for now then, and if somebody from Core opens an issue requesting a change, we can consider it :)

i18n/locales/es.json
AugustinMauroy: @ovflowd So with do you think ?

components/Common/ShellBox/__tests__/index.test.tsx
```diff
@@ -0,0 +1,43 @@
...
+describe('ShellBox component', (): void => {
+  it('renders correctly', (): void => {
+    const textToCopy = 'text to be copy';
+    const { container } = render(
+      <ShellBox textToCopy={textToCopy}>mock-children-code</ShellBox>
+    );
+    expect(container).toMatchSnapshot();
+  });
+
+  it('renders correctly', async () => {
```
ovflowd: Why the two tests are "renders correctly"?

components/Common/ShellBox/index.stories.tsx
```diff
@@ -0,0 +1,12 @@
+import ShellBox from '.';
+
+export default {
+  title: 'CommonComponents/ShellBox',
+  component: ShellBox,
+};
+
+export const Default = () => (
+  <ShellBox textToCopy="echo 'Hello World'">
+    echo &rsquo;Hello World&rsquo;
```
ovflowd: I think instead of making this static, this should be part of the Storybook args. (Aka user can change the content within the story)

components/Common/ShellBox/index.stories.tsx
```diff
@@ -0,0 +1,12 @@
+import ShellBox from '.';
+
+export default {
+  title: 'CommonComponents/ShellBox',
+  component: ShellBox,
+};
+
+export const Default = () => (
+  <ShellBox textToCopy="echo 'Hello World'">
```
ovflowd: Same here, this should be editable by the user.

components/Common/ShellBox/index.tsx
```diff
@@ -0,0 +1,31 @@
+import React from 'react';
+import { FormattedMessage } from 'react-intl';
+import { useCopyToClipboard } from '../../../hooks/useCopyToClipboard';
+import styles from './index.module.scss';
+
+interface Props {
+  textToCopy: string;
```
ovflowd: I think instead of having a `textToCopy` prop, we should infer the textToCopy based on the children, which in theory should be doable with `ReactDomServer.renderToString` (you can calculate this value as memoised value (`useMemo`) when the children changes

i18n/locales/en.json
```diff
@@ -35,5 +35,6 @@
   "components.pagination.next": "Newer | ",
   "components.pagination.previous": "Older",
   "layouts.blogPost.author.byLine": "{author, select, null {} other {By {author}, }}",
-  "layouts.blogIndex.currentYear": "News from {year}"
+  "layouts.blogIndex.currentYear": "News from {year}",
+  "components.codeBox.copy": "{copied, select, true {copied}other {copy}}"
```
ovflowd: Please remove the translation key from all languages except `en.json`. Thank you!

Also, the translation key should be `components.common.shellBox.copy` to actually properly refer to its canonical path.

i18n/locales/es.json
ovflowd: `components.common.shellBox.copy` is still the way I'd recommend, as this is its canonical path, and what we're doing on all other component.s

ovflowd (commented):
Left a few comments, but good work so far :)

i18n/locales/es.json
AugustinMauroy: Ok i will fix that 😁

components/Common/ShellBox/__tests__/index.test.tsx
AugustinMauroy: I don't really understand you question?

components/Common/ShellBox/index.stories.tsx
AugustinMauroy: Ok why not

components/Common/ShellBox/index.tsx
AugustinMauroy: I had answers you [here](https://github.com/nodejs/nodejs.org/pull/5234#discussion_r1159950100)

components/CommonComponents/ShellBox/index.tsx
ovflowd: Just because it was done like that doesn't mean we cannot change it.

I believe the textToCopy could be an optional argument and by default it goes to the approach I mentioned before? Wdyt

components/Common/ShellBox/__tests__/index.test.tsx
ovflowd: Both the test cases have the same title.

components/CommonComponents/ShellBox/index.tsx
AugustinMauroy: For a usage with children and textToCopy are same. I thinks we should use CodeBox.

components/Common/ShellBox/__tests__/index.test.tsx
AugustinMauroy: Ho yeah it's from nodejs.dev. I'll change that.

components/Common/ShellBox/__tests__/index.test.tsx
```diff
@@ -0,0 +1,43 @@
+import React from 'react';
+import userEvent from '@testing-library/user-event';
+import { render, screen } from '@testing-library/react';
+
+import ShellBox from '../index';
+
+Object.assign(navigator, {
```
ovflowd: Instead of assigning this here, you can create a mock on the beforeEach of each test and restore the navigator to its original value on the afterEach. The reasoning why is that depending on the Jest installation, this could interfere other test suites (it shouldn't by default) but it's also a good design to not assign/modify global variables on the main thread.

components/Common/ShellBox/__tests__/index.test.tsx
```diff
@@ -0,0 +1,43 @@
...
+  clipboard: {
+    writeText: jest.fn(),
+  },
+});
+
+const navigatorClipboardSpy = jest.spyOn(navigator.clipboard, 'writeText');
+
+afterEach(() => {
```
ovflowd: This should be inside the describe

components/Common/ShellBox/__tests__/index.test.tsx
```diff
@@ -0,0 +1,43 @@
...
+
+import ShellBox from '../index';
+
+Object.assign(navigator, {
+  clipboard: {
+    writeText: jest.fn(),
+  },
+});
+
+const navigatorClipboardSpy = jest.spyOn(navigator.clipboard, 'writeText');
```
ovflowd: This should be inside the describe

hooks/useCopyToClipboard.ts
```diff
@@ -0,0 +1,29 @@
...
+    return Promise.resolve(false);
+  }
+
+  return navigator.clipboard
+    .writeText(value)
+    .then(() => true)
+    .catch(() => false);
+};
+
+// eslint-disable-next-line no-unused-vars
```
ovflowd: Why is eslint complaining here?

hooks/useCopyToClipboard.ts
```diff
@@ -0,0 +1,29 @@
...
+export const useCopyToClipboard = (): [boolean, (text: string) => void] => {
```
ovflowd: No need of explicit type here.

hooks/useCopyToClipboard.ts
```diff
@@ -0,0 +1,29 @@
...
+
+  useEffect(() => {
+    if (copied) {
+      const timerId = setTimeout(() => setCopied(false), 3000);
+      return () => clearTimeout(timerId);
+    }
+    return undefined;
+  }, [copied]);
+
+  return [copied, copyText];
```
ovflowd: Add a `as const` in the end.

hooks/useCopyToClipboard.ts
```diff
@@ -0,0 +1,29 @@
...
+
+// eslint-disable-next-line no-unused-vars
+export const useCopyToClipboard = (): [boolean, (text: string) => void] => {
+  const [copied, setCopied] = useState(false);
+
+  const copyText = (text: string) => copyToClipboard(text).then(setCopied);
+
+  useEffect(() => {
+    if (copied) {
+      const timerId = setTimeout(() => setCopied(false), 3000);
```
ovflowd, line 22:
```suggestion
      const timerId = setTimeout(() => setCopied(false), 3000);

```

hooks/useCopyToClipboard.ts
```diff
@@ -0,0 +1,29 @@
...
+      return () => clearTimeout(timerId);
+    }
+    return undefined;
```
ovflowd, line 27:
```suggestion

    return undefined;
```

components/CommonComponents/ShellBox/index.tsx
ovflowd: What's the difference of CodeBox and ShellBox then? Just the style?

ovflowd (commented):
We're almost there. Left some more comments and I'm curious about the similarities of ShellBox and CodeBox and if the two should be just one component with two different flavours

AugustinMauroy:
> What's the difference of CodeBox and ShellBox then? Just the style?

No, codebox it's use to display code an hightlight it. ShellBox it's just an box for an shell command.

components/CommonComponents/ShellBox/index.tsx
AugustinMauroy: No, codebox it's use to display code an hightlight it. ShellBox it's just an box for an shell command.

ovflowd: You do realise semantically these are the same. Both display code, both allow code to be copied. Both have syntax highlighting to a sort.

Hence virtually both are the same. The only real difference are the layouts. Hence in UX that's what we call "flavours"

I do understand you have a different view of whatever you think you know, but it's kinda irritating from you to reject/disagree with these points when you have no real understanding of these concepts.

Instead of simply saying no, what about researching about these concepts and reflect a little about it. Because honestly, for me it makes sense that both are the same component but with different styles.

The same applies to the textToCopy issue. You simply negated my comment without a proper reasoning of why we should keep the prop. I'd recommend for you to actually start reflecting on your decisions and if I make a suggestions it must be because there's a point on it. I wouldn't recommend something just out of the blue. I don't want to sound rude/arrogant, but it's kinda tiring to have these kind of discussions because they're not productive.

hooks/useCopyToClipboard.ts
AugustinMauroy: `text` is unused var it's can be fix if we remove type

components/CommonComponents/ShellBox/index.tsx
AugustinMauroy: I had test `macosPanel` component with `textToCopy` using `children.toString()`. And the copied text is an `JSX` expressions.
To conclude we have to keep `textToCopy` in props

Of topics:
Thanks you Claudio for our time ad our review 😃.

ovflowd: I didn't say to use `children.toString()` (It was just an example), you should probably use `ReactDomServer.renderToString`.

> To conclude we have to keep textToCopy in props

What kind of conclusion is that? Listen, if a simple attempt doesn't work out, you can ask for alternatives, and not again, simply reject the feedback and conclude that the `prop` need to be kept because on your first attempt it didn't work :/

AugustinMauroy: > I didn't say to use children.toString() (It was just an example), you should probably use ReactDomServer.renderToString.

This case can work but. we will have this code copied:
`$ curl -o- https://raw.githubusercontent.com/nvm-sh/nvm/v0.39.3/install.sh | bash`

AugustinMauroy: The idea of having `textToCopy` is to have the copied text different from the displayed text.

ovflowd: I know what the component is about. Again, I think you're facing some creativity issues here and not understanding the broader goal of what I'm talking about having sub components / flavours.

I mentioned before that we he textToCopy could be an optional parameter and would default to the children.

Also, the only issue there is that it would copy the "$" symbol. This is easily solvable if we remove the "$" symbol from the original children and simply use CSS's :before or :after selectors to decorate each line of the ShellBox with the $ symbol.

The TL;DR is that there are many ways of achieving what I'm referring to, and you're simply creating a storm on a glass of water, by saying this is impossible.

AugustinMauroy: > I mentioned before that we he textToCopy could be an optional parameter and would default to the children.

I can do this shure.

ovflowd: To be fair, I know that you might not know these methods. And that's fine. I'm not expecting you to know all possible ways of implementing this as a Junior Dev/Newcomer in Web Dev.

What I ask is that instead of having a rejecting attitude, to have an "open to ideas" attitude. (Which maybe you have, but the tone on the messages hints the opposite, which could be due to language barriers 😅)

.eslintignore
```diff
@@ -8,6 +8,7 @@ public/static/documents
 public/static/legacy
 
 # We don't want to lint/prettify the Coverage Results
+tests
```
ovflowd: I think this got accidentally added on your rebase

.prettierignore
```diff
@@ -8,6 +8,7 @@ public/static/documents
 public/static/legacy
 
 # We don't want to lint/prettify the Coverage Results
+tests
```
ovflowd: I think this got accidentally added on your rebase

components/Common/ShellBox/__tests__/index.test.tsx
```diff
@@ -0,0 +1,51 @@
...
+import { render, screen } from '@testing-library/react';
+import { LocaleProvider } from '../../../../providers/localeProvider';
+import userEvent from '@testing-library/user-event';
+import { AppProps } from '../../../../types';
+import ShellBox from '../index';
+
+const mockWriteText = jest.fn();
+const i18nData = { currentLocale: { code: 'en' } } as AppProps['i18nData'];
+
+Object.defineProperty(window, 'navigator', {
```
ovflowd: Again, I feel like we should assign this on `beforeEach` and restore the original value on `afterEach`

components/Common/ShellBox/__tests__/index.test.tsx
```diff
@@ -0,0 +1,51 @@
+import React from 'react';
+import { render, screen } from '@testing-library/react';
+import { LocaleProvider } from '../../../../providers/localeProvider';
```
ovflowd: We shouldn't use `LocaleProvider` here, just the plain `IntlProvider`

components/Common/ShellBox/index.tsx
```diff
@@ -0,0 +1,37 @@
...
+  children: React.ReactNode;
+  textToCopy?: string;
+}
+
+const ShellBox = ({ children, textToCopy }: React.PropsWithChildren<Props>) => {
+  const [copied, copyText] = useCopyToClipboard();
+
+  const handleCopyCode = async (event: React.MouseEvent<HTMLButtonElement>) => {
+    event.preventDefault();
+    const text = textToCopy? textToCopy : ReactDomServer.renderToString(children);
```
ovflowd: I think this could be an `useCallback`, as we don't want to recalculate the textToCopy or the children with `ReactDomServer` on each click.

hooks/__test__/useCopyToClipboard.test.tsx
```diff
@@ -0,0 +1,50 @@
...
+import { render, fireEvent, screen } from '@testing-library/react';
+import { FormattedMessage } from 'react-intl';
+import { useCopyToClipboard } from '../useCopyToClipboard';
+import { LocaleProvider } from '../../providers/localeProvider';
+import { AppProps } from '../../types';
+
+const mockWriteText = jest.fn();
+const i18nData = { currentLocale: { code: 'en' } } as AppProps['i18nData'];
+
+Object.defineProperty(window, 'navigator', {
```
ovflowd: Same here from my previous comments

ovflowd (commented):
Really good stuff! I think we're getting closer to the final version of this 🎉 

Thank you for your continuous work, Augustin :)

components/Common/ShellBox/__tests__/index.test.tsx
AugustinMauroy: I will test

.eslintignore
```diff
@@ -16,4 +16,4 @@ coverage
 pages/**/*.md
 
 # We shouldn't lint statically generated Storybook files
-storybook-static
+storybook-static
```
ovflowd: @AugustinMauroy can you simply remove the changes you've made in these files? (I have the feeling you're still not doing rebasing correctly) as both this file, prettierignore and the package-lock.json were updated.

As mentioned before (including by others) you can simply cherry-pick your commits. There are even unrelated (unverified) commits in this PR from completely unrelated things.

Note that this PR cannot be merged before you fix your commit history.

components/Common/ShellBox/__tests__/index.test.tsx
```diff
@@ -0,0 +1,58 @@
+import React from 'react';
+import { render, screen } from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
+import { IntlProvider } from 'react-intl';
+import ShellBox from '../index';
+
+const mockWriteText = jest.fn();
+const originalNavigator = { ...window.navigator };
```
ovflowd, line 7: Nice!

AugustinMauroy: Thanks sweetie 😁

ovflowd: What?

AugustinMauroy: You say nice, I say thanks.

ovflowd: Ahm... gotcha. Appreciate that, but I'm not your "sweetie" 😅

hooks/__test__/useCopyToClipboard.test.tsx
```diff
@@ -0,0 +1,48 @@
...
+});
+
+describe('useCopyToClipboard', () => {
+  const TestComponent = ({ textToCopy }: { textToCopy: string }) => {
+    const [copied, copyText] = useCopyToClipboard();
+
+    return (
+      <IntlProvider locale="en">
+        <button onClick={() => copyText(textToCopy)} type="button">
+          <FormattedMessage id="components.codeBox.copy" values={{ copied }} />
```
HinataKah0: I haven't read through the code (so take it with a grain of salt) but...
Shouldn't this be `components.common.shellBox.copy`? Since I don't see any reference to `components.codeBox.copy`.

AugustinMauroy: it's fix.An oversight on my part

components/Common/ShellBox/__tests__/index.test.tsx
```diff
@@ -0,0 +1,59 @@
...
+
+  afterEach(() => {
+    Object.defineProperty(window, 'navigator', {
+      value: originalNavigator,
+    });
+  });
+
+  it('should render', () => {
+    const { container } = render(
+      <IntlProvider locale="en" messages={messages}>
```
shanpriyan: Remove `messages` from here (as per what we discussed) and pass `onError={() => {}}` to silence the errors from appearing in console.

components/Common/ShellBox/__tests__/index.test.tsx
```diff
@@ -0,0 +1,59 @@
...
+
+    Object.defineProperty(window.navigator, 'clipboard', {
+      writable: true,
+      value: {
+        writeText: navigatorClipboardWriteTextSpy,
+      },
+    });
+
+    render(
+      <IntlProvider locale="en" messages={messages}>
```
shanpriyan: Remove messages from here and also from other occurences of the test files.

components/Common/ShellBox/index.tsx
```diff
@@ -0,0 +1,43 @@
+import React, { useCallback } from 'react';
+import { FormattedMessage } from 'react-intl';
+import { useCopyToClipboard } from '../../../hooks/useCopyToClipboard';
+import styles from './index.module.scss';
+import ReactDomServer from 'react-dom/server';
+
+interface Props {
+  children: React.ReactNode | undefined;
```
shanpriyan:
```suggestion
```
I think there is no need to pass `children` type explicitly when using `React.PropsWithChildren<Props>`

components/Common/ShellBox/index.tsx
```diff
@@ -0,0 +1,43 @@
...
+
+const ShellBox = ({ children, textToCopy }: React.PropsWithChildren<Props>) => {
+  const [copied, copyText] = useCopyToClipboard();
+
+  const handleCopyCode = useCallback(
+    async (event: React.MouseEvent<HTMLButtonElement>) => {
+      event.preventDefault();
+      const text = textToCopy
+        ? textToCopy
+        : ReactDomServer.renderToString(<>children</>);
```
shanpriyan: Usage of `react-dom/server` should be avoided in client side since it will increase the bundle size of the app

[React DOCS](https://react.dev/reference/react-dom/server/renderToString#removing-rendertostring-from-the-client-code)

shanpriyan (dismissed):

util/jest-environment-jsdom.ts
```diff
@@ -0,0 +1,21 @@
+'use strict';
```
ovflowd: This sounds like overkill to create a new environment. As provided in the slack, pretty much the TextEncoder issue is coming from `isomorphic-dompurify`, we should probably just mock that dependency.

AugustinMauroy: Yeah why not.

AugustinMauroy: fix with shanpriyan review (about react server)

components/Common/ShellBox/index.tsx
```diff
@@ -0,0 +1,44 @@
...
+const ShellBox = ({ children, textToCopy }: React.PropsWithChildren<Props>) => {
+  const [copied, copyText] = useCopyToClipboard();
+
+  const handleCopyCode = useCallback(
+    async (event: React.MouseEvent<HTMLButtonElement>) => {
+      event.preventDefault();
+      const text = textToCopy
+        ? textToCopy
+        : // @ts-ignore - createRoot is not in the type definition
+          (createRoot(document.body).render(children) as string);
```
ovflowd: This needs to be corrected. Creating a root on `document.body` is really dangerous.

Let's simplify things here, I think we're overdoing ourselves here. If what we want is to copy the inner content of the ShellBox, this can be accomplished easily through references:

```tsx
  // ... remaining of the code of the element

  const shellBoxRef = useRef<HTMLCodeElement>();

  // ....
  // when copying content:
  // const text = textToCopy || shellBoxRef.current?.innerHTML;

  return (
    <pre className={styles.shellBox}>
      <div className={styles.top}>
        <span>SHELL</span>
        <button type="button" onClick={handleCopyCode}>
          <FormattedMessage
            id="components.common.shellBox.copy"
            values={{ copied }}
          />
        </button>
      </div>
      <code ref={shellBoxRef}>{children}</code>
    </pre>
  );
```

hooks/__test__/useCopyToClipboard.test.tsx
ovflowd: Please, this still needs to be fixed. It would be best if you did the same as from here: https://github.com/nodejs/nodejs.org/pull/5234/files#diff-71105d987a7c5eebb71d80e5a58ec56567f571a11a7f13aee6ef3d4aecfc0895R8

i18n/locales/en.json
```diff
@@ -37,6 +37,6 @@
   "layouts.blogPost.author.byLine": "{author, select, null {} other {By {author}, }}",
   "layouts.blogIndex.currentYear": "News from {year}",
   "components.common.banner.button.text": "Read More",
-  "components.article.author.githubLinkLabel": "{username} Github - opens in new tab",
```
ovflowd: Also it looks like you're deleting a translation key from here 🤔

AugustinMauroy: Fails from rebase -_-

components/Common/ShellBox/index.tsx
```diff
@@ -0,0 +1,41 @@
...
+
+const ShellBox = ({ children, textToCopy }: React.PropsWithChildren<Props>) => {
+  const [copied, copyText] = useCopyToClipboard();
+
+  const shellBoxRef = useRef<HTMLElement>(null);
+
+  const handleCopyCode = useCallback(
+    async (event: React.MouseEvent<HTMLButtonElement>) => {
+      event.preventDefault();
+      const text = textToCopy || shellBoxRef.current?.innerHTML;
```
ovflowd: Maybe we don't need a constant here, just `await copyText(textToCopy || shellBoxRef.current?.innerHTML);`

This also doesn't need to be an `useCallback` anymore as it doesn't do any complex computing anymore.

components/Common/ShellBox/index.stories.tsx
```diff
@@ -0,0 +1,12 @@
+import ShellBox from '.';
+
+export default {
```
ovflowd: Don't forget when refactoring to follow the Storybook example (default export goes after stories)

hooks/useCopyToClipboard.ts
```diff
@@ -0,0 +1,34 @@
+import { useEffect, useState } from 'react';
+
+const copyToClipboard = (value: string | undefined) => {
+  if (!value) {
+    return Promise.resolve(false);
+  }
+  if (typeof navigator === 'undefined') {
```
ovflowd: Just make it `if(!value || typeof navigator === 'undefined')` for simplicity

ovflowd (approved):
LGTM! We got there! Just left some minor comments here :)

components/Common/ShellBox/index.stories.tsx
```diff
@@ -0,0 +1,44 @@
...
+    children: (
+      <span>
+        <strong>$</strong>hello world
+      </span>
+    ),
+    textToCopy: 'hello world',
+  },
+};
+
+export default {
```
ovflowd: Can it be just `export default { component: ShellBox } as Meta;`?

components/Common/ShellBox/index.stories.tsx
```diff
@@ -0,0 +1,44 @@
...
+};
+
+export const WithoutTextToCopyJsx: Story = {
+  args: {
+    children: (
+      <span>
+        <strong>$</strong>hello world
+      </span>
+    ),
+    textToCopy: 'hello world',
```
ovflowd: This Story still has `textToCopy` 🤔

components/Common/ShellBox/index.stories.tsx
```diff
@@ -0,0 +1,44 @@
...
+  args: {
+    children: 'Hello World',
+  },
+};
+
+export const WithTextToCopyJsx: Story = {
+  args: {
+    children: (
+      <span>
+        <strong>$</strong>echo hello worl
```
ovflowd:
```suggestion
        <strong>$</strong>echo hello world
```

components/Common/ShellBox/index.tsx
```diff
@@ -0,0 +1,41 @@
...
+import styles from './index.module.scss';
+import { useCopyToClipboard } from '../../../hooks/useCopyToClipboard';
+import type { FC, PropsWithChildren, MouseEvent, ReactNode } from 'react';
+
+type ShellBoxProps = {
+  children: string | ReactNode;
+  textToCopy?: string;
+};
+
+const ShellBox: FC<ShellBoxProps> = ({
```
ovflowd: This should be:

```tsx
const ShellBox: FC<PropsWithChildren<ShellBoxProps>> = ({
  children,
  textTopCopy,
}) => { /* code of the component */ }
```

ovflowd: The Prop type you're setting on `FC` mismatches the one after the object. 
Also when you use `FC` you don't need to set the types of the props object :)

components/Common/ShellBox/index.tsx
```diff
@@ -0,0 +1,41 @@
+import { useRef } from 'react';
+import { FormattedMessage } from 'react-intl';
+import styles from './index.module.scss';
+import { useCopyToClipboard } from '../../../hooks/useCopyToClipboard';
+import type { FC, PropsWithChildren, MouseEvent, ReactNode } from 'react';
+
+type ShellBoxProps = {
+  children: string | ReactNode;
```
ovflowd, line 8: We shouldn't allow plain strings. Just ReactNode's (which in theory supports strings). But I guess this is fine...

ovflowd (approved):
Ladies and gentlemen. After 102 comments, we finally got this PR ready. It took you a while (and a lot of my time), but good stuff :)

:shipit:

ovflowd:
@AugustinMauroy I have the feeling styles are broken? https://nodejs-org-stories-git-fork-augustinmauroy-featshellbox-openjs.vercel.app/?path=/story/common-shellbox--default

I don't think the Component should look like that?

ovflowd:
I guess I found the culprit. I have made a small commit directly on `major/website-redesign` as a hot-fix.

mikeesto:
I don't have time to pull this down locally right now - but the styling still does not look right on what I think is the latest preview deployment? :S  https://nodejs-org-stories-92i3x8nba-openjs.vercel.app/?path=/story/common-shellbox--default

ovflowd:
Indeed the style looks very broken. I'm going to investigate but well, I guess this cannot be merged until the style gets fixed

AugustinMauroy:
> Indeed the style looks very broken. I'm going to investigate but well, I guess this cannot be merged until the style gets fixed

I know why it's broke.

1. They are confilct with old style (form site in handelbar)
2. An part of the style is an on over component

ovflowd:
@AugustinMauroy can you update your branch? That should fix the issues with the design (I guess)

AugustinMauroy:
Yes I can!

components/Common/ShellBox/__tests__/index.test.tsx
```diff
@@ -0,0 +1,57 @@
...
+      },
+    });
+
+    render(
+      <IntlProvider locale="en" onError={() => {}}>
+        <ShellBox textToCopy="test">test</ShellBox>
+      </IntlProvider>
+    );
+    const button = screen.getByRole('button');
+    userEvent.click(button);
```
ovflowd: Your tests are failing because this is an asynchronous method. 

This should pretty much be:

```tsx
// beginning of test
const user = userEvent.init();

// .. rest of test

const button = screen.getByRole('button');
await user.click(button);
```

components/Common/ShellBox/index.module.scss
```diff
@@ -0,0 +1,75 @@
...
+  border-radius: 0.4rem;
+  box-sizing: border-box;
+  display: flex;
+  flex-direction: column;
+  font-family: var(--mono);
+  padding: 0 0 var(--space-48) var(--space-16);
+  position: relative;
+
+  code {
+    color: var(--pink5);
```
ovflowd, line 12: On `nodejs.dev`, we have extra styles that allow some optional colouring... Any reason you didn't migrate those styles? They would need to be migrated differently.

**But pretty much you could do:**

`strong` and each `span` inside `code` to have some right margin, and make maybe certain elements to have different colour...

**So if I had:**

```tsx
<strong>$</strong> <span>choco</span> install nvm
```

It could create similar results from the images below.

image
image

AugustinMauroy: the "second style" from component named "OSpannel" eg: "MacosPannel".
So I don't thing we should migrate se second "style"

ovflowd: I didn't get your answer, sorry?

components/Common/ShellBox/__tests__/index.test.tsx
```diff
@@ -0,0 +1,58 @@
+import { render, screen } from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
+import { IntlProvider } from 'react-intl';
+import ShellBox from '../index';
+
+const mockWriteText = jest.fn();
+const originalNavigator = { ...window.navigator };
+const user = userEvent.setup();
```
ovflowd: This usually should be set inside each `it` (at the beginning of the test case)

components/Common/ShellBox/index.module.scss
AugustinMauroy: I did not migrate the additional styles, because they are part of the `OSpannel` component [Example](https://github.com/nodejs/nodejs.dev/tree/main/src/components/DownloadComponents/InstallTabs).

So I don't think we should add this style with the migration of the componen

ovflowd: If you say so. Then sure... We can always migrate this at a later time.

ovflowd:
cc @nodejs/website can we get another approval here? :)

shanpriyan (commented):
can we add a `cursor: pointer;` to the copy button 
Screenshot 2023-04-25 at 1 12 49 AM

shanpriyan (approved):
LGTM! :shipit: 

Nice contribution @AugustinMauroy 🌟

mikeesto (commented):
I'm just leaving a few comments of things that I have noticed here. There still seems to be some differences between this implementation and the one on nodejs.dev. I am mindful that this PR has been open for a while so perhaps some of these issues can be tackled in a subsequent PR.

1. When I press copy, it copies the command prompt symbol ($)
2. There is no space between the $ and the command
3. I'm not sure what the story "Without Text to Copy" is for. When I click copy, it still copies the text. Is this intended?
4. There doesn't seem to be any syntax highlighting. For example, `echo` on nodejs.dev would be coloured white whereas here everything is pink

ovflowd:
Yeah, @mikeesto your concerns are exactly the ones I had. (I think item 3 the Story meant to be that the text to copy comes from the `children` instead the `textToCopy` prop)

But for the other items, I honestly wish they were worked on. As I'm not the only one having concerns about them.

AugustinMauroy:
If I have understood correctly, I have answered your questions @mikeesto 😁

mikeesto:
Thanks @AugustinMauroy . Sorry to be a pain but the `Without Text to Copy Jsx` story doesn't seem quite right. When I copy that, I get `<span><strong>$</strong>hello world</span>` on my clipboard?

Could you also confirm if the syntax highlighting is meant to be part of this PR?

AugustinMauroy:
Yup they are a problem with JSX and copy. and style is not an highlight it's style adde by [InstallTabs](https://github.com/nodejs/nodejs.dev/tree/main/src/components/DownloadComponents/InstallTabs). 
But for the JSX copy I don't have any idea. RenderToString from react-server so it will increase the bundle size so no good. createRoot was not good because it's create a root 😆. So I'm void of idea.

ovflowd:
> story doesn't seem quite right. When I copy that, I get <span><strong>$</strong>hello world</span> on my clipboard?

The purpose of that Story is just to remove the "textToCopy" and use the children as the "text" to copy. So the Story is working correctly, of course the children could be updated to remove the $, but it's just an example.

components/Common/ShellBox/__tests__/index.test.tsx
```diff
@@ -0,0 +1,58 @@
+import { render, screen } from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
+import { IntlProvider } from 'react-intl';
+import ShellBox from '../index';
+
+const mockWriteText = jest.fn();
+const originalNavigator = { ...window.navigator };
+describe('ShellBox', () => {
+  const user = userEvent.setup();
```
ovflowd: This should be inside the beginning of each `it`

components/Common/ShellBox/index.tsx
```diff
@@ -0,0 +1,41 @@
...
+  children,
+  textToCopy,
+}: PropsWithChildren<ShellBoxProps>) => {
+  const [copied, copyText] = useCopyToClipboard();
+
+  const shellBoxRef = useRef<HTMLElement>(null);
+
+  const handleCopyCode = async (event: MouseEvent<HTMLButtonElement>) => {
+    event.preventDefault();
+    await copyText(textToCopy || shellBoxRef.current?.innerHTML);
```
ovflowd: If we want to make a "hack" to remove `$`

```suggestion
    // By default we want to use the textToCopy props but if it's absent
    // we allow the user to copy by getting the inner HTML content of the Element
    const _textToCopy = textToCopy || shellBoxRef.current?.innerHTML || '';

    await copyText(_textToCopy.replace('$', '');
```

ovflowd: I know @AugustinMauroy you said you had no idea how to accomplish this, but here's a simple solution.

ovflowd (approved):
LGTM! We're ready to get this merged, but I left two little comments if they could be addressed :)

mikeesto (approved):

AugustinMauroy:
I thinks we can merge

PR #5236, opened 2023-04-05
feat(DarkModeToggler): Migrate and add stories to theme toggler 🎉

shanpriyan:
## Description

- Migrated `DarkThemeToggler` from nodejs.dev
- Added story to the component
- Updated the test file

## Related Issue

#5193


.eslintrc
```diff
@@ -13,7 +13,7 @@
     },
     {
       "files": ["**/*.{ts,tsx}"],
-      "globals": { "globalThis": false }
+      "globals": { "globalThis": false, "JSX": true }
```
ovflowd: I think the "jsx: true" rule should be on the `"files": ["**/*.tsx"],` rule not this one.

.storybook/main.js
```diff
@@ -5,6 +5,7 @@ const config = {
     '@storybook/addon-links',
     '@storybook/addon-essentials',
     '@storybook/addon-interactions',
+    'storybook-addon-sass-postcss',
```
ovflowd: Is this needed?

components/Common/DarkModeToggle/__tests__/index.test.tsx
```diff
@@ -0,0 +1,39 @@
+import React from 'react';
+import { render, screen } from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
+import DarkModeToggle from '../index';
+
+let mockCurrentTheme = '';
```
ovflowd, line 7:
```suggestion
let mockCurrentTheme = '';

```

components/Common/DarkModeToggle/index.stories.tsx
```diff
@@ -0,0 +1,22 @@
...
+const meta: Meta<typeof DarkModeToggle> = {
+  title: 'DarkModeToggle',
+  component: DarkModeToggle,
+};
+
+export default meta;
+
+type Story = StoryObj<typeof DarkModeToggle>;
+
+export const Toggle: Story = {
```
ovflowd: Is there a benefit on this approach over `https://github.com/nodejs/nodejs.org/blob/major/website-redesign/components/Header.stories.ts#L3`?

I suppose because our `app.tsx` doesn't has a `themeProvider` yet. Wouldn't it make sense to add the ThemeProvider here? https://github.com/nodejs/nodejs.org/blob/major/website-redesign/.storybook/preview.js#L26

components/Common/DarkModeToggle/index.tsx
```diff
@@ -0,0 +1,29 @@
...
+const DarkModeToggle = () => {
+  const [theme, toggleTheme] = useTheme();
+
+  const isDark = theme === 'dark';
+
+  return (
+    <button
+      type="button"
+      className={styles.darkModeToggle}
+      onClick={() => {
```
ovflowd: No need of `{}` here. Can simply be `onClick={() => toggleTheme()}`. Also, any reason why the `?.` Can the function be somehow null?

providers/themeprovider.tsx
```diff
@@ -0,0 +1,61 @@
+import React, { createContext, useEffect, useState } from 'react';
```
ovflowd: What about using this? https://www.npmjs.com/package/next-themes, sounds like a good abstraction layer for the theme provider?

manishprivet: LGTM

components/Common/DarkModeToggle/index.tsx
shanpriyan: > Also, any reason why the ?. Can the function be somehow null?

Yeah ,ts compiler is throwing error, function could be `undefined` because of default context value

providers/themeprovider.tsx
shanpriyan: Let me look into it

components/Common/DarkModeToggle/index.stories.tsx
shanpriyan: my bad will add it there 😅

.storybook/main.js
shanpriyan: taken from #5234 , thought it is required to load Sass files for stories

providers/themeprovider.tsx
shanpriyan: `next-themes` is changing the color scheme and data-theme attribute of HTML only. We have handled all the dark theme styles with `body` tag with `.dark` className.

if we want to use `next-themes`  either we update the body tag with dark/light className when theme is getting toggled or change the existing stylesheet to select theme from `data-theme` attribute. Also I don't see a problem with current theme provider which suits our need.

ovflowd: > if we want to use next-themes either we update the body tag with dark/light className when theme is getting toggled or change the existing stylesheet to select theme from data-theme attribute. Also I don't see a problem with current theme provider which suits our need.

I think that's fine, honestly speaking. Let's switch to next-themes :)

.eslintrc
```diff
@@ -17,6 +17,9 @@
     },
     {
       "files": ["**/*.tsx"],
+      "globals": {
```
ovflowd: We don't need this, please remove 🙇

.storybook/preview.js
```diff
@@ -3,6 +3,7 @@ import App from '../pages/_app.mdx';
 import { pageProps } from './constants';
 import '../styles/styles.scss';
 import '../styles/tokens.scss';
+import { ThemeProvider } from 'next-themes';
```
ovflowd: This import should happen before the next/image import

components/Common/DarkModeToggle/index.tsx
```diff
@@ -0,0 +1,57 @@
...
+import BrightnessMediumIcon from '@mui/icons-material/BrightnessMedium';
+import styles from './index.module.scss';
+import { useTheme } from 'next-themes';
+
+const THEME = {
+  DARK: 'dark',
+  LIGHT: 'light',
+} as const;
+
+function updateBodyClass(theme: string) {
```
ovflowd: Let's convert this into a React Hook :) so that we don't fill this component with document logic

components/Common/DarkModeToggle/index.tsx
```diff
@@ -0,0 +1,57 @@
+import React, { useEffect } from 'react';
+import ModeNightIcon from '@mui/icons-material/ModeNight';
+import BrightnessMediumIcon from '@mui/icons-material/BrightnessMedium';
+import styles from './index.module.scss';
+import { useTheme } from 'next-themes';
+
+const THEME = {
```
ovflowd: Do we need a const? Isn't just a type enough? I think just a type would be enough. Can you check?

components/Common/DarkModeToggle/index.tsx
```diff
@@ -0,0 +1,57 @@
...
+    if (theme) updateBodyClass(theme);
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, []);
+
+  function toggleTheme(isKeyPress?: boolean) {
+    if (isKeyPress) {
+      return;
+    }
+
+    const newTheme = theme === THEME.DARK ? THEME.LIGHT : THEME.DARK;
```
ovflowd:
```suggestion
    const newTheme = theme === THEME.DARK ? THEME.LIGHT : THEME.DARK;

```

components/Common/DarkModeToggle/index.tsx
```diff
@@ -0,0 +1,57 @@
...
+  const { theme, setTheme } = useTheme();
+
+  const isDark = theme === 'dark';
+
+  useEffect(() => {
+    if (theme) updateBodyClass(theme);
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, []);
+
+  function toggleTheme(isKeyPress?: boolean) {
```
ovflowd: This should be an arrow function.

package.json
```diff
@@ -34,7 +34,11 @@
     "build-storybook": "storybook build"
   },
   "dependencies": {
+    "@emotion/react": "^11.10.6",
+    "@emotion/styled": "^11.10.6",
     "@mdx-js/react": "^2.3.0",
+    "@mui/icons-material": "^5.11.11",
+    "@mui/material": "^5.11.16",
```
ovflowd: @HinataKah0 here @shanpriyan is adding MUI Icons. Would MUI Icons also work for you?

package.json
```diff
@@ -34,7 +34,11 @@
     "build-storybook": "storybook build"
   },
   "dependencies": {
+    "@emotion/react": "^11.10.6",
```
ovflowd: Any reason why we need to add emotion here? I don't think we need it. Or is it a requirement of next-themes?

styles/_base.scss
```diff
@@ -106,3 +106,9 @@ blockquote {
 iframe {
   border: none;
 }
+
+
```
ovflowd:
```suggestion
```

ovflowd, line 110:
```suggestion
```

package.json
HinataKah0: Yes, MUI Icons will also work for `EditLink`.

However, I tried finding the usages of `@fortawesome`, it looks like it's consistently used across article related components (including "View as JSON"). "View as JSON" uses `faRobot` which doesn't seem to be available in MUI Icons. So it looks like we are going to need Font Awesome anyway unless we are changing others.

In this case, I would vote for keeping `EditLink` to use Font Awesome.

.eslintrc
shanpriyan: https://github.com/nodejs/nodejs.org/pull/5244#issuecomment-1503663833

components/Common/DarkModeToggle/index.tsx
shanpriyan: what's wrong with a normal `function` 🙃

package.json
shanpriyan: it's a peer of `@mui` icons

components/Common/DarkModeToggle/index.tsx
shanpriyan: `const` here is making the object **_readonly_** and prevent from mutating it accidentally.

.eslintrc
ovflowd: Sorry, but that is no argument. Please remove this from your PR, this is an unrelated change.

components/Common/DarkModeToggle/index.tsx
ovflowd: That's not what I meant. I meant why we need a constant. This should be a type, at best.

components/Common/DarkModeToggle/index.tsx
ovflowd: It doesn't make sense to write a named function within a react component when you're using a functional component with an arrow function approach. It's an antipattern.

package.json
ovflowd: I also see that MuiIcons don't have branded icons, such as GitHub Icon, which we need. @shanpriyan It might be better if we removed MUI Icons, and just keep fortawesome, or an Open Source alternative, but not sure if there's one, so I guess we can just use fortawesome.

.eslintrc
nschonni: Maybe your ESLint cache is bad. I just pulled the branch and ran `npm run lint:js` after `npm ci` and there is no error

package.json
shanpriyan: @ovflowd I believe you are referring to font awesome and not fort awesome (seems like fort awesome is a paid one)

ovflowd: No, I'm talking about FortAwesome. Please double check.

FortAwesome is the company/org behind FontAwesome ;)

.eslintrc
ovflowd: His branch doesn't use JSX types, he was trying to incorporate this "general fix" on his branch, hence why I said these are unrelated changes.

shanpriyan: @nschonni the error is not present in the existing branch, it appeared when we tried to migrate components from nodejs.dev. I believe removing this line will throw an error in CI, we will try an alternative fix.

ovflowd: Removing this line will not cause an issue once this PR gets merged. And I know I am repeating myself, but I already mentioned that I'm not letting the JSX global PR land. Not in the way it is.

I honestly believe the PR introducing the JSX.Element type usages should remove that explicit JSX type, and that's it.

components/Common/DarkModeToggle/index.tsx
shanpriyan: instead of writing hard coded strings , I used enums

ovflowd: Not sure what you mean with "hard coded" strings. Also this is not really an "enumerator" per definition. A TypeScript type would fit better here as

`type THEME = 'DARK' | 'LIGHT'`

package.json
shanpriyan: How about [react-icons](https://react-icons.github.io/react-icons/search?q=github)? it does have branded icons like Github and Twitter icons

ovflowd: Why not? :)

ovflowd: But it seems to include many other icon libraries within itself. How does it stack for tree shaking, load optimisation, etc. Is there any benchmark comparison online?

ovflowd:
@shanpriyan we need a rebase here

package.json
ovflowd: @shanpriyan lemme know once you have an update here :)

ovflowd:
@shanpriyan I did a recreation of the branch and clean cherry-pick for you :)

ovflowd:
I have no idea why builds are failing. I've just cleared the GitHub Actions cache for this PR... Retrying build now :)

ovflowd:
Okay, the `package-lock` for this branch was kinda broken, fixed it and that fixed the builds also.

shanpriyan:
| Dark | Light |
| ---- | ---- |
| Screenshot 2023-04-15 at 7 47 03 PM | image |

with react-icons package

shanpriyan:
@ovflowd Done 🚀

components/Common/DarkModeToggle/index.tsx
```diff
@@ -0,0 +1,47 @@
...
+
+  return (
+    <button
+      type="button"
+      className={styles.darkModeToggle}
+      onClick={() => toggleTheme()}
+      onKeyPress={() => toggleTheme(true)}
+      aria-pressed={isDark}
+    >
+      <span className="sr-only">Toggle Dark Mode</span>
```
ovflowd: This should be a translation key.

components/Common/DarkModeToggle/__tests__/index.test.tsx
```diff
@@ -0,0 +1,40 @@
...
+    render(<DarkModeToggle />);
+    const toggle = await screen.findByText('Toggle Dark Mode');
+    await userEvent.click(toggle);
+    expect(mockCurrentTheme).toBe('light');
+  });
+
+  it('switches light theme to dark theme', async () => {
+    mockCurrentTheme = 'light';
+    render(<DarkModeToggle />);
+    const toggle = await screen.findByText('Toggle Dark Mode');
```
ovflowd: Instead of checking `findByText` (as text can change) you can check by the presence of the `light-mode-only` or `dark-mode-only` or the `aria-pressed`

(Same on the other test above)

hooks/useUpdateBodyClass.ts
```diff
@@ -0,0 +1,13 @@
+export const useUpdateBodyClass = () => {
```
ovflowd: This shouldn't be a hook; it can be a simple util mainly because it uses 0 Reactive APIs.

Now, the name of this function is a little bit misleading; even tho it accepts a "string" it pretty much has specific logic for the "dark/light" theme.

**Also, remember this will not work correctly with Next.js when built. Next.js will override the body class list on re-renders.** (This is usually not noticeable with dev mode, but we shouldn't modify the body/document directly as this is an anti-pattern for React applications).

**If you want to properly manage the theme based on the state of the theme, you should:**

- Use the `useTheme` within the `theme.tsx` file
- Add the className for `.dark` or `.light` in [this](https://github.com/nodejs/nodejs.org/blob/main/theme.tsx#L53) line

That would be a clean solution that doesn't involve dom manipulation.

styles/_base.scss
```diff
@@ -106,3 +106,14 @@ blockquote {
 iframe {
   border: none;
 }
+
+body:not(.dark) .dark-mode-only,
+.dark .light-mode-only {
+  display: none;
+}
+
+.light-mode-only,.dark-mode-only{
```
ovflowd: This doesn't seem to be Stylelint conforming. There's missing spacing/padding in this whole section, and a missing file-line ending.

ovflowd (changes requested):
I've made a few comments of things that pretty much need to be changed before we proceed with this :)

components/Common/DarkModeToggle/__tests__/index.test.tsx
ovflowd: Just so you know, you shouldn't add the English messages on the IntlProvider for unit testing. See: https://github.com/nodejs/nodejs.org/pull/5279

shanpriyan: So the snapshots will contain the keys(`components.common.myComponen.copyButton.title`) now instead of the `en` lang translation?

ovflowd: Exactly!

hooks/useUpdateBodyClass.ts
shanpriyan: >This shouldn't be a hook; it can be a simple util mainly because it uses 0 Reactive APIs.

 It was a normal function previously, I moved it into a hook after your suggestion


>If you want to properly manage the theme based on the state of the theme, you should:
>
>Use the useTheme within the theme.tsx file
>Add the className for .dark or .light in [this](https://github.com/nodejs/nodejs.org/blob/main/theme.tsx#L53) line
>That would be a clean solution that doesn't involve dom manipulation.

Sounds great 👍🏻, but I tried to change the theme using `data-theme` attribute which has been updated by the `next-themes` and it works great.

ovflowd: Oh yes, I suggested to make it a hook because I thought you would move the whole useTheme to there 🙃

But it was my bad to not actually look carefully for what the function was doing 😅

For Next.js we want to avoid doing manual DOM mutations as Next rewrites stuff.

components/Common/DarkModeToggle/__tests__/index.test.tsx
```diff
@@ -0,0 +1,58 @@
+import React from 'react';
+import { render, screen } from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
+import { LocaleProvider } from '../../../../providers/localeProvider';
+import DarkModeToggle from '../index';
+import { AppProps } from '../../../../types';
+
+const i18nData = {
```
ovflowd: **Reminder:** Remove the `i18nData`. You don't need it, neither you need the `LocaleProvider`, you can simply just import `IntlProvider` (so that it becomes agnostic to our internal logic)

components/Common/DarkModeToggle/index.tsx
```diff
@@ -0,0 +1,40 @@
+import React from 'react';
+import { useTheme } from 'next-themes';
+import { FormattedMessage } from 'react-intl';
+import {
+  MdLightMode as IconLightMode,
```
ovflowd: Do we need to import with `as`, the current names sound good enough, no?

styles/layout/_dark-theme.scss
```diff
@@ -161,4 +177,4 @@ body.dark {
   .header-background-fill {
     fill: $dark-black2;
   }
-}
+}
```
ovflowd: Missing line ending here 🤔

ovflowd: Looks like Stylelint is misconfigured!

components/Common/DarkModeToggle/index.tsx
shanpriyan: `MdNightLight` was bit confusing. whether it's a night/light, I thought it would be better to rename it 🤔

styles/layout/_dark-theme.scss
ovflowd: @shanpriyan I made a PR for fixing stylelint.

components/Common/DarkModeToggle/index.tsx
ovflowd: I guess it's better to keep the original names just for reference. Otherwise if we start arbitrarily changing names well, that's actually confusing.

ovflowd:
@shanpriyan could you rebase your branch? With the Storybook and Stylelint fixes merged, you can add the ThemeProvider to the `preview.tsx` file and then run `npm run format` to fix your SCSS automagically

shanpriyan:
@ovflowd All good to go 🚀

components/Common/DarkModeToggle/index.stories.tsx
```diff
@@ -0,0 +1,15 @@
...
+  render: () => <DarkModeToggle />,
```
ovflowd: Do you need to provide this render method? You're already providing the `component`

I guess this can be changed to `export const Default: Story = {}`

components/Common/DarkModeToggle/__tests__/index.test.tsx
```diff
@@ -0,0 +1,53 @@
...
+jest.mock('next-themes', () => ({
+  useTheme: () => {
+    return { theme: mockCurrentTheme, setTheme: mockToggleTheme };
+  },
+}));
+
+describe('DarkModeToggle Component', () => {
+  it('render dark mode toggle', () => {
+    const { container } = render(
+      <IntlProvider locale="en">
```
ovflowd: We also need the `onError={() => {}}` stuff here 👀

ovflowd (approved):
LGTM! I again just left one more comment and forgot to mention the global styles are missing new lines 🤔

components/Common/DarkModeToggle/__tests__/index.test.tsx
shanpriyan: oops! will add it

ovflowd (approved):
LGTM! Really good stuff! Thank you, @shanpriyan!!

cc @nodejs/website for more approvals/reviews

components/Common/DarkModeToggle/__tests__/index.test.tsx
```diff
@@ -0,0 +1,53 @@
...
+import DarkModeToggle from '../index';
+import { IntlProvider } from 'react-intl';
+
+let mockCurrentTheme = '';
+
+const mockToggleTheme = jest.fn().mockImplementation(() => {
+  mockCurrentTheme = mockCurrentTheme === 'dark' ? 'light' : 'dark';
+});
+
+// // mock dark mode module for controlling dark mode HOC behavior
```
HinataKah0: Nit: duplicate `//`

ovflowd:
```suggestion
// Mock dark mode module for controlling dark mode HOC behaviour
```

mikeesto (approved):

PR #5238, opened 2023-04-07
feat: migrate AnimatedPlaceholder component (#5238

HinataKah0:
Migrate AnimatedPlaceholder component from nodejs.dev and create a new Story.

Related to: https://github.com/nodejs/nodejs.org/issues/5193

### DIFF
--- a/components/Common/AnimatedPlaceholder/index.module.scss
+++ b/components/Common/AnimatedPlaceholder/index.module.scss
@@ -1,0 +1,47 @@
+@keyframes placeHolderShimmer {
+  0% {
+    background-position: -468px 0;
+  }
+  100% {
+    background-position: 468px 0;
+  }
+}
+
+.animatedBackground,
+%animated-background {
+  animation-duration: 1.25s;
+  animation-fill-mode: forwards;
+  animation-iteration-count: infinite;
+  animation-name: placeHolderShimmer;
+  animation-timing-function: linear;
+  background: #f6f6f6;
+  background: linear-gradient(to right, #f6f6f6 8%, #f0f0f0 18%, #f6f6f6 33%);
+  background-size: 800px 104px;
+  height: 96px;
+  position: relative;
+}
+
+.placeholder {
+  display: flex;
+  width: 100%;
+
+  &Image {
+    @extend %animated-background;
+
+    height: 40px;
+    margin-right: 5px;
+    min-width: 40px;
+  }
+
+  &Text {
+    width: 100%;
+  }
+
+  &TextLine {
+    @extend %animated-background;
+
+    height: 10px;
+    margin: 4px 0;
+    width: 100%;
+  }
+}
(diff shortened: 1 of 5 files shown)


HinataKah0:
Tested:

Screenshot 2023-04-07 at 5 22 37 PM
Screenshot 2023-04-07 at 5 23 33 PM

mikeesto:
Thanks for your contribution! Please have a look at the lint error: https://github.com/nodejs/nodejs.org/actions/runs/4637041766/jobs/8205572868?pr=5238

rodion-arr (commented):
We need to update PR description and replace `Fixes: #5193` with `Related to #5293` so it will not close entire linked issue after merge because it tracks multiple components migration

components/Common/AnimatedPlaceholder/index.tsx
```diff
@@ -0,0 +1,29 @@
...
+  height?: number;
+}
+
+const AnimatedPlaceholder = ({ children, width = 400, height = 50 }: Props) => (
+  <div
+    className={styles.placeholder}
+    style={{ width: `${width}px`, height: `${height}px` }}
+  >
+    {/* Prefer external skeleton structure or render default in case not passed */}
+    {children}
```
manishprivet:
```suggestion
    {children || (
      <>
        <div className={styles.placeholderImage} />
        <div className={styles.placeholderText}>
          <div className={styles.placeholderTextLine} />
          <div className={styles.placeholderTextLine} />
        </div>
      </>
    )}
```

manishprivet: We can probably write it like this to prevent double checks on children

HinataKah0: Thanks, it looks neater! Tested as well.

HinataKah0:
Ah,
I saw the merge commit is being overwritten.

FYI, I usually use Rebase workflow rather than adding more commits if it is atomic enough.

components/Common/AnimatedPlaceholder/index.tsx
```diff
@@ -0,0 +1,28 @@
...
+  width?: number;
+  height?: number;
+}
+
+const AnimatedPlaceholder = ({ children, width = 400, height = 50 }: Props) => (
+  <div
+    className={styles.placeholder}
+    style={{ width: `${width}px`, height: `${height}px` }}
+  >
+    {/* Prefer external skeleton structure or render default in case not passed */}
```
ovflowd: Not sure we need an inline comment here.

components/Common/AnimatedPlaceholder/index.module.scss
```diff
@@ -0,0 +1,47 @@
...
+  background-size: 800px 104px;
+  height: 96px;
+  position: relative;
+}
+
+.placeholder {
+  display: flex;
+  width: 100%;
+
+  @keyframes placeHolderShimmer {
```
ovflowd: Can we remove the keyframes from this level? As keyframes in theory should not be contained within a class?

ovflowd (approved):
LGTM! Thank you for your first contribution, @HinataKah0! Left two small comments that you can address at your own discretion :)

mikeesto (approved):

shanpriyan (approved):

PR #5240, opened 2023-04-09
Migrate Dropdown Component

harshitphoenix:
Migrate Dropdown component from nodejs.dev and created a new story.

Related to: https://github.com/nodejs/nodejs.org/issues/5193


hooks/useElementPositionAndSize.ts
```diff
@@ -0,0 +1,40 @@
+import { useState, useEffect } from 'react';
+
+export interface ElementPositionAndSize {
```
manishprivet, line 3: This should go into the `types` folder

manishprivet:
@harshitphoenix Thanks for the first contribution 🙌🏼 

LGTM, just added one comment. Also seems like you will have to run `npm run lint` and `npm run build` to check and fix the lint and build errors

components/Common/Dropdown/Dropdown.stories.tsx
```diff
@@ -0,0 +1,5 @@
+import Dropdown from ".";
+
+export default { component: Dropdown };
+
+export const Default = {};
```
ovflowd: Hmm, git is accusing there's no ending line here. Could you add an empty line here=

components/Common/Dropdown/__tests__/Dropdown.test.tsx
```diff
@@ -0,0 +1,50 @@
+import React, { RefObject } from 'react';
+import { render, screen } from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
+import Dropdown from '..';
+
+describe('Dropdown component', () => {
+  const ref = {
```
ovflowd, line 7: Is all this extra casting needed here? I feel there are ways we could simplify the casting here 🤔 Maybe we could use the `satisfy` operator? https://dev.to/ayc0/typescript-49-satisfies-operator-1e4i

components/Common/Dropdown/index.module.scss
```diff
@@ -0,0 +1,30 @@
+.dropdown-list {
```
ovflowd, line 1: I think we should use camelCase classnames here

hooks/useElementPositionAndSize.ts
ovflowd: And probably be a type and not interface

components/Common/Dropdown/index.tsx
```diff
@@ -0,0 +1,62 @@
...
+import { DropdownItem } from '../../../types';
+import styles from './index.module.scss';
+
+interface Props<T> {
+  items: Array<DropdownItem>;
+  shouldShow: boolean;
+  elementRef: React.RefObject<T>;
+}
+
+const Dropdown = <T extends HTMLElement>({
```
ovflowd, line 13: I think on the other components we did not destructure props on component declaration?

components/Common/Dropdown/index.tsx
```diff
@@ -0,0 +1,62 @@
...
+  shouldShow,
+  elementRef,
+}: Props<T>): JSX.Element | null => {
+  const outerElementPosition = useElementPositionAndSize(elementRef);
+
+  const dropdownRef = useRef<HTMLUListElement>(null);
+
+  const mappedElements = items.map(item => {
+    const extraStyles = {
+      fontWeight: (item.active ? 'bold' : 'normal') as 'bold' | 'normal',
```
ovflowd, line 24: Is a casting of `as bold | normal` needed here?

ovflowd (approved):
Nice stuff! Amazing first contribution! Thank you 🙇 

Besides of the few comments I've left, I think we're good to go (after they get addressed)

ovflowd:
@manishprivet actually the command for automatically fixing the lint errors is `npm run format`.

hooks/useElementPositionAndSize.ts
```diff
@@ -0,0 +1,40 @@
...
+      const x = elementRef.current.offsetLeft || 0;
+      const y = elementRef.current.offsetTop || 0;
+      const width = elementRef.current.offsetWidth || 0;
+      const height = elementRef.current.offsetHeight || 0;
+
+      setPosition({ x, y, width, height });
+    }
+  };
+
+  useEffect(() => getPosition(), [elementRef.current]);
```
shanpriyan, line 32:
```suggestion
  useEffect(() => getPosition(), []);
```
Adding ref in deps wouldn't call the useEffect again

hooks/useElementPositionAndSize.ts
```diff
@@ -0,0 +1,40 @@
+import { useState, useEffect } from 'react';
```
shanpriyan:
```suggestion
import React, { useState, useEffect } from 'react';
```

SInce you are using `React` in following lines

shanpriyan:
- Run `npm run format` to fix prettier lint errors


- image

Check `.eslintrc` in #5236 and add it here (or may be we should raise a separate PR for that cc: @ovflowd , @manishprivet )

shanpriyan (approved):
LGTM! Thanks for the contribution ⭐️

I have left a few comments to fix the lint errors.

components/Common/Dropdown/index.tsx
```diff
@@ -0,0 +1,62 @@
...
+  items: Array<DropdownItem>;
+  shouldShow: boolean;
+  elementRef: React.RefObject<T>;
+}
+
+const Dropdown = <T extends HTMLElement>({
+  items,
+  shouldShow,
+  elementRef,
+}: Props<T>): JSX.Element | null => {
```
ovflowd: You don't need `JSX.Element` here. You can remove this explicit return type from here :)

hooks/useElementPositionAndSize.ts
ovflowd: Are we importing react on all other components? If yes, we should enforce the React import then :)

hooks/useElementPositionAndSize.ts
ovflowd: But, I think we should actually pass this reference, as we should recalculate if the current element changes, no?

shanpriyan: but even if we pass the ref in deps, it will not trigger the useEffect and re-calculate on new ref. The result will be the same. `getPosition` will be invoked only after mounting by useEffect. The only way the useEffect gets triggered is a change in ref value and then a re-render due to state change(PS: even the eslint rule advises the same to remove the ref from deps)

Ref: https://codesandbox.io/s/xenodochial-snowflake-hhgr6?fontsize=14&hidenavigation=1&theme=dark

I believe the recalculation is being happened by the `resize` listener below.

ovflowd: ESLint advising on unused props or missing props is not a mechanism of verifying if a prop is needed or not.

You can pass whatever you want on the useEffect "dependencies", as the "dependencies" tell when the effect should be re-run or not.

Sure, I have not checked if they actually use the reference there or if recalculating due to reference change would make sense here, but it's a common practice that if the element you're using as a ref changed and you're making a calculation (even of an element not directly related to that ref) to update something else.

ovflowd: Note.: I have no idea if a hook is even needed, I haven't checked what these effects and hooks are doing.

hooks/useElementPositionAndSize.ts
shanpriyan: This import is for `React.RefObject` usage.

ovflowd: Yeah I saw that, when I rebased your PR I saw that.

aymen94 (approved):

ovflowd:
@shanpriyan does this Dropdown component conflict with the one you created?

shanpriyan:
> @shanpriyan does this Dropdown component conflict with the one you created?

It's not conflicting I believe

components/Common/Dropdown/index.module.scss
shanpriyan, line 1:
```suggestion
.dropdownList {
```

ovflowd:
But @shanpriyan you made your own Dropdown component. If that gets merged first, what are the benefits of this one? It will be a direct conflict, no?

shanpriyan:
> But @shanpriyan you made your own Dropdown component. If that gets merged first, what are the benefits of this one? It will be a direct conflict, no?

@ovflowd  I'm not sure which DropDown Component you are referring to, As far as I remember I have not worked on any DropDown component 😄

ovflowd:
> > But @shanpriyan you made your own Dropdown component. If that gets merged first, what are the benefits of this one? It will be a direct conflict, no?
> 
> @ovflowd I'm not sure which DropDown Component you are referring to, As far as I remember I have not worked on any DropDown component 😄

Actually my bad, who did another Dropdown component was @ktssr, so I wonder if we merge that one or this one. I'm not sure what the differences are.

@ktssr this PR has a Hook that calculates positioning and stuff, did you not need to add that stuff there? What are your opinions? If the language selector PR gets merged it would probably kill the reason for this PR to exist to begin with 😅

ktssr:
@ovflowd, I didn't include the positioning hook in my changes because it seemed unnecessary. However, if you feel it's necessary, you can merge this PR first and I'll  make necessary changes accordingly.

ovflowd:
Hey @ktssr I don't know if it is necessary, and it might not be. I think we can proceed with your implementation, which is simpler and if we feel that something is missing in the future, we can always revisit the Dropdown component with what we saw in this PR.

I also feel like the author of this PR kinda abandoned the PR as he didn't reply to any of our comments yet lol

ovflowd:
Closing as another Dropdown component got merged.

PR #5244, opened 2023-04-10
chore(eslint): add JSX in eslint globals

shanpriyan:
In case of `JSX.Element` type, `JSX` is linted as not defined (`no-undef` rule) by eslint. To prevent that I have added it as globals

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -7,6 +7,13 @@
       "env": { "node": true, "es6": true }
     },
     {
+      "files": [
+        "**/**.test.{js,jsx,ts,tsx}"
+      ],
+      "extends": ["plugin:testing-library/react"],
+      "env": { "jest": true, "node": true, "browser": true }
+    },
+    {
       "files": ["**/*.{ts,tsx}"],
       "globals": { "globalThis": false }
     },
--- a/.eslintrc
+++ b/.eslintrc
@@ -7,9 +7,7 @@
       "env": { "node": true, "es6": true }
     },
     {
-      "files": [
-        "**/**.test.{js,jsx,ts,tsx}"
-      ],
+      "files": ["**/**.test.{js,jsx,ts,tsx}"],
       "extends": ["plugin:testing-library/react"],
       "env": { "jest": true, "node": true, "browser": true }
     },
@@ -74,6 +72,12 @@
         "no-unused-expressions": "off",
         "no-unused-vars": "off",
         "prefer-promise-reject-errors": "off"
+      }
+    },
+    {
+      "files": ["**/*.stories.{ts,tsx}"],
+      "rules": {
+        "import/no-anonymous-default-export": "off"
       }
     }
   ]
--- a/.eslintrc
+++ b/.eslintrc
@@ -17,6 +17,9 @@
     },
     {
       "files": ["**/*.tsx"],
+      "globals": {
+        "JSX": true
+      },
       "rules": {
         "react/no-unused-prop-types": "off",
         "react/require-default-props": "off",
--- a/.eslintignore
+++ b/.eslintignore
@@ -6,6 +6,8 @@ node_modules
 public/en/user-survey-report
 public/static/documents
 public/static/legacy
+
+# Test files
 tests
 coverage
 
@@ -13,3 +15,6 @@ coverage
 # be done in the future as we don't want to update the Markdown file
 # contents right now
 pages/**/*.md
+
+# Storybook
+storybook-static
--- a/.eslintignore
+++ b/.eslintignore
@@ -6,10 +6,8 @@ node_modules
 public/en/user-survey-report
 public/static/documents
 public/static/legacy
-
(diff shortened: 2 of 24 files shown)


AugustinMauroy (approved):
LGTM

AugustinMauroy:
Fast strack ?

nschonni:
Don't think this is the correct approach, see https://stackoverflow.com/a/36189032

.eslintrc
```diff
@@ -17,6 +17,9 @@
     },
     {
       "files": ["**/*.tsx"],
+      "globals": {
+        "JSX": true
```
ljharb, line 21: why is this a global?

AugustinMauroy: It's use to do this:
```tsx
const Components = ():JSX.Element =>{...}
```

shanpriyan: In case of `JSX.Element` type, `JSX` is linted as not defined (`no-undef` rule) by eslint, to prevent that

shanpriyan:
I would like to fast-track this since lot of people are working on migrating nodejs.dev components. 

cc: @nodejs/website

shanpriyan:
> Don't think this is the correct approach, see https://stackoverflow.com/a/36189032

This is not related to the PR

.eslintrc
ljharb: but that's a TS type, not a runtime value

ovflowd:
I need to agree with @ljharb here. There's no need for getting this merged. Why are we making this PR at all?

Is there any reference on why this is needed? Please add details.

Seeing that both @ljharb and @nschonni are against this change and myself included, fast-tracking is not possible.

nschonni:
> > Don't think this is the correct approach, see https://stackoverflow.com/a/36189032
> 
> This is not related to the PR

The answer in that link shows the normal way of declaring these variables in ESLint, rather than declaring JSX as a global

.eslintrc
ovflowd: Exactly as @ljharb you shouldn't be changing this file. If something needs to be changed is the tsconfig to have "jsx" as "detect" or "next".

ovflowd:
What I wonder is why do we need to make this change at all?

The setting that @nschonni mentioned should already be part of the default extended "next" configuration. So I wonder what are we missing here and which files are giving errors. I would love to have more context here.

.eslintrc
shanpriyan: True, the ts type has been identified as undefined by eslint

ljharb: are you using the TS parser?

shanpriyan:
> What I wonder is why do we need to make this change at all?
> 
> The setting that @nschonni mentioned should already be part of the default extended "next" configuration. So I wonder what are we missing here and which files are giving errors. I would love to have more context here.

When we try to use `JSX.Element` as return type for components , vscode shows error from eslint that 

`JSX` is not defined (from no-undef) 

Ref: https://stackoverflow.com/a/65828055

The link @nschonni shared is not related to the issue I'm talking about and is irrelevant to this PR.

ljharb:
I don't understand why the `no-undef` rule would be looking at type annotation nodes at all.

shanpriyan:
> I don't understand why the `no-undef` rule would be looking at type annotation nodes at all.

image

cc: @ovflowd, @nschonni

ljharb:
That doesn't seem possible to me; eslint core doesn't know about type annotation nodes. What happens on the command line? That's the real source of truth; if it only happens in the editor then the editor is broken.

ovflowd:
1. Adding `JSX` to globals is no fix. You're simply making ESLint ignore the fact the JSX is not defined.
2. `JSX.Element` is a type annotation; it should not be handled by the default ESLint parser. 
3. If the `typescript` parser is being used correctly, multiple things might be at play here. It could even be that we don't have the JSX-type definitions bundled.
4. We shouldn't create explicit return types as `JSX.Element` when using React. Either compose your Functional Component with `React.FC` on the declaration of your constant or return as a `React.ReactElement` (which should be avoided as TypeScript should make implicit inference here).

.eslintrc
ovflowd: @ljharb I suppose we are. (https://github.com/vercel/next.js/blob/canary/packages/eslint-config-next/index.js#L93) but we do not have the JSX types imported. (It might be that React is not exposing the JSX namespace anymore)

shanpriyan:
> That doesn't seem possible to me; eslint core doesn't know about type annotation nodes. What happens on the command line? That's the real source of truth; if it only happens in the editor then the editor is broken.

image

after running `npm run lint`

ljharb:
That is very bizarre :-/ no-undef isn't capable of visiting type annotation nodes, so I'm not sure what's going on.

ovflowd:
My comment still stands, we shouldn't be using JSX annotation types, regardless.

PR #5261, opened 2023-04-12
chore(workflow): remove dependabot workflow

AugustinMauroy:


ovflowd (approved):
LGTM! Let's remove Dependabot for now.

.github/dependabot.yml
```diff
@@ -1,12 +0,0 @@
-version: 2
-updates:
-  - package-ecosystem: github-actions
-    directory: '/'
-    schedule:
-      interval: weekly
-    open-pull-requests-limit: 10
-  - package-ecosystem: npm
```
ovflowd, line 8: Can we only remove the `npm` one and keep the `github-actions` one?

AugustinMauroy: Shure. We can also put evrythings on comment.

ghost (commented):
LTGM!

ghost (dismissed):

mikeesto (approved):
For context: I think the reason we are temporarily disabling Dependabot is because of the significant migration work happening on the `/major/website-redesign` branch, and trying to reduce the likelihood of dependency conflicts when it lands. A decrease in notifications within our GitHub inboxes is an added bonus too...

.github/dependabot.yml
ovflowd: I'd rather just remove. We can always add back by checking git history.

AugustinMauroy: Ok, lets go to remove.

ovflowd:
@MaledongGit dependabot alerts are about the security alerts feature. Not the PRs for dependency updatee.

aymen94 (approved):

ghost:
> @MaledongGit dependabot alerts are about the security alerts feature. Not the PRs for dependency updatee.

@ovflowd：

I submitted a PR to make a feature request to the dependabot, however the author of GitHub's waiting for my response, and I don't have the authentication to verify whether it works? Can any of you check it? I was told not only security alerts but updates as well……

Refs：https://github.com/dependabot/dependabot-core/issues/7072

If this still doesn't work, I'll submit and give him a reason why, otherwises I'll close this feature.

ovflowd:
@MaledongGit dependabot security alerts are an entirely different feature from dependabot automatic dependency updates, as I mentioned before.

Edit: It seems that we're able indeed to disable it via options, but I'm unable to find that option. Will give it another try.

.github/dependabot.yml
ovflowd: @AugustinMauroy I'm still waiting.

ovflowd (changes requested):
Still waiting for the unrelated changes to be undone 👀 (referring to remove the github-actions config for dependabot)

ghost:
> @MaledongGit dependabot security alerts are an entirely different feature from dependabot automatic dependency updates, as I mentioned before.
> 
> Edit: It seems that we're able indeed to disable it via options, but I'm unable to find that option. Will give it another try.

Yes, It would be better if we can keep the file :)

ovflowd:
I'm closing this PR as there's a few controversial things happening at play. I'll reach out GitHub support.

ghost:
Just a tip only (no need to change any more but just let you what to do if we meet such problems): 

https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuring-dependabot-version-updates#disabling-dependabot-version-updates

Just a very simple sample XD XD XD

PR #5269, opened 2023-04-13
Update theme.tsx

CodingRule:
created a H component to handle the different heading updated the highlightAll and startLegacyApp calls in Content to use the useEffect hook updated the ThemeProps interface to include a more specific type for the frontMatter property in pageOpts

should improve the codes readability maintainability and safety

### DIFF
--- a/theme.tsx
+++ b/theme.tsx
@@ -16,24 +16,36 @@ type LayoutProps = React.PropsWithChildren<{
   pageOpts: NextraThemeLayoutProps['pageOpts'];
 }>;
 
+type HProps = React.PropsWithChildren<{
+  level: number;
+}>;
+
+const H = ({ level, children }: HProps) => {
+  const Component = `h${level}` as keyof JSX.IntrinsicElements;
+  return <AnchoredHeading level={level} as={Component}>{children}</AnchoredHeading>;
+};
+
 const mdxComponents: MDXComponents = {
-  NodeApiVersionLinks: NodeApiVersionLinks,
-  h1: props => <AnchoredHeading level={1} {...props} />,
-  h2: props => <AnchoredHeading level={2} {...props} />,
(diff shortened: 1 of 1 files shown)


CodingRule:
I'm not the best But I think this could be good? if it's not atleast I tried Lol!

theme.tsx
```diff
@@ -16,24 +16,36 @@ type LayoutProps = React.PropsWithChildren<{
   pageOpts: NextraThemeLayoutProps['pageOpts'];
 }>;
 
+type HProps = React.PropsWithChildren<{
```
manishprivet, line 19: What is the purpose of this element? What problem does it solve?

theme.tsx
```diff
@@ -16,24 +16,36 @@ type LayoutProps = React.PropsWithChildren<{
...
+  h6: (props) => <H level={6} {...props} />,
   blockquote: ({ children }) => <div className="highlight-box">{children}</div>,
 };
 
 const Content = ({ children }: LayoutProps) => {
   const { asPath } = useRouter();
 
-  // Re-highlights the pages on route change
-  useEffect(() => highlightJs.highlightAll(), [asPath]);
+  useEffect(() => {
```
manishprivet, line 42: > updated the highlightAll and startLegacyApp calls in Content to use the useEffect hook

This seems like you just re-formatted it

theme.tsx
```diff
@@ -42,7 +54,6 @@ const Content = ({ children }: LayoutProps) => {
   );
 };
 
-// @TODO: Nextra should provide better customization to FrontMatter Props
```
manishprivet, line 45: Why are we removing this here?

ovflowd (changes requested):
These changes sadly bring no value. Also not sure what the H component is doing and why it is different.

This is very likely not going to land. Closing this PR.

PR #5271, opened 2023-04-14
feat: migrate EditLink component

HinataKah0:
Migrate EditLink component from nodejs.dev
and create a new Story.

Related to: https://github.com/nodejs/nodejs.org/issues/5242


components/Article/EditLink/index.tsx
```diff
@@ -0,0 +1,42 @@
...
+
+const EditLink = ({ relativePath, editPath, absolutePath }: Props) => {
+  const { currentLocale } = useLocale();
+
+  if (!relativePath && !editPath && !absolutePath) return null;
+
+  const href =
+    absolutePath ||
+    (relativePath
+      ? `${baseURL}/pages/${currentLocale.code}/${relativePath}`
```
HinataKah0: Need input for this.

Previous URL is `/content/learn/` and new URL is `/pages/{2-letters-language-code}`. Currently I am using `currentLocale`, not sure if it is the correct behavior.

ghost:
@HinataKah0 : Plz first to run the code formation check to verify your codes :)

Thank you for your 1st contribution!

```node
npm run prettier:fix
```

components/Article/EditLink/index.tsx
```diff
@@ -0,0 +1,42 @@
...
+
+interface Props {
+  relativePath?: string;
+  editPath?: string;
+  absolutePath?: string;
+}
+
+// TODO(HinataKah0): Change this
+const baseURL =
+  'https://github.com/nodejs/nodejs.org/edit/major/website-redesign';
```
MaledongGit, line 16: What will happen if we don't have translated files here? If we use this URL, this will directly guide us to change the English ones, and this won't be a right one.

If we don't have a translated file, the URL should be something like this:

https://github.com/nodejs/nodejs.org/new/major/website-redesign （Change "edit" to "new" in this URL）.

ghost (dismissed):
There're still many things I want to mention for the "Edit on GitHub", and this problem has happened in "nodejs.dev"：

## “Edit on GitHub”'s Problem

First let's see a bug (Sorry I cannot send it directly to the nodejs.dev, bcoz it's now archived, and no-one has reported the bug yet)：

1）Open the link something like this : https://nodejs.dev/zh-cn/about/

2）Click "Edit on GitHub"：
网页捕获_15-4-2023_103945_nodejs dev

3）Notice we don't have a simplified translated file here yet, so this will return you a 404 page instead!
网页捕获_15-4-2023_105748_github com


## Reason
We don't have real translated file yet, so we cannot use "edit" in the baseURL, as what I've mentioned in the code-review.

## Several Solutions:

### Solution 1：Smart Jump
Any languages with the "Edit on GitHub", when clicking it, it will jump to either the language itself (if we really have the page), or jump to the English page (to edit in English). Because we use translations in Crowdin instead of in GitHub.

### Solution 2："Edit" with an "Adding" function when without the translated page
1. If we don't the page, we use "new" in the base URL. Then an empty page will be returned to you in GitHub, where you can do whatever you can.
2. If we have the translated page, just keep "edit" in the base URL instead.

**PS: Do we really need "Edit on GitHub"?  Bcoz we've translated files in Crowdin instead of in GitHub, and if we do that, the default one will be in GitHub, is it suitable?**

### Solution 3："Edit on GitHub" in English and "Edit on Crowdin" for other translations
I tend to cope with this, bcoz we're translating and editing in Crowdin mainly. And another advantage is that no matter whether we've got the page or not, we can directly open the file in the English version, and choose the language you wanna to translate into. To simplifiy the problem ,we can just make the "Edit On GitHub"'s link something like this following:

https://crowdin.com/project/nodejs-website/(your current language prefix here)

Take the Simplified Chinese as an example:
https://crowdin.com/project/nodejs-website/zh-CN

Notice：Crowdin doesn't include English itself, maybe for English, we still need "Edit on GitHub", but "Edit on Crowdin" for other translations?

## Others
How do check the code just depends on how we think of the "Edit On GitHub"'s problem. And that's the reason why I use "Request changes" here. I'm not saying the code is "right" or "wrong", it should rely on discussions bcoz there're many things maybe we've missed before.

ghost:
Maybe "Edit On GitHub" for translations (for English pages/other translation pages) won't be too easy to migrate, it's nice for us to have a discussion here before deciding how to do XD

components/Article/EditLink/index.tsx
manishprivet: I believe it should be `{2-letters-language-code}/relativePath`

ovflowd:
Hey @HinataKah0 I've re-created your branch with the updated `major/website-redesign` so that you didn't need to rebase xD

Anyhow, can we use `react-icons` here instead? We're going to leave `mui-icons` and `fort-awesome` in favour of `react-icons`

HinataKah0:
@MaledongGit 

About lint issue:
It was resolved after merging `major/website-redesign`.

About redirection URL:
Thanks for bringing it up, it's a valid point.
+1 on it, I think let's wait for a while to hear inputs from others who understand the translation workflow as well.
I don't think it's worth to over-engineer the redirection logic. So I prefer to make it simple but enough to cater to our translation workflow.
Based on the `README`, initial development "usually" happens in English. So +1 for `en` -> "Edit on GitHub" and `xx` -> "Edit on Crowdin" currently.


@ovflowd 

Thanks! Sure, `react-icons` works for me.

It looks like I'll change my `copy-paste`-able (I am lazy in general) commands to not rebase to make it similar with everyone's workflow 😝

components/Article/EditLink/index.tsx
HinataKah0: Sorry my comment was a bit shallow & unclear. But my intention was similar to the discussion below about the redirection behavior.

ghost:
> So +1 for `en` -> "Edit on GitHub" and `xx` -> "Edit on Crowdin" currently.

Before posting this suggestions, I didn't anything related to it on README, is it you that is voting for this idea? Still on discussion or……?

But no matter what we do, if all of you can agree with the same idea or maybe a better one, that's nice, and I've cancelled my review now.

HinataKah0:
> is it you that is voting for this idea?

"+1" means I vote for the idea.

Actually, I think this is easily "revert"-able. So we can try any "easy-to-implement" idea first, then change it in the future if we see it doesn't fit well.

ovflowd:
I'm fine with whatever y'all decide :)

package.json
```diff
@@ -34,6 +34,8 @@
     "build-storybook": "storybook build"
   },
   "dependencies": {
+    "@fortawesome/free-solid-svg-icons": "^6.4.0",
```
HinataKah0: I'll change to `react-icons` after #5236 has been merged.

components/Article/EditLink/__tests__/__snapshots__/index.test.tsx.snap
```diff
@@ -43,18 +41,16 @@ exports[`EditLink component Translate renders correctly 1`] = `
...
       >
         <path
-          d="M410.3 231l11.3-11.3-33.9-33.9-62.1-62.1L291.7 89.8l-11.3 11.3-22.6 22.6L58.6 322.9c-10.4 10.4-18 23.3-22.2 37.4L1 480.7c-2.5 8.4-.2 17.5 6.1 23.7s15.3 8.5 23.7 6.1l120.3-35.4c14.1-4.2 27-11.8 37.4-22.2L387.7 253.7 410.3 231zM160 399.4l-9.1 22.7c-4 3.1-8.5 5.4-13.3 6.9L59.4 452l23-78.1c1.4-4.9 3.8-9.4 6.9-13.3l22.7-9.1v32c0 8.8 7.2 16 16 16h32zM362.7 18.7L348.3 33.2 325.7 55.8 314.3 67.1l33.9 33.9 62.1 62.1 33.9 33.9 11.3-11.3 22.6-22.6 14.5-14.5c25-25 25-65.5 0-90.5L453.3 18.7c-25-25-65.5-25-90.5 0zm-47.4 168l-144 144c-6.2 6.2-16.4 6.2-22.6 0s-6.2-16.4 0-22.6l144-144c6.2-6.2 16.4-6.2 22.6 0s6.2 16.4 0 22.6z"
-          fill="currentColor"
+          d="M497.9 142.1l-46.1 46.1c-4.7 4.7-12.3 4.7-17 0l-111-111c-4.7-4.7-4.7-12.3 0-17l46.1-46.1c18.7-18.7 49.1-18.7 67.9 0l60.1 60.1c18.8 18.7 18.8 49.1 0 67.9zM284.2 99.8L21.6 362.4.4 483.9c-2.9 16.4 11.4 30.6 27.8 27.8l121.5-21.3 262.6-262.6c4.7-4.7 4.7-12.3 0-17l-111-111c-4.8-4.7-12.4-4.7-17.1 0zM124.1 339.9c-5.5-5.5-5.5-14.3 0-19.8l154-154c5.5-5.5 14.3-5.5 19.8 0s5.5 14.3 0 19.8l-154 154c-5.5 5.5-14.3 5.5-19.8 0zM88 424h48v36.3l-64.5 11.3-31.1-31.1L51.7 376H88v48z"
```
manishprivet, line 55: Would be great if you could add a screenshot for this 😅

HinataKah0:
Edit (English):
Screenshot 2023-04-20 at 8 21 21 PM

Translate (Non-English):
Screenshot 2023-04-20 at 8 20 05 PM

components/Article/EditLink/__tests__/__snapshots__/index.test.tsx.snap
HinataKah0: react-icons:
Screenshot 2023-04-20 at 8 25 13 PM

Font Awesome:
https://fontawesome.com/v4/icon/pencil

How it looks like, you can see screenshots below.

ovflowd:
@HinataKah0 no need to add screenshots of Storybooks here. Vercel Bot already adds preview link for Storybook 👀

components/Article/EditLink/EditLink.stories.tsx
```diff
@@ -0,0 +1,34 @@
...
+  i18nMockDataNonEnglish,
+} from './mockDataConstants';
+
+export default { component: EditLink };
+
+export const Edit = {
+  args: {
+    relativePath: exampleRelativePath,
+  },
+  render: (args: any) => {
```
ovflowd: Can we use a decorator instead of render? https://storybook.js.org/docs/react/writing-stories/decorators

components/Article/EditLink/index.tsx
```diff
@@ -0,0 +1,54 @@
+import React from 'react';
```
ovflowd: After you rebase your branch this will start to fail/complain. Also, re-read the CONTRIBUTING guidelines as we updated a few things there ;d (Like the Props type name and being a type instead of interface)

components/Article/EditLink/EditLink.stories.tsx
```diff
@@ -0,0 +1,41 @@
...
+type Story = StoryObj<typeof EditLink>;
+type Meta = MetaObj<typeof EditLink>;
+
+export const Edit: Story = {
+  args: {
+    relativePath: exampleRelativePath,
+  },
+  decorators: [
+    Story => (
+      <LocaleProvider i18nData={i18nMockDataEnglish}>
```
ovflowd: We already have a constants file for all this. Check `.storybook/constants.js`.

Also we already inject i18n data by default on Storybook. No need to do this again.

components/Article/EditLink/EditLink.stories.tsx
ovflowd, line 1: This file should be `index.stories.tsx`

components/Article/EditLink/mockDataConstants.tsx
ovflowd, line 1: This file should not exist. The mocks for the URLs for the Stories should pretty much be on the same file as the Story.

components/Article/EditLink/index.tsx
```diff
@@ -0,0 +1,58 @@
...
+const EditLink: FC<EditLinkProps> = ({
+  absolutePath,
+  relativePath,
+  editPath,
+}) => {
+  const { currentLocale } = useLocale();
+
+  if (!relativePath && !editPath && !absolutePath) return null;
+
+  let href;
```
ovflowd: Instead of making `let` you can outsource this whole thing to a function e.g: `getEditLink`.

The function returns an object with `{ translationKey, href }`, and inside the function, you have only a switch statement with `return` statements. (So no need for a `let`).

Also `+=` on the translationKey isn't a good approach honestly.

ovflowd (commented):
Thanks for the work! I left some review comments :)

ghost:
Some tiny suggestions may be better....

1. Not all the characters are in the upper case?
2. "Non-English" can be ignored, just show what it is translated actually?

Something like this IMO：
English
NonEng

components/Article/EditLink/EditLink.stories.tsx
HinataKah0: `EditLink` logic relies heavily on the selected language.
Initially I wanted to make it explicit that we have language logic here by mocking the `i18nData` which is shared between Story and tests.

Did you mean something like this?
```jsx
<LocaleProvider i18nData={{ pageProps.i18nData..., currentLocale: { code: 'en' } }}>
<LocaleProvider i18nData={{ pageProps.i18nData..., currentLocale: { code: 'xx' } }}>
```
If yes, then I agree it is cleaner. Then for tests, I'll mock it inside `index.test.tsx`.

@MaledongGit Since it's related, let's discuss here.
I think let's keep it to English for now for both implementation simplicity and universality. Actually if I don't know the translated meaning, I'll assume they're also the same as "Edit on GitHub" which is incorrect.

components/Article/EditLink/EditLink.stories.tsx
HinataKah0: Will change

FYI, `Banner` and `AnimatedPlaceholder` are still using `ComponentName.stories.tsx` which is inconsistent with the others.

components/Article/EditLink/EditLink.stories.tsx
ovflowd: Honestly, we should not test different languages on the Storybook.

Otherwise, we need to add a global toggle on Storybook that allows you to switch languages.

And I guess the best way of doing so https://storybook.js.org/docs/react/essentials/controls would somewhat be to extend the preview or something.

But for now, I instead remove this from here. Also, having nested Providers isn't a good approach here.

components/Article/EditLink/index.tsx
HinataKah0: ~~About `+=`, are you referring to string concatenation performance issue?
Initially I was referring to [this](https://stackoverflow.com/questions/16696632/most-efficient-way-to-concatenate-strings-in-javascript). But there's a conflicting [here](https://moduscreate.com/blog/javascript-performance-tips-tricks/).~~

~~Actually I am not too sure myself. Do you mind suggesting?~~

Edit:
Don't mind this 😛

components/Article/EditLink/EditLink.stories.tsx
HinataKah0: > But for now, I instead remove this from here.

Sorry, I am just making sure that I got it correct.
So, you are suggesting to remove it, right?

I am okay with it since tests will cover the testing and if other people want to preview it on Storybook, they can just change the `currentLocale.code` value in `.storybook/constants.js`.

components/Article/EditLink/index.tsx
HinataKah0: Done, thanks!

Initially I was a bit hesitant because of the small amount of logic but it looks cleaner now. Having `break` statement there is quite ugly.

components/Article/EditLink/EditLink.stories.tsx
ovflowd: Yup, I'm fine with removing :)

components/Article/EditLink/index.tsx
```diff
@@ -0,0 +1,74 @@
...
+};
+
+const EditLink: FC<EditLinkProps> = ({
+  absolutePath,
+  relativePath,
+  editPath,
+}) => {
+  const { currentLocale } = useLocale();
+
+  if (!relativePath && !editPath && !absolutePath) return null;
```
ovflowd:
```suggestion
  if (!relativePath && !editPath && !absolutePath) {
    return null;
  }
```

components/Article/EditLink/mockDataConstants.tsx
ovflowd, line 1: Like I mentioned before, you don't need to mock any of this. We already provide a `LocaleProvider` directly on `.storybook/preview.tsx`

Please, if possible, delete this file :)

components/Article/EditLink/mockDataConstants.tsx
```diff
@@ -0,0 +1,26 @@
+import type { AppProps } from '../../../types';
+
+export const exampleAbsolutePath =
```
ovflowd: These should be set directly on the Storybook Args. Not as constants

components/Article/EditLink/mockDataConstants.tsx
HinataKah0: Thanks for reminding
Was about to remove
Done!

ovflowd (approved):
LGTM! Tremendous work :) Thank you for this amazing contribution <3

HinataKah0:
@MaledongGit Sorry I missed to address your comment.

About all uppercase, I did comparison:
Screenshot 2023-04-22 at 6 49 03 PM
Screenshot 2023-04-22 at 6 49 28 PM

Styling is always opinionated 😛 
TBH, I prefer the 2nd one as well (aesthetic). However, the 2nd one seems weaker in size compared to the paragraph above it which makes it less noticeable. So, if we want to change the style, I think we might need to increase the font size a bit or keep it as it is for now.

I think let's keep it as it is since it is very much a matter of taste (different people will have different opinions).

ghost:
> TBH, I prefer the 2nd one as well (aesthetic). However, the 2nd one seems weaker in size compared to the paragraph above it which makes it less noticeable.

Yes, I must admin it looks not noticeable, mixed with other Latin-words. 
Maybe prettier it with some styles (to change the colors of the font).
But that's not an important case. Just a suggestion left here.

ovflowd:
Can someone give the extra approval that this PR requires? :) cc @nodejs/website

mikeesto (approved):
LGTM, thanks for your contribution

PR #5277, opened 2023-04-15
feat: migrate AuthorList component and add story 🎉 

shanpriyan:
## Description
- MIgrated `AuthorList` component from nodejs.dev
- Added a new story for the component

## Issues
#5242


shanpriyan:
lint  and build fails due to 
 
Screenshot 2023-04-15 at 11 59 44 PM

ovflowd:
Well if that's the issue, just remove the JSX type references. (Build is only failing because Next.js runs ESLint at build time too)

shanpriyan:
> Well if that's the issue, just remove the JSX type references. (Build is only failing because Next.js runs ESLint at build time too)

Yeah, what should we use as an alternative 🤔 for JSX.Element type, ReactElement or ReactNode ?

ovflowd:
> Well if that's the issue, just remove the JSX type references. (Build is only failing because Next.js runs ESLint at build time too)

> Yeah, what should we use as an alternative 🤔 for JSX.Element type, ReactElement or ReactNode ?

I don't think we need explicit return types for functional components. TypeScript should infer them implicitly.

At best what you want is:
```js
const MyComponent: React.FC<MyProps> = props => someJSXCodeHere;
```

shanpriyan:
Yeah removed the explicit types.

components/Article/AuthorList/Author/__tests__/index.test.tsx
```diff
@@ -0,0 +1,32 @@
+import React from 'react';
+import { render } from '@testing-library/react';
+import { AppProps } from '../../../../../types';
+import { LocaleProvider } from '../../../../../providers/localeProvider';
+import messages from '../../../../../i18n/locales/en.json';
+import Author from '..';
+
+const i18nData = {
```
ovflowd: We shouldn't pass i18n data to unit tests. We shouldn't add translations/English text to unit tests and only render the translation keys.

Otherwise, we need to update the snapshots every time the translations get updated.

If we want to test the actual translations and fitting with the components, see the visual regression tests issue.

components/Article/AuthorList/Author/index.tsx
```diff
@@ -0,0 +1,46 @@
+import React from 'react';
+import { injectIntl, WrappedComponentProps } from 'react-intl';
+import styles from './index.module.scss';
+import authorPlaceholderImg from './placeholder-img.png';
```
ovflowd: This image should be placed in the public folder and be loaded with `next/image`. Also, if possible, an SVG would be preferred.

ovflowd: (I know you're using background attributes, but I rather using `next/image` as you can also use `placeholder` images with it.

components/Article/AuthorList/__tests__/authors-list.test.tsx
```diff
@@ -0,0 +1,23 @@
+import React from 'react';
+import { render } from '@testing-library/react';
+import { LocaleProvider } from '../../../../providers/localeProvider';
+import { AppProps } from '../../../../types';
+import messages from '../../../../i18n/locales/en.json';
+import AuthorsList from '..';
+
+const i18nData = {
```
ovflowd: Same as from: https://github.com/nodejs/nodejs.org/pull/5277/files#r1167790949

components/Article/AuthorList/index.stories.tsx
```diff
@@ -0,0 +1,9 @@
+import AuthorList from '.';
+
+export default { component: AuthorList };
+
+export const Default = {
+  args: {
+    authors: ['shanpriyan', 'ovflowd', 'manishprivet', 'unavailable-author'],
```
ovflowd: Can we put random people from GitHub instead of us? Just pick a few random users. I don't like that our names get listed just because we're the people doing this code.

i18n/locales/ar.json
```diff
@@ -34,6 +34,8 @@
   "components.header.buttons.toggleDarkMode": "Toggle dark/light mode",
   "components.pagination.next": "التالي | ",
   "components.pagination.previous": "السابق",
+  "components.author.githubLinkLabel": "{username} Github - opens in new tab",
   "layouts.blogPost.author.byLine": "{author, select, null {} other {By {author}, }}",
-  "layouts.blogIndex.currentYear": "News from {year}"
+  "layouts.blogIndex.currentYear": "News from {year}",
+  "containers.authorList.title": "Article Authors"
```
ovflowd: This should be `components.article.authorList.title`

i18n/locales/ar.json
```diff
@@ -34,6 +34,8 @@
   "components.header.buttons.toggleDarkMode": "Toggle dark/light mode",
   "components.pagination.next": "التالي | ",
   "components.pagination.previous": "السابق",
+  "components.author.githubLinkLabel": "{username} Github - opens in new tab",
```
ovflowd: This should also be added in the bottom. Any new translation key should be added in the bottom.

ovflowd: Also note, please only keep `en.json` as we discussed earlier. We don't want to add the new translation keys on the other language files. Crowdin will do that eventually for us :)

ovflowd (commented):
Good stuff! I left a few comments, but I think you're in the right path :)

components/Article/AuthorList/Author/index.tsx
shanpriyan: In case of unavailable users, placeholder is not being displayed as fallback, Is this fine?

image

shanpriyan: when using bg attribute, placeholder image will be rendered in case 404 authors

components/Article/AuthorList/index.stories.tsx
shanpriyan: Shall I pick the random users from `nodejs.dev` learn section

ovflowd: Sure!

components/Article/AuthorList/Author/index.tsx
ovflowd: I think on `next/image` you can set:

```jsx
onError={() => setImageSrcToPlaceHolder()}
```

Pretty much if the image gives 404 or any error you set the source to the fallback one 👀

components/Article/AuthorList/Author/__tests__/index.test.tsx
```diff
@@ -0,0 +1,25 @@
+import React from 'react';
+import { render } from '@testing-library/react';
+import Author from '..';
+import { IntlProvider } from 'react-intl';
+
+describe('Author component', () => {
+  it('renders correctly', () => {
+    const username = 'test-author';
+    const { container } = render(
+      <IntlProvider locale="en">
```
ovflowd: I forgot to mention, but @AugustinMauroy brought that to our attention, you might want to add `onError={() => {}}` so that the console.errors that are being emitted stop being emitted because of ReactIntl

components/Article/AuthorList/Author/index.tsx
ovflowd: Note that in that case the `src` must become a `useState`

ovflowd: @shanpriyan did it work well?

components/Article/AuthorList/Author/index.tsx
```diff
@@ -0,0 +1,52 @@
...
+        key={username}
+        target="_blank"
+        rel="noopener noreferrer"
+      >
+        <Image
+          alt=""
+          src={authorImg}
+          placeholder="blur"
+          blurDataURL="/placeholder-img.png"
+          width={Number(size)}
```
ovflowd: If we're doing number conversion here, shouldn't the prop just be a number?

components/Article/AuthorList/Author/index.tsx
```diff
@@ -0,0 +1,52 @@
...
+      <a
+        className={styles.link}
+        href={githubLink}
+        aria-label={translation}
+        key={username}
+        target="_blank"
+        rel="noopener noreferrer"
+      >
+        <Image
+          alt=""
```
ovflowd: The `alt` tag could be the username of the user

ovflowd (approved):
Amazing stuff! Just added my last two comments, after that we can merge this finally! 🎉

ovflowd:
BTW @shanpriyan looks like snapshots need to be updated!

components/Article/AuthorList/Author/index.tsx
shanpriyan: Yes, it did

shanpriyan: image

components/Article/AuthorList/Author/index.tsx
shanpriyan: updated it with number

ovflowd (approved):
Let's get this merged! cc @nodejs/website please give your reviews/approvals!

mikeesto (approved):

aymen94 (approved):

PR #5323, opened 2023-04-22
feat(blog) Migrate BlogCard component

Harkunwar:
This PR aims to migrate the BlogCard component as mentioned in #5308


mikeesto:
Thanks, see if you can fix the lint issue in `components/Blog/BlogCard/index.module.scss` https://github.com/nodejs/nodejs.org/actions/runs/4775331933/jobs/8489525157?pr=5323

i18n/locales/ca.json
```diff
@@ -35,5 +35,6 @@
   "components.pagination.next": "Següent | ",
   "components.pagination.previous": "Anterior",
   "layouts.blogPost.author.byLine": "{author, select, null {} other {By {author}, }}",
-  "layouts.blogIndex.currentYear": "News from {year}"
+  "layouts.blogIndex.currentYear": "News from {year}",
+  "components.blog.blogCard.author.by": "by"
```
ovflowd: You should only migrate the English translation key :)

Can you remove all other translations, pretty please 🙇

components/Blog/BlogCard/index.tsx
```diff
@@ -0,0 +1,46 @@
...
+import LocalizedLink from '../../LocalizedLink';
+import navigation from '../../../navigation.json';
+import type { BlogPost } from '../../../types';
+import type { FC } from 'react';
+
+const getBlogCategoryUrl = (category: string): string =>
+  `${navigation.blog.link}/${category}/`;
+
+const getBlogPostUrl = (slug: string) =>
+  slug.endsWith('/') ? slug : `${slug}/`;
```
ovflowd: AFAIK we don't use trailing slashes on links here, so no need to have this function

components/Blog/BlogCard/index.tsx
```diff
@@ -0,0 +1,46 @@
...
+const BlogCard: FC<Props> = ({ title, author, date, category, slug }) => {
+  const blogCategoryUrl = getBlogCategoryUrl(category);
+
+  return (
+    <div className={styles.blogCard}>
+      <div className={styles.title}>
+        <LocalizedLink href={getBlogPostUrl(slug)}>{title}</LocalizedLink>
+        <div className={styles.metadata}>
+          {category && (
+            <LocalizedLink className={styles.category} href={blogCategoryUrl}>
```
ovflowd: `getBlogCategoryUrl` can be added directly here, so you won't need the return statement and brackets surrounding the function.

components/Blog/BlogCard/index.tsx
```diff
@@ -0,0 +1,46 @@
...
+import type { BlogPost } from '../../../types';
+import type { FC } from 'react';
+
+const getBlogCategoryUrl = (category: string): string =>
+  `${navigation.blog.link}/${category}/`;
+
+const getBlogPostUrl = (slug: string) =>
+  slug.endsWith('/') ? slug : `${slug}/`;
+
+type Props = {} & Omit<BlogPost, 'file'>;
```
ovflowd: As per guidelines this should be named `type BlogCardProps` and it can be simplified to `type BlogCardProps = Omit<BlogPost, 'file'>`

ovflowd:
From the Stories, it looks like that:

- The colour of the title and font weights seem to be wrong
- The date formatting needs to be corrected? I guess you should use `FormattedDateTime` with something that makes "Month Day, Year" (e.g. "June, 21, 2022)
- After category, there should be a read time span

Example how it should look like:

image

Demo: https://nodejs.dev/en/blog

Harkunwar:
I've made the requested changes

components/Blog/BlogCard/__tests__/index.test.tsx
```diff
@@ -0,0 +1,39 @@
...
+import BlogCard from '..';
+
+jest.mock('next/router', () => ({
+  useRouter: jest.fn().mockReturnValue({}),
+}));
+
+describe('BlogCard component', () => {
+  it('renders correctly', () => {
+    const { container } = render(
+      <LocaleProvider
```
ovflowd: You should not provide the LocaleProvider here. Just se how other tests did testing here for i18n.

The tl;dr is just import the IntlProvider from react-intl, but without actually passing any data, just the onError callback. (See other tests)

The reasoning behind is that we don't want to have translations (even sources) of any kind on our unit tests as copy-write tends to change over time...

types/blog.ts
```diff
@@ -4,6 +4,7 @@ export interface BlogPost {
   date: string;
   category: string;
   slug: string;
+  readingTime: string;
```
ovflowd: We don't actually provide this from the sources of the blog post yet, right?

So maybe make this optional (nullable) and add a todo to verify in the future about the blog generation.

ovflowd (approved):
We're almost there. But impressive work!!

components/Blog/BlogCard/__tests__/index.test.tsx
Harkunwar: I tried using `IntlProvider` in the unit tests before but it breaks because `LocalizedLink` uses LocaleProvider. What I can do is mock LocaleProvider, which might solve the issue, so should I rather mock it?

Harkunwar: I've changed it to mocking `useLocale`. Let me know if that's sufficient.

types/blog.ts
Harkunwar: Updated!

Harkunwar:
@ovflowd I feel like there's a lot of duplication in the way we're writing unit tests and stories. If we could directly import stories in our unit tests, we'll have less duplication. Storybook already allows it, so would it be worth doing that?

Harkunwar:
If I add this to our `jest.setup.mjs`, it will allow us to use decorators defined in `.storybook/preview`.
```tsx
import { setProjectAnnotations } from '@storybook/react';
import * as globalStorybookConfig from './.storybook/preview';

setProjectAnnotations(globalStorybookConfig);

Object.defineProperty(window, 'matchMedia', {
    writable: true,
    value: jest.fn().mockImplementation((query) => ({
        matches: false,
        media: query,
        onchange: null,
        addListener: jest.fn(),
        removeListener: jest.fn(),
        addEventListener: jest.fn(),
        removeEventListener: jest.fn(),
        dispatchEvent: jest.fn(),
    })),
});
```

Then we won't need to use `IntlProvider` anymore and our snapshot tests could become as simple as:
```tsx
import { render } from '@testing-library/react';
import * as stories from '../index.stories';
import { composeStories } from '@storybook/react';

const { Default } = composeStories(stories);

jest.mock('next/router', () => ({
  useRouter: jest.fn().mockReturnValue({}),
}));

describe('BlogCard component', () => {
  it('renders correctly', () => {
    const { container } = render(<Default />);
    expect(container).toMatchSnapshot();
  });
});
```

And we can start using the stories in other tests as well instead of duplicating them multiple times.

ovflowd:
@Harkunwar feel free to create an issue and work on a PR updating all tests to the approach you're suggesting :)

shanpriyan (approved):
Looks great! 🚀

PR #5324, opened 2023-04-23
chore: migrate releases types

HinataKah0:
Migrate all types in types/releases.ts from nodejs.dev

See https://github.com/nodejs/nodejs.org/issues/5303#issuecomment-1518947513

### DIFF
--- a/types/releases.ts
+++ b/types/releases.ts
@@ -1,8 +1,6 @@
-type ReleaseType = 'Current' | 'LTS' | 'Maintenance' | 'End-of-life';
-
 export interface UpcomingReleaseData {
   releaseDate: string;
-  releaseType: ReleaseType;
+  releaseType: 'Current' | 'LTS' | 'Maintenance' | 'End-of-life';
   alreadyReleased: boolean;
 }
 
--- a/types/releases.ts
+++ b/types/releases.ts
@@ -1,11 +1,4 @@
-export const ReleaseTypes = {
-  current: 'Current',
-  lts: 'LTS',
-  maintenance: 'Maintenance',
-  endOfLife: 'End-of-life',
-} as const;
-
-type ReleaseType = (typeof ReleaseTypes)[keyof typeof ReleaseTypes];
+type ReleaseType = 'Current' | 'LTS' | 'Maintenance' | 'End-of-life';
 
 export interface UpcomingReleaseData {
   releaseDate: string;
--- a/types/releases.ts
+++ b/types/releases.ts
@@ -1,0 +1,37 @@
+// TODO(HinataKah0): Remove this, see https://github.com/nodejs/nodejs.org/issues/5303#issuecomment-1518947513
+/* eslint-disable no-unused-vars */
+
+export enum ReleaseTypes {
+  current = 'Current',
+  lts = 'LTS',
+  maintenance = 'Maintenance',
+  endoflife = 'End-of-life',
+}
+
+export interface UpcomingReleaseData {
+  releaseDate: string;
+  releaseType: ReleaseTypes;
+  alreadyReleased: boolean;
+}
+
+export interface UpcomingRelease {
+  title: string;
+  releases: UpcomingReleaseData[];
+}
+
+export interface NodeReleaseData {
+  version: string;
+  fullVersion: string;
+  codename: string;
+  isLts: boolean;
+  status:
+    | 'Maintenance LTS'
+    | 'Active LTS'
+    | 'Current'
+    | 'End-of-life'
+    | 'Pending';
+  initialRelease: string;
+  ltsStart: string | null;
+  maintenanceStart: string | null;
+  endOfLife: string;
+}
--- a/types/releases.ts
+++ b/types/releases.ts
@@ -1,16 +1,15 @@
-// TODO(HinataKah0): Remove this, see https://github.com/nodejs/nodejs.org/issues/5303#issuecomment-1518947513
-/* eslint-disable no-unused-vars */
+export const ReleaseTypes = {
+  current: 'Current',
+  lts: 'LTS',
+  maintenance: 'Maintenance',
+  endOfLife: 'End-of-life',
+} as const;
 
-export enum ReleaseTypes {
-  current = 'Current',
-  lts = 'LTS',
-  maintenance = 'Maintenance',
-  endoflife = 'End-of-life',
-}
+type ReleaseType = (typeof ReleaseTypes)[keyof typeof ReleaseTypes];
 
 export interface UpcomingReleaseData {
   releaseDate: string;
-  releaseType: ReleaseTypes;
+  releaseType: ReleaseType;
   alreadyReleased: boolean;
 }
 
--- a/types/releases.ts
+++ b/types/releases.ts
@@ -5,7 +5,7 @@ export interface UpcomingReleaseData {
 }
 
 export interface UpcomingRelease {
-  title: string;
+  name: string;
   releases: UpcomingReleaseData[];
 }
 
--- a/types/index.ts
+++ b/types/index.ts
@@ -15,6 +15,7 @@ export * from './layouts';
 export * from './navigation';
 export * from './nodeVersions';
 export * from './prevNextLink';
+export * from './releases';
 
 export interface AppProps {
   i18nData: Pick<LocaleContext, 'localeMessages' | 'currentLocale'>;


types/releases.ts
```diff
@@ -0,0 +1,37 @@
+// TODO(HinataKah0): Remove this, see https://github.com/nodejs/nodejs.org/issues/5303#issuecomment-1518947513
+/* eslint-disable no-unused-vars */
+
+export enum ReleaseTypes {
```
ovflowd: Do we need these types? Don't we already have types here? If not that's fine :D

types/releases.ts
```diff
@@ -0,0 +1,37 @@
+// TODO(HinataKah0): Remove this, see https://github.com/nodejs/nodejs.org/issues/5303#issuecomment-1518947513
+/* eslint-disable no-unused-vars */
```
ovflowd: Do we really need this eslint ignore?

ovflowd (approved):
LGTM!

types/releases.ts
HinataKah0: Yes, if we define the enum values, it will upset ESLint because they aren't used.

types/releases.ts
HinataKah0: After digging, it seems like it's only used in `UpcomingReleasesItem` and `UpcomingReleasesPanel`. Initially I want to move the types to component level but it looks like we can't because the data is propagated from page level.

Quick search, I can't find any similar enum.

ovflowd: What I mean to say is that we already have download pages? How are we populating these? I guess all statically or something?

But yeah if these types are necessary and used in more than one component then it's okay to leave them here (types/) folder

mikeesto (approved):

types/releases.ts
HinataKah0: The download page we have is the old one (same with current nodejs.org).
Should be all statically (GraphQL query once in the top level).
Since we are migrating the new download page, so we need these.

I hope I didn't misunderstand you 🤣

aymen94 (approved):

types/releases.ts
shanpriyan: instead of enum, can we try

```js
const ReleaseTypes = {
current: 'Current',
  lts: 'LTS'
} as const
```

eslint wouldn't bother this and also better than enum.

shanpriyan (approved):
LGTM!

types/releases.ts
ovflowd: I also think we could just make a TypeScript type instead of an enum, no?

HinataKah0: Yes, it also works. I've tested it and ESLint isn't upset.
I read [this](https://sergiocarracedo.es/typescript-enums/). It looks like Typescript doc also recommends using `as const`.
Thanks for the input!
```diff
diff --git a/types/releases.ts b/types/releases.ts
index 86766f77..1666d7d7 100644
--- a/types/releases.ts
+++ b/types/releases.ts
@@ -1,19 +1,24 @@
-// TODO(HinataKah0): Remove this, see https://github.com/nodejs/nodejs.org/issues/5303#issuecomment-1518947513
-/* eslint-disable no-unused-vars */
+export const ReleaseTypes = {
+  current: 'Current',
+  lts: 'LTS',
+  maintenance: 'Maintenance',
+  endOfLife: 'End-of-life',
+} as const;
 
-export enum ReleaseTypes {
-  current = 'Current',
-  lts = 'LTS',
-  maintenance = 'Maintenance',
-  endoflife = 'End-of-life',
-}
+type ReleaseType = (typeof ReleaseTypes)[keyof typeof ReleaseTypes];
 
 export interface UpcomingReleaseData {
   releaseDate: string;
-  releaseType: ReleaseTypes;
+  releaseType: ReleaseType;
   alreadyReleased: boolean;
 }
 
+let testNoError = {
+  releaseDate: '',
+  releaseType: ReleaseTypes.current,
+  alreadyReleased: true,
+} as UpcomingReleaseData;
+
 export interface UpcomingRelease {
   title: string;
   releases: UpcomingReleaseData[];
```

types/releases.ts
```diff
@@ -0,0 +1,37 @@
...
+  endoflife = 'End-of-life',
+}
+
+export interface UpcomingReleaseData {
+  releaseDate: string;
+  releaseType: ReleaseTypes;
+  alreadyReleased: boolean;
+}
+
+export interface UpcomingRelease {
```
HinataKah0, line 7: I just realized this.
Nit: I am thinking of renaming this to be `UpcomingReleases`. Do we have any naming convention?
What do you think?

mikeesto: `UpcomingReleases` sounds fine to me if you wish to change it

types/releases.ts
ovflowd: > Should be all statically (GraphQL query once in the top level).

Just a note, we don't intend to have GraphQL at all on the new node.js website 😅 

> Since we are migrating the new download page, so we need these.

I understand we need the new types for the Components themselves, that's fine.

types/releases.ts
ovflowd: I think what I'd suggest is to not create a variable at all. Just a TypeScript type.

E.g., we don't need an enumerator/variable for this. Just a type.

So if in the past you had something like:

```tsx
type Props = {
 releaseType: ReleaseTypes // this is enum
}

// usage of component
<ReleaseWhatever releaseType={ReleaseTypes.LTS} />
```

you would have the following with a plain type:

```tsx
type ReleaseType = 'LTS' | 'Current' | ...

type Props = {
 releaseType: ReleaseType // this is a type
}

<ReleaseWhatever releaseType="LTS" />
```

types/releases.ts
ovflowd: Where was this `UpcomingRelease` being used? Because it has a title and releases type?

types/releases.ts
HinataKah0: Ah, missed this 🤯 
I searched and saw many others do the same as well, thanks!

Edit:
Same style is used in `status` exactly below too 😓 
I checked the usages, it should be similar with how `status` in `NodeReleaseData` is being used

types/releases.ts
HinataKah0: It is used in the utility function [`getUpcomingReleases`](https://github.com/nodejs/nodejs.dev/blob/main/src/util/getUpcomingReleases.ts#L9-L12) and the component [`UpcomingReleases`](https://github.com/nodejs/nodejs.dev/blob/main/src/components/DownloadComponents/UpcomingReleases/index.tsx#L7-L9) alongside its' parent [`DownloadReleases`](https://github.com/nodejs/nodejs.dev/blob/main/src/components/DownloadComponents/DownloadReleases/index.tsx#L8-L11) (both usages are Prop's type).

`UpcomingReleaseData` is derived from `NodeReleaseData`. `UpcomingRelease` is simply a collection of `UpcomingReleaseData`.

I just realized it conflicts with the component name if we name it to be `UpcomingReleases` 😅 
An alternative will be `UpcomingReleasesData` so it's consistent with others within the same file

ovflowd: I mean why the type has an array of releases and a title as its type definition.

I'm more questioning the current composition of the type. Maybe we don't even need this type at all or we can change how the components are when we migrate them.

Feels kinda weird an "UpcomingRelease" with a title and multiple releases on it. Even if it is "UpcomingReleases" (what is the meaning of the title type?)

types/releases.ts
ovflowd: Feel free to resolve this conversation once you've made these changes :)

types/releases.ts
HinataKah0: Ah, sorry I misunderstood you.

Screenshot 2023-04-28 at 5 50 31 PM

`title` is actually a bit misleading :)
It is actually the major release version. So, `UpcomingRelease` basically groups all the releases by their major version.

~~`UpcomingMajorVersionReleases` or something along that line.~~

ovflowd: Can we rename this to `upcomingRelease.name` then?

HinataKah0: I think it's better to be more explicit `upcomingRelease.majorVersion`?

ovflowd: It's kinda implicit that the `name` is the name of a release. (major)

Kinda verbose naming it `majorVersion` (IMHO)

HinataKah0: > name of a release

Hmm, yeah, it actually makes sense.
+1 for `name`

HinataKah0: After looking at the implementation, I think `UpcomingRelease` is actually making sense :)
I tried thinking of many different names but they don't really make sense for me.
Thanks for the inputs!

richardlau: I don't want to derail this PR, but IMO the whole https://nodejs.dev/en/download/ page is a worse UX than the existing https://nodejs.org/en/download. Getting to the actual downloads (if you need something other than the three prominent buttons at the top of the page) requires more clicks and is after a section that overlaps with https://nodejs.dev/en/about/releases/.

ovflowd: @richardlau I'd say that's a topic for a Discussion (Discussions Tab). I definitely agree that the new redesigned Downloads page could show other platforms like we do on nodejs.org, but maybe on the bottom.

That's something to design/spec-out but not on this PR 😅

ovflowd: > Getting to the actual downloads (if you need something other than the three prominent buttons at the top of the page) requires more clicks and is after a section that overlaps with

Yeah that page should be removed tbh. But I disagree that it requires more clicks. You just need to scroll down to see more platforms at the bottom. But we should add even more platforms there and useful links.

ovflowd:
@HinataKah0 could you rebase?

HinataKah0:
Rebased, shall we proceed? I think it should be ready 😄

PR #5330, opened 2023-04-25
feat: migrate codebox component

araujogui:
Issue #5242
Migrate Codebox component


ovflowd:
@araujogui please run `npx turbo lint` your branch has linting issues 🤔

.storybook/preview.tsx
```diff
@@ -23,6 +23,7 @@ const preview: Preview = {
         basePath: '',
       },
     },
+    backgrounds: { disable: true },
```
ovflowd, line 26: Just a note here: This change was done because we have our own Theming System. We might want to integrate Storybook to support our `next/themes` stuff

.browserslistrc
```diff
@@ -1,14 +1,10 @@
...
 [production]
 
 cover 95%
 not dead
 
 [development]
 
-defaults and supports es6-module
-maintained node versions
+defaults
```
ovflowd: This fixes `storybook` local development mode. @araujogui could you make another PR just with the changes to this file? That'd be great!

components/Article/Codebox/InlineCode.tsx
```diff
@@ -0,0 +1,8 @@
+import styles from './index.module.scss';
+import type { PropsWithChildren } from 'react';
+
+const InlineCode = ({ children }: PropsWithChildren) => (
```
ovflowd: Could you change this to `const InlineCode: React.FC<PropsWithChildren> = ({ children }) =>`

components/Article/Codebox/__tests__/index.test.tsx
```diff
@@ -0,0 +1,121 @@
...
+    expect(replaceLanguages('language-cjs')).toBe('language-js');
+    expect(replaceLanguages('language-javascript')).toBe('language-js');
+    expect(replaceLanguages('language-console')).toBe('language-bash');
+    expect(replaceLanguages('language-shell')).toBe('language-bash');
+  });
+});
+
+describe('Codebox component (one lang)', (): void => {
+  const code = 'const a = 1;';
+  const textToCopy = sanitize(highlight(code, languages.js, 'js'));
```
ovflowd: I think this should be removed, we're double-encoding the content

components/Article/Codebox/__tests__/index.test.tsx
```diff
@@ -0,0 +1,121 @@
...
+    render(
+      <IntlProvider locale="en" onError={() => {}}>
+        <Codebox>
+          <code className="language-cjs|language-mjs">{code}</code>
+        </Codebox>
+      </IntlProvider>
+    );
+
+    const buttonElement = screen.getByText('cjs');
+    userEvent.click(buttonElement);
```
ovflowd: This should be async with `const user = userEvent.setup()`

components/Article/Codebox/__tests__/index.test.tsx
```diff
@@ -0,0 +1,121 @@
...
+    render(
+      <IntlProvider locale="en" onError={() => {}}>
+        <Codebox>
+          <code className="language-async|language-sync">{code}</code>
+        </Codebox>
+      </IntlProvider>
+    );
+
+    const buttonElement = screen.getByText('async');
+    userEvent.click(buttonElement);
```
ovflowd: Same here about userEvent

components/Article/Codebox/index.module.scss
```diff
@@ -0,0 +1,131 @@
...
+  tab-size: 4;
+  text-align: left;
+  white-space: pre;
+  word-break: normal;
+  word-spacing: normal;
+  word-wrap: normal;
+}
+
+.code {
+  @extend %baseStyles;
```
ovflowd:
```suggestion
  @extend %baseStyles;

```

components/Article/Codebox/index.stories.tsx
```diff
@@ -0,0 +1,11 @@
+import Codebox from './index';
+
+export default { component: Codebox };
+
+const code = 'const a = 1;';
+
+export const Default = {
+  args: {
+    children: <code className="language-js">{code}</code>,
```
ovflowd:
```suggestion
    children: <pre className="language-js">{code}</pre>,
```

ovflowd: If it is `code` then it should be the InLineCode. Also could you add another story here for the InLine code?

components/Article/Codebox/index.tsx
```diff
@@ -0,0 +1,78 @@
...
+  language
+    .replace(/mjs|cjs|javascript/i, 'js')
+    .replace(/console|shell/i, 'bash');
+
+const Codebox: FC<CodeBoxProps> = ({ children: { props } }) => {
+  const [parsedCode, setParsedCode] = useState('');
+  const [copied, copyText] = useCopyToClipboard();
+  const [langIndex, setLangIndex] = useState(0);
+
+  const className = props.className || 'text';
```
ovflowd, line 27:
```suggestion
  const className = props.className || 'text';

```

components/Article/Codebox/index.tsx
```diff
@@ -0,0 +1,78 @@
...
+  // Language Matches in class
+  const matches = className.match(/language-(?<lang>.*)/);
+  const languageOptions = (matches?.groups?.lang || 'text').split('|');
+  const language = languageOptions[langIndex];
```
ovflowd, line 33:
```suggestion
  const language = languageOptions[langIndex];

```

hooks/useCopyToClipboard.ts
```diff
@@ -0,0 +1,31 @@
+import { useEffect, useState } from 'react';
+
+const copyToClipboard = (value: string) => {
```
ovflowd: Oh, I guess Augustin was also adding this same hook to their PR.

hooks/useCopyToClipboard.ts
```diff
@@ -0,0 +1,31 @@
...
+  if (typeof window === 'undefined') {
```
ovflowd:
```suggestion
  if (typeof navigator === 'undefined') {
```

Since we use `navigator` call down there

hooks/useCopyToClipboard.ts
```diff
@@ -0,0 +1,31 @@
...
+  }
+
+  return navigator.clipboard
+    .writeText(value)
+    .then(() => true)
+    .catch(() => false);
+};
+
+// eslint-disable-next-line no-unused-vars
+export const useCopyToClipboard = (): [boolean, (text: string) => void] => {
```
ovflowd: Could you remove this explicit return type and then remote the eslint-disable statement

```suggestion
export const useCopyToClipboard = () => {
```

hooks/useCopyToClipboard.ts
```diff
@@ -0,0 +1,31 @@
...
+    if (!copied) {
+      return undefined;
+    }
+
+    const timerId = setTimeout(() => setCopied(false), 3000);
+
+    return () => clearTimeout(timerId);
+  }, [copied]);
+
+  return [copied, copyText];
```
ovflowd:
```suggestion
  return [copied, copyText] as const;
```

hooks/useCopyToClipboard.ts
```diff
@@ -0,0 +1,31 @@
...
+};
+
+// eslint-disable-next-line no-unused-vars
+export const useCopyToClipboard = (): [boolean, (text: string) => void] => {
+  const [copied, setCopied] = useState(false);
+
+  const copyText = (text: string) => copyToClipboard(text).then(setCopied);
+
+  useEffect(() => {
+    if (!copied) {
```
ovflowd: Change the logic to:

```jsx
if (copied) {
  const timerId = setTimeout(() => setCopied(false), 3000);

  return () => clearTimeout(timerId);
}

return undefined;
```

ovflowd (commented):
Really good stuff! Thank you for the contribution, @araujogui!!

Left a bunch of comments that I'd like to have them addressed before we proceed :)

hooks/useCopyToClipboard.ts
araujogui: An early return sounds cleaner to me

components/Article/Codebox/__tests__/index.test.tsx
araujogui: `userEvent.setup` is conflicting our `navigator` mocks
```
  ● Codebox component (multiple langs) › switch between languages

    TypeError: Cannot redefine property: clipboard
        at Function.defineProperty (<anonymous>)
```

ovflowd: Then we don't need a mock if we do userEvent.setup(). And probably there's an API from testing library to test this navigator API.

components/Article/Codebox/__tests__/index.test.tsx
```diff
@@ -0,0 +1,116 @@
...
+    fireEvent.click(buttonElement);
+
+    expect(navigatorClipboardWriteTextSpy).toHaveBeenCalledTimes(1);
+    expect(navigatorClipboardWriteTextSpy).toHaveBeenCalledWith(code);
+  });
+});
+
+describe('Codebox component (multiple langs)', (): void => {
+  const code = `const http = require('http');
+  -------
```
ovflowd: This sounds wrong. Even the snapshot is wrong

```tsx
const code = `const http = require('http');
-------
import http from 'http';`;
```

That's how this should look like.

components/Article/Codebox/__tests__/index.test.tsx
```diff
@@ -0,0 +1,116 @@
...
+
+    const buttonElement = screen.getByText('cjs');
+    userEvent.click(buttonElement);
+
+    await screen.findByText('cjs');
+
+    expect(screen.getByText('cjs')).toBeInTheDocument();
+  });
+
+  it('switch between sync and async', async () => {
```
ovflowd: Also, remove this test. This is kind of garbage. Already covered by previous test.

components/Article/Codebox/index.tsx
```diff
@@ -0,0 +1,80 @@
...
+
+  const codeArray = props.children
+    ? props.children.toString().split('--------------\n')
+    : [''];
+
+  const handleCopyCode = async (event: MouseEvent<HTMLButtonElement>) => {
```
ovflowd: This doesn't need to be `async`. There's no async operation happening here.

hooks/useCopyToClipboard.ts
ovflowd: Negated conditions are usually a bad practice if you can do the same condition without a negated operator. It's usually easier to read.

We often do negated conditions in JSX, because well, adding the `return null` after a gigantic block of JSX hinders the visibility of that statement.

jest.setup.mjs
```diff
@@ -1 +1,5 @@
+import { TextEncoder, TextDecoder } from 'util';
 import '@testing-library/jest-dom/extend-expect';
+
+global.TextEncoder = TextEncoder;
```
ovflowd: I honestly don't like to simply inject Node's TextEncoder API to the `globals`. That's a hack and shouldn't be necessary.

You can simply mock `isomorphic-dompurify` and that will have the same effect.

Since there's nothing to be "purified" on our unit-test. (We don't have malicious code)

components/Article/Codebox/index.tsx
araujogui: Actually, `copyText` is a async operation. Maybe an `await` here would be a better option?

ovflowd: Well that makes sense. But what will the onClick operator await for? Isn't onClick itself synchronous? (If I recall it does indeed in some certain blocking operations wait for something actually)

components/Article/Codebox/__tests__/index.test.tsx
```diff
@@ -0,0 +1,93 @@
...
+    const buttonElement = screen.getByText('components.codeBox.copy');
+    await user.click(buttonElement);
+
+    expect(navigatorClipboardWriteTextSpy).toHaveBeenCalledTimes(1);
+    expect(navigatorClipboardWriteTextSpy).toHaveBeenCalledWith(code);
+  });
+});
+
+describe('Codebox component (multiple langs)', (): void => {
+  const code = `const http = require('http');
```
ovflowd, line 55: Snapshot still seems wrong. Can you try `const http = "require('http');\n--------------\nimport http from 'http';`

ovflowd: Ah I know why. It needs to be `--------------`

ovflowd: I wonder how these tests are passing if they never then actually tested what this was supposed to test -.-

(Well they are, just not testing whatever they were supposed to test)

components/Article/Codebox/__tests__/index.test.tsx
```diff
@@ -0,0 +1,93 @@
...
+        <Codebox>
+          <pre className="language-cjs|language-mjs">{code}</pre>
+        </Codebox>
+      </IntlProvider>
+    );
+
+    const buttonElement = screen.getByText('cjs');
+    await user.click(buttonElement);
+
+    await screen.findByText('cjs');
```
ovflowd: Once you update that snippet. This should probably start to fail. As it should switch to `await screen.findByText('mjs');` and `expect(screen.getByText('mjs')).toBeInTheDocument();`

components/Article/Codebox/index.tsx
```diff
@@ -0,0 +1,80 @@
...
+
+const Codebox: FC<CodeBoxProps> = ({ children: { props } }) => {
+  const [parsedCode, setParsedCode] = useState('');
+  const [copied, copyText] = useCopyToClipboard();
+  const [langIndex, setLangIndex] = useState(0);
+
+  const className = props.className || 'text';
+
+  // Language Matches in class
+  const matches = className.match(/language-(?<lang>.*)/);
```
ovflowd: This is ultimately broken. This should honestly be:

```tsx
const language = useMemo(() => {
  let language = 'text';

  if (props.className && props.className.length) {
    language = props.className.split('|')[langIndex];
  }

  if (language.startsWith('language-')) {
    language = language.replace('language-', '');
  }

  return language;
}, [props.className, langIndex]);
```

components/Article/Codebox/__tests__/index.test.tsx
```diff
@@ -0,0 +1,93 @@
...
+      <IntlProvider locale="en" onError={() => {}}>
+        <Codebox>
+          <pre className="language-js|language-js">{code}</pre>
+        </Codebox>
+      </IntlProvider>
+    );
+    expect(container).toMatchSnapshot();
+  });
+
+  it('switch between languages', async () => {
```
ovflowd, line 59: Can we also add a snapshot for before and after click?

ovflowd (approved):
Now we're talking! This PR not only migrates the component but also fixed many invisible bugs we didn't even notice.

Amazing job, @araujogui! Thank you for being attentive :)

ovflowd:
Note: I feel that the current Storybook args is kinda not optimal.. If we could somehow allow people to actually write the content and map that to the story component?

E.g.:

- Both stories have a Decorator https://storybook.js.org/docs/react/writing-stories/decorators/
  - This decorator actually adds the `<pre>` tag and className
- Both stories then have args for:
  - Name of programming language as array
  - Content as array

Then for each programming language array item we join them with `|` and for each content we join them with that big `-------------`.

Could you do that? If it's not clear let me know!

__mocks__/isomorphic-dompurify.js
```diff
@@ -0,0 +1,5 @@
+const mock = {
```
ovflowd: The mock doesn't need to be done here. You can do it directly on the test. Since it's the only place we need to mock it for now :)

(I actually have the mock actually be done there, as it allows me to remember why this mock exists)

components/Article/Codebox/InlineCode.tsx
ovflowd, line 1: One last thing. Can you:

- Move this component to `components/Article/InlineCode`
- Move also the stories
- Move all the styles that are only for the `.code` class (including dark styles, light ones and etc cetera) also to the Inline code
- Add unit test just that it is rendering fine

That's pretty much my final ask :)

ovflowd:
Also @araujogui I know it's not of the scope of this PR but it would be great if the currently selected programming language is highlighted either by a different colour or something...

image

Right now I can't really see what one is selected. You could add a new class to the button called selected and add this className if the current button index === selectedIndex

components/Article/Codebox/inline.stories.tsx
```diff
@@ -0,0 +1,19 @@
...
+
+export default meta;
+
+type Story = StoryObj<typeof InlineCode>;
+
+const code = 'const a = 1;';
+
+export const Default: Story = {
+  args: {
+    children: <code>{code}</code>,
```
ovflowd: Also do the same with here (with the decorator) so we can actually customise the code xD

components/Article/Codebox/InlineCode.tsx
araujogui: Can I remove `.code` classes from Codebox, or should I leave it duplicated?

ovflowd: Remove. Not used there :)

components/Article/Codebox/index.module.scss
```diff
@@ -0,0 +1,105 @@
...
+    .copy {
+      align-items: center;
+      cursor: pointer;
+      display: inherit;
+      font-size: var(--font-size-code);
+      height: 23px;
+      justify-content: center;
+      width: 86px;
+    }
+    .langBox {
```
ovflowd, line 24:
```suggestion

    .langBox {
```

components/Article/InlineCode/index.module.scss
```diff
@@ -0,0 +1,58 @@
+%baseStyles {
```
ovflowd: I think these base styles instead of duplicated can be outsourced to somewhere else. Like maybe styles/code.scss. And maybe renamed to %codeBaseStyles I dunno

components/Article/InlineCode/index.stories.tsx
```diff
@@ -0,0 +1,30 @@
+import InlineCode from './index';
+import type { FC } from 'react';
+import type { Meta, StoryObj } from '@storybook/react';
+
+type InlineCodeArgs = {
+  code: string;
+};
+
+const meta: Meta<FC<InlineCodeArgs>> = {
+  title: 'InlineCode',
```
ovflowd: No need for title

components/Article/Codebox/index.stories.tsx
```diff
@@ -0,0 +1,48 @@
...
+
+type CodeBoxProps = {
+  language: string[];
+  code: string[];
+};
+
+type ICodebox = FC<CodeBoxProps>;
+
+const meta: Meta<ICodebox> = {
+  title: 'Codebox',
```
ovflowd: Same here, no title please. Otherwise we cannot group these components per folder

components/Article/InlineCode/index.stories.tsx
```diff
@@ -0,0 +1,30 @@
...
+};
+
+export default meta;
+
+type Story = StoryObj<FC<InlineCodeArgs>>;
+
+const code = 'const a = 1;';
+
+export const Default: Story = {
+  args: {
```
ovflowd: args: { code } 

(Just in one line)

ovflowd (commented):
Forgot to post my comments. Last review ;D

components/Article/InlineCode/index.stories.tsx
araujogui: A story should contain at least a title or a component
```
Error: CSF: missing title/component /Users/araujo/Documents/Projects/nodejs.org/components/Article/InlineCode/index.stories.tsx (line 1, col 0)
```

ovflowd: You should pass the component then, but not title. We're not using title anywhere :)

hooks/__tests__/useCopyToClipboard.test.tsx
```diff
@@ -0,0 +1,83 @@
...
+import type { AppProps } from '../../types';
+
+const i18nData = { currentLocale: { code: 'en' } } as AppProps['i18nData'];
+
+describe('useCopyToClipboard', () => {
+  const HookRenderer = ({ text }: { text: string }) => {
+    const [copied, copyText] = useCopyToClipboard();
+
+    return (
+      <LocaleProvider i18nData={i18nData}>
```
ovflowd: Can you remove the locale provider from here? (No need to use any translation key. Inside the unit test component you can even write "copy text test"

components/Article/InlineCode/index.stories.tsx
```diff
@@ -0,0 +1,28 @@
+import InlineCode from './index';
+import type { FC } from 'react';
+import type { Meta, StoryObj } from '@storybook/react';
+
+type InlineCodeArgs = {
```
ovflowd: This typing feels weird. Please check the CONTRIBUTING guidelines (there should be a Meta type)

components/Article/InlineCode/index.stories.tsx
```diff
@@ -0,0 +1,28 @@
...
+  decorators: [
+    (Story, context) => (
+      <InlineCode>
+        <code>{context.args.code}</code>
+      </InlineCode>
+    ),
+  ],
+};
+
+export default meta;
```
ovflowd: This export should be below everything and the naming of the export should be consistent as from the template on the CONTRIBUTING guidelines (same for the other story)

hooks/__tests__/useCopyToClipboard.test.tsx
araujogui: It's a must
```
[React Intl] Could not find required `intl` object. <IntlProvider> needs to exist in the component ancestry.
```

ovflowd: Yes because you're passing FormattedMessage there. You can simply remove that altogether lol

ovflowd (approved):
@nodejs/website can y'all also review/approve this?

mikeesto:
Has something happened with our Vercel storybook previews here? All the links are 404s for me

ovflowd:
> Has something happened with our Vercel storybook previews here? All the links are 404s for me

Misconfiguration on Vercel-side, but it was fixed already. I can redeploy this build.

ovflowd:
@araujogui there are some conflicts here

ovflowd:
@araujogui can you add a litle hover effect (darker colour) when you hover on the copy and language selection buttons?

ovflowd:
cc @nodejs/website also for review

mikeesto (approved):
The hover effect looks good. This LGTM, just a couple of conflicts to walk through

shanpriyan (approved):
LGTM! 🚀 


Suggestion: 
Can we add a **green** `border-bottom` to the selected language
image

aymen94 (approved):

ovflowd:
@araujogui updates?

ovflowd:
@araujogui can you run the linting commands and update the test snapshots here? @Harkunwar is anything else needed here? (Like Unit Tests or etc)

ovflowd:
@araujogui can you please run build/lint/test before committing? PR is still failing (build failing) with missing types?

components/Article/Codebox/__tests__/index.test.tsx
```diff
@@ -0,0 +1,99 @@
...
+    expect(navigatorClipboardWriteTextSpy).toHaveBeenCalledWith(code);
+  });
+});
+
+describe('Codebox component (multiple langs)', (): void => {
+  const code = `const http = require('http');
+--------------
+import http from 'http';`;
+
+  it('renders correctly', (): void => {
```
ovflowd: @araujogui this test can be removed, as Storybooks now have their own snapshot testing. You can also remove this snapshot,

Don't forget to run `npm run test:storybook:snapshot` just to double check all Storybooks have Snapshots

PR #5340, opened 2023-04-29
test(snapshot): Migrate snapshot tests to storybook

Harkunwar:
This PR aims to migrate existing snapshot tests written in jest, and instead automates it using the existing storybook stories. The test are also run in Playwright now, instead of JSDOM, so it's more accurate. 
#5329


.github/workflows/pull-request.yml
```diff
@@ -83,7 +83,13 @@ jobs:
...
 
       - name: Run Unit Tests
-        run: npx turbo test:ci
+        run: npx turbo test:unit:ci
+        env:
+          TURBO_TOKEN: ${{ secrets.TURBO_TOKEN }}
+          TURBO_TEAM: ${{ secrets.TURBO_TEAM }}
+
+      - name: Run Storybook Tests
+        run: npx turbo test:storybook:ci
```
Harkunwar: New step for running storybook tests. Currently only does snapshot but will use the same command to do interaction tests in the future.

.storybook/preview.tsx
```diff
@@ -39,7 +39,9 @@ export const decorators = [
                 }
               `}
             </style>
-            <Story />
+            <div id="story-root">
```
Harkunwar: Added this to easily identify the div to take the snapshot of, and not the whole tree

components/Article/BlockQuote/index.stories.tsx
```diff
@@ -10,6 +10,8 @@ export const Default: Story = {
   },
 };
 
+export const Empty: Story = {};
```
Harkunwar, line 13: Migrated all tests to stories if they didn't already exist in the stories file. The above is just one example. There should be no loss of snapshots

package.json
```diff
@@ -26,10 +26,14 @@
...
-    "test:watch": "npm run test -- --watch",
-    "test:ci": "npm run test -- --coverage",
-    "test:snapshot": "npm run test -- --updateSnapshot",
+    "test": "npm run test:unit && npm run test:storybook",
+    "test:unit": "cross-env NODE_NO_WARNINGS=1 jest --passWithNoTests",
+    "test:unit:watch": "npm run test:unit -- --watch",
+    "test:unit:ci": "npm run test:unit -- --coverage",
+    "test:storybook": "npx concurrently -P -k -s first -n \"SB,TEST\" -c \"magenta,blue\" \"npm run storybook -- --ci\" \"npx wait-on http://localhost:6006 && npx test-storybook {@}\"",
+    "test:storybook:snapshot": "npm run test:storybook -- -- --updateSnapshot",
+    "test:storybook:watch": "npm run test:storybook -- -- --watch",
```
Harkunwar, line 34: There are double `--` here because I need to pass the command two levels deeper, once to the `test:storybook` npm command, and then inside `concurrently`.

AugustinMauroy:
I am not a fan of this method. I prefer jest. (IMHO)

Harkunwar:
@AugustinMauroy Any reasons for preferring Jest over Storybook? I'm not sure if that's the right question.

- Storybook Test Runner is built over Jest, just uses playwright instead of JSDOM, which helps us use a real browser in our tests instead of using JSDOM in which tons of things don't work and you need to mock them.
- Since Storybook Test Runner uses Jest, you can use all the jest functionality inside the test-runner.
- Since we already use storybook, we're basically rewriting the same component twice, once in stories, another time in Jest to capture snapshot. This PR aims to eliminate that code duplication.
- Writing tests in JSDOM is tedious because you can never see what you built. What you want could be working in a real browser, but not in JSDOM.
- If we move over to storybook interaction testing, which also uses jest, you'll also get the power to run those tests in the browser instead of JSDOM, making debugging and writing tests way easier.

ovflowd:
@Harkunwar I don't think Augustin really knows what he's talking about. I'd recommend Augustin to read your original issue.

For what we we're doing here Storybook is definitely the choice to be done for the kind of testing we're doing.

ovflowd:
Also I haven't checked the PR yet, but don't forget to update the CONTRIBUTION guidelines of how to do React Unit Testing and templates for these tests and etc

Harkunwar:
I'll update the CONTRIBUTION guidelines. Definitely missed those.

ovflowd:
Also, @AugustinMauroy, I don't want to be dismissive of your opinions, but I'd be more productive if you explained what you don't like about this method and why Jest would be better.

There was a whole issue regarding this change; you could have voiced your opinions there.

.github/workflows/pull-request.yml
ovflowd: Coolio!

.storybook/preview.tsx
ovflowd: Could we use maybe `data-test-id`? If this is only used for testing

.storybook/test-runner.ts
```diff
@@ -0,0 +1,12 @@
+// TODO: Convert this to ESM once the bug is resolved
+// https://github.com/storybookjs/test-runner/issues/293
+
+const config = {
+  async postRender(page, _context) {
+    const elem = await page.$('#story-root');
+    const html = await elem?.innerHTML();
```
ovflowd: If elem is null we would be awaiting on null, which probably would cause an error 🤔

.storybook/test-runner.ts
```diff
@@ -0,0 +1,12 @@
+// TODO: Convert this to ESM once the bug is resolved
+// https://github.com/storybookjs/test-runner/issues/293
+
+const config = {
+  async postRender(page, _context) {
+    const elem = await page.$('#story-root');
```
ovflowd: The `$` syntax, is this by any chance jQuery? :O

components/Article/EditLink/index.stories.tsx
```diff
@@ -12,4 +14,28 @@ export const Edit: Story = {
...
+        currentLocale: {
+          code: 'xx',
+        },
+        localeMessages: {
+          'components.article.editLink.title.translate':
+            'Interested to help with translations?',
+        },
+      } as unknown as AppProps['i18nData'];
+      return (
+        <LocaleProvider i18nData={i18nDataTranslateMode}>
```
ovflowd: As I mentioned before we shouldn't need to pass localeProvider here. Feel free to delete this.

Storybook is already wrapped on LocaleProvider that contains these exact translation keys.

package.json
ovflowd: Interesting. I doubt we can use turbo here for the `test:storybook:**` commands, but could you add a new `turbo` entry (turbo.json) for the `test:unit` and `test:unit:ci` commands?

ovflowd (approved):
Overall, LGTM. Impressive work :)

components/Article/EditLink/index.stories.tsx
Harkunwar: I was migrating the second test from here:
https://github.com/nodejs/nodejs.org/pull/5340/files#diff-d769fad9f33af5e222b1b2b7c5f92f5351ef4bba7036bef93607c593ea5ce1f9

Harkunwar: The original test was this:
```typescript
const i18nDataTranslateMode = {
  currentLocale: {
    code: 'xx',
  },
  localeMessages: {
    'components.article.editLink.title.translate':
      'Interested to help with translations?',
  },
} as unknown as AppProps['i18nData'];

describe('EditLink component', () => {
  it('translate mode renders correctly', () => {
    const { container } = render(
      <LocaleProvider i18nData={i18nDataTranslateMode}>
        <EditLink relativePath={relativePath} />
      </LocaleProvider>
    );
    expect(container).toMatchSnapshot();
  });
```

.storybook/test-runner.ts
Harkunwar: Ideally it should never be null since we're declaring it as a decorator and it should always exist.

Although awaiting something that's not a Promise, awaiting should return itslef.

```typescript
const x = await null; // x is null
const y = await undefined; // y is undefined
const z = await 23; // z is 23
```

The actual problem here would be that if `elem` is undefined, it would resolve `elem.innerHTML` to be undefined and `elem.innerHTML()` to be an error since you can't call a function on an undefined. That can however be fixed using `elem.innerHTML?.()`

components/Article/EditLink/index.stories.tsx
ovflowd: Yeah, but we don't need this, to be honest.

.storybook/test-runner.ts
Harkunwar: It's not JQuery, It's the page object from Playwright.
https://playwright.dev/docs/api/class-page

I just read while checking that the `$` syntax is deprecated so I'll switch to their newer function called `locator`

.storybook/test-runner.ts
ovflowd: This probably happens due to how the `await` is internally converted to a Promise, as pretty much it becomes either Promise.resolve(null) or new Promise(resolve => resolve(null);

Yet, I'd rather have we throwing an error if element is null as it pretty much is a precondition for our test-runner.

.storybook/test-runner.ts
ovflowd: Thanks! Appreciated :) (Nothing against jQuery just surprised lol).

components/Article/EditLink/index.stories.tsx
Harkunwar: Removed.

.storybook/test-runner.ts
Harkunwar: Updated to `locator`

.storybook/test-runner.ts
Harkunwar: Added an error check

.storybook/test-runner.ts
```diff
@@ -0,0 +1,15 @@
+// TODO: Convert this to ESM once the bug is resolved
+// https://github.com/storybookjs/test-runner/issues/293
+
+const config = {
+  async postRender(page, _context) {
+    const elem = await page.locator('[data-test-id="story-root"]');
+    if (!elem) {
```
ovflowd: Hm I was actually talking about making the test fail instead of throwing an error 🙈  Like first assert that the element exists :)

ovflowd:
```tsx
const rootElementId = '[data-test-id="story-root"]';

const rootElement = await page.locator(rootElementId);

expect(rootElement).toBeDefined();

if (rootElement) {
  const content = await rootElement.innerHTML();

  expect(html).toMatchSnapshot();
}
```

ovflowd: Also not sure if this `await rootElement.innerHTML()` is part of this API you mentioned, but usually on the DOM API, `rootElement.innerHTML` should be sync and a property 🤔

Harkunwar: Sounds good, my bad!

Harkunwar: Although the `if` shouldn't be needed since we're already checking if `rootElement` is defined.

Harkunwar: @ovflowd It's part of the api of `page` that playwright provides. I used innerHTML() as well and was shocked it didn't work and had to add the `await` later.

ovflowd: The `if` is needed as the assertion failing doesn't stop the test from running :)

Harkunwar: I believe `jest` stops execution when the first assertion fails.
Here's a simple example
https://stackblitz.com/edit/jest-example-josuki?file=sum.test.js

ovflowd: I'm not sure that's how it works (by personal experience). If you could actually test it is doing what you expect by manually making it fail, let me know.

Harkunwar: In the above case, seems like `toBeDefined` is doing nothing because the `elem` will always be defined even if it doesn't exist
image
which is why the test continues.

Harkunwar: And `content` is returned as `undefined` when the `elem` doesn't exist without causing any errors
image

ovflowd: Then we don't need the if checks

But let's add a toBeDefined after the await of the innerHTML

Harkunwar: Updated!

CONTRIBUTING.md
```diff
@@ -102,7 +102,10 @@ We also offer other commands that offer you assistance during your local develop
   - **Usage:** `npx turbo scripts:release-post -- --version=vXX.X.X --force`
 - `npx turbo storybook` starts Storybook's local server
 - `npx turbo storybook:build` builds Storybook as a static web application for publishing
-- `npx turbo test` runs jest (unit-tests) locally
+- `npx turbo test` runs all tests locally
```
ovflowd, line 105: I think you should also on the unit testing section of the guidelines mention that unit testing of components rendering as expected is done via Storybook. And add maybe a link to some reference/docs

ovflowd (approved):
Honestly, really good stuff! Thank you for the contribution!

CONTRIBUTING.md
Harkunwar: Sounds good! I'll update those tomorrow. Almost 5am here.

AugustinMauroy:
To answer you.  I don't really like storybook. And I love Jest. So yes there are advantages. But I find that they will add "difficulty" for newcomers.

ovflowd:
Storybook is essential for the migration and visual testing of components. Jest is an automated testing framework that allows us to pretty much test and cover the functionalities of our codebase.

With Storybook we can ensure that the design and interactions of a Component are working as expected. Storybooks are mostly used while we are creating Components.

And with Jest and testing-library we can ensure that certain behaviours and functionality work as expected and that we can easily identify it something breaks.

But for just testing that a Component rendered correctly, we don't need to create a skeleton test on Jest because that's pretty much code duplication.

And here's where Storybook enters the play, by allowing us to reuse our Stories for these exact tests that we were one by one manually writing.

Finally, this doesn't create any barrier to Newcomers as they will not need to create new tests for hat are already covered by Storybook.

And Storybook is needed for the development/migration of Components, so I don't see the friction here. Not to mention that making a Storybooks is relatively easy and can be quickly accomplished by following the CONTRIBUTING guidelines.

It is newcomer friendly. I don't know why you think this PR would make the repository any less newcomer friendly, so I'd appreciate if you could elaborate on that.

CONTRIBUTING.md
ovflowd: Okay I'll be waiting for that :)

ovflowd:
cc @nodejs/website for more reviews :)

ovflowd:
FYI @Harkunwar you might need to update the PR after more components get merged

CONTRIBUTING.md
Harkunwar: Working on it! Got sick over the weekend.

Harkunwar: Updated!

Harkunwar:
I'm really confused why the unit tests under windows are failing. I've tried to following approach of changing the name of the script

`test:unit` - works
`test:unit -- --coverage` - works
`test:unit:ci` - breaks
`test:unit:cid` - works
`test:unit:c` - works
`test:storybook:ci` - works

I don't know what's special with the name `test:unit:ci`.

ovflowd:
@Harkunwar the windows error could be related to some funky GitHub CI cache... Let's see what happens without `:ci` but usually we do want to separate ci from non-ci commands.

package.json
```diff
@@ -26,10 +26,12 @@
     "prettier:fix": "npm run prettier -- --write",
     "format": "npm run prettier:fix && npm run lint:fix",
     "scripts:release-post": "NODE_NO_WARNINGS=1 node scripts/release-post/index.mjs",
-    "test": "cross-env NODE_NO_WARNINGS=1 jest --passWithNoTests",
-    "test:watch": "npm run test -- --watch",
-    "test:ci": "npm run test -- --coverage",
-    "test:snapshot": "npm run test -- --updateSnapshot",
+    "test": "npm run test:unit && npm run test:storybook",
+    "test:unit": "cross-env NODE_NO_WARNINGS=1 jest --passWithNoTests",
```
ovflowd, line 30: If we don't make a specific `:ci` command we need to add `--coverage` generation on the regular `test:unit`

package.json
```diff
@@ -26,10 +26,12 @@
...
+    "test:unit:watch": "npm run test:unit -- --watch",
+    "test:storybook": "npx concurrently -P -k -s first -n \"SB,TEST\" -c \"magenta,blue\" \"npm run storybook -- --ci\" \"npx wait-on http://localhost:6006 && npx test-storybook {@}\"",
+    "test:storybook:snapshot": "npm run test:storybook -- -- --updateSnapshot",
```
ovflowd, line 33: Maybe also add a `test:unit:snapshot` for all individual unit tests with snapshots? (might exist)

turbo.json
```diff
@@ -1,19 +1,17 @@
...
       ],
       "outputMode": "full"
     },
-    "test:ci": {
+    "test:storybook": {
       "inputs": [
-        "{components,hooks,layouts,providers,types,util}/**/*.{ts,tsx}",
-        "jest.config.mjs",
-        "jest.setup.mjs"
+        "{components,hooks,layouts,providers,types,util}/**/*.{ts,tsx}"
```
ovflowd, line 14: Storybook doesn't use the regular jest config file I suppose?

package.json
Harkunwar: The error is happening with windows when there's two `:` present in the name. Maybe an issue with turborepo. I tried several cases in my private repo. .

I updated the `pull_request.yml` to include `-- --coverage`

turbo.json
Harkunwar: It doesn't

package.json
Harkunwar: I removed all existing snapshot tests.

You can search `toMatchSnapshot` and there's only one instance in `test-runner`

package.json
Harkunwar: If we use a different naming scheme, we can add the `ci` commands. I tried `test:unit-ci` and it was working.

ovflowd:
**This PR got accidentally closed because the branch got deleted. Please feel free to reopen the PR.**

ovflowd:
@Harkunwar fyi that your branch is also desynced, and has 69 commits showing here. You probably need to recreate your branch and cherry pick your commits.

Harkunwar:
@ovflowd That should be fine since if you look at the files changed, it's correct. I did merge instead of rebase, which is why those commits show up. Once we squash merge this branch, there will be only a single commit on the `major/website-redesign` branch.

ovflowd:
I mean sure, since we're squashing the commits the outcome will be the same, but just wanted to let you know that your PR history is polluted.

PR #5370, opened 2023-05-11
Migrate JsonLink component

Ve33y:
## Description
Bringing the JsonLink component over. There was a faRobot icon I left out since I was not sure if it's necessary and it would presumably need @fortawesome/react-fontawesome but let me know if we'd like to keep it and that can be brought in as well.

## Related Issues
https://github.com/nodejs/nodejs.org/issues/5309

### Check List

- [x] I've covered new added functionality with unit tests if necessary. (storybook story)

### DIFF
--- a/components/Api/JsonLink/index.stories.ts
+++ b/components/Api/JsonLink/index.stories.ts
@@ -1,0 +1,14 @@
+import JsonLink from './index';
+import type { Meta as MetaObj, StoryObj } from '@storybook/react';
+
+type Story = StoryObj<typeof JsonLink>;
+type Meta = MetaObj<typeof JsonLink>;
+
+export const Default: Story = {
+  args: {
+    version: 'v18',
+    fileName: 'documentation',
+  },
+};
+
+export default { component: JsonLink } as Meta;
--- a/components/Api/JsonLink/index.tsx
+++ b/components/Api/JsonLink/index.tsx
@@ -1,4 +1,6 @@
 import { FormattedMessage } from 'react-intl';
+import { FaRobot } from 'react-icons/fa';
+import { IconContext } from 'react-icons';
 import styles from './index.module.scss';
 import type { FC } from 'react';
 
@@ -13,6 +15,9 @@ const JsonLink: FC<JsonLinkProps> = ({ fileName, version }) => (
       href={`https://nodejs.org/docs/latest-${version}.x/api/${fileName}.json`}
     >
       <FormattedMessage id="components.api.jsonLink.title" tagName="span" />
+      <IconContext.Provider value={{ className: styles.FaRobotIcon }}>
+        <FaRobot />
+      </IconContext.Provider>
     </a>
   </div>
 );
--- a/components/Api/JsonLink/index.tsx
+++ b/components/Api/JsonLink/index.tsx
@@ -1,6 +1,5 @@
 import { FormattedMessage } from 'react-intl';
 import { FaRobot } from 'react-icons/fa';
-import { IconContext } from 'react-icons';
 import styles from './index.module.scss';
 import type { FC } from 'react';
 
@@ -15,9 +14,7 @@ const JsonLink: FC<JsonLinkProps> = ({ fileName, version }) => (
       href={`https://nodejs.org/docs/latest-${version}.x/api/${fileName}.json`}
     >
       <FormattedMessage id="components.api.jsonLink.title" tagName="span" />
-      <IconContext.Provider value={{ className: styles.FaRobotIcon }}>
-        <FaRobot />
-      </IconContext.Provider>
+      <FaRobot className={styles.FaRobotIcon} />
     </a>
   </div>
 );
--- a/components/Api/JsonLink/index.tsx
+++ b/components/Api/JsonLink/index.tsx
@@ -1,0 +1,20 @@
+import { FormattedMessage } from 'react-intl';
+import styles from './index.module.scss';
+import type { FC } from 'react';
+
+type JsonLinkProps = {
+  fileName: string;
+  version: string;
+};
+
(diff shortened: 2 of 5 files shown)


Harkunwar:
Hello @Ve33y,
Thank you for contributing to nodejs.org repo towards our website redesign initiative.

components/Api/JsonLink/index.stories.ts
```diff
@@ -0,0 +1,14 @@
...
+type Meta = MetaObj<typeof JsonLink>;
+
+export const Default: Story = {
+  args: {
+    version: 'v18',
+    fileName: 'documentation',
+  },
+};
+
+export default { component: JsonLink } as Meta;
```
Harkunwar, line 14: I see that you've added a story for the component. Thanks for doing that. We're also saving snapshots of our stories so you can run `npm run test:storybook:snapshot` to generate that, and you can commit that too. That's the reason the tests are failing for your commit.

components/Api/JsonLink/index.tsx
```diff
@@ -0,0 +1,20 @@
...
+  fileName: string;
+  version: string;
+};
+
+const JsonLink: FC<JsonLinkProps> = ({ fileName, version }) => (
+  <div className={styles.json}>
+    <a
+      href={`https://nodejs.org/docs/latest-${version}.x/api/${fileName}.json`}
+    >
+      <FormattedMessage id="components.api.jsonLink.title" tagName="span" />
```
AugustinMauroy, line 16: The icon is missing

AugustinMauroy: try use that
```suggestion
      <FormattedMessage id="components.api.jsonLink.title" tagName="span" /><FaRobot />
```
and import it like that:
```js
import { FaRobot } from 'react-icons/fa';
```

components/Api/JsonLink/index.stories.ts
Ve33y: Thank you - totally missed that, added now.

components/Api/JsonLink/index.tsx
Ve33y: Wasn't sure if it should be moved but this was helpful and have now added the robot icon - thanks!

ovflowd: Pretty much what @AugustinMauroy mentioned, @Ve33y, you can use the package we already gave (`react-icons`) for the Rbot Icon :)

shanpriyan (approved):
LGTM! 🚀 

Thanks for your contribution

components/Api/JsonLink/index.tsx
```diff
@@ -0,0 +1,25 @@
...
+      <IconContext.Provider value={{ className: styles.FaRobotIcon }}>
+        <FaRobot />
```
AugustinMauroy: and that not working I will be little bit better tan an provider.
What do you think
```suggestion
        <FaRobot className={styles.FaRobotIcon} />
```

shanpriyan: > and that not working I will be little bit better tan an provider. What do you think


This should work I think 🤔 , +1 for this change

Ve33y: hi @AugustinMauroy - just curious what you mean by it not working since I see the storybook story rendering as expected? I used the suggested configuration here https://github.com/react-icons/react-icons#configuration but if we usually pass the styles directly to the icon, happy to switch to that. I don't believe there is a difference in performance between the two approaches, one benefit of the provider, in general, would be that styles specified within it would apply to all icon child components, although that might not be making a big difference in this case since there are probably no additional icons that will be needed. Thanks!

Screenshot 2023-05-12

AugustinMauroy: When I say "that not working".
I mean "hey my proposal should work"

ovflowd: He meant that the current approach is missing the grey-coloured Icon styles. I'm fine with either ways (keeping it green or making the icon grey)

components/Api/JsonLink/index.tsx
```diff
@@ -0,0 +1,25 @@
...
+  version: string;
+};
+
+const JsonLink: FC<JsonLinkProps> = ({ fileName, version }) => (
+  <div className={styles.json}>
+    <a
+      href={`https://nodejs.org/docs/latest-${version}.x/api/${fileName}.json`}
+    >
+      <FormattedMessage id="components.api.jsonLink.title" tagName="span" />
+      <IconContext.Provider value={{ className: styles.FaRobotIcon }}>
```
ovflowd: Question, why do we need this IconContext Provider here?

components/Api/JsonLink/index.tsx
Ve33y: It's actually grey by default and green on hover - sorry just took the snapshot on hover state.

components/Api/JsonLink/index.tsx
Ve33y: As I was saying to Augustin - I used the suggested configuration here https://github.com/react-icons/react-icons#configuration but if we usually pass the styles directly to the icon, happy to switch to that. In fact, will just push an update with passing the styles directly. Thanks!

ovflowd: I have the feeling this Context Provider should probably not be defined on each individual component. Ill give a read to these docs later just to double check

Ve33y: I removed it - just in case 🙂 (but could be brought back, although probably not likely to need to do that) Thanks once more!

ovflowd (approved):
LGTM! Thank you for your contribution, @Ve33y!!

Really appreciate it! 🎉

PR #5419, opened 2023-06-05
feat: Added generic and SearchBar and Orama powered SearchBar.

ShogunPanda:
## Description

This PR introduces a Search Bar component and it also introduces a [Orama](https://oramasearch.com) based implementation of the same component.

### Check List


tniessen:
Is this a client-side index that is generated at build time? I am not familiar with Orama or Next.js but I believe @ovflowd suggested multiple alternatives and wanted to avoid client-side processing of search indices.

Refs: https://github.com/nodejs/admin/issues/766#issuecomment-1431560083

ovflowd:
I also haven't checked this PR yet. But I asked the Orama team for something that allowed us leverage both world. A client-side only on first build but if Server-side is available to leverage that 🤔

Will check if this PR is implementing this 👀

ShogunPanda:
This PR is opened to both possibilities.
At the moment I'm not generating any index and I'm just providing the component for storybook.
In the future the index can be generated on the server and downloaded via `fetch` in the `setup` call. Or can be generated in the browser (always in the `setup` call).

components/Common/OramaSearchBar/index.module.scss
```diff
@@ -0,0 +1,24 @@
+.oramaSearchBarFooter {
```
ovflowd, line 1: The style might need some changes for dark mode

image

components/Common/OramaSearchBar/index.tsx
```diff
@@ -0,0 +1,91 @@
+import { create, load, search } from '@orama/orama';
+import { useTheme } from 'next-themes';
+import { useCallback, useMemo, useState } from 'react';
+import { FormattedMessage } from 'react-intl';
+import { OramaLogoDark, OramaLogoLight } from './OramaLogo';
+import styles from './index.module.scss';
```
ovflowd, line 6: styles usually should be the last import 👀 

After all other imports (so before the `DocumentMapper` type

components/Common/OramaSearchBar/index.tsx
```diff
@@ -0,0 +1,91 @@
...
+import SearchBar from '../SearchBar';
+
+import type { Document, Orama, RawData, Schema } from '@orama/orama';
+import type { FC } from 'react';
+import type { SearchFunction, SearchResult } from '../../../types';
+
+type DocumentMapper = (doc: Document) => SearchResult;
+
+type FooterProps = {
```
ovflowd: This can be a one-line type `type FooterProps = { ... }`

components/Common/OramaSearchBar/index.tsx
```diff
@@ -0,0 +1,91 @@
...
+  theme?: string;
+};
+
+type OramaSearchBarProps = {
+  schema: Schema;
+  index: RawData;
+  documentMapper?: DocumentMapper;
+};
+
+function identityDocumentMapper(doc: Document): SearchResult {
```
ovflowd: We prefer arrow functions within the codebase.

components/Common/OramaSearchBar/index.tsx
```diff
@@ -0,0 +1,91 @@
...
+  mapper: DocumentMapper,
+  term: string
+): Promise<SearchResult[]> {
+  const results = await search(database, { term });
+
+  return results.hits.map(h => mapper(h.document));
+}
+
+const Footer: FC<FooterProps> = ({ theme }: FooterProps) => {
+  return (
```
ovflowd, line 19: No need of a return statement here

components/Common/OramaSearchBar/index.tsx
```diff
@@ -0,0 +1,91 @@
...
+    <div className={styles.oramaSearchBarFooter}>
+      <span className={styles.oramaSearchBarFooterLabel}>
+        <FormattedMessage id="components.oramaSearchBar.search.poweredBy" />
+      </span>
+      <a
+        href={'http://oramajs.io'}
```
ovflowd:
```suggestion
        href="http://oramajs.io"
```

components/Common/OramaSearchBar/index.tsx
```diff
@@ -0,0 +1,91 @@
...
+        target="_blank"
+        rel="noopener noreferrer"
+        className={styles.oramaSearchBarFooterLink}
+      >
+        {theme === 'dark' ? (
```
ovflowd: This can be memoized I guess. Also probably you want to use `useTheme` hook for the theme :)

components/Common/OramaSearchBar/index.tsx
ovflowd, line 1: I think all the application logic, like performing search and etc should be done in a Hook and utils files.

components/Common/OramaSearchBar/index.tsx
```diff
@@ -0,0 +1,91 @@
...
+  schema: Schema;
+  index: RawData;
+  documentMapper?: DocumentMapper;
+};
+
+function identityDocumentMapper(doc: Document): SearchResult {
+  return doc as unknown as SearchResult;
+}
+
+async function performSearch(
```
ovflowd: This should be on a dedicated place just for search logic, like an util or just a separate file within `OramaSearchBar`

components/Common/SearchBar/index.module.scss
```diff
@@ -0,0 +1,135 @@
...
+
+    .searchResultCode {
+      border-radius: 0.3rem;
+      font-weight: var(--font-weight-light);
+      padding: 0 6px;
+    }
+  }
+}
+
+[data-theme='light'] .searchResultCode {
```
ovflowd, line 128: I don't think we're using `data-theme` for themes anymore. Search for "dark" and "light"

ovflowd (commented):
I haven't gone through the SearchBar and useKeypress hook yet, but this are my comments so far :)

AugustinMauroy:
Hey!
You should update NewHeader with search feat.

ovflowd:
No, @AugustinMauroy he doesn't need to. He (or anybody else) can do that on a follow-up PR. As this is out of scope of this PR.

components/Common/OramaSearchBar/index.module.scss
ShogunPanda: Solved. I was using `theme` instead of `resolvedTheme`.

components/Common/OramaSearchBar/index.tsx
ShogunPanda: I think I fixed it now. I let VS code fix it :)

components/Common/OramaSearchBar/index.tsx
ShogunPanda: You were right. Anyway I removed this after last refactor.

components/Common/OramaSearchBar/index.tsx
ShogunPanda: Fixed!

components/Common/OramaSearchBar/index.tsx
ShogunPanda: You were right. It is now after component refactoring :)

components/Common/OramaSearchBar/index.tsx
ShogunPanda: Fixed.

components/Common/OramaSearchBar/index.tsx
ShogunPanda: I was already using `useTheme` but in the parent component for another stupid reason I'm to embarassed to mention now.
Anyway this is fixed now. :)

components/Common/OramaSearchBar/index.tsx
ShogunPanda: Done. What about now?

components/Common/OramaSearchBar/index.tsx
ShogunPanda: See above :)

components/Common/SearchBar/index.module.scss
ShogunPanda: I checked and I couldn't find anything.

Instead I found [this](https://github.com/nodejs/nodejs.org/blob/main/styles/_base.scss#L45) which lead me to this implementation.
And in the code (at least in storybook) I see `[data-theme]`. 
Am I missing anything?

ovflowd:
FYI I still need to review the remaining code once I get some time 👀

ShogunPanda:
No worries. Let me know if I have to do anything elsee

components/Common/SearchBar/index.module.scss
shanpriyan: @ovflowd we're currently using `data-theme` for theming, since the `next-theme` introduction.

components/Common/OramaSearchBar/index.tsx
```diff
@@ -0,0 +1,63 @@
...
+import SearchBar from '../SearchBar';
+
+import type { FC } from 'react';
+import type { DocumentMapper, OramaSearchBarProps } from './search';
+import type { SearchFunction } from '../../../types';
+
+const Footer: FC = () => {
+  const { resolvedTheme: theme } = useTheme();
+
+  const isDark = useMemo(() => theme === 'dark', [theme]);
```
shanpriyan: No need of useMemo here, since there is no heavy computation stuff
```suggestion
  const isDark = theme === 'dark';
```

components/Common/OramaSearchBar/index.tsx
```diff
@@ -0,0 +1,63 @@
...
+    return documentMapper ?? identityDocumentMapper;
+  }, [documentMapper]);
+
+  // Create a setup function to deserialize the database
+  const setup = useCallback(async () => {
+    const database = await create({ schema });
+    await load(database, index);
+
+    setSearchFunction(() => performSearch.bind(null, database, mapper));
+  }, [schema, index, mapper, setSearchFunction]);
```
shanpriyan: I think we can use the `documentMapper` directly inside the  setup function

```suggestion
  // Create a setup function to deserialize the database
  const setup = useCallback(async () => {
    const database = await create({ schema });
    await load(database, index);

    setSearchFunction(() => performSearch.bind(null, database, documentMapper ?? identityDocumentMapper));
  }, [schema, index, documentMapper, setSearchFunction]);
```

components/Common/SearchBar/index.tsx
```diff
@@ -0,0 +1,343 @@
...
+    setExpanded(false);
+  }, [setQuery, setExpanded]);
+
+  const selectItem = useCallback(
+    (index: number) => {
+      if (!listRef.current) {
+        return;
+      }
+
+      console.log(index);
```
shanpriyan:
```suggestion
```

components/Common/SearchBar/index.tsx
```diff
@@ -0,0 +1,343 @@
...
+}: SearchBarProps) => {
+  const [isLoading, setIsLoading] = useState<boolean>(true);
+  const searchInputRef = useRef<HTMLInputElement | null>(null);
+  const listRef = useRef<HTMLUListElement | null>(null);
+  const activeResultRef = useRef<number>(-1);
+
+  const [isExpanded, setExpanded] = useState(false);
+  const [query, setQuery] = useState('');
+  const [results, setResults] = useState<SearchResult[]>([]);
+  const isEmpty = useMemo(() => !results || results.length === 0, [results]);
```
shanpriyan:
```suggestion
  const isEmpty = !results || results.length === 0;
```

components/Common/SearchBar/index.tsx
```diff
@@ -0,0 +1,343 @@
...
+    height: '3em',
+  },
+};
+
+const SearchBar: FC<SearchBarProps> = ({
+  setup,
+  search,
+  footer: Footer,
+}: SearchBarProps) => {
+  const [isLoading, setIsLoading] = useState<boolean>(true);
```
shanpriyan:
```suggestion
  const [isLoading, setIsLoading] = useState(true);
```

components/Common/SearchBar/index.tsx
```diff
@@ -0,0 +1,343 @@
...
+  // Focus the field when expanding the bar
+  useEffect(() => {
+    if (isExpanded) {
+      searchInputRef.current?.focus();
+    }
+  }, [isExpanded, searchInputRef]);
+
+  // Clear results when the query is emptied
+  useEffect(() => {
+    if (!query && results.length) {
```
shanpriyan: Can we do this in `onQueryChangeHandler` and remove this useEffect 🤔 ?

This useEffect will result in another render cycle after query has been updated

```js
if(!e.target.value) setResults([])
```

i18n/locales/en.json
```diff
@@ -32,6 +32,11 @@
   "components.downloadReleasesTable.docs": "Docs",
   "components.header.buttons.toggleLanguage": "Toggle Language",
   "components.header.buttons.toggleDarkMode": "Toggle dark/light mode",
+  "components.searchBar.placeholder": "Search",
```
shanpriyan: Please add the  new keys at the bottom of the file

https://github.com/nodejs/nodejs.org/blob/7a4e29c8090caa8980f51a5676731ea43888f729/TRANSLATION.md?plain=1#L54

package.json
```diff
@@ -36,8 +36,10 @@
   },
   "dependencies": {
     "@mdx-js/react": "^2.3.0",
+    "@mui/icons-material": "^5.11.16",
```
shanpriyan: We are currently using `react-icons` for icons package, update the icons from mui to react-icons

shanpriyan (changes requested):

components/Common/OramaSearchBar/index.tsx
ShogunPanda: 👍

components/Common/OramaSearchBar/index.tsx
ShogunPanda: 👍

components/Common/SearchBar/index.tsx
ShogunPanda: 👍

i18n/locales/en.json
ShogunPanda: 👍

package.json
ShogunPanda: 👍

ovflowd:
FYI gonna finally review the remaining of the PR this week.

ovflowd:
Relates to https://github.com/nodejs/nodejs.org/issues/5204

components/Common/OramaSearchBar/OramaLogo.tsx
```diff
@@ -0,0 +1,731 @@
...
+      x="0px"
+      y="0px"
+      viewBox="0 0 4207.11 1155.11"
+      className={className}
+      style={{ enableBackground: 'new 0 0 4207.11 1155.11' } as any}
+      xmlSpace="preserve"
+    >
+      <style
+        type="text/css"
+        dangerouslySetInnerHTML={{ __html: logoDarkStyle }}
```
Harkunwar: Could we move the styles to a proper css file so we don't have to use `dangerouslySetInnerHTML`

components/Common/SearchBar/index.tsx
```diff
@@ -0,0 +1,337 @@
...
+      e.preventDefault();
+
+      // Clear results when the query is emptied
+      if (!e.target.value) {
+        setResults([]);
+      }
+
+      setQuery(e.target.value);
+    },
+    [setQuery, setResults]
```
Harkunwar, line 99:
```suggestion
    []
```
We don't need to add these as any setter resulting from `useState` will never change.

components/Common/OramaSearchBar/index.tsx
```diff
@@ -0,0 +1,64 @@
...
+
+const OramaSearchBar: FC<OramaSearchBarProps> = ({
+  schema,
+  index,
+  documentMapper,
+}: OramaSearchBarProps) => {
+  const [searchFunction, setSearchFunction] = useState<SearchFunction>();
+
+  // Create a setup function to deserialize the database
+  const setup = useCallback(async () => {
```
Harkunwar, line 48: Is there any reason why we're passing the setup function to the child instead of just  doing the setup here. It feels more complicated.

components/Common/SearchBar/index.tsx
```diff
@@ -0,0 +1,337 @@
...
+    (index: number) => {
+      if (!listRef.current) {
+        return;
+      }
+
+      const el = listRef.current.children[index]!;
+      el.querySelector('a')?.focus();
+      el.scrollIntoView({ behavior: 'smooth', block: 'nearest' });
+    },
+    [listRef]
```
Harkunwar, line 75: Refs never change and don't need to be in deps array.

components/Common/SearchBar/index.tsx
```diff
@@ -0,0 +1,337 @@
...
+}: SearchBarProps) => {
+  const [isLoading, setIsLoading] = useState(true);
+  const searchInputRef = useRef<HTMLInputElement | null>(null);
+  const listRef = useRef<HTMLUListElement | null>(null);
+  const activeResultRef = useRef<number>(-1);
+
+  const [isExpanded, setExpanded] = useState(false);
+  const [query, setQuery] = useState('');
+  const [results, setResults] = useState<SearchResult[]>([]);
+  const isEmpty = !results || results.length === 0;
```
Harkunwar, line 46: I feel like this Component is doing to many things. The interactions  between a lot of the state objects is making it complicated. Can we try to use maybe `useReducer` to see if we can clean things up here?

components/Common/OramaSearchBar/index.tsx
ShogunPanda: I designed `SearchBar` to be backend agnostic, therefore it accepts a `setup` function to prepare (and thus not allow searching before setup has completed.)
If we do as you suggested, then `useCallback` will become a `useEffect`, and then I'll have to change SearchBar to accept a boolean for `isLoading`. I see little benefit here as the setup part is separated from the component anyway.

components/Common/SearchBar/index.tsx
ShogunPanda: Here and below: while I agree on that, I tend to pass all values (including immutable one) to useCallback/useMemo in order to have a quick view of what is used in the function. I don't think it affect performance or are there any cons in doing this I'm not aware of?

components/Common/SearchBar/index.tsx
ShogunPanda: See https://github.com/nodejs/nodejs.org/pull/5419#discussion_r1251636993

components/Common/SearchBar/index.tsx
ShogunPanda: I'm -1 on this.

`query` and `results` are basically changed at the same time.
`expanded` is triggered independently.
The only other state is `isLoading` which is only changed once.

I feel adding adding a reducer would be slightly overengineered.

components/Common/OramaSearchBar/OramaLogo.tsx
ShogunPanda: I tried. And I discover none of the styles were actually used. Lol.
I've removed everything.

components/Common/OramaSearchBar/index.tsx
ovflowd: Let me re-review 👀

ovflowd:
And here I go reviewing this, sorry for the super significant delay, @ShogunPanda!

components/Common/OramaSearchBar/index.module.scss
```diff
@@ -0,0 +1,24 @@
...
+  align-items: center;
+  column-gap: 1ch;
+  display: flex;
+  justify-content: end;
+  margin: 5px;
+
+  .oramaSearchBarFooterLabel {
```
ovflowd, line 8: You can use `&_` statements here, so you don't need to add the whole prefix again. (Actually in all the child Sass styles here)

Eg.:

```
.oramaSearchBarFooter {

  &_label {

  }

}
```

components/Common/OramaSearchBar/index.tsx
```diff
@@ -0,0 +1,64 @@
...
+import { OramaLogoDark, OramaLogoLight } from './OramaLogo';
+import styles from './index.module.scss';
+import { identityDocumentMapper, performSearch } from './search';
+import SearchBar from '../SearchBar';
+
+import type { FC } from 'react';
+import type { OramaSearchBarProps } from './search';
+import type { SearchFunction } from '../../../types';
+
+const Footer: FC = () => {
```
ovflowd, line 14: Can this become a separate component `OramaSearchBarFooter.tsx`? Just because our approach here is to have one top-level Component per file.

components/Common/OramaSearchBar/index.tsx
```diff
@@ -0,0 +1,64 @@
...
+      <span className={styles.oramaSearchBarFooterLabel}>
+        <FormattedMessage id="components.oramaSearchBar.search.poweredBy" />
+      </span>
+      <a
+        href="http://oramajs.io"
+        target="_blank"
+        rel="noopener noreferrer"
+        className={styles.oramaSearchBarFooterLink}
+      >
+        {isDark ? (
```
ovflowd, line 30: Maybe we can simplify this to:

```tsx

const OramaLogo = theme === 'dark' ? OramaLogoDark : OramaLogoLight;

...

<OramaLogo className={styles.oramaSearchBarFooterLogo} />
```

components/Common/OramaSearchBar/index.tsx
```diff
@@ -0,0 +1,64 @@
...
+    setSearchFunction(() =>
+      performSearch.bind(
+        null,
+        database,
+        documentMapper ?? identityDocumentMapper
+      )
+    );
+  }, [schema, index, documentMapper, setSearchFunction]);
+
+  return <SearchBar setup={setup} search={searchFunction} footer={Footer} />;
```
ovflowd, line 61: I assume that the `SearchBar` awaits for the setup function?

components/Common/OramaSearchBar/search.ts
```diff
@@ -0,0 +1,26 @@
+import { search } from '@orama/orama';
+
+import type { Document, Orama, RawData, Schema } from '@orama/orama';
+import type { SearchResult } from '../../../types';
+
+export type DocumentMapper = (doc: Document) => SearchResult;
+
+export type OramaSearchBarProps = {
```
ovflowd, line 8: I believe types should be solely defined on a `types.ts` file just to separate type definitions from actual `search` logic

components/Common/OramaSearchBar/search.ts
```diff
@@ -0,0 +1,26 @@
...
+
+export const identityDocumentMapper = (doc: Document): SearchResult => {
+  return doc as unknown as SearchResult;
+};
+
+export const performSearch = async (
+  database: Orama,
+  mapper: DocumentMapper,
+  term: string
+): Promise<SearchResult[]> => {
```
ovflowd, line 22: I feel that this could be transformed into a promise. Could we do that?

components/Common/OramaSearchBar/search.ts
```diff
@@ -0,0 +1,26 @@
...
+
+export type DocumentMapper = (doc: Document) => SearchResult;
+
+export type OramaSearchBarProps = {
+  schema: Schema;
+  index: RawData;
+  documentMapper?: DocumentMapper;
+};
+
+export const identityDocumentMapper = (doc: Document): SearchResult => {
```
ovflowd, line 14: Any reason why this function exists? It seems just to be casting one type to another.

components/Common/SearchBar/index.tsx
ovflowd, line 1: I'm sadly negatively impressed and shocked with the pure size of this Component.

It is so freaking complex. We should discard this component and do something from scratch.

We don't need to keyboard listeners, handlers and all these things now. We can iterate later.

I know this Component is part of the migration, but since it is so complex, it makes it harder to know what's being migrated and what is not.

If I might suggest, let's discard this component and its styles and feel free to make a new one that allows you to type and render the results.

It would be nice if styles, at least of the search container, could be similar, but the content itself, how it collapses/expands, and how it renders the results up to you.

I'm afraid I have to disagree with a Component with many callbacks, effects, conditions and more. Even the `useEffect`'s super complicated.

For example, the rendering of the list (https://github.com/nodejs/nodejs.org/pull/5419/files#diff-cc0fb6b25cb6dad14226e6daa7b1819a859f6c9e5aa102d9557c52b2db485f67R292) is highly complex; we have two effects (https://github.com/nodejs/nodejs.org/pull/5419/files#diff-cc0fb6b25cb6dad14226e6daa7b1819a859f6c9e5aa102d9557c52b2db485f67R157, https://github.com/nodejs/nodejs.org/pull/5419/files#diff-cc0fb6b25cb6dad14226e6daa7b1819a859f6c9e5aa102d9557c52b2db485f67R173) that are also super complex.

I want also to avoid mixing async/await logic on effects, etc., as much as possible. Otherwise, we're creating way too many side-effects.

TL;DR I will find it hard to approve this PR with this Component as it's way too complex. It will on-the-go scare anyone to modify/contribute to this component and it is massively hard to read and understand what it is doing 😅

hooks/useKeypress.ts
ovflowd, line 1: Let's remove this hook from now. It is complex, and the code quality is arguable.

We can think about accessibility on a second pass for this PR. Let's get the basics done, and then we think about invoking the search bar with keystrokes or moving through entries with keystrokes.

components/Common/SearchBar/index.tsx
ovflowd: I would also mention that the SearchBar component should probably be broken down in 3 components:

- Search Container
  - This is the actual wrapper where you type your query
- Search Results
  - The container that renders conditionally when there are results
- Search Result / Search Entry
  - The individual element for each result can have different styles

I would also say that SearchResults would have the reference of what is the current highlighted result and handle the keystrokes from going up and down.

And Search Container would handle the keystrokes for expanding/collapsing the search container.

+ on that, I would also say we should make our key-stroke hook very generic once we implement it. But again, all keystroke logic can be done on a follow-up PR.

types/search.ts
```diff
@@ -0,0 +1,12 @@
+export type SearchResult = {
```
ovflowd, line 1: Usually, on the `types` folder we use interfaces whenever possible. So this one should be an interface.

ovflowd (changes requested):
Overall fantastic work done here, @ShogunPanda.

My review has some minor comments and significant concerns, so I'm sadly requesting changes. Mostly are tied to the SearchBar component.

Thank you so much for your effort here! I hope we're able to address these review concerns so we can move forward 💪

components/Common/OramaSearchBar/index.module.scss
ShogunPanda: Cool, I'll update this

components/Common/OramaSearchBar/index.tsx
ShogunPanda: I guess so, it's pretty self contained. Will move it soon.

components/Common/OramaSearchBar/index.tsx
ShogunPanda: 👍

components/Common/OramaSearchBar/index.tsx
ShogunPanda: Yes, see on line 157.

components/Common/OramaSearchBar/search.ts
ShogunPanda: Not an issue. Will move them.

components/Common/OramaSearchBar/search.ts
ShogunPanda: Ehm... it is already a promise. What do you mean exactly?

components/Common/OramaSearchBar/search.ts
ShogunPanda: The default one is technically a no-op.

Similarly to `SearchBar`, I designed those functions to be use-agnostic. 
Depending on when you're rendering results, you might need to map Orama results to something else. I left room for a simple mapping function to be passed, rather than implementing this inside `useEffect` or similar.

components/Common/OramaSearchBar/search.ts
ovflowd: You’re mixing async with Promises. I’m referring to use a pure Promise approach for this function

ShogunPanda: Ehm... `Promise` it's just the return type. Are looking for something like this?

```
export const performSearch = (
  database: Orama,
  mapper: DocumentMapper,
  term: string
): Promise<SearchResult[]> => {
  return search(database, { term }).then(results => results.hits.map(h => mapper(h.document)));
};
```

components/Common/SearchBar/index.tsx
ShogunPanda: Ok, I'll try to split components as you suggested.
Will follow up on this I hope within this week.

components/Common/OramaSearchBar/search.ts
ovflowd: Yup, that’s what I meant. I also guess we don’t need a return statement as this is an one liner.

Yeah, I meant native promises rather than async/await.

components/Common/SearchBar/index.tsx
ovflowd: Yeah I think by doing that, and simplifying the logic and removing everything that is not necessary we make this whole solution way easier.

ovflowd: But let’s also remove unnecessary callbacks, effects and try to simplify them as much as possible. I haven’t gone through the whole component but I guess there’s room for improvements there.

ovflowd:
Hey @ShogunPanda have you had some time to follow-up on this? Do you need any assistance! Thanks :3

ShogunPanda:
Nope, I don't need any help at the moment.
I plan to to work on this next week. I hope to have some updates soon!

tuanggo (commented):
open

ShogunPanda:
I mistakenly deleted my fork. Will send a new PR (which fixes all the suggestions outlined here) soon. Sorry for the mess.

PR #5425, opened 2023-06-12
chore: merge shellbox into codebox

araujogui:
## Description

Merge Shellbox into Codebox

### Check List


ovflowd:
Can on dark theme the button to have this background and colour? 

image

Also, the "copy" text should be "copy" not "Copy" or we can use a Copy emoji/icon from React Icons (I rather the emoji/icon!)

ovflowd:
Also what about the selected option also have a darker colour? (Screenshot below with the change)

image

The colour is `--dark5`

ovflowd:
Also if we use an icon for the copy/paste button instead of text (which I think is better) we need to remove the translation. (And also change the width of the button)

AugustinMauroy:
Personally -1 for the icons on button. Copy is more explicit that an icons. Then there's nothing specific about the design system.

ovflowd:
> Personally -1 for the icons on button. Copy is more explicit that an icons. Then there's nothing specific about the design system.

A copy icon is a very standard practice. Is actually the most common thing done out there. (Example from Next.js docs)

image

ovflowd:
Like something like this 
image

But maybe floating or hmmm

AugustinMauroy:
With more padding and small increase of size and ok

ovflowd:
My screenshot was just an example. Ideally the way of the Next.js docs would be great. But we don't have a designer, so...

AugustinMauroy:
I'm learning design with Figma. Are you interested by I mock-up ?

ovflowd:
> I'm learning design with Figma. Are you interested by I mock-up ?

Thanks, but we don't need that for now. The Foundation is also in talks on providing a Designer, so that's awesome news.

shanpriyan:
> Also, the "copy" text should be "copy" not "Copy" or we can use a Copy emoji/icon from React Icons (I rather the emoji/icon!)

**+1** for using an copy emoji

hooks/useCopyToClipboard.ts
```diff
@@ -19,7 +19,7 @@ export const useCopyToClipboard = () => {
 
   useEffect(() => {
     if (copied) {
-      const timerId = setTimeout(() => setCopied(false), 3000);
+      const timerId = setTimeout(() => setCopied(false), 2000);
```
shanpriyan: Any particular reason for changing in duration ? I believe 3000ms is sufficient enough

mikeesto (commented):
Nice work. The unit tests are failing atm, I think we just need a new snapshot.

It's also kind of annoying that the `$ ` is copied for shell. If it's not fixed in this PR, I think it would be good to create a new issue for it.

ovflowd:
cc @araujogui any updates here?

ovflowd:
Also, @araujogui can we create a prop that allows us to disable the "header" of the Codebox (like `hideHeader`), which hides the language label? Like the screenshot below

image

Also when the header is hidden the copy button should only appear when hovering and should be inline with the code like in the screenshot below,

image

ovflowd:
Also on Light Mode, the copy button background feels weird. It doesn't fit. It should be a lighter green variation.

image

AugustinMauroy (changes requested):
We should do something for lang button. Because now in "floating" it's look strange. Before that better (when we just have the selected in bold)

ovflowd:
> We should do something for lang button. Because now in "floating" it's look strange. Before that better (when we just have the selected in bold)

I don't think it looks strange, many Codeboxes do this approach of having a separated header.

AugustinMauroy:
Capture d’écran 2023-06-28 à 13 32 07

Look up at this example. I really thinks it's will better to not have padding.

ovflowd:
> Look up at this example. I really thinks it's will better to not have padding.

We can experiment with no padding between the headers and a simple "|" (division) between each entry. But I'm fine also with the current approach. But maybe the other approach would feel better.

components/Article/Codebox/index.module.scss
```diff
@@ -3,97 +3,143 @@
 @use 'styles/mixins/components';
 
 .pre {
-  @include components.code;
+  border-radius: 0.3rem;
```
ovflowd: Can we use a var here?

components/Article/Codebox/index.tsx
```diff
@@ -1,13 +1,18 @@
 import { useEffect, useState } from 'react';
-import { FormattedMessage } from 'react-intl';
 import { highlight, languages } from 'prismjs';
 import classnames from 'classnames';
+import { FaRegCopy, FaCheck } from 'react-icons/fa';
```
ovflowd: These imports are unused

ovflowd: Please double check your code for unused imports :)

components/Article/Codebox/index.tsx
```diff
@@ -1,13 +1,18 @@
...
+import { TbCopy, TbCheck } from 'react-icons/tb';
+import { useIntl } from 'react-intl';
 import styles from './index.module.scss';
```
ovflowd: Move style import to after all imports on separated line

components/Article/Codebox/index.tsx
```diff
@@ -1,13 +1,18 @@
...
 import { useCopyToClipboard } from '../../../hooks/useCopyToClipboard';
 import type { FC, PropsWithChildren, ReactElement, MouseEvent } from 'react';
+import 'prismjs/components/prism-bash';
```
ovflowd: Why are we importing this here?

components/Article/Codebox/index.tsx
```diff
@@ -48,11 +64,34 @@ const Codebox: FC<CodeBoxProps> = ({ children: { props } }) => {
       highlight(codeArray[langIndex], prismLanguage, parsedLanguage)
     );
     // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [langIndex]);
+  }, [langIndex, codeArray]);
+
+  const ariaLabelText = intl.formatMessage(
```
ovflowd: We don't need this `aria-label` for now. screen readers will not use a "copy" button, this button is not "accessible" right now

components/Article/Codebox/index.tsx
```diff
@@ -48,11 +64,34 @@ const Codebox: FC<CodeBoxProps> = ({ children: { props } }) => {
...
+    { copied }
+  );
+
+  const copyButton = (
+    <button
+      type="button"
+      className={styles.copy}
+      onClick={handleCopyCode}
+      aria-label={ariaLabelText}
+      data-testid="copy"
```
ovflowd: I'd remove this `data-testid" as it's not being used. Remove `aria-label` and add `aria-hidden="true"`v

components/Article/InlineCode/index.module.scss
```diff
@@ -3,7 +3,10 @@
 .code {
   @include components.code;
 
+  border-radius: 0.3rem;
```
ovflowd: Again this standardise border-radius can become a var for border radius

i18n/locales/en.json
```diff
@@ -58,7 +58,6 @@
   "components.home.nodeFeatures.openSource.description": "Node.js is open source and actively maintained by contributors all over the world",
   "components.home.nodeFeatures.everywhere.title": "Everywhere",
   "components.home.nodeFeatures.everywhere.description": "Node.js has been adapted to work in a wide variety of places",
-  "components.common.shellBox.copy": "{copied, select, true {copied}other {copy}}",
   "components.codeBox.copy": "{copied, select, true {copied}other {copy}}",
```
ovflowd: This can be removed as aria is getting gone. Also this should have ben `compoonents.common.codeBox.copy` 😢

components/Article/Codebox/index.tsx
araujogui: Actually I am using it in the unit tests.

components/Article/Codebox/index.tsx
araujogui: Prism.js doesn't come with syntax highlighting for bash or zsh by default, and we need it in `Shellbox`

components/Article/Codebox/index.tsx
ovflowd: I didn't see a reference to this in the unit tests.

araujogui: `const buttonElement = await screen.findByTestId('copy');`

components/Article/Codebox/__tests__/index.test.tsx
```diff
@@ -37,12 +37,37 @@ describe('Codebox component (one lang)', (): void => {
...
+        </Codebox>
+      </IntlProvider>
+    );
+
+    const navigatorClipboardWriteTextSpy = jest.spyOn(
+      navigator.clipboard,
+      'writeText'
+    );
+
+    const buttonElement = await screen.findByTestId('copy');
```
ovflowd: I thought the usage of `screen` was discouraged.

components/Article/Codebox/index.tsx
ovflowd: Can you give an example of what you meant by not having syntax highlighting for bash?

ovflowd: What bothers me is that we're loading this at bundle time even if we don't need it.

I would if possible, create a dynamic loader (utility file) that loads prismsjs modules on demand.

```tsx
export const loadPrismLanguages = async (language: string) => {
  switch (language) {
    case 'bash':
    case 'zsh':
      await import('prismjs/components/prism-bash');
      break;
    default:
      break;
  }
}
```

Something like that. We can do a call later today on Discord, wdyt?

components/Article/Codebox/index.tsx
```diff
@@ -48,11 +61,27 @@ const Codebox: FC<CodeBoxProps> = ({ children: { props } }) => {
...
     // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [langIndex]);
+  }, [langIndex, codeArray]);
+
+  const copyButton = (
+    <button
+      type="button"
+      className={styles.copy}
+      onClick={handleCopyCode}
+      aria-hidden
```
ovflowd: This should probably be `aria-hidden={true}`

components/Article/Codebox/index.tsx
ovflowd: This should probably be:

```tsx
const { container } = render(<MyComponent />);

const button = container.querySelector('button');
```

For this simple example, no need for a test-id

components/Article/Codebox/__tests__/index.test.tsx
bmuenzenmeyer: this tripped up our team for a while too. but the maintainer actually says screen is peferred from a code organization standpoint. less keeping destructures up to date per test

https://kentcdodds.com/blog/common-mistakes-with-react-testing-library#not-using-screen

ovflowd (approved):
LGTM! Fantastic work you've done here. Also, the pair programming sessions were super insightful!

This component looks way better and the new design is stunning!

mikeesto (approved):
Great work

bmuenzenmeyer:
> It's also kind of annoying that the `$ ` is copied for shell. If it's not fixed in this PR, I think it would be good to create a new issue for it.

Yeah I've seen a technique where it's visually present but not copy-able. And in case people don't know (because I didn't forever), the prompt does serve some purpose, if however small, in showing that `sudo` is not needed.

ovflowd:
Well this PR fixed the $ issue but only for $.

Maybe what we need in the future is a field called prefix that allows us to add non-cooyable pieces

AugustinMauroy:
I’m still not fan of Lang button. I propose to glue them and use an separator like |

ovflowd:
> I’m still not fan of Lang button. I propose to glue them and use an separator like |

Please open the latest version of Storybook before commenting this.

AugustinMauroy:
In lasted version of storybook it’s isn’t glue

ovflowd:
@araujogui a tiny bug the third one should not have left border-radius and the second one should.

image

ovflowd:
> In lasted version of storybook it’s isn’t glue

image

Yes, they are. Or what do you mean with "glue" if this is not what you're referring to?

AugustinMauroy:
7B674B7D-0818-43A9-B21A-DE9C1E6D6911

idk but it’s render like that on my phone

ovflowd:
image

It is working as expected on my "mobile" viewport. @araujogui are you able to reproduce this?

@AugustinMauroy can you tell me which browser you're using?

AugustinMauroy:
Safari lasted on iPhone 7 (I’m on lasted storybook deploy)

ovflowd:
> Safari lasted on iPhone 7 (I’m on lasted storybook deploy)

I can't reproduce this on Safari macOS, nor on Safari on my iPhone 14 Pro (iOS 17)

ovflowd:
I'm merging this as it is now, but @araujogui, can you make a follow-up PR with the tiny bug fix I've mentioned? This PR just had been sitting for so long, and we can reiterate the minor bugs on follow-up PRs.

PR #5480, opened 2023-07-08
docs: remove dependencies page and link to existing resource

bmuenzenmeyer:
## Description

Discussion within #5185 evolved to suggest directly linking to a [source within nodejs/node](https://github.com/nodejs/node/blob/main/doc/contributing/maintaining/maintaining-dependencies.md) instead of attempting to sync the content.

- ✅ there is precedent in doing this on the docs site already, such as https://github.com/nodejs/node/blob/HEAD/SECURITY.md#security
- ✅ the proposed change routes to a file that is already well-maintained and kept up to date during actual dependency updates

❓ I removed the translated content as part of this PR. I wasn't sure what the process was for _removing_ translations. Happy to drop 604408dff2beffe8f027eade12b9b6185ae578e2 if needed.

❓ I suspect we should create redirects from [locale]/docs/meta/topics/dependencies to the new content, but I do not know how to do that.

## Validation

From the preview, click on Docs > Dependencies

It should take you to https://github.com/nodejs/node/blob/main/doc/contributing/maintaining/maintaining-dependencies.md

image

## Related Issues

closes #5185

### Check List


bmuenzenmeyer:
will work on rebasing this

mikeesto:
On https://nodejs.org/en/docs there is a brief outline of the different types of documentation available that line up with the left navigation options. I'm wondering whether we should put a brief addition here for "dependencies"? For completeness more than anything else (?)

bmuenzenmeyer:
> For completeness more than anything else (?)

Good question. I didn't even take too much notice to the content on that page, but it has some problems. 

First, the easy thing, adding a blurb as you described for completion's sake.

image

The second problem (to me), is the order of the side nav vs the content itself.

In the left nav, "E6 and beyond" needs to go below the "API documentation", according to the content and content TOC.

image

---

@nodejs/website is this getting to far into the realm of content? According to [content vs code](https://github.com/nodejs/nodejs.org/blob/main/CONTENT_VS_CODE.md), other folks might have a say.

ovflowd:
> https://github.com/orgs/nodejs/teams/website is this getting to far into the realm of content? According to [content vs code](https://github.com/nodejs/nodejs.org/blob/main/CONTENT_VS_CODE.md?rgh-link-date=2023-07-08T16%3A52%3A42Z), other folks might have a say.

I would say maybe. We usually don't have much of a say in content, but we can raise issues and discuss them. And if no other team has any particular opinion, we can work on that.

I would just advise us to not spend too much time on this as all the current pages will cease to exist, as we're going to adopt the pages from nodejs.dev, and no such page as "Dependencies" exists there. Maybe at the time, they already considered this.

bmuenzenmeyer:
> I would just advise us to not spend too much time on this as all the current pages will cease to exist

Good point. I keep forgetting that. To me that's an argument for less concern and more expediency in fixing things. When I have more time later I'll make the changes and ask for fastrack

bmuenzenmeyer:
This is now ready for final review

mikeesto (approved):
LGTM, thanks

ovflowd (approved):
LGTM!

PR #5506, opened 2023-07-13
chore: add validation instructions to pr template

bmuenzenmeyer:
## Description

Pull requests like https://github.com/nodejs/nodejs.org/pull/5500 created confusion as to what the actual bug and fix were. Reviewers had to compare old and new version of the code to determine the efficacy of the change.

This PR updates our pull request template to add a new section, `Validation`. It provides brief guidance to the submitter as an HTML comment:

- How do you know this is working? 
- What should a reviewer look for? 
- Provide a screenshot if your change is visual.

### Check List


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -6,6 +6,10 @@ Please read the [Code of Conduct](https://github.com/nodejs/nodejs.org/blob/main
 
 <!-- Write a brief description of the changes introduced by this PR -->
 
+## Validation
```
HinataKah0, line 9: IMO the term `Validation` is a bit too broad, it can mean any Validation.
I would say something like "Test Plan" suits better in this case.
The benefit of having a Test Plan is that we encourage people to test the changes first before Code Review.

~~However, I am wondering if it's a necessary step. Why? Because a Preview will be auto deployed (we can see the Preview instead). And after seeing the PR you're referencing, I think that the description can be made clearer hence there's no need to have a Test Plan in the first place.~~

HinataKah0: Adding more...

But if we have something that can't be covered by Preview, e.g. result from running a command. Then, having a Test Plan as a necessary step will benefit us.

Edit:
Maybe having a before and after comparison is a good idea for bug fixes.

bmuenzenmeyer: happy to rename it. the goal is to make things WAY more clear what was broke/changed, how you fixed it, and how reviewers will know its working/fixed. The automated deployments arent enough. See https://github.com/nodejs/nodejs.org/pull/5500#pullrequestreview-1525353617

The burden of proof that a change is valid is on the pull request author. We all have too much to do as reviewers to guess.

HinataKah0: > The burden of proof that a change is valid is on the pull request author.

Agree on this. I prefer not to review a non-working PR since it's still subjected to change.

It helps the reviewers to recall the necessary contexts as well.
In some places, it's actually a mandatory to write the Validation/Test Plan.

HinataKah0 (approved):
+1 on this for better workflow (although let's wait for other opinions as well).

A small nit is to rename it to `Test Plan` but let's see how other people think about it.
Thanks for bringing it up!

bmuenzenmeyer:
@nodejs/website any other opinions on `Validation` vs `Test Plan`?

mikeesto:
I don't have a strong view on validation vs test plan. At the end of the day, it boils down to how well the contributor communicates the changes in their PR and I'm not sure the wording of the section is going to change that.

I think it would be helpful if visual changes are accompanied by screenshots. I would probably actually prefer a section that very explicitly says 

```
Screenshots (if applicable)
<-- If your changes involve visual modifications to the website, please include screenshots that demonstrate the before and after state. --> 
```

bmuenzenmeyer:
lets go with this for now and see how it feels over time

PR #5518, opened 2023-07-14
Improvement banner#5510

BahugunaSaumya:
Description
A Fix to Improvement: Banner issue [5510](https://github.com/nodejs/nodejs.org/issues/5510)
mostly all the changes are committed but changes are as follows:-
Added a div to the react component and added a class called banner content to make the div centred and add the banner background to it.

Newer:- Changed the colour of the button matched values with the Figma prototype
added all the changes mentioned in the above comment except the button size increase

One thing to review:- I tried fitting the banner by 100% to the sides but due to padding put by StoryBook div, I had to put a margin = -1 in the banner which helped solve the issue but I am not sure if it is the sustainable solution, so can you have a quick look and let me know if I need to change that - [It is in index. module.scss, line 27 , inside.bannerContent class](https://github.com/BahugunaSaumya/nodejs.org/blob/dab3b37594403f9e0a804a77d76e1edefbe77abf/components/Common/Banner/index.module.scss#L27)

Related Issues
Check List
 - [x] I have read the Contributing Guidelines and made commit messages that follow the guideline.
 - [x] I have run npx turbo lint to ensure the code follows the style guide. And run npx turbo lint:fix to fix the style errors if necessary.
 - [x] I have run npx turbo format to ensure the code follows the style guide.
 - [x] I have run npx turbo test to check if all tests are passing, and/or npx turbo test:snapshot to update snapshots if I created and/or updated React Components.

### DIFF
--- a/components/Common/Banner/index.module.scss
+++ b/components/Common/Banner/index.module.scss
@@ -25,7 +25,6 @@
     justify-content: center;
     line-height: var(--line-height--display5); /* 150% */
     margin: -1rem;
-    max-width: 150% !important;
     padding: var(--space-13);
     width: 100%;
   }
--- a/components/Common/Banner/index.module.scss
+++ b/components/Common/Banner/index.module.scss
@@ -16,7 +16,7 @@
 
   .bannerContent {
     background-color: var(--color-fill-banner);
-
+    border-radius: 0;
     color: var(--black-900, #2c3437);
     display: flex;
     font-family: --var(--sans);
--- a/components/Common/Banner/index.module.scss
+++ b/components/Common/Banner/index.module.scss
@@ -1,16 +1,33 @@
 .banner {
   color: var(--color-text-primary);
   margin: 0 auto;
-  max-width: 90vw;
-  padding-top: var(--space-08);
+  max-width: 90vw; /*Can be changed based on design requirement*/
+  padding-top: var(--space-08); /* Add additional margin if needed */
   position: relative;
+  display: flex; /* Aligns child content */
+  justify-content: center; /* Centers child content horizontally */
+  align-items: center; /* Centers child content vertically */
+
 
   &,
   p {
     font-size: var(--font-size-body1);
     font-weight: var(--font-weight-bold);
   }
-
+  
+  .bannerContent {
+    align-items: center;
+    background-color:  var(--color-fill-banner);
+    border-radius: .2px;
+    display: flex;
+    justify-content: center;
+    margin: 0;
+    padding: var(--space-12);
+    width: 100%;
+    max-width: 90vw;
+    margin: 0 auto;
+  }
+  
   a {
     align-items: center;
     color: var(--color-text-primary);
@@ -31,7 +48,7 @@
     &.bannerBtn {
       background: var(--purple5);
       border: 1px solid transparent;
-      border-radius: 5.6rem;
+      border-radius: .2rem;
       color: var(--color-fill-top-nav);
       font-family: var(--sans);
       font-size: 1rem;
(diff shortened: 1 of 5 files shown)


BahugunaSaumya:
Hey, Is there anything wrong or something different than what you had in your mind in my PR?

ovflowd:
The font size is wrong. It should be 14px and not 18px. Also, the height should be `3rem` (`var(--line-height-subheading)`)

You also manually set the colour property of the text, which is wrong. It would be best if you used the variables. Otherwise, on the dark theme, the colour looks completely wrong. Also, the colour on the hover should not be `var(--color-fill-app)` otherwise, on dark the hover also feels incorrect.

components/Common/Banner/index.module.scss
```diff
@@ -11,6 +14,21 @@
     font-weight: var(--font-weight-bold);
   }
 
+  .bannerContent {
+    background: var(--color-fill-canvas);
+    border-radius: 0;
+    color: var(--color-text-primary);
+    display: flex;
+    font-family: --var(--sans);
```
ovflowd: This line should be removed. (There's no such font-family)

ovflowd (commented):
Overall there are many styles that are repeated or improper/should be removed. I would recommend to please cleanup some of these styles if possible.

Also, the font-size is wrong. 1rem is way too small. Please try to check other font-size variables that we have that better match what would 14px look like :)

components/Common/Banner/index.module.scss
```diff
@@ -29,24 +46,26 @@
...
+      border-radius: var(--border-radius-3);
+      color: var(--color-brand-primary);
+      flex-shrink: 0;
+      font-size: var(--font-size-body2);
       font-style: normal;
       font-weight: var(--font-weight-semibold);
-      line-height: var(--line-height-subheading);
-      margin-right: var(--space-32);
-      padding: 0 var(--space-16);
+      height: var(--space-32);
```
ovflowd: No need to have a height actually, it calculates by itself correctly already

components/Common/Banner/index.module.scss
```diff
@@ -29,24 +46,26 @@
...
+      margin-right: var(--space-12);
+      padding: 0 var(--space-32);
       position: relative;
       text-decoration: none;
       white-space: nowrap;
+      width: var(--space-92);
```
ovflowd: This variable `--space-92` does not exist. Also no need to have a width setting here

mikeesto:
@BahugunaSaumya Do you need to update the snapshots? I ran the PR checks for you and storybook tests are not passing atm

BahugunaSaumya:
Hey Michael,


I ran it but it just says remote cachin disabled,

mikeesto:
@BahugunaSaumya If you run prettier now, that should fix up the formatting in `components/Common/Banner/index.module.scss` and then hopefully we can land this

ovflowd (approved):
LGTM! Thank you for your first contribution! Super appreciate the effort put into this! 🙇

ovflowd (changes requested):
FYI I've just noticed the DOM Validation fails on the UNIT tests.

Apparently, you have `p` elements inside other `p` elements. This is happening in this test `components/Common/Banner/__tests__/index.test.tsx`

I would say the Banner should be a `div` element and not a paragraph element. So if you can revert the `p` change that be great!

BahugunaSaumya:
I've done that now, thankyou for the help

ovflowd:
> I've done that now, thankyou for the help

Did you also update the snapshots after changing the tags? 👀

BahugunaSaumya:
Yes I did that this time :-)

BahugunaSaumya (commented):
https://github.com/nodejs/nodejs.org/pull/5518#pullrequestreview-1542254270 Done that now

BahugunaSaumya:
https://github.com/nodejs/nodejs.org/pull/5518#pullrequestreview-1542254270 finished that

ovflowd (approved):

Issue #5492, opened 2023-07-09
Add Lighthouse-CI


bmuenzenmeyer:
Add the Lighthouse CI GitHub action

https://github.com/GoogleChrome/lighthouse-ci

We want to be more intentional about the impact development does to real-user metrics, and choose to accept any changes over time to a baseline.

This is somewhat broken out of a conversation within #5489

Chandupuram:
Hi @bmuenzenmeyer Can I pick this issue and start working on it?

bmuenzenmeyer:
sure thing, let me know if you have any questions

gaurav15gkgk:
Hey @bmuenzenmeyer, Can I pick this feature request and work on it?

bmuenzenmeyer:
> Hey @bmuenzenmeyer, Can I pick this feature request and work on it?

@gaurav15gkgk no, sorry. @Chandupuram asked first. You are welcome to review other pull requests or pick up something else

Chandupuram:
@bmuenzenmeyer 
ERROR:
23 Unable to automatically determine the location of static site files.
24 Use the CLI flag --collect.staticDistDir or a lighthouserc config file to tell Lighthouse CI where your HTML files are located.
I'm facing this error. Can u point me to the directory where we can find the static html files in our repo?

bmuenzenmeyer:
can you follow this guide with success? it's for Next.js apps https://dev.to/joerismits/ensure-your-nextjs-apps-performance-is-top-notch-with-lighthouse-ci-and-github-actions-4ne8

Chandupuram:
After trying the method posted in the blog I'm getting this error message
 "runtimeError": {
    "code": "ERRORED_DOCUMENT_REQUEST",
    "message": "Lighthouse was unable to reliably load the page you requested. Make sure you are testing the correct URL and that the server is properly responding to all requests. (Status code: 404)"
  },
  "runWarnings": [
    "Lighthouse was unable to reliably load the page you requested. Make sure you are testing the correct URL and that the server is properly responding to all requests. (Status code: 404)"
  ],
But when I start my local the URL is working and the same url is not working for Lighthouse CI I tried various ways but of no use Can u help me on this

bmuenzenmeyer:
https://github.com/GoogleChrome/lighthouse-ci/blob/main/docs/configuration.md#startserverreadypattern should be defaulting to `listen|ready`, which the site outputs on `npm run serve`

[...]

PR #5521, opened 2023-07-16
Add Lighthouse-CI #5492

Chandupuram:
## Description

This PR consists of the infra setup for the Lighthouse CI for this project. This helps us to check and monitor the performance of the pages and various statistics after each and every push done to the repo.

## Related Issues
Add Lighthouse-CI #5492

### Check List


.github/workflows/lighthouse-ci.yml
```diff
@@ -0,0 +1,14 @@
+name: CI
+on: [push]
+jobs:
+  lighthouseci:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/setup-node@v3
+        with:
+          node-version: 18
```
bmuenzenmeyer: We can eliminate a magic string here by following the guidance of the existing workflow file at https://github.com/nodejs/nodejs.org/blob/main/.github/workflows/pull-request.yml

```suggestion
          node-version-file: '.nvmrc'
```

.lighthouserc.json
```diff
@@ -0,0 +1,39 @@
...
+        "http://localhost:3000/ja",
+        "http://localhost:3000/ka",
+        "http://localhost:3000/ko",
+        "http://localhost:3000/pt-br",
+        "http://localhost:3000/ro",
+        "http://localhost:3000/ru",
+        "http://localhost:3000/tr",
+        "http://localhost:3000/uk",
+        "http://localhost:3000/zh-cn",
+        "http://localhost:3000/zh-tw"
```
bmuenzenmeyer: I suspect we only need to test `/en` - the rest will exhibit similar experiences but with different content.

.github/workflows/lighthouse-ci.yml
```diff
@@ -0,0 +1,14 @@
+name: CI
```
bmuenzenmeyer: this should have a more descriptive name

package.json
```diff
@@ -71,6 +71,7 @@
     "vfile": "^5.3.7"
   },
   "devDependencies": {
+    "@lhci/cli": "^0.12.0",
```
bmuenzenmeyer: per https://github.com/nodejs/nodejs.org/blob/main/DEPENDENCY_PINNING.md#when-pinning-dependencies this should be tilde-versions

```suggestion
    "@lhci/cli": "~0.12.0",
```

.github/workflows/lighthouse-ci.yml
```diff
@@ -0,0 +1,14 @@
...
+on: [push]
+jobs:
+  lighthouseci:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/setup-node@v3
+        with:
+          node-version: 18
+      - run: npm install && npm install -g @lhci/cli@0.11.x
```
bmuenzenmeyer: - we run `npm ci`
- is the global install necessary if we also have this in `devDependencies`?
- this version of @lhci/cli is different than what is declared in `package.json`. I suggest we use the `package.json` only, or perhaps https://github.com/treosh/lighthouse-ci-action - did you look at that at all?

bmuenzenmeyer (changes requested):
1. I wonder if anyone else thinks this functionality should exist in this file as a separate workflow, or integrate into one of the existing ones at https://github.com/nodejs/nodejs.org/tree/main/.github/workflows. I can see both ways making sense, and not clear precedent. `build-and-analysis` is rather new.
2. do you have validation from your fork that we can look at?

.github/workflows/lighthouse-ci.yml
Chandupuram: Changed to Light House CI this make more sense.

.github/workflows/lighthouse-ci.yml
Chandupuram: Done

.lighthouserc.json
Chandupuram: Changed to more useful URLs

package.json
Chandupuram: Removed this dependency I had added it to test forgot to remove while pushing It is done now

.github/workflows/lighthouse-ci.yml
Chandupuram: npm install was changed to npm ci

.github/workflows/lighthouse-ci.yml
ovflowd, line 1: We shouldn't create yet another GitHub Actions Workflow, this should be added on the `build-and-analysis.yml` Workflow, as a separate job. It can just be called "lighthouse"

Also, it'd be great to give a `name:` field as "Google Lighthouse"`

.lighthouserc.json
```diff
@@ -0,0 +1,26 @@
+{
+  "ci": {
+    "collect": {
+      "startServerCommand": "npm run serve",
+      "startServerReadyPattern": "Server running at http://localhost:3000/",
+      "url": [
+        "http://localhost:3000/en",
```
ovflowd, line 7: I'm not a fan of checking hardcoded URLs.

Let's check all pages besides of blog-posts. Also, do you know how the report gets generated?

We were adding a bot that would comment on the PR?

.lighthouserc.json
```diff
@@ -0,0 +1,26 @@
...
+        "http://localhost:3000/en/get-involved",
+        "http://localhost:3000/en/blog"
+      ],
+      "numberOfRuns": 1,
+      "settings": {
+        "preset": "desktop"
+      }
+    },
+    "upload": {
+      "target": "temporary-public-storage"
```
ovflowd, line 20: What is this upload target?

.github/workflows/lighthouse-ci.yml
```diff
@@ -0,0 +1,14 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/setup-node@v3
+        with:
+          node-version-file: '.nvmrc'
+      - run: npm ci && npm install -g @lhci/cli@0.11.x
+      - run: lhci autorun
+        env:
+          LHCI_GITHUB_APP_TOKEN: ${{ secrets.LIGHTHOUSE_APP_TOKEN }}
```
ovflowd: Where does this token come from? Where are the docs for this? And why do we need an extra token? And how can we get one?

ovflowd:
I also share the views fo @bmuenzenmeyer is there any test repository or fork you've enabled this so we can see the results?

Also, how does this PR achieves `. This helps us to check and monitor the performance of the pages and various statistics after each and every push done to the repo.` from your PR description?

Could you describe how we are able to actually get the insights and analysis? Is the bot/CI supposed to comment on the PR or what?

Chandupuram:
> 1. I wonder if anyone else thinks this functionality should exist in this file as a separate workflow, or integrate into one of the existing ones at https://github.com/nodejs/nodejs.org/tree/main/.github/workflows. I can see both ways making sense, and not clear precedent. `build-and-analysis` is rather new.
> 2. do you have validation from your fork that we can look at?

https://github.com/Chandupuram/nodejs.org/pull/2
Report for /about page: https://storage.googleapis.com/lighthouse-infrastructure.appspot.com/reports/1689488571961-82303.report.html
look at this PR commits and the detail reports generated by LightHouse CI

ovflowd:
> look at this PR commits and the detail reports generated by LightHouse CI

It's kinda annoying that Lighthouse generates one status-check per page. As I mentioned we probably would like to generate reports for each major page, but that seems probably not feasible. What bothers me is how these scores are so low, when I see these when running Lighthouse by myself 🤔 

image

I was genuinely looking forward to something like this: https://github.com/OskarAhl/Lighthouse-github-action-comment, but not sure if this creates a new comment on every run or what.

I also wonder where are the custom settings/thresholds at? Did you configure them somewhere? I don't see that in this PR.

ovflowd:
Also kinda important to mention that the audit should **not** run against `https://nodejs.org` but against the current build of the PR, otherwise Lighthouse has 0 value here.

bmuenzenmeyer:
> I was genuinely looking forward to something like this: https://github.com/OskarAhl/Lighthouse-github-action-comment, but not sure if this creates a new comment on every run or what.

most good actions will be designed to keep updating the same comment on each new push

.github/workflows/lighthouse-ci.yml
Chandupuram: I will move this to build and analysis.yml and also the name

.lighthouserc.json
Chandupuram: I was also looking for a way to not hardcode the URLs. Check this link to get the full view of how the reports gets generated : https://developer.chrome.com/docs/lighthouse/overview/

.lighthouserc.json
Chandupuram: This helps us to store the generated reports we can access them after the Audit through the URLs in the report at the end.

.github/workflows/lighthouse-ci.yml
Chandupuram: You can go through this link : https://github.com/apps/lighthouse-ci for setting up the token

Chandupuram:
> Also kinda important to mention that the audit should **not** run against `https://nodejs.org` but against the current build of the PR, otherwise Lighthouse has 0 value here.

It will run against the local development not for the https://nodejs.org. The thresholds we were using were the Lighthouse CI recommended values. You can check the assert part in .json file.

bmuenzenmeyer:
@Chandupuram my initial notes and suggested tooling were not clear enough and that's led us down an incorrect path for a bit. 

as stated in earlier review, lets make sure to:

- add this new work to `.github/workflows/build-and-analysis.yml`
- find a lighthouse ci tool that adds comments to GitHub PRs. I am curious if https://github.com/treosh/lighthouse-ci-action is better. Reading more: Though https://github.com/treosh/lighthouse-ci-action/issues/44 seems to indicate there is no comment  functionality. Apologies to all, this seems like an haphazardly supported solution
- https://github.com/OskarAhl/Lighthouse-github-action-comment looks to just be a custom workflow script https://github.com/OskarAhl/Lighthouse-github-action-comment/blob/main/.github/workflows/lighthouse-on-vercel-preview-url.yml - we could write our own.
- At this point, @ovflowd I like the idea of leaning on a tool we dont have to maintain. are you okay with https://github.com/treosh/lighthouse-ci-action ?

.lighthouserc.json
ovflowd: Can we supply the "url" field via CLI? Or environment variables?

.lighthouserc.json
ovflowd: Thanks for explaining!

ovflowd:
> It will run against the local development not for the [nodejs.org](https://nodejs.org/). The thresholds we were using were the Lighthouse CI recommended values. You can check the assert part in .json file.

Are we sure about that? What does the "URL" field then even mean? How can it run against the local environment / current running-CI if no web server is running? That doesn't sound correct to me.

ovflowd:
> At this point, @ovflowd I like the idea of leaning on a tool we dont have to maintain. are you okay with [treosh/lighthouse-ci-action](https://github.com/treosh/lighthouse-ci-action) ?

Is this still a recommendation?

bmuenzenmeyer:
Here's what I think we should do.

Let's first clear up requirements:

- we would prefer to avoid status check increases like this, there will simply be too many
image
- have a sticky PR comment instead of status checks
   image
  - use this script as a basis https://github.com/OskarAhl/Lighthouse-github-action-comment/blob/main/.github/workflows/lighthouse-on-vercel-preview-url.yml
  - check each composed action to see if it's the latest version. most look old. pin them using the pattern found in our other workflow files 
     - https://github.com/marocchino/sticky-pull-request-comment
     - https://github.com/aaimio/vercel-preview-url-action (might be more up to date than the fork)
     - https://github.com/actions/checkout
     - https://github.com/treosh/lighthouse-ci-action
     - https://github.com/actions/github-script 
- rely on an existing tool first, before developing or adapting custom tooling
   - > Is this still a recommendation?
   - I've come around. https://github.com/treosh/lighthouse-ci-action simply isn't enough to achieve our goals
- start with ONLY the home page. This mirrors the vercel preview. I'd rather us incrementally adopt other pages

---

@Chandupuram do you feel up for this pivot? If I don't hear from one in a couple days I will start it

Chandupuram:
> 

I think you can start working on this @bmuenzenmeyer

.github/workflows/lighthouse-ci.yml
```diff
@@ -0,0 +1,71 @@
...
+                `| ${score(result.performance)} Performance | ${result.performance} |`,
+                `| ${score(result.accessibility)} Accessibility | ${result.accessibility} |`,
+                `| ${score(result['best-practices'])} Best practices | ${result['best-practices']} |`,
+                `| ${score(result.seo)} SEO | ${result.seo} |`,
+                `| ${score(result.pwa)} PWA | ${result.pwa} |`,
+                ' ',
+                `*Lighthouse ran on [${Object.keys(links)[0]}](${Object.keys(links)[0]})*`
+            ].join('\n')
+
+             core.setOutput("comment", comment);
```
ovflowd, line 62: The indentation of this line is wrong.

.github/workflows/lighthouse-ci.yml
```diff
@@ -0,0 +1,71 @@
+name: Google Lighthouse
+
+on:
```
ovflowd, line 3: I'm not sure this is correct, I believe the `on` here should be of `pull_request`, since this should be updated on every Pull Request.

.github/workflows/lighthouse-ci.yml
```diff
@@ -0,0 +1,71 @@
...
+jobs:
+  generate_lighthouse_audit:
+    timeout-minutes: 30
+    runs-on: ubuntu-latest
+    steps:
+      - name: Add comment to PR
+        id: loading_comment_to_pr
+        uses: marocchino/sticky-pull-request-comment@v2
+        with:
+          GITHUB_TOKEN: ${{ secrets.LIGHTHOUSE_APP_TOKEN }}
```
ovflowd, line 16: This should not be set. By default it will use `GITHUB_TOKEN` which is already an Environment Variable

bmuenzenmeyer:
> > 
> 
> I think you can start working on this @bmuenzenmeyer

citing the above, we will be closing this to take another run at it in the future

PR #5599, opened 2023-08-01
fix: sitemap url for production deployments

styfle:
## Description

The sitemap for production deployments point to the wrong url.

You can see this by visiting https://nodejs.org/sitemap.xml 

image

## Validation

You'll need to set the `NEXT_PUBLIC_BASE_URL` env var and then visit `/sitemap.xml` to see the impact.

You can set the env var in the Vercel dashboard and redeploy, or you can try it locally via `next dev` by creating `.env.local`.

## Related Issues

https://twitter.com/mayankistyping/status/1686288768326914048

### Check List

### DIFF
--- a/next.constants.mjs
+++ b/next.constants.mjs
@@ -32,11 +32,12 @@ export const ENABLE_STATIC_EXPORT =
  *
  * @see https://vercel.com/docs/concepts/projects/environment-variables/system-environment-variables#framework-environment-variables
  */
-export const BASE_URL = process.env.VERCEL_ENV === 'production'
-  ? 'https://nodejs.org'
-  : process.env.NEXT_PUBLIC_VERCEL_URL
-  ? `https://${process.env.NEXT_PUBLIC_VERCEL_URL}`
-  : process.env.NEXT_PUBLIC_BASE_URL || 'https://nodejs.org';
+export const BASE_URL =
+  process.env.VERCEL_ENV === 'production'
+    ? 'https://nodejs.org'
+    : process.env.NEXT_PUBLIC_VERCEL_URL
+    ? `https://${process.env.NEXT_PUBLIC_VERCEL_URL}`
+    : process.env.NEXT_PUBLIC_BASE_URL || 'https://nodejs.org';
 
 /**
  * This is used for any place that requires the Node.js distribution URL (which by default is nodejs.org/dist)
--- a/next.constants.mjs
+++ b/next.constants.mjs
@@ -32,7 +32,9 @@ export const ENABLE_STATIC_EXPORT =
  *
  * @see https://vercel.com/docs/concepts/projects/environment-variables/system-environment-variables#framework-environment-variables
  */
-export const BASE_URL = process.env.NEXT_PUBLIC_VERCEL_URL
+export const BASE_URL = process.env.VERCEL_ENV === 'production'
+  ? 'https://nodejs.org'
+  : process.env.NEXT_PUBLIC_VERCEL_URL
   ? `https://${process.env.NEXT_PUBLIC_VERCEL_URL}`
   : process.env.NEXT_PUBLIC_BASE_URL || 'https://nodejs.org';
 
--- a/next.constants.mjs
+++ b/next.constants.mjs
@@ -32,12 +32,11 @@ export const ENABLE_STATIC_EXPORT =
  *
  * @see https://vercel.com/docs/concepts/projects/environment-variables/system-environment-variables#framework-environment-variables
  */
-export const BASE_URL =
-  process.env.VERCEL_ENV === 'production'
-    ? 'https://nodejs.org'
-    : process.env.NEXT_PUBLIC_VERCEL_URL
-    ? `https://${process.env.NEXT_PUBLIC_VERCEL_URL}`
-    : process.env.NEXT_PUBLIC_BASE_URL || 'https://nodejs.org';
+export const BASE_URL = process.env.NEXT_PUBLIC_BASE_URL
+  ? process.env.NEXT_PUBLIC_BASE_URL
+  : process.env.NEXT_PUBLIC_VERCEL_URL
+  ? `https://${process.env.NEXT_PUBLIC_VERCEL_URL}`
+  : 'https://nodejs.org';
 
 /**
  * This is used for any place that requires the Node.js distribution URL (which by default is nodejs.org/dist)


next.constants.mjs
```diff
@@ -32,9 +32,12 @@ export const ENABLE_STATIC_EXPORT =
  *
  * @see https://vercel.com/docs/concepts/projects/environment-variables/system-environment-variables#framework-environment-variables
  */
-export const BASE_URL = process.env.NEXT_PUBLIC_VERCEL_URL
-  ? `https://${process.env.NEXT_PUBLIC_VERCEL_URL}`
-  : process.env.NEXT_PUBLIC_BASE_URL || 'https://nodejs.org';
+export const BASE_URL =
+  process.env.VERCEL_ENV === 'production'
```
ovflowd: I'm not sure if hardcoding the nodejs.org here is a good idea.

In the original issue we were thinking in simply giving higher priority to NEXT_PUBLIC_BASE_URL (which we can easily configure on our Vercel Dashboard)

TO L;DR swap the position of the NEXT_PUBLIC_VERCEL_URL with the BASE_URL one

ovflowd: (Also we already have a constant for VERCEL_ENV)

ovflowd (commented):
I appreciate the effort, but I'm not keen on the approach 👀

next.constants.mjs
styfle: My PR doesn't require any configuration, but if you prefer to configure a new production env var with `NEXT_PUBLIC_BASE_URL=https://nodejs.org`, that would work too.

If you want to do that, free free to close this PR since I don't have permission to add env vars to your project.

styfle: I suppose another option is to add a `.env.production` file with `NEXT_PUBLIC_BASE_URL=https://nodejs.org`.

It would be assigned every time your run `next build` which would be for all Vercel deployments (both preview and production).

Do you want me to add that?

ovflowd: Would that file also be served and available to the serverless functions?

I'm not much keen on .env files on the main repo, and simply let whoever that consumes this to configure it the way they want.

Hence why for me is more convenient setting up a env variable on Vercel itself.

But Im all ears 🙃

ovflowd: I understand your PR initial changeset would fix the current symptom, and require no actionables, but it does in depriment of allowing any customisation of the base url on production vercel environments.

styfle: > Would that file also be served and available to the serverless functions?

Yes, it works the same as normal env vars. 

Learn more: https://nextjs.org/docs/pages/building-your-application/configuring/environment-variables

styfle: I assumed this was a high priority PR since a broken sitemap means new pages might not be indexed by google and other search engines.

(the current sitemap urls have `X-Robots-Tag: noindex` header since those are not using a production domain)

ovflowd: Well, kinda yes. But that doesn't mean we can't provide a future-proof solution given the nature and complexity of this; + setting an env variable on vercel is as fast as, something really fast (for the lack of words) 😅

Don't get me wrong, not criticising your approach, I'm simply making a review 🙃

styfle: Ok sounds like you want the env var in the dashboard solution so I pushed [da7cb21](https://github.com/nodejs/nodejs.org/pull/5599/commits/da7cb21488f6dd6ae34fb0c5b20691f9130089a8)

ovflowd: I mean, I'm fine with both solutions, I've just mentioned the variable one is convenient for me 🙃

ovflowd (approved):
LGTM!

bmuenzenmeyer (approved):
works for me, and FANTASTIC validation instructions. This was the exact thing I was hoping to unlock by introducing that.

bmuenzenmeyer:
I agree that this appears to be a production bug - but not knowing the status of the env var or having access to vercel, I'd rather @ovflowd merge it in the morning

ovflowd:
Yup that's right! Deploying now :)

PS: I didn't merge back then because I was already going to sleep!

styfle:
@ovflowd Did you set the env var yet? I’m still seeing the bug on the latest deployment.

ovflowd:
Of course I dit. And it was working when I made the deployment. Now when I open this (https://nodejs.org/sitemap.xml), it seems gone.

Is Vercel applying my Env Variables to the Serverless Functions? I have no idea why it would work then stop working.

ovflowd:
image

ovflowd:
Triggering a redeployment. What is even weirder is that the URL on the current sitemap, is coming from a Preview Deployment? (nodejs-lp1va9y4q) 

image

styfle:
> What is even weirder is that the URL on the current sitemap, is coming from a Preview Deployment?

That sounds like its the issue. If you promoted a Preview deployment, then it wouldn't get Production env vars.

ovflowd:
> That sounds like its the issue. If you promoted a Preview deployment, then it wouldn't get Production env vars.

To clarify, we never promoted any "preview" build to production. We use default settings. Only "main" branch should go to "production". Is this by any chance a bug on Vercel-side?

ovflowd:
It's fixed now, but pretty much because I did a "rebuild" of a production build. And this seems to be a thing that is recurring (from time to time these .vercel.app URLs from preview deployments seem to somehow get there)

Could aos by any chance Turborepo Cache be influencing here?

styfle:
Ah yes it could be turborepo cache.

The deployment from PR #5595 would be considered Preview and then once merged, since there were no code changes, you might see "FULL TURBO" in the Production build logs meaning it would just reuse the built output from the PR.

You might want to add all the env vars you depend on to the `turbo.json` like this: https://turbo.build/repo/docs/core-concepts/caching/environment-variable-inputs

I'll ask the turborepo team if they have any better suggestions.

ovflowd:
This is definitely relevant knowledge, right there! (Thank you @styfle !!)

ovflowd:
Hot-fix: https://github.com/nodejs/nodejs.org/commit/97821bc35485e2c35eedc2cb27160ddba70c9763

styfle:
@ovflowd I spoke to the turborepo team and they said this is the expected behavior because default is "loose" env var handling. Adding the env vars like you did is a good solution 👍  

If you wanted to catch this type of bug in the future, you could enable "strict" env var handling, which would prevent any env var from being used without explicitly marking it as part of the cache key hash or marking it as a pass-through.

https://turbo.build/repo/docs/core-concepts/caching/environment-variable-inputs#loose--strict-environment-modes

ovflowd:
Thanks @styfle! Would you be kind enough and make a PR for strict env? 🙇

styfle:
I have't tried it yet, so I'll have to update my own repos before this one 😆

ovflowd:
Np! If you prefer I can handle the PR.

PR #5780, opened 2023-09-10
feat(storybook): viewport

AugustinMauroy:
## Description

Adding of viewport plugins to storybook. It's allow us to see how component will render in different device.

## Related Issues

No related issues

### Check List


ovflowd (changes requested):
Im not sure if a plugin is needed for something that can easily be done with a browser viewport functionality.

Im usually against adding extra dependencies, and this one doesn't really briny much value, since browsers can do that by themselves.

There's a reason why I removed this plugin to begin with ;)

I'm neutral to be honest, but any extra dependency we always need to ponder if it's worth it. As it's more install time, more transient and peer dependencies, bigger bundle and more chances of having security issues.

AugustinMauroy:
I agree with you that it's addictive.
The installation time isn't too bad because the project takes 24s to download (with NPM or PNPM).
Why use this package, because it allows you to change the size of the component without touching the gui. In the example below you can see what happens when you use the integrated tool.

Capture d’écran 2023-09-10 à 13 33 10

here's when you use the plugins (it doesn't modify the storybook layout)

Capture d’écran 2023-09-10 à 13 33 43

bmuenzenmeyer:
I can see this being useful for upcoming nav bar work. If you feel really strongly @ovflowd that's fine, but honestly I'm surprised this functionality isn't standard storybook behavior. (We built this into Patternlab.io 7+ years ago)

ovflowd:
I steel feel that the Browser DevTools is sufficient. But I dont have strong feelings, feel free to approve and dismiss my review.

ovflowd (approved):

PR #5783, opened 2023-09-11
feat: add `Button` component

### LINKED ISSUE #5758: Create `Button` component
Create a `Button` component using the new [figma design (direct link)](https://www.figma.com/file/pu1vZPqNIM7BePd6W8APA5/Node.js?type=design&node-id=423-28019&mode=design&t=q5ACW63em8GjhzrV-4). 

---

### Before You Start...

---

### Details

image

- Create a new directory called `Button` within the [`Common` directory](https://github.com/nodejs/nodejs.org/tree/main/components/Common) that contains all the new code.
- Use the new tailwind color and typography system to assign properties. For example, `Primary, Light, Default` uses `Green/600` per the figma.
- Add a storybook story file `components/Common/Button/index.stories.tsx` which exercises each of the button's states.

There are 24 states to capture within styles and stories:

#### Light

- Primary, Light Default
- Primary, Light, Hover
- Primary, Light, Focus
- Primary, Light, Disabled
- Secondary, Light, Default
- Secondary, Light, Hover
- Secondary, Light, Focus
- Secondary, Light, Disabled

#### Dark

- Primary, Dark, Default
- Primary, Dark, Hover
- Primary, Dark, Focus
- Primary, Dark, Disabled
- Secondary, Dark, Default
- Secondary, Dark, Hover
- Secondary, Dark, Focus
- Secondary, Dark, Disabled

#### Special

- Primary, Special, Default
- Primary, Special, Hover
- Primary, Special, Focus
- Primary, Special, Disabled
- Secondary, Special, Default
- Secondary, Special, Hover
- Secondary, Special, Focus
- Secondary, Special, Disabled

umairraza96:
## Description

Add new dependency for storybook for pseudo states

Write stories for 24 button states

## Related Issues

Addresses #5758

### Check List


components/Common/Button/index.stories.tsx
```diff
@@ -0,0 +1,294 @@
+import type { Meta, StoryObj } from '@storybook/react';
```
AugustinMauroy: Wouldn't it be easier to use storybook controls instead of lot of stories ?

umairraza96: I do agree with that, less code, easier to maintain. but the only down side is I couldn't figure out a way to display hover and focus states with that.

AugustinMauroy: for how and focus do an dedicated stories

components/Common/Button/index.stories.tsx
ovflowd, line 1: You don't need to make dedicated stories for hover and visited states. I don't think you should add another Storybook library for these things.

You don't need to create dedicated stories for disabled state either. IMO you only need a single story.

Dark Mode and Light mode should be controlled via [this](https://github.com/nodejs/nodejs.org/pull/5781/files#r1321595086) which @HinataKah0 is going to introduce.

hover should be triggered via natural hover action; disabled state via controls; primary or secondary button also via controls.

I don't see the value of having multiple stories of the same button that mutates over control changes.

ovflowd: Note that maybe it's worth having stories only for the variants of the button styles so we can do proper regression testing.

I know that not having stories for states of the element (as it is an interactable element) might be error-prone, but I'm not sure what would be a good approach here, as I do not fancy having individual tests for states. BUT I am neutral on that, if you deem needed feel free to have tests for all variants (primary, secondary, primary hover, secondary hover, primary disabled, secondary disabled)

But for all other sort of variations (clicked, dark, light, etc; you don't need stories)

.storybook/main.ts
```diff
@@ -3,7 +3,11 @@ import type { StorybookConfig } from '@storybook/nextjs';
 
 const config: StorybookConfig = {
   stories: ['../components/**/*.stories.tsx'],
-  addons: ['@storybook/addon-controls', '@storybook/addon-interactions'],
+  addons: [
+    '@storybook/addon-controls',
+    '@storybook/addon-interactions',
+    'storybook-addon-pseudo-states',
```
ovflowd: How well maintained is this storybook-addon-pseudo-stories?

components/Common/Button/index.module.scss
```diff
@@ -0,0 +1,59 @@
...
+      &:disabled {
+        @apply opacity-50;
+      }
+    }
+  }
+
+  &.secondary {
+    @apply text-neutral-800 rounded-lg disabled:opacity-50 dark:text-neutral-200/100;
+
+    &:hover {
```
ovflowd: There shouldn't be class names for states.

You should use `:focus` and `:hover` selectors.

components/Common/Button/index.module.scss
```diff
@@ -0,0 +1,59 @@
...
+    &:focus {
+      @apply bg-green-700 border-green-700;
+    }
+
+    &:disabled {
+      @apply bg-green-600 opacity-50;
+    }
+
+    &.special {
+      @apply bg-neutral-100/0 bg-gradient-to-r from-green-600/30 via-green-500/20 to-green-600/10 shadow-sm  border border-green-600/30;
```
ovflowd:
```suggestion
      @apply bg-neutral-100/0 bg-gradient-to-r from-green-600/30 via-green-500/20 to-green-600/10 shadow-sm border border-green-600/30;
```

components/Common/Button/index.stories.tsx
```diff
@@ -0,0 +1,294 @@
...
+    variant: ['primary', 'secondary'],
+  },
+} satisfies Meta<typeof Button>;
+type Story = StoryObj<typeof meta>;
+
+/**
+ * Primary
+ */
+
+export const Primary: Story = {
```
ovflowd, line 7: These stories (all of them) are violating some of our guidelines. Using the `render` method is not recommended. Please prefer to use `component` and `args`

https://storybook.js.org/docs/react/writing-stories/args

components/Common/Button/index.stories.tsx
```diff
@@ -0,0 +1,294 @@
+import type { Meta, StoryObj } from '@storybook/react';
+
+import Button from './index';
+
+const meta = {
+  component: Button,
+  argTypes: {
+    variant: ['primary', 'secondary'],
+  },
+} satisfies Meta<typeof Button>;
```
ovflowd: Let's not use satisfiest here, and follow the conventions from the Collab guide

.storybook/main.ts
umairraza96: This library is actively maintained, last update was 23 days ago

It has 147k weekly downloads on npm

Socket Report
Screenshot_20230913_201445

AugustinMauroy: it's maintained by chromatic
https://github.com/chromaui/storybook-addon-pseudo-states

components/Common/Button/index.module.scss
umairraza96: You mean like this `hover:text-neutral-800` ?

ovflowd: No. `:hover` and `:focus` pseudo selectors. 

(https://developer.mozilla.org/en-US/docs/Web/CSS/Pseudo-classes, https://developer.mozilla.org/en-US/docs/Web/CSS/:hover)

But I just realised you are already using them. I read `&.hover` instead of `&:hover`, my bad!

types/button.ts
```diff
@@ -0,0 +1,6 @@
+import type { ButtonHTMLAttributes } from 'react';
+
+export interface ButtonProps extends ButtonHTMLAttributes<HTMLButtonElement> {
```
ovflowd: As mentioned before, these types should be in the same file as the Button component.

components/Common/Button/index.tsx
```diff
@@ -0,0 +1,32 @@
...
+};
+
+const Button: FC<ButtonProps> = ({
+  variant = 'primary',
+  special = false,
+  children,
+  className,
+  ...props
+}) => {
+  const buttonStyles = classNames(
```
araujogui: I feel we should have tailwind-merge here, to remove the conflicts between the className prop and the button's default styles.

umairraza96: We have already discussed it here https://github.com/nodejs/nodejs.org/issues/5758

ovflowd: I assume tailwind does the optimisations by itself? Since tailwind actually "compiles" its classnames during build/runtime; hence I'd imagine it already merging things?

But this is just a hypothesis

ovflowd (approved):
LGTM

ovflowd (approved):
LGTM! Thank you so much for the contribution!

ovflowd:
@bmuenzenmeyer can you check the storybooks for the Button looks correct? Just wanted to have a 2nd opinion to see if the designs are matching

components/Common/Button/index.tsx
araujogui: No, it doesn't. We can try to use the [Important modifier](https://tailwindcss.com/docs/configuration#important-modifier) or [tailwind-merge](https://github.com/dcastil/tailwind-merge). https://github.com/dcastil/tailwind-merge/blob/v1.14.0/docs/when-and-how-to-use-it.md#using-tailwinds-important-modifier

bmuenzenmeyer: @araujogui can you show us what that would look like in this component?

I am also reading through https://github.com/dcastil/tailwind-merge/blob/v1.14.0/docs/what-is-it-for.md etc in the meantime

components/Common/Button/index.module.scss
```diff
@@ -0,0 +1,59 @@
...
+    &:focus {
+      @apply bg-green-700 border-green-700;
+    }
+
+    &:disabled {
+      @apply bg-green-600 opacity-50;
+    }
+
+    &.special {
+      @apply bg-neutral-100/0 bg-gradient-to-r from-green-600/30 via-green-500/20 to-green-600/10 shadow-sm border border-green-600/30;
```
ovflowd: Can you separate the tailwind selectors to one per line? As shown on the Collaborator Guide? Eg.:


```scss
@apply bg-neutral-100/0
 bg-gradient-to-r
 from-green-600/30
 ...
 border-green-600/30;
```

ovflowd: Do that on all the `@apply` calls

umairraza96: Sure, I'll do that

umairraza96:
> LGTM! Thank you so much for the contribution!

It is my first time ever writing stories with storybook, so learned quite a lot!

components/Common/Button/index.module.scss
```diff
@@ -0,0 +1,85 @@
+.button {
+  @apply px-[18px]
+  py-[10px]
```
ovflowd: If you see on our example, we add a space after the first line containing `@apply` to kinda hint that these tokens are part of the apply statement (on my review comment I also kinda mentioned that)

Could you make this final change? 🙇

components/Common/Button/index.tsx
ovflowd: @bmuenzenmeyer checked the docs and it seems that using tailwind-merge isn't recommended for our use case. Brian has more details that he can go over.

components/Common/Button/index.module.scss
umairraza96: Well I missed it unfortunately.

components/Common/Button/index.tsx
araujogui: Oh, I've just seen [it](https://github.com/dcastil/tailwind-merge/blob/v1.14.0/docs/recipes.md#extracting-classes-with-tailwinds-apply) too so I believe we can use the important modifier when we need to.

araujogui (approved):

components/Common/Button/index.tsx
umairraza96: Interesting didn't know that `tailwind-merge` doesn't play nice with `@apply` modifier

ovflowd (changes requested):
Hey @umairraza96, the styles of the Button don't really match the Figma,

Also, "special" should be one of the variations, not a "boolean". 

So pretty much we should have **only:**

- primary
- secondary
- special

We do not have "tertiary"

I see that you made a `.dark` class for the "secondary" button, but that should not exist.

Note that the "special" button should have the same style on dark/light as per Figma, but the style of the "special" button here is different from Figma.

So pretty much if you can update these styles, that'd be great! Hover and Focus looks alright, so the only issues (again TL;DR) are:

- Special should be a variation, and should have the same styles regardless of light/dark
- There's a `.dark` selector that is not being used anywhere
- There's no such thing as a "tertiary" variation

umairraza96:
> Hey @umairraza96, the styles of the Button don't really match the Figma,
> 
> Also, "special" should be one of the variations, not a "boolean".
> 
> So pretty much we should have **only:**
> 
> * primary
> * secondary
> * special
> 
> We do not have "tertiary"
> 
> I see that you made a `.dark` class for the "secondary" button, but that should not exist.
> 
> Note that the "special" button should have the same style on dark/light as per Figma, but the style of the "special" button here is different from Figma.
> 
> So pretty much if you can update these styles, that'd be great! Hover and Focus looks alright, so the only issues (again TL;DR) are:
> 
> * Special should be a variation, and should have the same styles regardless of light/dark
> * There's a `.dark` selector that is not being used anywhere
> * There's no such thing as a "tertiary" variation

Sorry for that misinterpretation, however I would like to ask for styles of special button as it is not clear on figma, I tried to make it as similar as possible, may be I am missing few points.

ovflowd:
> Sorry for that misinterpretation, however I would like to ask for styles of special button as it is not clear on figma, I tried to make it as similar as possible, may be I am missing few points.

image

This is the special style.

- https://www.figma.com/file/pu1vZPqNIM7BePd6W8APA5/Node.js?type=design&node-id=423-28013&mode=design&t=fkLA2i1umUyMu961-4 special:default
- https://www.figma.com/file/pu1vZPqNIM7BePd6W8APA5/Node.js?type=design&node-id=695-15154&mode=design&t=fkLA2i1umUyMu961-4 special:hover
- https://www.figma.com/file/pu1vZPqNIM7BePd6W8APA5/Node.js?type=design&node-id=695-15192&mode=design&t=fkLA2i1umUyMu961-4 special:focus
- https://www.figma.com/file/pu1vZPqNIM7BePd6W8APA5/Node.js?type=design&node-id=695-15230&mode=design&t=fkLA2i1umUyMu961-4 special:disabled

(special should look like this both on dark/light) even tho in general the "special" button should only be used in dark themes.

umairraza96:
> > Sorry for that misinterpretation, however I would like to ask for styles of special button as it is not clear on figma, I tried to make it as similar as possible, may be I am missing few points.
> 
> image
> This is the special style.
> 
> * https://www.figma.com/file/pu1vZPqNIM7BePd6W8APA5/Node.js?type=design&node-id=423-28013&mode=design&t=fkLA2i1umUyMu961-4 special:default
> * https://www.figma.com/file/pu1vZPqNIM7BePd6W8APA5/Node.js?type=design&node-id=695-15154&mode=design&t=fkLA2i1umUyMu961-4 special:hover
> * https://www.figma.com/file/pu1vZPqNIM7BePd6W8APA5/Node.js?type=design&node-id=695-15192&mode=design&t=fkLA2i1umUyMu961-4 special:focus
> * https://www.figma.com/file/pu1vZPqNIM7BePd6W8APA5/Node.js?type=design&node-id=695-15230&mode=design&t=fkLA2i1umUyMu961-4 special:disabled
> 
> (special should look like this both on dark/light) even tho in general the "special" button should only be used in dark themes.

Took me a while to figure this out because button has a transparent background so if you switch its background it will change the way it looks

ovflowd:
I see, I guess since this button is only meant to be used on dark, you could apply the dark background used there as part of the gradient

components/Common/Button/index.module.scss
```diff
@@ -0,0 +1,82 @@
...
+     border
+     border-green-600/30;
+
+    &::before {
+      @apply absolute
+       h-full
+       w-full
+       top-0
+       right-0;
+      background: radial-gradient(
```
ovflowd: I wonder if using before is the right way of doing this 🤔

Also the style changes a little bit on hover and on disabled

components/Common/Button/index.stories.tsx
```diff
@@ -0,0 +1,15 @@
...
+
+type Story = StoryObj<typeof Button>;
+type Meta = MetaObj<typeof Button>;
+
+export const Default: Story = {
+  args: {
+    variant: 'primary',
+    children: 'Button',
+    disabled: false,
+    special: false,
```
ovflowd: Shouldn't this be removed?

umairraza96: oh yeah I forgot about that, I was so occupied with the styles

components/Common/Button/index.module.scss
umairraza96: Is there a better way ? let me check.

ovflowd: So I've looked into it, and the current layout looks great.

Can I only ask you to export this gradient into a variable?

Within `styles/new/mixins.scss` and then define a mixin for this gradient? https://sass-lang.com/documentation/at-rules/mixin/

(And the import the mixins.scss within `styles/new/index.scss`

components/Common/Button/index.stories.tsx
```diff
@@ -0,0 +1,28 @@
...
+
+export const Default: Story = {
+  args: {
+    variant: 'primary',
+    children: 'Download Node (LTS)',
+    disabled: false,
+  },
+  decorators: [
+    (Story, { args }) => (
+      <>
```
ovflowd: Not sure the `<>` is needed here

components/Common/Button/index.stories.tsx
```diff
@@ -0,0 +1,28 @@
...
+        <div
+          className={classNames({
+            'bg-[#000000] w-fit rounded-md': args.variant === 'special',
```
ovflowd, line 19: This doesn't seem to have worked on Figma afaik

ovflowd: Can we make the whole container black? Just to see border effect looking correct?

ovflowd (commented):
Really good stuff! Now we're almost there!

ovflowd:
Also, I've noticed one small aspect of the button is missing, which is this effect from the image below

image

You can accomplish that with something like the snippet below as a pseudo element (`::after`)

```css
background: var(--gradient-glow, linear-gradient(90deg, rgba(65, 126, 56, 0.00) 0%, #417E38 50%, rgba(65, 126, 56, 0.00) 100%));
content: '';
width: 100px;
height: 1px;
position: absolute;
top: -1px;
right: -5px;
```

Available here: https://www.figma.com/file/pu1vZPqNIM7BePd6W8APA5/Node.js?node-id=422%3A6591&mode=dev

components/Common/Button/index.stories.tsx
ovflowd: Nvm, I'm adding this here: https://github.com/nodejs/nodejs.org/pull/5875

ovflowd: Can we remove then this className part for the special button?

components/Common/Button/index.module.scss
```diff
@@ -0,0 +1,106 @@
...
+        --gradient-glow,
+        linear-gradient(
+          90deg,
+          rgba(65, 126, 56, 0) 0%,
+          #417e38 50%,
+          rgba(65, 126, 56, 0) 100%
+        )
+      );
+      content: '';
+      height: 1px;
```
ovflowd: Note, that you should still use tailwind variables for this; such as `height-[1px]`

We shouldn't do any manual CSS here if possible

components/Common/Button/index.module.scss
```diff
@@ -0,0 +1,106 @@
...
+      content: '';
+    }
+
+    &::after {
+      background: var(
+        --gradient-glow,
+        linear-gradient(
+          90deg,
+          rgba(65, 126, 56, 0) 0%,
+          #417e38 50%,
```
ovflowd: I believe instead of using `#417e38` we should use one of our colour variables

components/Common/Button/index.module.scss
```diff
@@ -0,0 +1,106 @@
...
+     border
+     border-green-600/30;
+
+    &::before {
+      @apply absolute
+       h-full
+       w-full
+       top-0
+       right-0;
+      @include glow-backdrop(8em, #5fa04e);
```
ovflowd: Same here, if we could use one of our color variables? Not sure if it works tho

ovflowd (commented):
I left some minor comments, but we're really close!

Sorry for all the back and forth reviews! I super appreciate all the effort you're putting here ❤️

umairraza96:
> I left some minor comments, but we're really close!
> 
> Sorry for all the back and forth reviews! I super appreciate all the effort you're putting here ❤️

Don't be sorry, I should be thanking you if anything, I do understand that my commits are not perfect and I am still learning quite a few new things and I do appreciate your detailed and informative feedbacks. It helps me skill up. 
So Thank you!! <3

components/Common/Button/index.module.scss
umairraza96: you mean like `bg-green-600` ? or do you guys have color variables defined elsewhere

ovflowd: Yeah, and the color variables are all on tailwind.config.ts

ovflowd: Maybe you can use this https://tailwindcomponents.com/gradient-generator/

components/Common/Button/index.module.scss
umairraza96: It doesn't work here, we could create a new variables files in which we can re-declare tailwind color tokens in sass variables. what do you think about this approach ?

ovflowd: That's not needed, you can invoke tailwind theme colours directly from CSS https://stackoverflow.com/a/67281604

umairraza96: @ovflowd This doesn't seem to work with sass https://tailwindcss.com/docs/using-with-preprocessors#using-sass-less-or-stylus

ovflowd: You need to use @theme instead of just theme when using with scss

ovflowd: It should work, if you don't combine with sass functions. Because sass gets compiled first, so the alternative here is to not use mixins?

Let's try to see in another perspective, why can't we use tailwinds gradient generator for the gradients that we need?

ovflowd: Maybe another approach is to not use Sass at all and only use PostCSS as otherwise we are using two preprocessors at once.. 

I will think about this

umairraza96: Because tailwind doesn't support radial gradient, and I am using radial gradient

ovflowd: It does not, but you can add radial-gradients on Tailwind config by yourself https://github.com/tailwindlabs/tailwindcss/discussions/2599#discussioncomment-102667

ovflowd: This exact one should work: https://play.tailwindcss.com/kQvzcR8OF8 (check the config tab)

ovflowd:
@umairraza96 once https://github.com/nodejs/nodejs.org/pull/5888 gets merged, you will be able to use the Tailwind `theme` statement

components/Common/Button/index.module.scss
umairraza96: We can add radial gradients this way but I can't figure out how to replicate the exact style, it seems it is different from what I used, I think it lacks control tbh.

ovflowd: That's fine, let's keep it the way it is.

ovflowd:
Hey @umairraza96 the PostCSS changes got merged. I guess you can rebase and do the following:

- update your mixin to use the `@define-mixin` API from `postcss-mixins` (and update the `@include` to be `@mixin`) (also they don't use `()` on both definition and usage)
- replace the plain Hex Colors on your styles with the Talwind `theme()` call

targos:
What kind of review do you need from releaser?

ovflowd:
I think that @umairraza96 messed up their rebase.

And now it is containing files that are owned by releasers (aka release blog posts)

This PR became a big mess...

umairraza96:
> I think that @umairraza96 messed up their rebase.
> 
> And now it is containing files that are owned by releasers (aka release blog posts)
> 
> This PR became a big mess...

I will try to fix it. Using 
`git reset --hard main` and then try to do things again hope it helps

ovflowd:
> > I think that @umairraza96 messed up their rebase.
> > And now it is containing files that are owned by releasers (aka release blog posts)
> > This PR became a big mess...
> 
> I will try to fix it. Using `git reset --hard main` and then try to do things again hope it helps

You need to keep in mind that your fork `main` branch is not the same as the upstream `main` branch. The source of this mess is probably that you tried to rebase based on your fork's main.

I recommend syncing your fork through GitHub UI or manual pull on your main from upstream's main.

umairraza96:
> > > I think that @umairraza96 messed up their rebase.
> > > And now it is containing files that are owned by releasers (aka release blog posts)
> > > This PR became a big mess...
> > 
> > 
> > I will try to fix it. Using `git reset --hard main` and then try to do things again hope it helps
> 
> You need to keep in mind that your fork `main` branch is not the same as the upstream `main` branch. The source of this mess is probably that you tried to rebase based on your fork's main.
> 
> I recommend syncing your fork through GitHub UI or manual pull on your main from upstream's main.

Exactly that was the issue, figured it out.

ovflowd (approved):
I've pushed a quick fix (accidentally you created a index.scss) and added the variant stories.

LGTM! Thank you soooooo much for your contribution, @umairraza96! Superb work, and thank you for the continuous effort here, even after all the extensive reviews 🙇

PR #5814, opened 2023-09-20
refactor(stylelint): Enfore camelCase for CSS Selectors.

Harkunwar:
## Description

This PR aims to update the stylelint to align with the new guidelines. This forces the usage of camel case to all selectors, and would fail with the following error if non camel case is used. These rules would also now be only enforced in the `components` directory, so we do not mess with any of the old CSS>
image
This PR also updates the existing CSS module files with some spacing alignments.

## Validation

You can try breaking the selector rules. 

### Check List


bmuenzenmeyer (approved):

AugustinMauroy (approved):

ovflowd (approved):
LGTM! But can we update the `package.json` command to still stylelint everything and simple add an ignore to `styles/old` for now?

RelxOff:
👍

PR #5916, opened 2023-09-29
feat(avatarGroup): introduce

### LINKED ISSUE #5910: Create `AvatarGroup` Component
Create an `AvatarGroup` component using the new [figma design (direct link)](https://www.figma.com/file/pu1vZPqNIM7BePd6W8APA5/Node.js?type=design&node-id=477-13668&mode=dev). 

This component is a group of multiple `Avatars`, with optional display of overflow.

---

### Before You Start...

---

### Details

image

- Create a new directory called `AvatarGroup` within the [`Common` directory](https://github.com/nodejs/nodejs.org/tree/main/components/Common) that contains all the new code.
- Use the new tailwind color and typography system to assign properties. For example, the background of the overflow `Avatar` uses `Neutral/100` in light mode per the figma.
- [Add a new dependency,`@radix-ui/avatar`](https://www.radix-ui.com/primitives/docs/components/avatar), for the basic `Avatar` functionality. Follow the [the dependency guide](https://github.com/nodejs/nodejs.org/blob/main/DEPENDENCY_PINNING.md)
- After `limit` `Avatars` are provided, an overflow `Avatar` displaying `+N` should render, where `N` is the length of the avatar collection minus `limit`. Make sure the math is correct before, at, and past the limit.
- Add a storybook story file `components/Common/AvatarGroup/index.stories.tsx` which exercises each of the component's states.

Suggested props include:

- `limit` (integer, default 10): dictates total `Avatars` to render
- `avatars` (object[]): that follows this general shape:

```json
[
  {
    "imageUrl": "https://avatars.githubusercontent.com/u/298435?v=4",
    "href": "https://github.com/bmuenzenmeyer"
  },
  ...
]
```

There are 6 states to capture within styles and stories:

- Light, under limit
- Light, over limit
- Light, at limit
- Dark, under limit
- Dark, over limit
- Dark, at limit

AugustinMauroy:
## Description

Introduce `avatarGroup`

## Validation

* It's should be same as [figma](https://www.figma.com/file/pu1vZPqNIM7BePd6W8APA5/Node.js?node-id=477%3A13668&mode=dev)
* If 404 put two charter of "alt props"
 
## Related Issues

Closes #5910

### Check List


ovflowd:
image

This implementation doesn't look right. It would only work if the images are perfect squares.

components/Common/AvatarGroup/avatar/index.module.css
```diff
@@ -0,0 +1,3 @@
+.avatar {
+  @apply -ml-2 rounded-full border-2 border-white bg-neutral-100 first:ml-0 dark:border-neutral-950 dark:bg-neutral-800;
```
ovflowd: Please read the collaborator guidelines of how to style components carefully.

components/Common/AvatarGroup/avatar/index.stories.tsx
```diff
@@ -0,0 +1,19 @@
...
+
+type Story = StoryObj<typeof Avatar>;
+type Meta = MetaObj<typeof Avatar>;
+export const Default: Story = {
+  args: {
+    src: userNameToAvatarUrl('AugustinMauroy'),
+  },
+};
+
+export const Failed: Story = {
```
ovflowd: I'd call this a Fallback

components/Common/AvatarGroup/avatar/index.stories.tsx
```diff
@@ -0,0 +1,19 @@
+import { userNameToAvatarUrl } from '@/util/github';
+import Avatar from './';
+import type { Meta as MetaObj, StoryObj } from '@storybook/react';
```
ovflowd: Type imports come before all other imports

components/Common/AvatarGroup/avatar/index.stories.tsx
```diff
@@ -0,0 +1,19 @@
...
+
+type Story = StoryObj<typeof Avatar>;
+type Meta = MetaObj<typeof Avatar>;
+export const Default: Story = {
```
ovflowd: Missing empty line to separate types

components/Common/AvatarGroup/avatar/index.module.css
ovflowd, line 1: Name of this folder should be `Avatar`, not `avatar`. Now that you committed it already, due to case insensitiveness, you will need to remove these files and add them again

components/Common/AvatarGroup/avatar/index.stories.tsx
AugustinMauroy: So I should call it fallback ?

util/__tests__/github.test.ts
ovflowd, line 1: Unit Tests should be done in `.mjs`, read collaborator guide 😅

components/Common/AvatarGroup/index.tsx
```diff
@@ -0,0 +1,41 @@
...
+
+const AvatarGroup: FC<AvatarGroupProps> = ({
+  limit = 10,
+  avatars,
+  forceShow,
+}) => {
+  const [showMore, setShowMore] = useState(forceShow?.valueOf() || false);
+
+  const handleShowMoreClick = () => {
+    if (forceShow) return;
```
ovflowd: What is forceShow? What is is used for?

components/Common/AvatarGroup/index.tsx
```diff
@@ -0,0 +1,41 @@
...
+  limit?: number;
+  forceShow?: boolean;
+};
+
+const AvatarGroup: FC<AvatarGroupProps> = ({
+  limit = 10,
+  avatars,
+  forceShow,
+}) => {
+  const [showMore, setShowMore] = useState(forceShow?.valueOf() || false);
```
ovflowd:
```suggestion
  const [showMore, setShowMore] = useState(forceShow || false);
```

components/Common/AvatarGroup/index.tsx
```diff
@@ -0,0 +1,41 @@
...
+
+  const handleShowMoreClick = () => {
+    if (forceShow) return;
+    setShowMore(true);
+  };
+
+  return (
+    <div className={styles.avatarGroup}>
+      {showMore
+        ? avatars.map((avatar, index) => <Avatar key={index} src={avatar} />)
```
ovflowd: This should be memoized and not done directly within the render of the method

components/Common/AvatarGroup/index.tsx
```diff
@@ -0,0 +1,41 @@
...
+        : avatars
+            .slice(0, limit)
+            .map((avatar, index) => <Avatar key={index} src={avatar} />)}
+      {avatars.length > limit && !showMore && (
+        <span className={styles.showMore} onClick={() => handleShowMoreClick()}>
```
ovflowd:
```suggestion
        <span className={styles.showMore} onClick={handleShowMoreClick}>
```

components/Common/AvatarGroup/index.stories.tsx
```diff
@@ -0,0 +1,61 @@
...
+    userNameToAvatarUrl('aymen94'),
+    userNameToAvatarUrl('shanpriyan'),
+    userNameToAvatarUrl('Wai-Dung'),
+    userNameToAvatarUrl('manishprivet'),
+    userNameToAvatarUrl('araujogui'),
+  ],
+};
+
+export const Default: Story = {
+  args: {
```
ovflowd:
```ts
args: { ...avatarListProps }
```

components/Common/AvatarGroup/index.stories.tsx
```diff
@@ -0,0 +1,61 @@
+import { userNameToAvatarUrl } from '@/util/github';
+import AvatarGroup from './';
+import type { Meta as MetaObj, StoryObj } from '@storybook/react';
+
+type Story = StoryObj<typeof AvatarGroup>;
+type Meta = MetaObj<typeof AvatarGroup>;
+
+const AvatarList = {
```
ovflowd: This is not a component, so no use of titlecase. Call it `defaultProps` or `avatarListProps`

components/Common/AvatarGroup/avatar/index.stories.tsx
ovflowd: Sure

ovflowd:
> We need to know what if image is 404. This is an illustration I did on figma. So it's not part of the design system

I'd vote for the two first letters of their github username :) or at your criteria, that's a common pattern and doesn't require us adding yet another img.

components/Common/AvatarGroup/avatar/index.tsx
```diff
@@ -0,0 +1,25 @@
...
+
+type AvatarProps = {
+  src: string;
+};
+
+const Avatar: FC<AvatarProps> = ({ src }) => {
+  const [url, setUrl] = useState(src);
+
+  return (
+    <Image
```
bmuenzenmeyer: should we use `Avatar` here ? https://www.radix-ui.com/primitives/docs/components/avatar 

this is mentioned in the scope of the issue https://github.com/nodejs/nodejs.org/issues/5910

it has built in support for things like Fallback

ovflowd: Yes, imho we should. I feel that Augustin didn't read the issue as a whole (or maybe just didn't notice them) + didn't read the latest changes to the collab guides...

util/__tests__/github.test.ts
AugustinMauroy: why this other test is in ts ?
https://github.com/nodejs/nodejs.org/tree/main/util/__tests__
BTW: i had change my to `.mjs`

components/Common/AvatarGroup/index.tsx
AugustinMauroy: to fort showing all avatars without taking attention to limit

AugustinMauroy:
> This implementation doesn't look right. It would only work if the images are perfect squares.

should I use `overflow-hidden`?

components/Common/AvatarGroup/avatar/index.tsx
AugustinMauroy: I read but I started the work in computer class and then finished at home. And so I forgot about this. _oops_ 😅

AugustinMauroy:
> I'd vote for the two first letters of their github username :) or at your criteria, that's a common pattern and doesn't require us adding yet another img.

done ! And FYI: GitHub add gray octocat if 404.

components/Common/AvatarGroup/index.tsx
AugustinMauroy: I don't have 100% understand but I think I have do what you propose 😅

components/Common/AvatarGroup/index.tsx
AugustinMauroy: remove because not useful adding complexity for nothing

components/Common/AvatarGroup/avatar/index.module.css
AugustinMauroy: Fix, out of date because we are using `RadixUI`.

components/Common/AvatarGroup/avatar/index.stories.tsx
AugustinMauroy: Fix, out of date because we are using RadixUI.

components/Common/AvatarGroup/avatar/index.module.css
AugustinMauroy: I had re-read normally it's good ! I don't have see/follow new rule 😅.

components/Common/AvatarGroup/index.tsx
ovflowd: Pretty muuch do the map within a `useMemo`

components/Common/AvatarGroup/avatar/index.tsx
ovflowd: If possible, open PRs as ready after going through your code twice or 3 times. I know sometimes we don't do that (I myself often rush with things, but when I do they are usually hot fixes)

For these sort of PRs, even tho as they are super small, focus on re-reading and reviewing your code yourself before opening a PR. Be mindful of the reviewers time.

ovflowd:
> > This implementation doesn't look right. It would only work if the images are perfect squares.
> 
> should I use `overflow-hidden`?

I assume Radix UI does that for you

AugustinMauroy:
> I assume Radix UI does that for you

nope Radix UI just provide Avatar. So for the box we need to style it

components/Common/AvatarGroup/index.tsx
AugustinMauroy: Ok i will test that !

components/Common/AvatarGroup/index.module.css
```diff
@@ -0,0 +1,11 @@
+.avatarGroup {
+  @apply flex items-center overflow-x-auto;
+
+  .avatarRoot {
+    @apply -ml-2 first:ml-0;
+  }
+
+  .avatar {
+    @apply flex h-8 w-8 items-center justify-center rounded-full border-2 border-white bg-neutral-100 text-sm  dark:border-neutral-950 dark:bg-neutral-900;
```
ovflowd: Please, re-read the collaborator guide. Especifically the Styling a Component part.

components/Common/AvatarGroup/index.module.css
```diff
@@ -0,0 +1,11 @@
+.avatarGroup {
+  @apply flex items-center overflow-x-auto;
+
+  .avatarRoot {
+    @apply -ml-2 first:ml-0;
```
ovflowd: Same over here

components/Common/AvatarGroup/index.module.css
```diff
@@ -0,0 +1,11 @@
+.avatarGroup {
+  @apply flex items-center overflow-x-auto;
```
ovflowd: And here

components/Common/AvatarGroup/index.stories.tsx
```diff
@@ -0,0 +1,62 @@
+import { userNameToAvatarUrl } from '@/util/github';
+import AvatarGroup from './';
+import type { Meta as MetaObj, StoryObj } from '@storybook/react';
```
ovflowd, line 1: As I mentioned (again) type imports happen before all imports

components/Common/AvatarGroup/index.stories.tsx
```diff
@@ -0,0 +1,62 @@
...
+  'mikeesto',
+  'bnb',
+  'benhalverson',
+  'aymen94',
+  'shanpriyan',
+  'Wai-Dung',
+  'manishprivet',
+  'araujogui',
+];
+const defaultProps = {
```
ovflowd: Mising empty line between these two constants (nit)

components/Common/AvatarGroup/avatar/index.tsx
AugustinMauroy: Thanks for mentioning it, I'll be careful next time.

components/Common/AvatarGroup/index.tsx
```diff
@@ -0,0 +1,62 @@
...
+};
+
+const AvatarGroup: FC<AvatarGroupProps> = ({ avatars, limit = 10 }) => {
+  const [showMore, setShowMore] = useState(false);
+
+  const toggleShowMore = () => {
+    setShowMore(!showMore);
+  };
+
+  console.log('avatars', avatars);
```
ovflowd: Seriously? C'mon.

components/Common/AvatarGroup/index.tsx
```diff
@@ -0,0 +1,62 @@
...
+    name: string;
+  }[];
+  limit?: number;
+};
+
+const userNameToFallBackText = (name: string) => {
+  // sanitize name before using it
+  name = name.toLowerCase().replace(' ', '-');
+  const words = name.split('-');
+  if (words.length > 1) {
```
ovflowd: This can be simplified to a ternary condition

components/Common/AvatarGroup/index.tsx
```diff
@@ -0,0 +1,62 @@
+'use client';
+import * as Avatar from '@radix-ui/react-avatar';
+import { useState } from 'react';
+import type { FC } from 'react';
+import styles from './index.module.css';
+
+type AvatarGroupProps = {
+  avatars: {
```
ovflowd: Make an individual type for Avatar.

components/Common/AvatarGroup/index.tsx
```diff
@@ -0,0 +1,62 @@
...
+  };
+
+  console.log('avatars', avatars);
+
+  const renderAvatars = showMore ? avatars : avatars.slice(0, limit);
+
+  return (
+    <div className={styles.avatarGroup}>
+      {renderAvatars.map((avatar, index) => (
+        <Avatar.Root key={index} className={styles.avatarRoot}>
```
ovflowd: You should make an individual component for Avatar, as the issue states. Even tho you're using RadixUI.

Please follow the issue description...

components/Common/AvatarGroup/index.tsx
```diff
@@ -0,0 +1,62 @@
...
+  avatars: {
+    src: string;
+    name: string;
+  }[];
+  limit?: number;
+};
+
+const userNameToFallBackText = (name: string) => {
+  // sanitize name before using it
+  name = name.toLowerCase().replace(' ', '-');
```
ovflowd: It is discouraged to reassign/mutate any value that comes from an argument.

components/Common/AvatarGroup/index.tsx
```diff
@@ -0,0 +1,62 @@
...
+
+type AvatarGroupProps = {
+  avatars: {
+    src: string;
+    name: string;
+  }[];
+  limit?: number;
+};
+
+const userNameToFallBackText = (name: string) => {
```
ovflowd:
```suggestion
const usrnameToFallbackText = (name: string) => {
```

components/Common/AvatarGroup/index.module.css
AugustinMauroy: "We require that you define one Tailwind Token per line, just as shown on the example above, since this improves readability" I missed that line really sorry. Can't the `npm run format` do it?

ovflowd (changes requested):
With all due respect, Augustin. I'm not sure for how long I'll keep reiterating myself, at this point I feel this is just plain disrespect towards reviewers.

You're not a newcomer, you've been contributing for moe than one year, and I still need to repeat myself in many things till today. Please, give a small step back and reflect on the things I'm saying.

You don't need to rush with things. Take time and care on perfecting your PRs.

components/Common/AvatarGroup/index.module.css
ovflowd: Sadly nope. There's no eslint rule that can do that. We could create our own ESLint rule, but that might be overkill

ovflowd:
> > > This implementation doesn't look right. It would only work if the images are perfect squares.
> > 
> > 
> > should I use `overflow-hidden`?
> 
> I assume Radix UI does that for you

(Just assuming, but my tip here, test with non rectangular images, until they all look alright)

Usually the trick here is to use images as background-img or to actually zoom the image to fit to the threshold that is asked. This is a good example: https://jsfiddle.net/aew4ztvm/

I also know that modern HTML Image features/CSS can do that for regular images, maybe even RadixUI has this builtin. Keep searching :)

Let's perfect this Component!

Edit: I can see this is not fixed with the current commit, so try to follow what I wrote above

components/Common/AvatarGroup/index.tsx
```diff
@@ -0,0 +1,62 @@
+'use client';
+import * as Avatar from '@radix-ui/react-avatar';
+import { useState } from 'react';
+import type { FC } from 'react';
+import styles from './index.module.css';
+
+type AvatarGroupProps = {
+  avatars: {
+    src: string;
```
ovflowd: Also being very honest, the Avatar component should be agnostic to "GitHub", hence I think the only prop they should receive is a SRC parameter.

At maximum an `alt` optional attribute (both for accessibility) and when the image gives 404 we can render the circle with the text, for example.

The Avatar Component should **not** be GitHub bound. Nor use GitHub usernames as parameters. This is overcomplexing the code and making it limited to GitHub avatars.

ovflowd:
image

Also this part is not as how the Figma is showing (the +X part)

components/Common/AvatarGroup/index.module.css
AugustinMauroy: Ok thanks for clarification. I reread it twice so that I can see the changes. I think that we should notify changes to the collaborator guide on slack.

components/Common/AvatarGroup/index.tsx
AugustinMauroy: so create a new variable call result.

components/Common/AvatarGroup/index.tsx
AugustinMauroy: I will take a look to that

components/Common/AvatarGroup/index.tsx
AugustinMauroy: in separated file ?

ovflowd: Nope

AugustinMauroy: OK 👍

bmuenzenmeyer:
i plan on reviewing this today - but i do notice some merge conflicts that could be fixed in the meantime.

@AugustinMauroy while you are waiting, if you could review the dependabot PRs, that'd be helpful!

components/Common/AvatarGroup/Avatar/index.tsx
```diff
@@ -0,0 +1,29 @@
...
+import styles from './index.module.css';
+
+type AvatarProps = {
+  src?: string;
+  alt?: string;
+  extraClass?: string;
+  onClick?: () => void;
+};
+
+const usrnameToFallbackText = (name?: string) => {
```
ovflowd: You should assume an alt is provided or simply do `name: string = ''`

So no need for `if (!name) return`

components/Common/AvatarGroup/Avatar/index.tsx
```diff
@@ -0,0 +1,29 @@
...
+  // sanitize name before using it
+  if (!name) return;
+  const sanitizeName = name.toLowerCase().replace(' ', '-');
```
ovflowd: This whole code can be extremely simplified to:

```ts
const fallbackAltAsAvatar = (alt: string = '') => 
  alt.match(/\b(\w)/g).join('').substring(0, 2)
```

The new name is also more inline with what this code does... As it's not grabbing an username, but the `alt` text.

components/Common/AvatarGroup/Avatar/index.tsx
```diff
@@ -0,0 +1,29 @@
...
+  const words = sanitizeName.split('-');
+  return words.length > 1 ? words[0][0] + words[1][0] : name.slice(0, 2);
+};
+
+const Avatar: FC<AvatarProps> = ({ src, alt, extraClass, onClick }) => (
+  <RadixAvatar.Root className={extraClass} onClick={onClick}>
+    <RadixAvatar.Image src={src} alt={alt} className={styles.avatar} />
+    <RadixAvatar.Fallback className={styles.avatar}>
+      {src ? usrnameToFallbackText(alt) : alt}
```
ovflowd: What is this condition doing? Shouldn't it always invoke the function?

util/github.ts
```diff
@@ -0,0 +1,3 @@
+export const userNameToAvatarUrl = (userName: string): string => {
```
ovflowd:
```suggestion
export const githubProfileAvatarUrl = (username: string): string => {
```

ovflowd: Also no need of `{}` and a `return` statement.

components/Common/AvatarGroup/Avatar/index.tsx
AugustinMauroy: it's will become to much complicated to read because prettier taking one line for each "dot". and regex possibly null.

components/Common/AvatarGroup/Avatar/index.tsx
AugustinMauroy: gotcha !

components/Common/AvatarGroup/Avatar/index.tsx
AugustinMauroy: this allowing to use `Avatar` for expending. So alt is `+X`. And it's allow to have one style for each element on avatarGroup.
So I thinks it's good

components/Common/AvatarGroup/index.tsx
```diff
@@ -0,0 +1,48 @@
...
+    alt?: string;
+  }[];
+  limit?: number;
+};
+
+const AvatarGroup: FC<AvatarGroupProps> = ({ avatars, limit = 10 }) => {
+  const [showMore, setShowMore] = useState(false);
+
+  const toggleShowMore = () => {
+    setShowMore(!showMore);
```
AugustinMauroy: should we put this directly on `onClick` ?

components/Common/AvatarGroup/Avatar/index.tsx
ovflowd: Wdym? The code I placed here does the exact same as your code above. It's just a simplification, please apply it, and/or add unit tests showing what is your expected behaviour and what is not working correctly.

You can always tweak the regex. And also you could "document" the regex if you want.

AugustinMauroy: OK i will deep dive (regex is my weak point)

components/Common/AvatarGroup/Avatar/index.tsx
ovflowd: Wait, you're using the same Avatar component for `+X`? I... gotcha. But implying that this should be done based on the presence or absence of a `src` is flakey logic.

No one besides the maintainers of this code will know that this implies that. Imagine I add a "src" on-demand? (lazy loading) then, my whole component would be just the "alt" text and I would not know why.

I discourage doing this within this component.

You should do the `+X` logic within `AvatarGroups` because that's where this makes sense. Avatar itself should not be aware of such behaviour.

**1st Recommendation:**

- Remove this `src` check. Always call the method
- Within `AAvatarGroup` you import the styles from the individual Avatar (as `import avatarStyles from './Avatar/index.css'`;
- Add the "+X" element with the Avatar styles there. That should be good enough IMO.

**2nd Recommendation:** (Preferred)

Or you simply remove the `usernameToFallbackText` (this is a poor method name) from the Avatar component. Which makes more sense, simply render the `alt`.

And let end-users manually call the function if they want their alt to be rendered that way. (ie: export this method to a string utility file) and simply call it outside of AvatarGroup `([username: 'ovflowd', alt: textToAcronym('ovflowd', 2)]`

And the method could become generic such as `textToAcronym` with two args: the input, and the length of the acronym (i.e. two digits)

components/Common/AvatarGroup/index.tsx
ovflowd: Sure

components/Common/AvatarGroup/Avatar/index.tsx
ovflowd: ChatGPT is your friend. "Make me a regex in JavaScript that does XXXX by using the String.match function"

components/Common/AvatarGroup/Avatar/index.tsx
AugustinMauroy: i have done something look clean for me. Tell me what do you think

components/Common/AvatarGroup/Avatar/index.module.css
```diff
@@ -0,0 +1,21 @@
+.avatar {
+  @apply flex
```
ovflowd, line 2: Forgot indentation 🙈

components/Common/AvatarGroup/Avatar/index.tsx
```diff
@@ -0,0 +1,25 @@
+import * as RadixAvatar from '@radix-ui/react-avatar';
+import type { FC } from 'react';
+
+import type { AvatarType } from '@/components/Common/AvatarGroup';
+
+import styles from './index.module.css';
+
+type AvatarProps = AvatarType;
+
+export const fallbackAltAsAvatar = (alt: string = '') =>
```
ovflowd: As I was saying, this should probably be removed from here, so you only render `alt`

And exported to the same GitHub utility or a string utility (I explained the thing about being a generic function that can create these acronyms with X characters:)

```tsx
export const getAcronymFromString = (str = '', length = 2) =>
 str.match(/\b(\w)/g).join('').toLocaleUpperCase().substr(0, length);
```

And then that's called by AvatarGroup, for example.

components/Common/AvatarGroup/index.tsx
```diff
@@ -0,0 +1,48 @@
...
+    [showMore, avatars, limit]
+  );
+
+  return (
+    <div className={styles.avatarGroup}>
+      {renderAvatars.map((avatar, index) => (
+        <Avatar src={avatar.src} alt={avatar.alt} key={index} />
+      ))}
+      {avatars.length > limit && (
+        // this is simulating Radix Avatar.root and Avatar.fallback
```
ovflowd: No need to add inline comments here

components/Common/AvatarGroup/index.tsx
```diff
@@ -0,0 +1,48 @@
...
+    () => (showMore ? avatars : avatars.slice(0, limit)),
+    [showMore, avatars, limit]
+  );
+
+  return (
+    <div className={styles.avatarGroup}>
+      {renderAvatars.map((avatar, index) => (
+        <Avatar src={avatar.src} alt={avatar.alt} key={index} />
+      ))}
+      {avatars.length > limit && (
```
ovflowd, line 34:
```suggestion

      {avatars.length > limit && (
```

Just to separate these two things

components/Common/AvatarGroup/index.tsx
```diff
@@ -0,0 +1,48 @@
+'use client';
+import type { FC } from 'react';
+import { useState, useMemo } from 'react';
+
+import Avatar from './Avatar';
+import avatarstyles from './Avatar/index.module.css';
+import styles from './index.module.css';
+
+export type AvatarType = {
```
ovflowd: Why do we need to export this? And why would AvatarGroup export the type of Avatar?

If you need on Storybook you can do it with:

```tsx
import type { ComponentProps } from 'react';

type AvatarProps = ComponentProps<typeof Avatar>;
```

components/Common/AvatarGroup/Avatar/index.tsx
ovflowd: (and yea this should not be exported from here 🙈)

components/Common/AvatarGroup/Avatar/index.tsx
```diff
@@ -0,0 +1,25 @@
...
+  alt
+    .match(/\b(\w)/g)
+    ?.join('')
+    .toLocaleUpperCase();
+
+const Avatar: FC<AvatarProps> = ({ src, alt }) => (
+  <RadixAvatar.Root className={styles.avatarRoot}>
+    <RadixAvatar.Image src={src} alt={alt} className={styles.avatar} />
+    <RadixAvatar.Fallback className={styles.avatar}>
+      {fallbackAltAsAvatar(alt)}
```
ovflowd:
```suggestion
      {alt}
```

components/Common/AvatarGroup/index.tsx
AugustinMauroy: so I juste remove it ?

components/Common/AvatarGroup/Avatar/index.module.css
AugustinMauroy: 🙈

components/Common/AvatarGroup/index.tsx
ovflowd: Yean, no need to have these inline comments :)

components/Common/AvatarGroup/Avatar/index.module.css
ovflowd, line 1: Not sure you noticed, but the indentation is still wrong.

components/Common/AvatarGroup/index.module.css
ovflowd, line 1: Also here...

components/Common/AvatarGroup/index.stories.tsx
```diff
@@ -0,0 +1,80 @@
...
+// an link to other image can also be used
+const defaultProps = {
+  avatars: names.map(name => ({
+    src: githubProfileAvatarUrl(name),
+    alt: name,
+  })),
+};
+
+// add falback text to avatars we are hardcoding link because github api add image for 404 user avatar
+defaultProps.avatars.push({
```
ovflowd: I don't think .push should be called here. If you want to add this to avatars on defaultProps... You can simply:

```tsx
const unknownAvatar = [{
  src: 'https://avatars.githubusercontent.com/u/',
  alt: 'unknown-avatar'
});

const defaultProps = {
  avatars: [unknownAvatar, ...avatars.map(....)],
};
```

components/Common/AvatarGroup/index.stories.tsx
```diff
@@ -0,0 +1,80 @@
...
+      </div>
+    ),
+  ],
+  args: {
+    ...defaultProps,
+    limit: 5,
+  },
+};
+
+// normally it's support by radix-ui
```
ovflowd: What is this comment supposed to mean?

components/Common/AvatarGroup/index.tsx
```diff
@@ -0,0 +1,51 @@
+'use client';
```
ovflowd: Why did you add `'use client';`? We're not using React Server-Side/Client-Side components yet. Adding this shouldn't be done at the moment.

components/Common/AvatarGroup/index.tsx
ovflowd: Still waiting you to make a dedicated type for this 👀

util/avatars.ts
```diff
@@ -0,0 +1,15 @@
+export const githubProfileAvatarUrl = (username: string): string =>
+  `https://avatars.githubusercontent.com/${username}`;
+
+export const getAcronymFromString = (str: string): string =>
```
ovflowd: This shouldn't be on a file called avatars, as this can be used for any sort of string.

util/avatars.ts
```diff
@@ -0,0 +1,15 @@
...
+  str || str.length > 0
```
ovflowd: I don't think this is correct... There's no need for an explicit return type or this overly complex logic. I recommend studying how `string.match()` works. Also, adding a `ts-ignore` is a no-go; in most cases, if you ever need to do a `// ts-ignore`, you're doing something wrong.

```tsx
export const getAcronymFromString = (str = '') =>
  [...str.trim().match(/\b(\w)/g) || ''].join().toUpperCase()
```

ovflowd: `.match()` can return null, hence why the `[...]` spread that ensures value becomes an array. And join() ensures it becomes a string, because `[null].join()` becomes an empty string. And `[...null || ''].join()` also becomes an empty string

ovflowd: Also, there's no need to call `locale` here. Plain uppercase is enough.

components/Common/AvatarGroup/Avatar/index.module.css
AugustinMauroy: What I don't see the problem. copy page indentation for other component so could you give more information ?

components/Common/AvatarGroup/index.tsx
AugustinMauroy: I'm adding that because if it's put on app dir it's will throw an error. I did this in anticipation of the future. I'm convinced that APP dir will replace page dir in future Nextjs updates. But if the website team doesn't think it's necessary, I'll remove it.

util/avatars.ts
AugustinMauroy: If you are talking about file name. you thinks we should have a file call `string` and a file `github` ?

components/Common/AvatarGroup/index.stories.tsx
AugustinMauroy: I had removed it but it's just mean "hey normally radix do something to support non-square"

components/Common/AvatarGroup/Avatar/index.module.css
ovflowd: Just look at the diff in this file. Please, be attentive,

image

components/Common/AvatarGroup/index.module.css
ovflowd: In case you also haven't noticed, Augustin. 

image

util/avatars.ts
```diff
@@ -0,0 +1,7 @@
+export const githubProfileAvatarUrl = (username: string): string =>
+  `https://avatars.githubusercontent.com/${username}`;
+
+export const getAcronymFromString = (str = '') =>
+  str.length > 0
```
ovflowd: Why are you checking `str.length`? The string should always have something. There's no point of someone using this just to pass an empty string. We could even remove the `(str = '') from the signature to enforce that a valid string is required.

components/Common/AvatarGroup/index.tsx
ovflowd: > I'm adding that because if it's put on app dir it's will throw an error. I did this in anticipation of the future.

Don't make unnecessary changes. We don't know when we're going to migrate to `app` router, and we definitely want to tackle this into another time.

Every other component is not using "use client".

> But if the website team doesn't think it's necessary, I'll remove it.

So, I'm not Website team? Since I asked you to remove it and you said "no" lol

util/avatars.ts
ovflowd: I think this specific one (the lower one) should be something like `stringUtils.ts` and the upper one (`githubUtils.ts`) or maybe just `strings.ts` and `github.ts`

AugustinMauroy: OK ! I agree with you

util/avatars.ts
AugustinMauroy: > Why are you checking str.length? 

`str.length` it's just to output something if string it's empty

components/Common/AvatarGroup/index.tsx
AugustinMauroy: > So, I'm not Website team? Since I asked you to remove it and you said "no" lol

😄 I mean yeah I will remove it but I just want to be sure you have my POV

components/Common/AvatarGroup/index.module.css
AugustinMauroy: Oh ooooppss

components/Common/AvatarGroup/Avatar/index.module.css
AugustinMauroy: Sorry but how did you display the dot ?

ovflowd: Which dot?

util/avatars.ts
ovflowd: But we shouldn't provide empty strings here, that's the point.

components/Common/AvatarGroup/index.tsx
ovflowd: I have your PoV. Still, we want to streamline the work here. So any work related to "app" router, even if it's one line, we want to properly do research in the future.

util/avatars.ts
AugustinMauroy: I don't really get your idea.
so you mean do something like that:
```js
export const getAcronymFromString = (str: string) =>
  str.length > 0 || str
    ? [...(str.trim().match(/\b(\w)/g) || '')].join('').toUpperCase()
    : 'NA';
```

components/Common/AvatarGroup/Avatar/index.module.css
AugustinMauroy: dot to display  indentation

ovflowd (changes requested):
Sadly the PR still have issues, for example. You haven't fixed the issue of non-square images. (Which is not something Radix should take care of), as Radix doesn't do Component Styling, just the actual rendering part.

I'm taking over this PR, because sadly I just can't keep going on a back and forth discussion here. This is blocking other work from getting done.

ovflowd (approved):
Now the PR is finally ready and spec complete.

cc @nodejs/nodejs-website if anyone else wants to give an extra approval/review.

ovflowd:
BTW @AugustinMauroy I just realised you forgot to add unit tests for the "showMore" interaction. Could you please do that?

AugustinMauroy:
Yeah, I completely forgot 😄

bmuenzenmeyer (approved):

bmuenzenmeyer:
I think this is just waiting on the unit test then

ovflowd (changes requested):
I don't think you've noticed, Augustin. But you literally reverted my changes. 😞

ovflowd (approved):
I've re-added my changes. Please, be more careful next time, Augustin. Pull branches instead of force-pushing.

ovflowd (approved):
Anyhow, overall, thank you for your contribution, Augustin! I appreciate your effort here.

As usual, I'd recommend to be more attentive to feedback, and more careful when pushing changes, and be more diligent (self-review) your changes before pushing them 🙇

PR #5928, opened 2023-10-01
feat(components): add `Breadcrumbs` component

### LINKED ISSUE #5914: Create `Breadcrumbs` Component
Create a `Breadcrumbs` component using the new [figma design (direct link)](https://www.figma.com/file/pu1vZPqNIM7BePd6W8APA5/Node.js?type=design&node-id=433-4349&mode=dev). 

---

### Before You Start...

---

### Details

image

- Create a new directory called `Breadcrumbs` within the [`Common` directory](https://github.com/nodejs/nodejs.org/tree/main/components/common) that contains all the new code. (This location may change in the future.)
- Use the new tailwind color and typography system to assign properties. For example, the active link uses Green/600 per the figma.
- The `@heroicons/react` module has the icons `chevron-right` and `home available for use
- Add a storybook story file `components/common/Breadcrumbs/index.stories.tsx` which exercises each of the component's states.

Suggested props include:

- `links`, an array of objects that looks like this:

```js
[
    {
        value: "Learn",
        url: "path/to/learn"
    },
    {
        value: "Getting Started",
        url: "path/to/learn/getting-started"
    },
    {
        value: "Introduction to Node.js,
        url: "path/to/learn/getting-started/intro"
    }
]
```

- The home icon is always rendered, and goes to the root
- The last record in the array is considered the active location.
  - Research if [Next.js useRouter](https://nextjs.org/docs/pages/api-reference/functions/use-router) would be useful for confirming the user is at the leaf node. Though I don't know it it's better to be fully static, fully dynamic, or a hybrid.

- Light
- Dark

### Open Questions, to Discuss

- Do breadcrumbs truncate at some point? The figma shows an ellipsis. I've asked @HaydenBleasel to clarify

image

junwen-k:
## Description

Adds `Breadcrumbs` component based on provided Figma design.

Instead of using width to determine if the breadcrumb items should be truncated, it is now specified by `maxLength` props for the sake of simplicity.

Generally, I would split the components into smaller ones (e.g, `BreadcrumbRoot`, `BreadcrumbList`, `BreadcrumbItem` etc) and uses compositional technique to build more complicated components. At the current state it might seem a bit overkill hence I've created the component in a higher level of abstraction where you specify a list of links to render, and the last item is always considered as the "active" item without any integration of Next router.

Citing from [w3.org](https://www.w3.org/WAI/ARIA/apg/patterns/breadcrumb/examples/breadcrumb/#rps_label)
> Applied to the last link in the set to indicate that it represents the current page.

As for the home icon, I am currently using the `Home` icon from [Heroicons](https://heroicons.com/) instead of the icon in the Figma design for consistency.

## Validation

Added `Breadcrumbs` component in Storybook.

`Default`

image

`Truncate`

image

`Hide Home` 

image

## Related Issues

Closes #5914.

### Check List


AugustinMauroy:
happy to see you back on website redesign !

components/Common/Breadcrumbs/index.module.css
```diff
@@ -0,0 +1,36 @@
+.wrapper {
+  .list {
+    @apply flex
+    items-center
```
umairraza96: tailwind tokens should be formatted with a space, see [Collaborator Guideline](https://github.com/nodejs/nodejs.org/blob/main/COLLABORATOR_GUIDE.md#styling-a-component)

AugustinMauroy (commented):
Design look same, spacing is right color good !
But I found nothing here or on issue about what if we click on "...". Does it expand?

components/Common/Breadcrumbs/index.module.css
AugustinMauroy: [Collaborator Guideline](https://github.com/nodejs/nodejs.org/blob/main/COLLABORATOR_GUIDE.md#example-of-a-css-module):
"We require that you define one Tailwind Token per line, just as shown on the example above, since this improves readability"
_It has changed recently but no worries. Even I didn't see the change when I reread it._

umairraza96: Yes, but if you look closely its also indented with a space like this:
```css
@apply flex
 items-center; 
```

ovflowd: That's not what @umairraza96 said, Augustin. (Umair is correct here)

The point is that after the first `@apply` line each next line must have an extra level of indentation. i.e:

```css
@apply flex
  items-center
  ...
```

components/Common/Breadcrumbs/index.tsx
```diff
@@ -0,0 +1,73 @@
...
+        <li className={styles.item}>
+          {/* NOTE: In the future, this could be expanded with Dropdown feature to allow selection of route */}
+          <button disabled>...</button>
+          <ChevronRightIcon
+            aria-hidden="true"
+            className={classNames(styles.icon, styles.separator)}
+          />
+        </li>
+      )}
+      {links
```
ovflowd: Could we extract this into an `useMemo`?

components/Common/Breadcrumbs/index.tsx
```diff
@@ -0,0 +1,73 @@
...
+};
+
+const Breadcrumbs: FC<BreadcrumbsProps> = ({
+  links,
+  maxLength = 3,
+  hideHome = false,
+}) => (
+  <nav aria-label="breadcrumb" className={styles.wrapper}>
+    <ol className={styles.list}>
+      {!hideHome && (
```
ovflowd, line 57: Can we extract this as a simple Subcomponent? (BreadcrumbRoot)

components/Common/Breadcrumbs/index.tsx
ovflowd, line 1: I feel like certain things here could become simpler, such as the Root (Home), the `...`, which would allow us to generify certain things and simple construct the whole things as an array

Where the first item is the Home, and then the nth-1 is the `...` or something of sort.

I like the idea of using a `maxLength` to truncate things, as we could use hooks such as `useMediaQuery` to generate on-demand props of what should the max-length be. Good stuff.

ovflowd (commented):
Overall, I like what I see! Excellent stuff :)

I'm going to give a look into the Stories to see if I've missed anything, but I also left a few comments with changes that'd be nice, and one on the CSS that is required.

BTW, your import order will update once https://github.com/nodejs/nodejs.org/pull/5929 gets merged, so I recommend a rebase after that.

components/Common/Breadcrumbs/index.module.css
junwen-k: Cool, is there a way to automate this process though? Not sure if prettier supports configuration that does just this. Normally I would expect that the formatter doing the heavy lifting so that we can just focus on the coding part.

I also recommend setting up a prettier sort import plugin, e.g, [prettier-plugin-sort-imports](https://github.com/trivago/prettier-plugin-sort-imports) to enforce consistent import style. :) @ovflowd

ovflowd: We already merged a consistent import style, on ESLint side of things. (A few hours ago)

No there's no way to automate this without a Stylelint/Prettier plugin. I find it okay-enough to just ask for people to abide by it.

junwen-k:
> Design look same, spacing is right color good ! But I found nothing here or on issue about what if we click on "...". Does it expand?

Not for the current implementation, a `Dropdown` component is needed if we were to follow the implementation of [Blueprintjs](https://blueprintjs.com/docs/#core/components/breadcrumbs) though.

components/Common/Breadcrumbs/index.tsx
```diff
@@ -0,0 +1,73 @@
...
+            className={classNames(styles.icon, styles.separator)}
+          />
+        </li>
+      )}
+      {links
+        .slice(links.length - maxLength, links.length)
+        .map((link, idx, _links) => {
+          const isLastItem = idx === _links.length - 1;
+          return (
+            <li key={link.href.toString()} className={styles.item}>
```
ovflowd: Interesting choice to use the href as a key.

junwen-k:
@ovflowd Hi, I've did some refactoring by splitting Breadcrumbs into smaller sub components.

Initially I thought of using CSS to control the active state of the link (last-child) selector, but since it has been separated to different sub component, right now it is controlled with Javascript. 

I've also thought of using 

```tsx
<ActiveLocalizedLink
  activeClassName={styles.active}
  href={link.href}
  {...(isLastItem && { 'aria-current': 'page' })}
>
  {link.label}
</ActiveLocalizedLink>
```

so that it follows the current active page instead of being always the last item.

There are some edge cases need to be handled, e.g, if `maxLength` is set to less than 1, some weird bugs will appear.

Nonetheless, I feel that this still be can improved as you suggested (perhaps the caller passing the entire array to be rendered and making this Breadcrumb component dumber, etc). I believe the component should be good enough for the current use case. :)

ovflowd:
> Nonetheless, I feel that this still be can improved as you suggested (perhaps the caller passing the entire array to be rendered and making this Breadcrumb component dumber, etc). I believe the component should be good enough for the current use case. :)

Feel free to keep iterating on this component as much as you want :)

We're in no rush to get this merged, and if you feel that you genuinely want to keep improving this, that's fine.

> There are some edge cases need to be handled, e.g, if maxLength is set to less than 1, some weird bugs will appear.

I'm fine on not having logic for this. I would rather consider this Application logic and delegate the caller to decide on what values should be passed (ie `useMediaQuery` if returning less than 1 should fallback to 1). This way we reduce extra logic on a Component, that should do only one job: Render a Breadrcumb.

components/Common/Breadcrumbs/index.tsx
```diff
@@ -0,0 +1,59 @@
...
+  maxLength?: number;
+  hideHome?: boolean;
+};
+
+const Breadcrumbs: FC<BreadcrumbsProps> = ({
+  links,
+  maxLength = 3,
+  hideHome = false,
+}) => {
+  const items = useMemo(
```
ovflowd, line 30: I think the whole `.map` could be memoized here, that was more my point

components/Common/Breadcrumbs/index.tsx
```diff
@@ -0,0 +1,59 @@
...
+    () =>
+      links.length > maxLength ? links.slice(links.length - maxLength) : links,
+    [links, maxLength]
+  );
+  return (
+    <BreadcrumbRoot hideHome={hideHome}>
+      {links.length > maxLength && (
+        <BreadcrumbItem>
+          {/* NOTE: In the future, this could be expanded with Dropdown feature to allow selection of route */}
```
ovflowd: Could we remove this comment? I don't think it brings value at the moment

components/Common/Breadcrumbs/index.tsx
```diff
@@ -0,0 +1,59 @@
+import classNames from 'classnames';
+import { type LinkProps } from 'next/link';
+import Link from 'next/link';
```
ovflowd: Can you replace this by `LocalizedLink`, we avoid using `next/link` and we only use LocalizedLink which extends `next/link`

components/Common/Breadcrumbs/index.tsx
```diff
@@ -0,0 +1,59 @@
...
+      {items.map((link, idx, arr) => {
+        const isLastItem = idx === arr.length - 1;
+        return (
+          <BreadcrumbItem key={link.href.toString()} hideSeparator={isLastItem}>
+            <Link
+              href={link.href}
+              className={classNames({
+                [styles.active]: isLastItem,
+              })}
+              {...(isLastItem && { 'aria-current': 'page' })}
```
ovflowd: I guess instead of this you could:

```suggestion
              aria-current={isLastItem ? 'page' : undefined}
```

ovflowd: (Removes the need of complex weird spread operators)

components/Common/Breadcrumbs/BreadcrumbRoot/index.tsx
```diff
@@ -0,0 +1,31 @@
...
+import LocalizedLink from '@/components/LocalizedLink';
+
+import styles from './index.module.css';
+
+type BreadcrumbRootProps = {
+  hideHome?: boolean;
+};
+
+const BreadcrumbRoot: FC<PropsWithChildren<BreadcrumbRootProps>> = ({
+  hideHome = false,
```
ovflowd: I usually prefer non-negated logic, like `showHome`, often negated logic can be easier to read, and it will make it sound as "home" should always be rendering except in some cases.

components/Common/Breadcrumbs/index.tsx
```diff
@@ -0,0 +1,62 @@
...
+  links,
+  maxLength = 3,
+  hideHome = false,
+}) => {
+  const items = useMemo(
+    () =>
+      (links.length > maxLength
+        ? links.slice(links.length - maxLength)
+        : links
+      ).map((link, idx, arr) => {
```
ovflowd:
```suggestion
      ).map((link, index, items) => {
```

No need to abbreviate these constants

components/Common/Breadcrumbs/BreadcrumbRoot/index.tsx
junwen-k: Thanks for commenting on this. Generally I tend to follow the prop naming convention by [MUI](https://mui.com/material-ui/guides/api/#prop-naming). I assumed that the Home button will be shown by default, hence we can do `hideHome` instead of `showHome={false}`

components/Common/Breadcrumbs/index.tsx
```diff
@@ -0,0 +1,62 @@
...
+};
+
+const Breadcrumbs: FC<BreadcrumbsProps> = ({
+  links,
+  maxLength = 3,
+  hideHome = false,
+}) => {
+  const items = useMemo(
+    () =>
+      (links.length > maxLength
```
ovflowd: For the sake of keeping the code clean, it does make sense to assign this to a constant,

also, you could simply do this:

```ts

const itemsToRender = links.slice(links.length > maxLength && maxLength || 0);
```

For all situations, this logic should work.

components/Common/Breadcrumbs/BreadcrumbRoot/index.tsx
ovflowd: I see, I thought for one second that this was the BreadcrumbItem component. Yea negating here makes sense

components/Common/Breadcrumbs/index.tsx
```diff
@@ -0,0 +1,62 @@
...
+      ).map((link, index, items) => {
+        const isLastItem = index === items.length - 1;
+        return (
+          <BreadcrumbItem key={link.href.toString()} hideSeparator={isLastItem}>
+            <LocalizedLink
+              href={link.href}
+              className={classNames({
+                [styles.active]: isLastItem,
+              })}
+              aria-current={isLastItem ? 'page' : undefined}
```
ovflowd:
```suggestion
              aria-current={isLastItem && 'page'}
```

Could you check if this works? (I fear that it would render a string `false`

components/Common/Breadcrumbs/index.tsx
```diff
@@ -0,0 +1,62 @@
...
+            >
+              {link.label}
+            </LocalizedLink>
+          </BreadcrumbItem>
+        );
+      }),
+    [links, maxLength]
+  );
+  return (
```
ovflowd, line 55:
```suggestion

  return (
```

We usually add an empty line between variable assignment and other operators just to keep it clean

components/Common/Breadcrumbs/index.tsx
```diff
@@ -0,0 +1,62 @@
...
+  hideHome = false,
+}) => {
+  const items = useMemo(
+    () =>
+      (links.length > maxLength
+        ? links.slice(links.length - maxLength)
+        : links
+      ).map((link, index, items) => {
+        const isLastItem = index === items.length - 1;
+        return (
```
ovflowd, line 39:
```suggestion

        return (
```

components/Common/Breadcrumbs/index.tsx
junwen-k: Yup, it would render a string `false`.

junwen-k: Yup, it will render a string with value `false`.

components/Common/Breadcrumbs/BreadcrumbRoot/index.tsx
```diff
@@ -0,0 +1,31 @@
...
+  children,
+}) => (
+  <nav aria-label="breadcrumb">
+    <ol className={styles.list}>
+      {!hideHome && (
+        <BreadcrumbItem>
+          <LocalizedLink href="/">
+            <HomeIcon aria-label="Home" className={styles.icon} />
```
ovflowd: This probably should be a i18n string (Home) in case

canerakdas:
Hey, @AugustinMauroy @ovflowd  Do we have SEO concerns throughout the project (I think it should be)? When I check the storybook preview on Markup Validator it doesn't see it as a valid Breadcrumb schema.

Validation: https://validator.schema.org/
Schema: https://schema.org/BreadcrumbList

ovflowd:
> Hey, @AugustinMauroy @ovflowd Do we have SEO concerns throughout the project (I think it should be)? When I check the storybook preview on Markup Validator it doesn't see it as a valid Breadcrumb schema.
> 
> Validation: [validator.schema.org](https://validator.schema.org/) Schema: [schema.org/breadcrumblist](https://schema.org/breadcrumblist)

Can you attach the output? And I assume you've added the iframe URL there? I think being SEO compatible would be great, but it depends on effort x result

ovflowd (approved):
SGTM! Thank you for the contribution 🎉 exceptional work done here :)

ovflowd:
I guess the PR needs rebasing and then `npm run format`

canerakdas:
> > Hey, @AugustinMauroy @ovflowd Do we have SEO concerns throughout the project (I think it should be)? When I check the storybook preview on Markup Validator it doesn't see it as a valid Breadcrumb schema.
> > Validation: [validator.schema.org](https://validator.schema.org/) Schema: [schema.org/breadcrumblist](https://schema.org/breadcrumblist)
> 
> Can you attach the output? And I assume you've added the iframe URL there? I think being SEO compatible would be great, but it depends on effort x result

I attached the iframe inner HTML as a code snippet. Output is here;
image

ovflowd:
> > > Hey, @AugustinMauroy @ovflowd Do we have SEO concerns throughout the project (I think it should be)? When I check the storybook preview on Markup Validator it doesn't see it as a valid Breadcrumb schema.
> > > Validation: [validator.schema.org](https://validator.schema.org/) Schema: [schema.org/breadcrumblist](https://schema.org/breadcrumblist)
> > 
> > 
> > Can you attach the output? And I assume you've added the iframe URL there? I think being SEO compatible would be great, but it depends on effort x result
> 
> I attached the iframe inner HTML as a code snippet. Output is here; image

I think this is expected as there's no microdata within the HTML. This schema validator only checks for schema errors and checks for microdata.

Afaik we're following the "No Markup" approach as described here https://schema.org/BreadcrumbList, so everything looks alright IMO.

Even on Google's own guide https://developers.google.com/search/docs/appearance/structured-data/breadcrumb#html_1 it seems that we're all abiding correctly.

ovflowd:
If we want it to detect Breadcrumbs, we must inject the microdata. Also, I prefer Google's validator (https://search.google.com/test/rich-results). It just has more details.

(Note: I'm all fine by adding Microdata here), since Breadcrumblist seems to be an excellent addition to SEO. And the changes seem trivial.

junwen-k:
> SGTM! Thank you for the contribution 🎉 exceptional work done here :)

Thanks! I've added unit test to the PR as well. Please feel free to take a look before merging :)

AugustinMauroy (approved):
LGTM !

components/Common/Breadcrumbs/__tests__/index.test.mjs
```diff
@@ -0,0 +1,147 @@
+import { render } from '@testing-library/react';
+
+import Breadcrumbs from '..';
+
+jest.mock('next/router', () => jest.requireActual('next-router-mock'));
+
+describe('Breadcrumbs', () => {
```
ovflowd: I usually like unit tests, but aren't the Storybooks already testing the exact same behaviour/functionality that is being done here?

I think we usually add unit tests for components that allow user interaction (such as clicks) and that have multiple states that Storybook can't test.

Since we use Visual Regression Tests on Storybook. Nothing against keeping these, but I wonder if this is needed. Because I don't think so.

package.json
```diff
@@ -57,6 +57,7 @@
     "lint-staged": "14.0.1",
     "next": "^13.5.2",
     "next-mdx-remote": "^4.4.1",
+    "next-router-mock": "^0.9.10",
```
ovflowd: Please don't add a new package just for mocking a package. You can easily mock the next router with a simple object.

Read the Collaborator Guidelines, we recommend avoiding any package as much as possible.

Also I believe this one can be removed, because I still believe that unit test is not necessary.

components/Common/Breadcrumbs/BreadcrumbRoot/index.tsx
```diff
@@ -0,0 +1,31 @@
...
+type BreadcrumbRootProps = {
+  hideHome?: boolean;
+};
+
+const BreadcrumbRoot: FC<PropsWithChildren<BreadcrumbRootProps>> = ({
+  hideHome = false,
+  children,
+}) => (
+  <nav aria-label="breadcrumb">
+    <ol className={styles.list}>
```
ovflowd: I see that you haven't added the microdata from schema.org yet, are you going to eventually add it?

components/Common/Breadcrumbs/BreadcrumbRoot/index.tsx
ovflowd: Check the TRANSLATING.md guideline for this part of the work

And this translation should probably be `Navigate to Home`

components/Common/Breadcrumbs/index.tsx
```diff
@@ -0,0 +1,75 @@
...
+}) => {
+  const items = useMemo(() => {
+    const itemsToRender = links.slice(
+      links.length > maxLength ? -maxLength : 0
+    );
+
+    return itemsToRender.map((link, index, items) => {
+      const isLastItem = index === items.length - 1;
+      let position = index + 1;
+      if (!hideHome) {
```
ovflowd: I'm not much of a fan of this hacky logic...

We could remove the `.slice()` call and do the map for all entries but have an `if` statement that well ... if the current index is more significant than maxLength, then just don't render anything.

**Now...**

Is hiding entries with JavaScript a good idea? Knowing this is SEO-related and that we're rendering different amounts of items will mess up SEO.

So I have the feeling we might want to switch over from a JavaScript approach (render all items) to a CSS approach (if current index > maxLength, add something that does `display: none`) to the items that should be hidden. (Like a class name)

This will guarantee that all SEO entries are always there (as we simply render all entries) and that we're still only displaying whatever `maxLength` dictates, but with CSS.

components/Common/Breadcrumbs/BreadcrumbRoot/index.tsx
```diff
@@ -0,0 +1,57 @@
...
+        {!hideHome && (
+          <BreadcrumbItem>
+            <LocalizedLink
+              itemScope
+              itemType="http://schema.org/Thing"
+              itemProp="item"
+              itemID="/"
+              href="/"
+            >
+              <span itemProp="name">
```
ovflowd: Does it need to be surrounded by a span? Quite sure you could just add itemProp to the SVG.

ovflowd:
Hey @junwen-k just weather checking if everything is alright or if you need any support here 🙇

junwen-k:
> Hey @junwen-k just weather checking if everything is alright or if you need any support here 🙇

Hi, sorry for the unresponsiveness.

I am currently caught up with work related issues and unable to make any significant changes from here at the moment.
Let me try to wrap this up asap so that this can be merged 😄.

Thanks!

ovflowd:
No worries! Take your time 🙇

aymen94 (approved):

aymen94:
LGTM @junwen-k thanks for your contribution 😄

components/Common/Breadcrumbs/index.tsx
junwen-k: Hi @ovflowd, I've gathered some insights from other design systems / Breadcrumb implementations.
To be honest, I am not much of an expert in accessibility / SEO in general, so I am not too sure what is the best way to implement an accessible + SEO friendly breadcrumbs, but here might provide you with some ideas:

## Atlassian

[Demo](https://atlassian.design/components/breadcrumbs/examples#long-breadcrumbs)

- Javascript rendering


## Carbon Design System

[Demo](https://react.carbondesignsystem.com/?path=/story/components-breadcrumb--breadcrumb-with-overflow-menu)

- Javascript rendering
- Render popover menu within the breadcrumb item
- Popover `ul` with `aria-label="Overflow menu in a breadcrumb"`

## React Spectrum (Adobe)

[Demo](https://react-spectrum.adobe.com/react-spectrum/Breadcrumbs.html#visible-items-overflow-behavior)

- Javascript rendering
- Additional Breadcrumbs rendered within a React Portal
- `…` breadcrumb item with `aria-label="…"`
- Menu popover with `aria-labelledby` the id of the button that triggers the popup.

## Dell

[Demo](https://vanilla.delldesignsystem.com/2.34.0/iframe.html?viewMode=docs&id=components-breadcrumb--basic#seo-guidelines)

- Hide truncated breadcrumbs with CSS
- `…` breadcrumb item with `aria-label="show 2 more navigation levels"`
- Expands when clicked on `…` breadcrumb item without popover menu

---

I believe there might be more examples out there but this should give some good insights.

Overall, everyone seem to implement it differently. `Dell` even provided some best practices for SEO / Accessibility concern that might worth a look.

Let me know what you think!

components/Common/Breadcrumbs/BreadcrumbRoot/index.tsx
junwen-k: I had to do this ugly hack

```diff
<HomeIcon
+ {...{ itemProp: 'name' as ComponentProps<'svg'> }}
- itemProp="name"
  ...
/>
```

else I am getting compiler error

```
Type '{ itemProp: string; title: string; "aria-label": string; className: string; }' is not assignable to type 'IntrinsicAttributes & Omit<SVGProps<SVGSVGElement>, "ref"> & { title?: string | undefined; titleId?: string | undefined; } & RefAttributes<...>'.
  Property 'itemProp' does not exist on type 'IntrinsicAttributes & Omit<SVGProps<SVGSVGElement>, "ref"> & { title?: string | undefined; titleId?: string | undefined; } & RefAttributes<...>'.ts(2322)
```

Not quite sure why `HomeIcon` type does not support adding `itemProp`.

components/Common/Breadcrumbs/BreadcrumbRoot/index.tsx
```diff
@@ -0,0 +1,56 @@
...
+          <BreadcrumbItem>
+            <LocalizedLink
+              itemScope
+              itemType="http://schema.org/Thing"
+              itemProp="item"
+              itemID="/"
+              href="/"
+            >
+              <HomeIcon
+                {...{ itemProp: 'name' as ComponentProps<'svg'> }}
```
ovflowd: I think what you can do is to do something like:

```
const HeroIconAsSVG = HeroIcon as SVGSVGElement;
```

On the beginning of the file. That should do it. Is dirty but less than this, and then you can add an inline comment saying why this was done like

```
// Apparently the type of HeroIcon doesn't support `itemProp` on their type
// @see https://github.com/tailwindlabs/heroicons/issues/XXXX

and you can open an issue on https://github.com/tailwindlabs/heroicons/issues and link it there

File: components/Common/Breadcrumbs/index.tsx (hunk 19 above)
Comment (ovflowd) [reply components/Common/Breadcrumbs/index.tsx]: Right, I'm not sure if what I mentioned made sense or was well explained.

What I'm trying to say is that we should remove the array slice, and hide the element via CSS.

Otherwise the Breadcrumbs will always be different which will confuse SEO. Because with JavaScript rendering and omitting certain entries they will simply not be in the DOM at all.

And we are only hiding those extra entries for "fitting in the screen purpose". But by doing this way, we're right now having this hacky "position += 1" and other parts of the logic that are honestly not cool at all.

So you can remove all this extra "garbage" code, and remove the array slice and everything. And simply render everything. And for the items you don't want to render (because currentIndex > maxLength) then you do:

```tsx
className({ "tailwind classname for hiding element": index > maxLength })
```

junwen-k: Yes, I understood what you meant in your first comment. I was just wondering if we want to explore further before making a decision as currently there are different implementations to the same issue.

ovflowd: Afaik, the code changes would not change how the component works (as initially described by the issue) and what is rendered and whatnot. I would rather have this simple change be done, and you can create a follow up issue with your concerns and different implementation ideas and we can go through it.

I even wonder if we could somehow "grab/reuse" the logic for conditional `...` that is being introduced on the Pagination component.

We can always iterate this again after getting the first version merged, and then we can verify if we should update the logic/simplify/improve etc

components/Common/Breadcrumbs/BreadcrumbRoot/index.tsx
junwen-k: > Microdata can only be used on [HTML elements](http://www.w3.org/TR/2013/NOTE-microdata-20131029/#html-elements) as defined by HTML5. According to HTML5, the [svg element](https://www.w3.org/TR/2014/REC-html5-20141028/embedded-content-0.html#svg) is not in the HTML namespace.

Apparently, `svg` does not support `itemProp` based on this [answer](https://webmasters.stackexchange.com/questions/56669/using-schema-org-logo-markup-with-svg-images/56670#56670). Even with simple element like

```tsx
<svg
  itemProp="name"
/>
```

Resulting in error

```
Type '{ itemProp: string; }' is not assignable to type 'SVGProps<SVGSVGElement>'.
  Property 'itemProp' does not exist on type 'SVGProps<SVGSVGElement>'.ts(2322)
```

So I think there is a high chance this is not an issue related to HeroIcons.

ovflowd: Let's then revert this and go back to span. Quite interesting. I wonder... why 🤔

junwen-k:
@ovflowd I've did some major changes to the structure of the component (further modularise them so that it is easier to use with a cleaner API).

Regarding to the logic for the rendering of Breadcrumbs... I've tried my best to make the API as clean as possible, however it seems that some "complex" logic still need to be implemented as the truncation is done on the middle part of the array.

Feel free to further improve / iterate the logic once this PR gets merged :)

[Results for Rich Results Test](https://search.google.com/test/rich-results/result?id=U3SZzrOjWh9HfURNmzYaDQ)

ovflowd:
Why disabling microdata for truncated items? Isn't the point that all items have microdata?

Truncated items are ONLY truncated because of stylistic reasons (ie, do not fit in the page) they all should have microdata imo

junwen-k:
> Why disabling microdata for truncated items? Isn't the point that all items have microdata?
> 
> Truncated items are ONLY truncated because of stylistic reasons (ie, do not fit in the page) they all should have microdata imo

I've tried that and the result shows as invalid Breadcrumb, that's why I removed it. Can you show an example of what you describing?

ovflowd:
> > Why disabling microdata for truncated items? Isn't the point that all items have microdata?
> > 
> > Truncated items are ONLY truncated because of stylistic reasons (ie, do not fit in the page) they all should have microdata imo
> 
> I've tried that and the result shows as invalid Breadcrumb, that's why I removed it. Can you show an example of what you describing? 

How invalid? (Can you share the errors?) It depends how you are hiding the items 👀

junwen-k:
> Why disabling microdata for truncated items? Isn't the point that all items have microdata?
> 
> Truncated items are ONLY truncated because of stylistic reasons (ie, do not fit in the page) they all should have microdata imo

I think you might misunderstood the component. The microdata is only disabled for that "..." Breadcrumb item, which should not be considered as an actual link.

You may check the attached result as that was copied directly from the rendered markup.

components/Common/Breadcrumbs/index.tsx
```diff
@@ -0,0 +1,68 @@
...
+  maxLength?: number;
+  hideHome?: boolean;
+};
+
+const Breadcrumbs: FC<BreadcrumbsProps> = ({
+  links = [],
+  maxLength = 5,
+  hideHome = false,
+}) => {
+  const totalLength = links.length + +!hideHome;
```
ovflowd, line 26: Huh interesting transforming boolean into a number, cool trick

ovflowd:
> I think you might misunderstood the component. The microdata is only disabled for that "..." Breadcrumb item, which should not be considered as an actual link.
> 
> You may check the attached result as that was copied directly from the rendered markup.

Oh!!! Right, you are right.

ovflowd (approved):
Solid work. Honestly speaking, tremendous work done here and extremely clean.

We're good to go!

shanpriyan (commented):
LGTM! 🚀

shanpriyan (approved):
LGTM! 🚀

PR #5954, opened 2023-10-02
style: Added `Anchor` styling

byt3h3ad:
## Description
Added styling for `Anchor` tag.
light mode

dark mode

## Related Issues
Addresses #5933 

### Check List

### DIFF
--- a/components/__design__/text.stories.tsx
+++ b/components/__design__/text.stories.tsx
@@ -1,6 +1,6 @@
 import type { Meta as MetaObj, StoryObj } from '@storybook/react';
 
-export const Text: StoryObj = {
+export const Anchor: StoryObj = {
   render: () => (
     <div>
       <p>
--- a/components/__design__/text.stories.tsx
+++ b/components/__design__/text.stories.tsx
@@ -1,0 +1,25 @@
+import type { Meta as MetaObj, StoryObj } from '@storybook/react';
+import '/styles/new/base.css'; // eslint-disable-line
+
+export const Text: StoryObj = {
+  render: () => (
+    <div>
+      <p>
+        The current Node.js security policy can be found at{' '}
+        <a href="https://github.com/nodejs/node/blob/HEAD/SECURITY.md#security">
+          https://github.com/nodejs/node/blob/HEAD/SECURITY.md#security
+        </a>
+        , including information on how to report a vulnerability in Node.js.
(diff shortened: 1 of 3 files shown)


components/__design__/text.stories.tsx
```diff
@@ -0,0 +1,25 @@
+import type { Meta as MetaObj, StoryObj } from '@storybook/react';
+import '/styles/new/base.css'; // eslint-disable-line
```
ovflowd: There's no need to import this here, you probably want to add `@import 'base.css'` within `styles/new/index.css` before the `@tailwind` statements.

byt3h3ad: ah thank you, that's exactly what I wanted to do. made the changes :)

components/__design__/text.stories.tsx
```diff
@@ -0,0 +1,24 @@
+import type { Meta as MetaObj, StoryObj } from '@storybook/react';
+
+export const Text: StoryObj = {
```
ovflowd: Let's rename this export as `Anchor` as we will add other stories in this file.

byt3h3ad: renamed!

ovflowd (approved):
LGTM! :shipit: thank you for your first contribution :)

Appreciate it!

byt3h3ad:
thank you! looking forward to more!

styles/new/base.css
ovflowd, line 1: FYI on Light theme the color is not applying.

image

styles/new/base.css
```diff
@@ -0,0 +1,4 @@
+a {
+  @apply text-green-600
+  dark:text-green-400;
```
ovflowd:
```suggestion
    dark:text-green-400;
```

styles/new/base.css
ovflowd: I think this is happening because, the 'base.css' import is happening before `@tailwind`

Can you move the two imports (mixins and base) after the `@tailwind statement` and see if it works alright?

Also check if by any chance Stylelint complains anything.

byt3h3ad: found the bug, it was because `a:link` and `a:visited` were not specified. should work alright now, pushing the fix.

byt3h3ad: running `npm run format` will catch any and all lint errors right?

PR #5957, opened 2023-10-03
Style: Add unordered and ordered list styling

danielselga:
## Description
This pull request introduces a visual enhancement to the styling of unordered `<ul/>` and ordered `<ol/>` HTML list elements based on the new Figma design, specifically designed for the pagelink component.

## Validation
Screenshot

## Related Issues
Implements #5935

### Check List


styles/new/base.css
```diff
@@ -0,0 +1,19 @@
+@tailwind base;
```
ovflowd: These should not be moved to base.css

styles/new/base.css
```diff
@@ -0,0 +1,19 @@
...
+@tailwind components;
+@tailwind utilities;
+
+ul {
+  @apply list-disc font-open-sans font-regular text-neutral-900
```
ovflowd: We usually have one token per line meaning:

```css
  @apply list-disc
    font-open-sans
    font-regular
    text-neutral-900
    ...
```

ovflowd: Also no need to add `font-open-sans` it is already the default font

styles/new/base.css
```diff
@@ -0,0 +1,19 @@
...
+    dark:text-white;
+
+  font-style: normal;
```
ovflowd: Please use Tailwind for these. There are Tailwind tokens for font-style and line-height (ie: https://tailwindcss.com/docs/line-height)

components/__design__/list.stories.tsx
```diff
@@ -0,0 +1,55 @@
+import type { Meta as MetaObj, StoryObj } from '@storybook/react';
+import type { CSSProperties } from 'react';
+
+export const Lists: StoryObj = {};
+
+const listContainer = {
+  padding: '1.5rem',
+} as CSSProperties;
+
+const listDarkContainer = {
```
ovflowd: You don't need to do this. Storybook has dark theme by default, you can simply switch from dark to light by pressing a button...

components/__design__/list.stories.tsx
```diff
@@ -0,0 +1,55 @@
+import type { Meta as MetaObj, StoryObj } from '@storybook/react';
+import type { CSSProperties } from 'react';
+
+export const Lists: StoryObj = {};
+
+const listContainer = {
```
ovflowd: No need to style this

components/__design__/list.stories.tsx
ovflowd, line 1: Instead of having one "export default", you should actually have something like:

```tsx
export const UnorderedLists: StoryObj = {
  render: () => .... // code for unordered lists
};

export const OrderedLists: StoryObj = {
  render: () => .... // code for ordered lists
};

export default { title: 'Design System/Lists' };
```

danielselga: Fixed!

styles/new/base.css
danielselga: Applyed.

styles/new/base.css
danielselga: @ovflowd There is a conflict here, if I keep the `@tailwind base;` inside of the index.css there is a property that set the `<ul>` and `<ol>`  to `list-style: none;`, since the import is before the './base' file is not overriding the current default value making impossible to keep tha changes inside the base.css file.

The only way I could solve it its to keep the declaration inside of index.css but this goes agains this comment: https://github.com/nodejs/nodejs.org/issues/5935#issuecomment-1742123787, the question is: how should I proceed?

byt3h3ad: importing `base` after `@tailwind base` will most probably fix this.

ovflowd: Exactly as @byt3h3ad mentioned.

danielselga: When I implement the suggestion that you guys are proposing this is the undesirable output that I have, If I move the `@tailwind base;` as before works as expected.

Screenshot from 2023-10-03 15-17-59
Screenshot from 2023-10-03 15-17-32

danielselga: And also the StyleLint triggers an error: `Unexpected invalid position @import rule (no-invalid-position-at-import-rule)Stylelint[no-invalid-position-at-import-rule](https://stylelint.io/user-guide/rules/no-invalid-position-at-import-rule)`

ovflowd: That's because you should move both imports after `@tailwind`

styles/new/base.css
```diff
@@ -0,0 +1,17 @@
+ul {
+  @apply list-disc
+    font-regular
+    not-italic
+    leading-6
+  text-neutral-900
```
ovflowd: Missing indentation here and down there

styles/new/base.css
ovflowd: I'm 100% sure that just moving the imports after `@tailwind` solves your issue, that's essential CSS. CSS follows rules based on a set of priorities. Also, another PR faced the same, and they fixed it by doing this. So you might want to investigate further.

danielselga: I've tried in the way as you suggest, the output is wrong and the linter is blocking me.
Screenshot from 2023-10-03 15-59-58

Even if I try to force the linter rule the style on the tailwind base persists.
Screenshot from 2023-10-03 16-01-52

ovflowd: Interesting, I've identified the issue. And pushed a hotfix. Can you rebase your branch?

byt3h3ad: @ovflowd what is the difference between `@import tailwind/` and `@tailwind`?

components/__design__/list.stories.tsx
ovflowd: Don't forget to remove this `listContainerStyle` (no real need of having this afaik)

components/__design__/list.stories.tsx
```diff
@@ -0,0 +1,36 @@
...
+          <li>with numbers</li>
+        </ol>
+      </div>
+    );
+  },
+};
+
+export const UnorderedLists: StoryObj = {
+  render: () => {
+    return (
```
ovflowd: No need of return statements and `{}` you can remove them :). (On the one above too)

components/__design__/list.stories.tsx
```diff
@@ -0,0 +1,36 @@
+import type { Meta as MetaObj, StoryObj } from '@storybook/react';
+import type { CSSProperties } from 'react';
+
+const listContainer = {
+  padding: '1.5rem',
+} as CSSProperties;
```
ovflowd:
```suggestion
} satisfies CSSProperties;
```

Technically speaking this object is **not** `CSSProperties` but it satisfies it.

components/__design__/list.stories.tsx
danielselga: Removed

components/__design__/list.stories.tsx
danielselga: Fixed

components/__design__/list.stories.tsx
danielselga: Perfect and updated!

ovflowd (approved):
LGTM! Thank you for your first contribution! And for bearing with me with all my feedback.

:shipit:

danielselga:
> LGTM! Thank you for your first contribution! And for bearing with me with all my feedback.
> 
> :shipit:

Thanks, I apreciate all the help :+1:

styles/new/base.css
ovflowd: `@tailwind` hooks directly into PostCSS engine, as `@tailwind` is a custom PostCSS plugin, this usually has faster loading and building of CSS and has some optimizations.

But for us at least the performance impact isn't noticeable. The regular import also imports the whole Tailwind styles instead of invoking Tailwind's plugin to simply add what we need (it prescans the codebase); The Tailwind PostCSS plugin will still do transformations and afterwards remove what we don't use (with regular imports)

shanpriyan (approved):
Looks good to me !

:shipit:

ovflowd:
Hmm, could you rebase your branch and I don't think we need font-weight and font-style attributes here, you can remove them. Also if you could add left padding of around 20px on both the `ul` and `ol` elements, that'd be great.

bmuenzenmeyer:
I could have been more clear in https://github.com/nodejs/nodejs.org/issues/5935 - but I'd like to see nested lists too. It would help validate @ovflowd's suggestion about left-padding

danielselga:
> Hmm, could you rebase your branch and I don't think we need font-weight and font-style attributes here, you can remove them. Also if you could add left padding of around 20px on both the `ul` and `ol` elements, that'd be great.

I pushed your requests.

ovflowd (approved):
LGTM!

Looks good :)

ovflowd:
Oh I didn't notice @mikeesto already ran the CI 🙈 

@danielselga thank you for your hard work and contributions! I appreciate your contribution 🙇

aymen94 (approved):

PR #5974, opened 2023-10-09
fix: Fedora/RHEL section for installing via Package Manager

ghost:
I was the one who updated this section last time, sadly it seems that NodeJS is NOT available as a module anymore and hasn't been for quite a while.

So, it should be directed to the `nodesource` installation that uses `yum`. This method still works and I've tested this with NodeJS v16, v18, and v20.

Old instructions no longer work:
image

### DIFF
--- a/pages/en/download/package-manager.md
+++ b/pages/en/download/package-manager.md
@@ -71,24 +71,7 @@ pacman -S nodejs npm
 
 ## CentOS, Fedora and Red Hat Enterprise Linux
 
-Node.js is available as a module called `nodejs` in CentOS/RHEL 8 and Fedora.
-
-```bash
-dnf module install nodejs:<stream>
-```
-
-where `<stream>` corresponds to the major version of Node.js.
-To see a list of available streams:
-
-```bash
-dnf module list nodejs
-```
-
-For example, to install Node.js 18:
-
-```bash
-dnf module install nodejs:18/common
-```
+Follow the [instructions here](https://github.com/nodesource/distributions#installation-instructions-1) on installing NodejS for Fedora/RHEL.
 
 ### Alternatives
 
--- a/pages/en/download/package-manager.md
+++ b/pages/en/download/package-manager.md
@@ -78,7 +78,6 @@ Follow the [instructions here](https://github.com/nodesource/distributions#insta
 These resources provide packages compatible with CentOS, Fedora, and RHEL.
 
 - [Node.js snaps](#snap) maintained and supported at https://github.com/nodejs/snap
-- [Node.js binary distributions](#debian-and-ubuntu-based-linux-distributions) maintained and supported by [NodeSource](https://github.com/nodesource/distributions)
 
 ## Debian and Ubuntu based Linux distributions
 


mikeesto:
On Fedora 38 I can see some node packages in the repos:

```
nodejs16.i686 : JavaScript runtime
nodejs16.x86_64 : JavaScript runtime
nodejs16-devel.i686 : JavaScript runtime - development headers
nodejs16-devel.x86_64 : JavaScript runtime - development headers
nodejs20.i686 : JavaScript runtime
nodejs20.x86_64 : JavaScript runtime
nodejs20-devel.i686 : JavaScript runtime - development headers
nodejs20-devel.x86_64 : JavaScript runtime - development headers
```
I was able to get node v16 going. v20 seems to have some dependency issues. I have no idea what's happened to v18. I'm not much of a Fedora user, and I'm mindful that we are the first hit on Google when searching for "how to install node on Fedora" - so I'm in favour of whatever helps folks get it running.

mikeesto (commented):
On Fedora 38 I can see some node packages in the repos:

```
nodejs16.i686 : JavaScript runtime
nodejs16.x86_64 : JavaScript runtime
nodejs16-devel.i686 : JavaScript runtime - development headers
nodejs16-devel.x86_64 : JavaScript runtime - development headers
nodejs20.i686 : JavaScript runtime
nodejs20.x86_64 : JavaScript runtime
nodejs20-devel.i686 : JavaScript runtime - development headers
nodejs20-devel.x86_64 : JavaScript runtime - development headers
```
I was able to get node v16 going. v20 seems to have some dependency issues. I have no idea what's happened to v18. I'm not much of a Fedora user though, and I'm mindful that we are the first hit on Google when searching for "How to install Node on Fedora" - so I'm in favour of whatever helps folks get up & running.

If we're going with Nodesource as the recommendation, then I think it should be removed from the alternatives list.

ghost:
I've tested this module list command on Fedora Workstation 36/37/38/39.

I went ahead and removed `nodesource` as an alternative.

sxa (commented):
I'd be -1 on this. At least for me nodejs20 in Fedora (38, aarch64) installs and runs ok:
```
sxa@test-equinix-ubuntu2004-armv8l-02:~$ docker run -it fedora
[root@00dbcbf002d0 /]# yum install nodejs20
[...]

Installed:
  libuv-1:1.46.0-1.fc38.aarch64              nodejs20-1:20.8.0-1.fc38.aarch64                  
  nodejs20-docs-1:20.8.0-1.fc38.noarch       nodejs20-full-i18n-1:20.8.0-1.fc38.aarch64        
  nodejs20-libs-1:20.8.0-1.fc38.aarch64      nodejs20-npm-1:10.1.0-1.20.8.0.1.fc38.aarch64     

Complete!
[root@00dbcbf002d0 /]# node --version
bash: node: command not found
[root@00dbcbf002d0 /]# node-20 --version
v20.8.0
[root@00dbcbf002d0 /]# 
```
What was the dependency issue you had trying to use the nodejs20 package?

~~Although if 18 is missing and that's the example then this definitely needs an update~~
Actually `dnf install nodejs` installs the latest LTS version, which is currently 18.18.0, so yeah we probably just need a doc update if that's the current Fedora policy.
```
[root@00dbcbf002d0 /]# node-18 --version
v18.18.0
[root@00dbcbf002d0 /]# 
```

mhdawson:
I'm -1 as well, we should likely figure out if there is an issue or not, but the modules should be available.

sgallagher:
Hi, Fedora's current Node.js package maintainer here. 

Fedora has decided to decommission the "Modularity" approach to packaging. As a result, I reworked the Node.js packages in Fedora to use a non-modular approach. On each Fedora release (F37, F38, F39, etc.) there will be one `nodejs` package providing the default `/usr/bin/node` binary and one or more `nodejsXX` packages, providing other LTS (or future LTS) versions of Node.js for these releases. (Note that since our development schedules are not perfectly aligned, the default Node.js version for a Fedora release may not yet be in LTS mode when that Fedora release hits GA).

For example, in Fedora 37 and 38, Node.js 18 is the default version, so that is packaged as `nodejs`, whereas in the upcoming Fedora 39 and 40, it will be Node.js 20 for the default. Upgrades should be handled appropriately as well.

The installation layouts of these versions on the Fedora system are isolated from one another, allowing for parallel-installation, but that comes at a cost: only one package can provide `/usr/bin/node` at a time, so all non-default version binaries have a `-XX` suffix (e.g. `/usr/bin/node-16`). If you are creating an OCI/Docker container derived from Fedora (e.g. `FROM fedora:38`) and want to use a Node.js 16 instead of 18, you will need to create a symlink for `/usr/bin/node -> node-16`.


RHEL 8 and RHEL 9 (and the CentOS Stream equivalents) are still using Modules and they should remain available for the lifetime of those releases for existing module streams, though it's possible they will adapt to the Fedora approach at some point down the road.

sxa:
Thanks @sgallagher for clarifying, and that seems in line with what I noticed - that 18 was special because it was the latest LTS. Does that mean 18 will go back to being `nodejs18` when the next LTS shows up or is the "default" version in the `nodejs` package fixed for the life of any given Fedora distribution?

> RHEL 8 and RHEL 9 (and the CentOS Stream equivalents) are still using Modules and they should remain available for the lifetime of those releases for existing module streams, though it's possible they will adapt to the Fedora approach at some point down the road.

Sounds like the right approach would be to decouple the instructions for RHEL/CS and Fedora at the moment.

mhdawson:
I think this is a key part:

>RHEL 8 and RHEL 9 (and the CentOS Stream equivalents) are still using Modules and they should remain available for the lifetime of those releases for existing module streams, though it's possible they will adapt to the Fedora approach at some point down the road.

So we likely need to split the documentation so there are different instructions. One part for RHEL (and maybe Centos?) and then one for Fedora so that we can have the right instructions for each.

sgallagher:
> Thanks @sgallagher for clarifying, and that seems in line with what I noticed - that 18 was special because it was the latest LTS. Does that mean 18 will go back to being `nodejs18` when the next LTS shows up or is the "default" version in the `nodejs` package fixed for the life of any given Fedora distribution?
> 

The default is fixed for the duration of that Fedora release. Each Fedora release hits EOL in approximately 13 months, so it really wouldn't make sense to try to upgrade to a newer LTS within that timeframe.

> > RHEL 8 and RHEL 9 (and the CentOS Stream equivalents) are still using Modules and they should remain available for the lifetime of those releases for existing module streams, though it's possible they will adapt to the Fedora approach at some point down the road.
> 
> Sounds like the right approach would be to decouple the instructions for RHEL/CS and Fedora at the moment.

I think I agree.

mhdawson:
I think a good next step would be to rename the section to be just  `## CentOS and Red Hat Enterprise Linux` instead of removing and then the next step would be to add a new section for Fedora.

@SudoVanilla would you like to update your PR to rename the section instead of removing it?

ghost:
> I think a good next step would be to rename the section to be just `## CentOS and Red Hat Enterprise Linux` instead of removing and then the next step would be to add a new section for Fedora.
> 
> @SudoVanilla would you like to update your PR to rename the section instead of removing it?

Could replace "CentOS" with "Rocky Linux" instead.

sxa:
> Could replace "CentOS" with "Rocky Linux" instead. 

I think "RHEL and compatible distributions" or similar would be best to avoid having an exhaustive list, especially since "CentOS" as it was is no longer maintained to modern versions, replaced by CentOS stream.
We should definitely expedite having the Fedora instructions in there too though, even if it's not in this PR

mhdawson:
I agree that `RHEL and compatible distributions` is probably what makes the most sense.

PR #5979, opened 2023-10-11
rename list and logos stories

bmuenzenmeyer:
## Description

Renamed two stories to better group the storybook design system.

Especially in Chromatic, the way they are currently named caused a strange grouping:

image
Only two

image
Others?

###

Before

image

After

image

One could argue that it looks better BEFORE, but I feel like it's oddly inconsistent with the other Design System stories and I'd rather we pick one pattern and go with it. I suppose, the crux of the problem is chormatics rendering, not local.

Chromatic... https://www.chromatic.com/library?appId=64c7d71358830e9105808652&branch=stories-fix

Is this better? Lemme know in the reviews - I won't be offended with "no" 😄 

### Check List

### DIFF
--- a/components/__design__/list.stories.tsx
+++ b/components/__design__/list.stories.tsx
@@ -24,4 +24,4 @@ export const UnorderedLists: StoryObj = {
   ),
 };
 
-export default { title: 'Design System/Lists' } as MetaObj;
+export default { title: 'Design System' } as MetaObj;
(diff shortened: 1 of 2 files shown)


mikeesto:
To be honest I think I always open the Storybook directly. This is the first time I've really looked closely at Chromatic's library page. I take your point though, it is a bit weird how it is currently grouped. I guess I don't have a strong opinion either way. I'll leave it to those who are using Chromatic in this way

ovflowd (approved):

AugustinMauroy (approved):
LGTM !

PR #5982, opened 2023-10-11
Component/language drop down

### LINKED ISSUE #5981: Create `LanguageDropDown` component
Create a `LanguageDropDown` component using the new [figma design (direct link)](https://www.figma.com/file/pu1vZPqNIM7BePd6W8APA5/Node.js?type=design&node-id=695-15031&mode=dev).  

This component is an icon-button that when activated shows a dropdown of all locales. This is the final port of the [LanguageDropDowncomponent recently deleted](https://github.com/nodejs/nodejs.org/pull/5724/files). Much of the functionality is the same to end users, but with different tech.

---

### Before You Start...

---

### Details
image
image

- Create a new directory called `LanguageDropDown` within the [`Common` directory](https://github.com/nodejs/nodejs.org/tree/main/components/Common) that contains all the new code.
- [Install `@radix-ui/react-dropdown-menu`](https://www.radix-ui.com/primitives/docs/components/dropdown-menu) for the accessible foundation and functionality we need.
  - Use `DropdownMenu.Trigger` to place the icon and button
  - The `@heroicons/react` module has the icon `language` available for use.
  - source the languages to display and the current value from the ['@/hooks/useLocale' hook](https://github.com/nodejs/nodejs.org/blob/main/hooks/useLocale.ts)
  - implement a no-op for the click handler
- Add a storybook story file `components/Common/LanguageDropDown/index.stories.tsx` which exercises each of the component's states.
- Delete https://github.com/search?q=repo%3Anodejs%2Fnodejs.org%20useClickOutside&type=code after a quick search of the codebase - this was leftover from the previous design
- The `aria-label` must be internationalized, and in English state `Choose Language`. Add a new key `components.common.languageDropdown.label` to the `i18n/locales/en.json` file. `useIntl` from `react-intl` should be used to reference the key. Look for examples elsewhere in the codebase or in the old PR.

- Light
- Dark

Rounak-stha:
## Description

This PR creates `LanguageDropDown` component. The figma design is followed and the component is tested visually.
This PR also removes the leftover files from the previus design as mentioned in the issue.

## Validation
Dark Mode:

image

Light Mode:

image

## Related Issues

Addresses #5981 

### Check List


components/Common/LanguageDropDown/index.module.css
```diff
@@ -0,0 +1,34 @@
+.iconWrapper {
+  @apply h-[36px] 
+	  w-[36px] 
+	  rounded-md 
+	  bg-neutral-100 
+	  p-2 
+	  text-neutral-700
+	  dark:bg-neutral-900
+	dark:text-neutral-300;
```
ovflowd: This last one didn't get indented...

components/Common/LanguageDropDown/index.module.css
```diff
@@ -0,0 +1,34 @@
+.iconWrapper {
+  @apply h-[36px] 
+	  w-[36px]
```
ovflowd: I'm almost sure, we already have predefined variables for sizes? @canerakdas I remember you mentioned something about that?

components/Common/LanguageDropDown/index.stories.tsx
```diff
@@ -0,0 +1,12 @@
+import type { Meta as MetaObj, StoryObj } from '@storybook/react';
+
+import LanguageDropDown from './index';
+
+type Story = StoryObj<typeof LanguageDropDown>;
+type Meta = MetaObj<typeof LanguageDropDown>;
+
+export const Default: Story = {
+  args: {},
```
ovflowd: I guess no args is needed here.

components/Common/LanguageDropDown/index.tsx
```diff
@@ -0,0 +1,49 @@
...
+        <DropdownMenu.Content
+          align="start"
+          className={styles.dropDownContent}
+          sideOffset={5}
+        >
+          {availableLocales.map(({ name, code }) => (
+            <DropdownMenu.Item
+              key={name + code}
+              onClick={() => {}}
+              className={[
```
ovflowd: You should use `classNames` package that we have, so you can transform this into:

```tsx
classNames(
  styles.dropDownItem,
  { [styles.currentDropDown]: code === currentLocale.code }
)
```

components/Common/LanguageDropDown/index.tsx
```diff
@@ -0,0 +1,49 @@
...
+
+      <DropdownMenu.Portal>
+        <DropdownMenu.Content
+          align="start"
+          className={styles.dropDownContent}
+          sideOffset={5}
+        >
+          {availableLocales.map(({ name, code }) => (
+            <DropdownMenu.Item
+              key={name + code}
```
ovflowd:
```suggestion
              key={code}
```

Just code should be enough

components/Common/LanguageDropDown/index.tsx
```diff
@@ -0,0 +1,49 @@
...
+              onClick={() => {}}
```
ovflowd: Can you make it a prop?

```tsx
type LanguageDropDownProps = {
  onClick: () => void;
}

const LanguageDropDown: FC<LanguageDropDownProps> = ({ onClick = () => {} }) => {
  ...
}
```

components/Common/LanguageDropDown/index.tsx
```diff
@@ -0,0 +1,49 @@
+import { LanguageIcon } from '@heroicons/react/24/outline';
+import * as DropdownMenu from '@radix-ui/react-dropdown-menu';
+import { useIntl } from 'react-intl';
+
+import { useLocale } from '@/hooks/useLocale';
+
+import styles from './index.module.css';
+
+const DropdownMenuDemo = () => {
```
ovflowd:
```suggestion
const LanguageDropDown = () => {
```

the name of the component should be LanguageDropdown

ovflowd:
image

Does the Radix component support a scroll/max amount of items to render / maxheight?

components/Common/LanguageDropDown/index.module.css
canerakdas: Yes, we can use `w-9` and `h-9` there its equal to  `2.25rem; /* 36px */`. 

@Rounak-stha  We prefer default classes because;

- Allows editing and formatting styles in a one-place
- Tailwind uses [rem for spacing](https://github.com/tailwindlabs/tailwindcss/blob/master/stubs/config.full.js#L815-L850), it would be annoying if this area remains fixed while the rest of the page is updated relative to the browser font size.

components/Common/LanguageDropDown/index.tsx
Rounak-stha: A click handler is required, but shouldn't it be an optional param if we're passing a default value?

Rounak-stha:
> Does the Radix component support a scroll/max amount of items to render / maxheight?

Yes, it does. I've set the max-height and here's what it looks like. 

In Firefox

image
image

In Chrome

image
I'm finding a way to change webkit colors in dark mode for webkit.

`Radix` also ships a [Scroll-area](https://www.radix-ui.com/primitives/docs/components/scroll-area) component if want to use that.

NOTE: The scrollbar width will be included in the width of the container.

ovflowd:
> I'm finding a way to change webkit colors in dark mode for webkit.

I think that is done with color-scheme. Not sure if Storybook is respecting that yet; But we also haven't configured it yet I supposed (on our base-styles) (We might want to add `color-scheme: dark` for when the `[data-theme="dark"]` global style being applied.

components/Common/LanguageDropDown/index.tsx
```diff
@@ -0,0 +1,49 @@
...
+          {availableLocales.map(({ name, code }) => (
+            <DropdownMenu.Item
+              key={name + code}
+              onClick={() => {}}
+              className={[
+                styles.dropDownItem,
+                code === currentLocale.code && styles.currentDropDown,
+              ].join(' ')}
+            >
+              <span className="text-sm">{name}</span>
```
canerakdas: Since using span here and the default value of the span element display property is `inline`, this makes the height of the element higher than it should be. It would be more efficient to give the font properties of this element from a parent element and remove unnecessary span. Additionally, when I check Figma, can you add the `font-medium` class with font-weight set to 500

Rounak-stha:
> I think that is done with color-scheme. Not sure if Storybook is respecting that yet; But we also haven't configured it yet I supposed (on our base-styles) (We might want to add `color-scheme: dark` for when the `[data-theme="dark"]` global style being applied.

I am trying this but it's not working. The media query for `preferred-color-scheme: dark` always fires as it's the default setting in my machine even when settng the color-scheme.

ovflowd:
> > I think that is done with color-scheme. Not sure if Storybook is respecting that yet; But we also haven't configured it yet I supposed (on our base-styles) (We might want to add `color-scheme: dark` for when the `[data-theme="dark"]` global style being applied.
> 
> I am trying this but it's not working. The media query for `preferred-color-scheme: dark` always fires as it's the default setting in my machine even when settng the color-scheme.

Don't worry about that. That is a Storybook-specific issue. That happens because Storybook itself, sets itself (its own theme) to what matches your system, so it actually overrides whatever we set as color-scheme.

We might be able to disable that on Storybook and I can investigate that, but for now. Don't worry about it not working on your Storybook. You can check the behaviour, if you inspect the iframe HTML element and remove the "color-scheme" that Storybook injects.

Rounak-stha:
I am using the `theme()` function to use tailwind colors but it results in error when linting (`npx turbo lint`) and formatting (`npx turbo format`) . 
Any suggestions?

ovflowd:
> I am using the `theme()` function to use tailwind colors but it results in error when linting (`npx turbo lint`) and formatting (`npx turbo format`) . Any suggestions?

Can you show how are you using it?

Rounak-stha:
> Can you show how are you using it?

Using `color-scheme` did not work for me even outside storybook.
Instead, I did this.

```css
[data-theme='dark'] .dropDownContent {
  --language-drop-scrollbar-bg: theme('colors.neutral.900');
  --language-drop-scrollthumb-bg: theme('colors.neutral.800');
}

[data-theme='dark'] .scrollbar {
  scrollbar-color: var(--language-drop-scrollthumb-bg)
  var(--language-drop-scrollbar-bg);
}
```

ovflowd:
> > Can you show how are you using it?
> 
> Using `color-scheme` did not work for me even outside storybook. Instead, I did this.
> 
> ```css
> [data-theme='dark'] .dropDownContent {
>   --language-drop-scrollbar-bg: theme('colors.neutral.900');
>   --language-drop-scrollthumb-bg: theme('colors.neutral.800');
> }
> 
> [data-theme='dark'] .scrollbar {
>   scrollbar-color: var(--language-drop-scrollthumb-bg)
>   var(--language-drop-scrollbar-bg);
> }
> ```

Why are you trying to manually style the scrollbar?

Rounak-stha:
> Why are you trying to manually style the scrollbar?

because the default is ugly
image

ovflowd:
> > Why are you trying to manually style the scrollbar?
> 
> because the default is ugly image

It changes browser by browser OS by OS. It's not part of our styling to style the scroll wheel.

components/Common/LanguageDropDown/index.tsx
Rounak-stha: Hey, I updated the component to receive the prop.  Currently, it's optional cause we're supplying a default value but please let me know if this is not what's expected.

ovflowd: Right, will re-review 🙇

ovflowd:
@Rounak-stha can you rebase your PR? 👀

Rounak-stha:
> @Rounak-stha can you rebase your PR? 👀

Hey, just did the rebase. 🤞

components/Common/LanguageDropDown/index.module.css
```diff
@@ -0,0 +1,39 @@
+.iconWrapper {
+  @apply h-9
+	w-9
```
ovflowd: Hmm, the indentation is wrong here again 👀  (on all these lines)

components/Common/LanguageDropDown/index.module.css
```diff
@@ -0,0 +1,39 @@
...
+	bg-neutral-100 
+	p-2 
+	text-neutral-700
+	dark:bg-neutral-900
+	dark:text-neutral-300;
+}
+
+.dropDownContent {
+  @apply max-h-80
+	w-[184px]
```
ovflowd: Is this manual value needed? Wouldn't something like w-20 or larger work here? Could you try out?

ovflowd (approved):
Everything else besides of those two minor things look ok :)

bmuenzenmeyer:
@ovflowd what do you think of wrapping the button in https://www.radix-ui.com/primitives/docs/utilities/accessible-icon ? I noticed this utility just now when creating the `ThemeToggle` component story

components/Common/LanguageDropDown/index.module.css
Rounak-stha: The width of the component in Figma is `184px`. Tailwind does not have a class that exactly fit this value. The 2 closest one to this are `w-44 = 176px` and `w-48 = 192px`.

image

components/Common/LanguageDropDown/index.module.css
Rounak-stha: Hey, the indentation is correct in my local branch which I pushed.

image

I am confused about what's happening.

aymen94 (approved):

ovflowd:
> @ovflowd what do you think of wrapping the button in [radix-ui.com/primitives/docs/utilities/accessible-icon](https://www.radix-ui.com/primitives/docs/utilities/accessible-icon) ? I noticed this utility just now when creating the `ThemeToggle` component story

Shouldn't the Icon be wrapped? Not the button? Or whatÄs the difference?

components/Common/LanguageDropDown/index.module.css
ovflowd: I have the feeling you used tabs here instead of spacing. Try to use just two spaces.

components/Common/LanguageDropDown/index.module.css
ovflowd: I'd say let's till with w-48 then.

components/Common/LanguageDropDown/index.module.css
Rounak-stha: Hmm, just viewed the prettierrc file and saw that the indentation with spaces is preferred but the css file uses tabs instead. with tab size of 2.
Viewing the file on github, it uses tab size of 2
image

But locally in my setup, its 4.

Rounak-stha: > I have the feeling you used tabs here instead of spacing. Try to use just two spaces.

Sure

Rounak-stha:
> @ovflowd what do you think of wrapping the button in https://www.radix-ui.com/primitives/docs/utilities/accessible-icon ? I noticed this utility just now when creating the `ThemeToggle` component story

Isn't the dropdown trigger button already accessible?

ovflowd:
> > @ovflowd what do you think of wrapping the button in [radix-ui.com/primitives/docs/utilities/accessible-icon](https://www.radix-ui.com/primitives/docs/utilities/accessible-icon) ? I noticed this utility just now when creating the `ThemeToggle` component story
> 
> Isn't the dropdown trigger button already accessible?

Yeah, I think so... @bmuenzenmeyer?

bmuenzenmeyer:
Keyboard, sure. It mentions screen reader a11y too. I can test if my copy of JAWS still works.

We can consider this non-blocking for this PR and I will look to address globally in another issue.

PR #6007, opened 2023-10-15
feat(themeToggle): introduce

### LINKED ISSUE #6005: Create `ThemeToggle` Component
Create a `ThemeToggle` component using the new [figma design (indirect link to NavBar)](https://www.figma.com/file/pu1vZPqNIM7BePd6W8APA5/Node.js?type=design&node-id=415-8491&mode=dev).  

This component is an icon-button that when activated toggles the current theme (light to dark, or dark to light). This is the final port of the [DarkModeToggle recently deleted](https://github.com/nodejs/nodejs.org/pull/5730). Much of the functionality is the same to end users, but with different tech.

---

### Before You Start...

---

### Details

image

> showing the whole navbar . The icon is only this icon button

image

- Create a new directory called `ThemeToggle` within the [`Common` directory](https://github.com/nodejs/nodejs.org/tree/main/components/Common) that contains all the new code.
- [Install `@radix-ui/react-accessible-icon`](https://www.radix-ui.com/primitives/docs/utilities/accessible-icon) for the accessible foundation and functionality we need.
- The `@heroicons/react` module has the icons `moon` and `sun` available for use.
- Add a storybook story file `components/Common/ThemeToggle/index.stories.tsx` which exercises each of the component's states.
- The [`label` prop](https://www.radix-ui.com/primitives/docs/utilities/accessible-icon#root) must be internationalized, and use the [existing key `components.header.buttons.toggleDarkMode`](https://github.com/nodejs/nodejs.org/blob/main/i18n/locales/en.json#L39) in the `i18n/locales/en.json` file. `useIntl` from `react-intl` should be used to reference the key. Look for examples elsewhere in the codebase or in the old PR.
- The component should set its default mode to the user's system setting. Tailwind [offers some guidance](https://tailwindcss.com/docs/dark-mode).

- Light Mode, which shows the `moon` icon to toggle to dark mode
- Dark Mode, which shows the `sun` icon to toggle to light mode

AugustinMauroy:
## Description

Introduction of `ThemeToggle` component.

I doesn't have add any padding or margin on style because I assume this styles will be cover by header's styles.

## Validation

* It's should be same as [figma design](https://www.figma.com/file/pu1vZPqNIM7BePd6W8APA5/Node.js?type=design&node-id=415-8491&mode=dev).
* it's should toggle dark mode when click on button. And switch the icon.
* unit test should pass and be suffisant

## Related Issues

#6005 

### Check List

### DIFF
--- a/.storybook/preview.tsx
+++ b/.storybook/preview.tsx
@@ -4,6 +4,7 @@ import { withThemeByDataAttribute } from '@storybook/addon-themes';
 import { SiteProvider } from '../providers/siteProvider';
 import { LocaleProvider } from '../providers/localeProvider';
 import { NotificationProvider } from '../providers/notificationProvider';
+import { ThemeProvider } from '../providers/themeProvider';
 import * as constants from './constants';
 import type { Preview, ReactRenderer } from '@storybook/react';
 
@@ -31,11 +32,13 @@ const preview: Preview = {
     Story => (
       <SiteProvider>
         <LocaleProvider>
-          <NotificationProvider viewportClassName="absolute top-0 left-0 list-none">
-            <div className={rootClasses}>
-              <Story />
-            </div>
-          </NotificationProvider>
+          <ThemeProvider>
+            <NotificationProvider viewportClassName="absolute top-0 left-0 list-none">
+              <div className={rootClasses}>
+                <Story />
+              </div>
+            </NotificationProvider>
+          </ThemeProvider>
         </LocaleProvider>
       </SiteProvider>
     ),
--- a/.storybook/preview.tsx
+++ b/.storybook/preview.tsx
@@ -4,7 +4,6 @@ import { withThemeByDataAttribute } from '@storybook/addon-themes';
 import { SiteProvider } from '../providers/siteProvider';
 import { LocaleProvider } from '../providers/localeProvider';
 import { NotificationProvider } from '../providers/notificationProvider';
-import { ThemeProvider } from '../providers/themeProvider';
 import * as constants from './constants';
 import type { Preview, ReactRenderer } from '@storybook/react';
 
@@ -32,13 +31,11 @@ const preview: Preview = {
     Story => (
       <SiteProvider>
         <LocaleProvider>
-          <ThemeProvider>
-            <NotificationProvider viewportClassName="absolute top-0 left-0 list-none">
-              <div className={rootClasses}>
-                <Story />
-              </div>
-            </NotificationProvider>
-          </ThemeProvider>
+          <NotificationProvider viewportClassName="absolute top-0 left-0 list-none">
+            <div className={rootClasses}>
+              <Story />
+            </div>
+          </NotificationProvider>
         </LocaleProvider>
       </SiteProvider>
     ),
--- a/components/Common/ThemeToggle/__tests__/index.test.mjs
+++ b/components/Common/ThemeToggle/__tests__/index.test.mjs
@@ -17,7 +17,7 @@ beforeEach(() => {
   mockCurrentTheme = 'light';
   render(
     <LocaleProvider>
-      <ThemeToggle onChange={toggleTheme} />
+      <ThemeToggle onClick={toggleTheme} />
     </LocaleProvider>
   );
   toggle = screen.getByRole('button');
@@ -32,11 +32,6 @@ describe('ThemeToggle component', () => {
 
   it('switches light theme to dark theme', async () => {
     await userEvent.click(toggle);
-    expect(mockCurrentTheme).toBe('dark');
-  });
-
-  it('switch when press enter', async () => {
-    await userEvent.type(toggle, '{enter}');
     expect(mockCurrentTheme).toBe('dark');
   });
 });
--- a/components/Common/ThemeToggle/__tests__/index.test.mjs
+++ b/components/Common/ThemeToggle/__tests__/index.test.mjs
@@ -42,4 +42,12 @@ describe('ThemeToggle component', () => {
     await userEvent.click(toggle);
     expect(mockCurrentTheme).toBe('dark');
   });
+
+  it('switch when press enter', async () => {
+    mockCurrentTheme = 'light';
+    renderThemeToggle();
+    const toggle = screen.getByRole('button');
+    await userEvent.type(toggle, '{enter}');
+    expect(mockCurrentTheme).toBe('dark');
+  });
 });
--- a/components/Common/ThemeToggle/__tests__/index.test.mjs
+++ b/components/Common/ThemeToggle/__tests__/index.test.mjs
@@ -11,19 +11,20 @@ const toggleTheme = () => {
   mockCurrentTheme = mockCurrentTheme === 'light' ? 'dark' : 'light';
 };
 
-let toggle;
+describe('ThemeToggle', () => {
+  let toggle;
 
-beforeEach(() => {
-  mockCurrentTheme = 'light';
-  render(
-    <LocaleProvider>
-      <ThemeToggle onClick={toggleTheme} />
-    </LocaleProvider>
-  );
-  toggle = screen.getByRole('button');
-});
+  beforeEach(() => {
+    mockCurrentTheme = 'light';
 
-describe('ThemeToggle component', () => {
+    render(
+      <LocaleProvider>
+        <ThemeToggle onClick={toggleTheme} />
+      </LocaleProvider>
+    );
+    toggle = screen.getByRole('button');
+  });
+
   it('switches dark theme to light theme', async () => {
     mockCurrentTheme = 'dark';
     await userEvent.click(toggle);
--- a/components/Common/ThemeToggle/__tests__/index.test.mjs
+++ b/components/Common/ThemeToggle/__tests__/index.test.mjs
@@ -5,26 +5,19 @@ import { LocaleProvider } from '@/providers/localeProvider';
 
 import ThemeToggle from '../';
 
+let mockCurrentTheme = 'light';
+
+const toggleTheme = () => {
+  mockCurrentTheme = mockCurrentTheme === 'light' ? 'dark' : 'light';
+};
+
 function renderThemeToggle() {
   return render(
     <LocaleProvider>
-      <ThemeToggle />
+      <ThemeToggle toggleTheme={toggleTheme} />
     </LocaleProvider>
   );
 }
-
-let mockCurrentTheme = 'light';
-
-const mockToggleTheme = jest.fn().mockImplementation(() => {
-  mockCurrentTheme = mockCurrentTheme === 'dark' ? 'light' : 'dark';
-});
-
-// Mock dark mode module for controlling dark mode HOC behaviour
-jest.mock('next-themes', () => ({
-  useTheme: () => {
-    return { theme: mockCurrentTheme, setTheme: mockToggleTheme };
-  },
-}));
 
 describe('ThemeToggle component', () => {
   it('switches dark theme to light theme', async () => {
--- a/components/Common/ThemeToggle/__tests__/index.test.mjs
+++ b/components/Common/ThemeToggle/__tests__/index.test.mjs
@@ -1,0 +1,45 @@
+import { render, screen } from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
+
+import { LocaleProvider } from '@/providers/localeProvider';
+
+import ThemeToggle from '../';
+
+function renderThemeToggle() {
+  return render(
+    <LocaleProvider>
+      <ThemeToggle />
+    </LocaleProvider>
+  );
+}
+
+let mockCurrentTheme = 'light';
+
+const mockToggleTheme = jest.fn().mockImplementation(() => {
+  mockCurrentTheme = mockCurrentTheme === 'dark' ? 'light' : 'dark';
+});
+
+// Mock dark mode module for controlling dark mode HOC behaviour
+jest.mock('next-themes', () => ({
+  useTheme: () => {
+    return { theme: mockCurrentTheme, setTheme: mockToggleTheme };
+  },
+}));
+
+describe('ThemeToggle component', () => {
+  it('switches dark theme to light theme', async () => {
+    mockCurrentTheme = 'dark';
+    renderThemeToggle();
+    const toggle = screen.getByRole('button');
+    await userEvent.click(toggle);
+    expect(mockCurrentTheme).toBe('light');
+  });
+
+  it('switches light theme to dark theme', async () => {
+    mockCurrentTheme = 'light';
+    renderThemeToggle();
+    const toggle = screen.getByRole('button');
+    await userEvent.click(toggle);
+    expect(mockCurrentTheme).toBe('dark');
+  });
+});
--- a/components/Common/ThemeToggle/__tests__/index.test.mjs
+++ b/components/Common/ThemeToggle/__tests__/index.test.mjs
@@ -11,35 +11,31 @@ const toggleTheme = () => {
   mockCurrentTheme = mockCurrentTheme === 'light' ? 'dark' : 'light';
 };
 
-function renderThemeToggle() {
-  return render(
+let toggle;
+
+beforeEach(() => {
+  mockCurrentTheme = 'light';
+  render(
     <LocaleProvider>
-      <ThemeToggle toggleTheme={toggleTheme} />
+      <ThemeToggle onChange={toggleTheme} />
     </LocaleProvider>
   );
-}
+  toggle = screen.getByRole('button');
+});
 
 describe('ThemeToggle component', () => {
   it('switches dark theme to light theme', async () => {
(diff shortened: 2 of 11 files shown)


.storybook/preview.tsx
```diff
@@ -31,11 +32,13 @@ const preview: Preview = {
     Story => (
       <SiteProvider>
         <LocaleProvider>
-          <NotificationProvider viewportClassName="absolute top-0 left-0 list-none">
-            <div className={rootClasses}>
-              <Story />
-            </div>
-          </NotificationProvider>
+          <ThemeProvider>
```
ovflowd: Why are you adding ThemeProvider on Storybook?

components/Common/ThemeToggle/index.module.css
```diff
@@ -0,0 +1,6 @@
+.icon {
+  @apply h-5
+      w-5
```
ovflowd: This indentation is wrong. You added double indentation...

components/Common/ThemeToggle/index.tsx
```diff
@@ -0,0 +1,44 @@
...
+import styles from './index.module.css';
+
+const ThemeToggle: FC = () => {
+  const intl = useIntl();
+  const { theme, setTheme } = useTheme();
+
+  const isDark = theme === 'dark';
+
+  const toggleTheme = (isKeyPress?: boolean) => {
+    if (isKeyPress) {
```
ovflowd: Why? I expect that key presses should also allow the user to toggle the theme, no?

components/Common/ThemeToggle/index.tsx
```diff
@@ -0,0 +1,44 @@
...
+    }
+
+    setTheme(isDark ? 'light' : 'dark');
+  };
+
+  return (
+    <button
+      type="button"
+      onClick={() => toggleTheme()}
+      onKeyDown={() => toggleTheme(true)}
```
ovflowd: Why do you have a function that does nothing if it's key press? What is the point here?

hooks/useNotification 2.ts
ovflowd, line 1: What is this file??

AugustinMauroy: IDK why it's push.But it's bug from iCloud. It's really annoying

ovflowd: iCloud???

.storybook/preview.tsx
AugustinMauroy: if it's not aded icon switch doesn't work

ovflowd: Please remove it. Why would you want the Theme Toggle button actually to toggle the Theme on a Storybook?

The components of Storybook are just to validate the design and interactions. Not things that are side-effects to the Component.

(I know this might not be a trivial question, but we should not test "real world" behavior within Storybook. That's the same reason we render the Notification Component statically, and why the Language Dropdown doesn't really do anything when clicked)

ovflowd: Also, I would really appreciate it if you're more attentive to the history of files and PRs being merged as a **member of this project**.

The ThemeProvider was removed for numerous reasons (including as breaking the Storybook theme functionality, and other things, and mostly because we don't want theme switching from the real App within Storybook)

hooks/useNotification 2.ts
ovflowd: Now I am really curious. What does iCloud has to do with this?

components/Common/ThemeToggle/index.tsx
```diff
@@ -0,0 +1,44 @@
...
+import { AccessibleIcon } from '@radix-ui/react-accessible-icon';
+import { useTheme } from 'next-themes';
+import type { FC } from 'react';
+import { useIntl } from 'react-intl';
+
+import styles from './index.module.css';
+
+const ThemeToggle: FC = () => {
+  const intl = useIntl();
+  const { theme, setTheme } = useTheme();
```
ovflowd: This Component should be generic and not wired up to `useTheme`. 

I would argue that this Component receives one prop:

`{ onClick: () => void `}

components/Common/ThemeToggle/index.tsx
```diff
@@ -0,0 +1,44 @@
...
+      onClick={() => toggleTheme()}
+      onKeyDown={() => toggleTheme(true)}
+    >
+      <AccessibleIcon
+        label={intl.formatMessage({
+          id: 'components.header.buttons.toggleDarkMode',
+        })}
+      >
+        {isDark ? (
+          <SunIcon className={styles.icon} />
```
ovflowd: I think the rendering of these icons should be done via CSS, in the sense that we should always render both but hide the one we're not showing.

```tsx
<SunIcon className="dark:hidden" />
<MoonIcon className="dark:block hiddne" />
```

Also, the styling shouldn't be done in the icon `styles.icon` but on the container. And then use something like:


```css
.className svg {}
```

for the actual SVG styles

components/Common/ThemeToggle/index.tsx
AugustinMauroy: > This is the final port of the https://github.com/nodejs/nodejs.org/pull/5730. Much of the functionality is the same to end users, but with different tech.

So I have base ma work on "old" themeToggle. So why not remove it. I will test when we focus on button en press enter if it's toggle

hooks/useNotification 2.ts
AugustinMauroy: Yeah the things running on my Macbook.
iCloud is sync all of my desktop on cloud. But sometime he duplicates file.

components/Common/ThemeToggle/index.tsx
AugustinMauroy: I agree with you that it would be simpler

components/Common/ThemeToggle/index.tsx
ovflowd: Right, but that just means that the underlying technology is still `next-theme` but the base component should be simple.

components/Common/ThemeToggle/index.tsx
AugustinMauroy: I don't really agree with you, because the component is designed to switch between themes. What you're proposing would be for a "feature toggle" component and you'd have props like that:
```ts
{
onClick: () => void ;
value: number
feature: ReactNode[] | string[]
}

File: components/Common/ThemeToggle/index.tsx (hunk 3 above)
Comment (AugustinMauroy) [reply components/Common/ThemeToggle/index.tsx]: OK go to remove that 👍

File: components/Common/ThemeToggle/index.tsx (hunk 5 above)
Comment (ovflowd) [reply components/Common/ThemeToggle/index.tsx]: Im not proposing a feature toggle. And the component should be generic.

Components in general should be as simple as possible.

It doesn't need any "value" logic.

It's still a theme toggle because it's contextually aware of the dark theme/light theme through Tailwind and has an onClick handler.

File: components/Common/ThemeToggle/index.tsx [hunk 7]
```diff
@@ -0,0 +1,38 @@
...
+      onClick={toggleTheme}
+      onKeyDown={toggleTheme}
+      className={styles.themeToggle}
+    >
+      <AccessibleIcon
+        label={intl.formatMessage({
+          id: 'components.header.buttons.toggleDarkMode',
+        })}
+      >
+        <>
```
AugustinMauroy: this is done because without the fragment it's create error boundary.
It's because radix use `React.cloneElement`
[source code](https://github.com/radix-ui/primitives/blob/c31c97274ff357aea99afe6c01c1c8c58b6356e0/packages/react/accessible-icon/src/AccessibleIcon.tsx#L15-L27)

components/Common/ThemeToggle/index.tsx
ovflowd: Also in no moment I mentioned any of the props you're shown on your comment.

.storybook/preview.tsx
AugustinMauroy: I had removed my bad idea sorry 😅. I had try to make something cool but ...

components/Common/ThemeToggle/index.tsx
AugustinMauroy: I got it !

components/Common/ThemeToggle/index.tsx
```diff
@@ -0,0 +1,36 @@
...
+
+type ThemeToggleProps = {
+  toggleTheme: () => void;
+};
+
+const ThemeToggle: FC<ThemeToggleProps> = ({ toggleTheme }) => {
+  const intl = useIntl();
+
+  return (
+    <button
```
ovflowd, line 19: Hmm, a button? @bmuenzenmeyer does the icon need to be a button?

components/Common/ThemeToggle/index.tsx
ovflowd: No, you're wrong. The issue is that AccessibleIcon should have only a single children.

So you should wrap AccessibleIcon on each one of these Icon Elements. By adding a `<>` you are pretty much destroying all the accessibility benefits this component creates.

components/Common/ThemeToggle/index.tsx
```diff
@@ -0,0 +1,36 @@
+import { MoonIcon, SunIcon } from '@heroicons/react/24/outline';
+import { AccessibleIcon } from '@radix-ui/react-accessible-icon';
+import type { FC } from 'react';
+import { useIntl } from 'react-intl';
+
+import styles from './index.module.css';
+
+type ThemeToggleProps = {
+  toggleTheme: () => void;
```
ovflowd: I think it is incorrect to call this toggleTheme. As this doesn't toggle the theme, hence why I said (and repeating myself again) to call it `onClick`.

Often I don't understand WHY when I ask you to do X, and you agree with X then you just do something completely different from asked?

components/Common/ThemeToggle/index.module.css
```diff
@@ -0,0 +1,8 @@
+.themeToggle {
+  svg {
+    @apply h-5
```
ovflowd: Don't forget the indentation... Please...

components/Common/ThemeToggle/index.tsx
ovflowd: @bmuenzenmeyer based on your comment on the LanguageDropdown PR. Should AccessibleIcon wrap the Button element or each Icon element?

Also on that, should AccessibleIcon handle the onClick or onKeyPresses or what?

components/Common/ThemeToggle/index.tsx
AugustinMauroy: because it's not really onClick it's also onKeydown so I call props toggleTheme

ovflowd: But it's not an `toggleTheme`, this Component does not toggle a theme.

You can just call it `onChange` or `inTeraction` or simply have two different handlers.

one called `onClick` and one called `onKeyDown` as props and simply expose them ;)

AugustinMauroy: > one called onClick and one called onKeyDown as props and simply expose them ;)

i prefer `onChange` it's less code but two props is more clear 👌🏻

components/Common/ThemeToggle/index.module.css
AugustinMauroy: Identation is not really visual on VScode with tailwind extension (it's pushing text when displaying color instead of floating).

components/Common/ThemeToggle/index.tsx
ovflowd: Right, but the thing is that hmm, nothing is really changing, right? And in theory we are "listening" to the user to interact with this Icon. Either by a keypress or a click (hence why I believe exposing two handlers is more appropriate).

Since you're being stubborn here (as usual), @bmuenzenmeyer what's your opinion in here? (if any)

components/Common/ThemeToggle/index.module.css
ovflowd: Right, but Extension has nothing to do with indentation. I guess how the colour can be confusing. But ultimately you can see these on git diffs.

AugustinMauroy: > git diffs

I never take really attention to that . Thank you for opening an eye for me

ovflowd: Of course :) here to help! 🙇

components/Common/ThemeToggle/index.tsx
bmuenzenmeyer: Sorry I didnt see this until now. My notifications were messed up somehow.

Let's look inward to help.  We are internally inconsistent about this. A non-exhaustive check of our current new components:

- [Select calls it `onChange`](https://github.com/nodejs/nodejs.org/blob/main/components/Common/Select/index.tsx#L33)
- [Notification calls it `onChange`](https://github.com/nodejs/nodejs.org/blob/main/components/Common/Notification/index.tsx#L10)
- [LanguageDropdown calls it `onClick`](https://github.com/nodejs/nodejs.org/blob/main/components/Common/LanguageDropDown/index.tsx#L16)

If it were me (one opinion), I'd probably call the function prop `onChange` or `handleChange`...but only wire it up to the `onClick` handler of the button. Button's being interactive elements, they should accept keyboard input automatically right? The tests could prove that. Also, `onkeydown` could spam events if someone holds it down right?

components/Common/ThemeToggle/index.tsx
bmuenzenmeyer: JAWS currently announces https://64c7d71358830e9105808652-bhljmtpybt.chromatic.com/?path=/story/common-themetoggle--default correctly

components/Common/ThemeToggle/__tests__/index.test.mjs
```diff
@@ -0,0 +1,46 @@
...
+import ThemeToggle from '../';
+
+let mockCurrentTheme = 'light';
+
+const toggleTheme = () => {
+  mockCurrentTheme = mockCurrentTheme === 'light' ? 'dark' : 'light';
+};
+
+function renderThemeToggle() {
+  return render(
```
araujogui: You can put `renderThemeToggle` inside a `beforeEach` callback. DRY

araujogui (dismissed):

components/Common/ThemeToggle/index.tsx
AugustinMauroy: >  Also, onkeydown could spam events if someone holds it down right?

Yes, so I can wrap `onChange` with an timeout ?

components/Common/ThemeToggle/__tests__/index.test.mjs
AugustinMauroy: Done !

components/Common/ThemeToggle/index.tsx
```diff
@@ -0,0 +1,40 @@
...
+  const intl = useIntl();
+
+  return (
+    <button
+      type="button"
+      onClick={onChange}
+      onKeyDown={onChange}
+      className={styles.themeToggle}
+    >
+      <AccessibleIcon
```
ovflowd: @bmuenzenmeyer just double checking, is `AccessibleIcon` needed for Buttons? (Just double-checking)...

Should we even use `onKeyDown`?

components/Common/ThemeToggle/index.tsx
ovflowd: As @bmuenzenmeyer said, we don't need an onKeyDown here, Augustin. accessible interactions are already handled by the `button` element.

Hence this should be called `onClick`.  Can you please just do as you're being asked?
And no, we shouldn't have a timeout, and no, please stop over-complicating this so simple PR!

components/Common/ThemeToggle/index.tsx
bmuenzenmeyer: perhaps not! I am defintely interested in their use in the footer icons however. I can test JAWS again in the morning.

bmuenzenmeyer:
I think this is nearly read - @AugustinMauroy do you want to convert it from draft to a full pull request?

AugustinMauroy:
I'm sorry @bmuenzenmeyer I din't get you message.

bmuenzenmeyer:
> I'm sorry @bmuenzenmeyer I din't get you message.

@AugustinMauroy I mean that if you are comfortable with the code now, to click this button

image

At least in my usual workflow, I don't look at PRs deeply until they are out of `Draft` or someone explicitly asks me to

ovflowd:
> > I'm sorry @bmuenzenmeyer I din't get you message.
> 
> 
> 
> @AugustinMauroy I mean that if you are comfortable with the code now, to click this button
> 
> 
> 
> image
> 
> 
> 
> At least in my usual workflow, I don't look at PRs deeply until they are out of `Draft` or someone explicitly asks me to

Right, I think this PR can be marked as ready but I still have open comments that need to be addressed.

ovflowd:
@AugustinMauroy can you please finish this PR if you have time? It's blocking us from proceeding with other issues. Otherwise I invite any other collaborator (cc @nodejs/nodejs-website) to take over the PR, and do the last requested changes.

ovflowd (approved):

PR #6014, opened 2023-10-16
feat(Preview): add Preview component

### LINKED ISSUE #6011: Create `Preview` Component
Create a `Preview` component using the [figma design (direct link)](https://www.figma.com/file/pu1vZPqNIM7BePd6W8APA5/Node.js?type=design&node-id=695-2355&mode=dev).  

This component that is a themed image with text that re-use in blogs and OpenGraph.

---

### Before You Start...

---

### Details

image

- Create a new directory called `Preview` within the [`Common` directory](https://github.com/nodejs/nodejs.org/tree/main/components/Common) that contains all the new code.
- Use the existing [js-white SVG](https://github.com/nodejs/nodejs.org/blob/main/public/static/images/logos/js-white.svg)
- Add the [Frame, Group 3, Group 1 hexagons from the Figma](https://www.figma.com/file/pu1vZPqNIM7BePd6W8APA5/Node.js?type=design&node-id=415-6717&mode=design&t=gghCgmQfCIVIEoO5-4) as a new SVG, not the Ellipse
- Use our tailwind tokens to help style the blurred effect. The Figma suggests the CSS if you need help
  - Announcement: Green/700
  - Release: Info/600
  - Vulnerability: Warning/600

#### Suggested Props

| Name | type | comments |
| - | - | - |
`title` | `string`  | example: `Changing the End-of-Life Date for Node.js 16 to September 11th, 2023` per Figma
`type` | enum `announcement` \| `release` \| `vulnerability` | default is `announcement`

#### Storybook

Add a storybook story file `components/Common/Preview/index.stories.tsx` which exercises each of the API's states.

- Default (Announcement)
- Announcement
- Release
- Vulnerability

junwen-k:
## Description

Add `Preview` component which renders the OpenGraph Image.

## Validation

Added stories `Preview` component.

### Announcement 
image

### Release
image

### Vulnerability
image

## Related Issues

Closes #6011

### Check List


ovflowd (approved):
This is such a clean implementation!

Thank you for your contribution! Super appreciate the effort here ✨

components/Common/Preview/index.module.css
```diff
@@ -0,0 +1,47 @@
+.root {
+  @apply relative
+    flex
+    h-[630px]
```
ovflowd: Hmm, I wonder if we could use tailwind predefined sizes here? Like `h-XX`
Maybe `h-80`? (You can keep increasing the numbers, and with the Tailwind VSCode Extension it should give you an approx number).

We also should avoid using `px` 👀

components/Common/Preview/index.module.css
```diff
@@ -0,0 +1,47 @@
...
+  & > .container {
+    @apply z-10
+      mx-auto
+      flex
+      max-w-xl
+      flex-col
+      gap-12
+      text-center
+      text-3xl
+      font-semibold
```
ovflowd, line 39: I wonder if the font-weight here is a little bit different from the Figma ones?

components/Common/Preview/index.module.css
junwen-k: This is actually a good question. The reason this is set to arbitrary `px` value (1200px x 630px) is because this is the [recommended size](https://developers.facebook.com/docs/sharing/webmasters/images/) for OpenGraph Image and I assumed that this component will be used in conjunction with the [ImageResponse API](https://nextjs.org/docs/app/api-reference/file-conventions/metadata/opengraph-image#generate-images-using-code-js-ts-tsx) to dynamically generate the OG image.

However, I am not sure if this component is used solely for OpenGraph only. One potential way is that maybe we can expose props like `width` and `height` so that the caller can control the size. 👀

components/Common/Preview/index.module.css
junwen-k: Hmm, it looks like it is indeed matching the specs from Figma. Maybe for some reason it appears different visually 👀 

image

components/Common/Preview/index.module.css
ovflowd: Right I like the idea of exposing width/height, but in general it would be solely used for OG for now.

junwen-k:
@ovflowd Hi, I've actually went back and updated the ellipse to use radial gradient instead and I think the color looks slightly more natural now. It is not 100% pixel perfect with the original design but I think it looks close enough though 😄

Let me know if I should revert it if that is not preferred :)

mikeesto:
The build failure in CI doesn't seem related to this PR - currently Google fonts is timing out

components/Common/Preview/index.module.css
bmuenzenmeyer: we will also use this within blog post cards

image

bmuenzenmeyer (approved):

aymen94 (approved):

PR #6027, opened 2023-10-19
Fix: render HTML tables in md files (#6002)

victorsptt:
## Description

Added the `rehype-raw` plugin to preserve HTML tables in md files. By default `@mdx-js/mdx` removes HTML from `md` files.

## Validation

I checked that the pages are rendering properly.

Screenshot from 2023-10-19 02-29-34

## Related Issues

#6002 

### Check List

- [ x ] I have read the [Contributing Guidelines](https://github.com/nodejs/nodejs.org/blob/main/CONTRIBUTING.md) and made commit messages that follow the guideline.
- [ x ] I have run `npx turbo lint` to ensure the code follows the style guide. And run `npx turbo lint:fix` to fix the style errors if necessary.
- [ x ] I have run `npx turbo format` to ensure the code follows the style guide.
- [ x ] I have run `npx turbo test` to check if all tests are passing.
- [ x ] I've covered new added functionality with unit tests if necessary.

### DIFF
--- a/next.dynamic.mjs
+++ b/next.dynamic.mjs
@@ -7,6 +7,7 @@ import remarkHeadings from '@vcarl/remark-headings';
 import { serialize } from 'next-mdx-remote/serialize';
 import rehypeAutolinkHeadings from 'rehype-autolink-headings';
 import rehypePrettyCode from 'rehype-pretty-code';
+import rehypeRaw from 'rehype-raw';
 import rehypeSlug from 'rehype-slug';
 import { getHighlighter } from 'shiki';
 import shikiNordTheme from 'shiki/themes/nord.json';
@@ -177,6 +178,7 @@ export const generateStaticProps = async (source = '', filename = '') => {
                 getHighlighter({ ...options, langs: SUPPORTED_LANGUAGES }),
             },
           ],
+          rehypeRaw,
         ],
         remarkPlugins: [remarkHeadings, remarkAutoLinkLiteral],
         format: filename.includes('.mdx') ? 'mdx' : 'md',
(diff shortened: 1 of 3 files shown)


ovflowd:
I'm not sure this is a good idea. By adding `rehype-raw` you're making the each document be parsed twice. That is horrible for memory.

HTML is not Markdown-compliant, and we shouldn't have raw HTML within our Markdown files. There's a a reason MDX is not parsing those, and that's because they're HTML.

MDX explicitly on their documentation state that mixing HTML is not recommended. (Of course there are some exceptions, but it's not recommended).

If you could just simply do what I've asked, which is transforming the HTML tables into Markdown ones, that'd be great :)

ovflowd:
But this might be the only solution, @victorspt can you provide some benchmarks?

ovflowd:
This PR breaks the website. Check the errors,  @victorspt. I feel that you should only add `rehype-raw` if the file extension is `md`. Can you update the `next.dynamic.mjs` with such logic? It should be simple as we already know the extension.

victorsptt:
> But this might be the only solution, @victorspt can you provide some benchmarks?

Yes, I will emit the benchmark.

ovflowd:
Hey @victorspt I'm going to make a PR to fix plain Markdown tables not being rendered/etc (adding the remark-gfm plugin for tables), then you can update this PR updating the HTML tables into plain ones. Does that sound good?

ovflowd:
Okay, this PR now is superseeded by https://github.com/nodejs/nodejs.org/pull/6039

victorsptt:
Thank you for fixing it Claudio, it looks great! 🎉

Em ter., 24 de out. de 2023 08:38, Claudio W ***@***.***>
escreveu:

PR #6037, opened 2023-10-20
Feat: create Common/BlogPostCard component

### LINKED ISSUE #6033: Create `Card` Component
Create a `Card` component using the [figma design (direct link)](https://www.figma.com/file/pu1vZPqNIM7BePd6W8APA5/Node.js?type=design&node-id=497-10522&mode=dev).  

This component is a card that previews other content, usually a blog post.

---

### Before You Start...

---

### Details

image

- Create a new directory called `Card` within the [`Common` directory](https://github.com/nodejs/nodejs.org/tree/main/components/Common) that contains all the new code.
- Use the new `Preview` component introduced within https://github.com/nodejs/nodejs.org/issues/6011
- Use the new `AvatarGroup` component introduced within https://github.com/nodejs/nodejs.org/issues/5910
- Use the new tailwind color and typography system to assign properties. For example, the subheading uses Green/400 per the figma.

#### Suggested Props

| Name | type | comments |
| - | - | - |
`heading` | `string`  | example: `Node.js March 17th Infrastructure Incident Post-mortem` per figma, passes through to `Preview`
`subheading` | `string`  | example: `Vulnerabilities` per figma
`type` | enum `announcement` \| `release` \| `vulnerability` | default is `announcement`, passes through to `Preview` - this enum is already in preview - reuse it
`supportingText` | `string` | excerpt of post content, limited to 3 lines via [line-clamp](https://caniuse.com/css-line-clamp). `Starting on March 15th and going through to March 17th (with much of the issue being mitigated on the 16th), users were receiving intermittent 404 responses when trying to download Node.js from nodejs.org, or even accessing parts of the website.` per figma
`avatars` | typeof `Avatar[]` | use the `alt` value from Avatar as the author's full name, such as `Hayden Bleasel` per figma
`date` | `Date` | formatted as `20 Jan 2024` per figma

#### Storybook

Add a storybook story file `components/Common/Cards/index.stories.tsx` which exercises each of the compnent's states.

- Light
- Dark

danielmarcano:
## Description

- Create the `Common/Card` component, based on these [Figma designs](https://www.figma.com/file/pu1vZPqNIM7BePd6W8APA5/Node.js?type=design&node-id=497-10522&mode=dev).
  - It takes the following properties:

```ts
type Author = {
  fullName: string;
  src: string;
};

export type CardProps = {
  title: ComponentProps<typeof Preview>['title'];
  type: Required<ComponentProps<typeof Preview>>['type'];
  description: string;
  authors: Author[];
  date: Date;
};
```

- The component uses the `Preview` and `AvatarGroup` components, and semantic HTML tags to keep it as accessible as possible.

The PR includes two minor fixes:

- `Common/AvatarGroup/Avatar`: Update component styles as it was creating a layout shift issue, which is not accessible, as it increases [Cumulative Layout Shift](https://web.dev/articles/cls), a Web Core Vital that is supposed to stay as low as possible. Here's a short video demonstrating this:

https://github.com/nodejs/nodejs.org/assets/93147810/e5d02413-75cc-479a-ae7a-521e8e509519

- `Common/Preview`: the `h1` tag has been replaced with an `h2` one, so as to avoid having more than one `h1` tag per page, as it's [not considered a good practice](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/Heading_Elements#avoid_using_multiple_h1_elements_on_one_page), and is also better for screen reader users.

## Validation

- Here's a short video showcasing the stories for this new component:

https://github.com/nodejs/nodejs.org/assets/93147810/47bb57e1-66d3-4707-9c17-9f5c397be5e8

- Here's a short video showing the styles fix on the `Avatar` component:

https://github.com/nodejs/nodejs.org/assets/93147810/f6c4264d-a86b-4413-8877-169d40d97885

- Unit tests have been added to prevent regression errors on the `Common/BlogPostCard` component

## Related Issues

Closes [#6033](https://github.com/nodejs/nodejs.org/issues/6033).

### Check List


ovflowd:
> Common/AvatarGroup/Avatar: The delayMs prop has been removed from the RedixAvatar.Fallback component, as it was creating a layout shift issue, which is not accessible, as it increases [Cumulative Layout Shift](https://web.dev/articles/cls), a Web Core Vital that is supposed to stay as low as possible. Here's a short video demonstrating this:

Could you let me know if you are sure about this? According to Radix documentation, `delayMs` exist exactly to reduce layout shift and reduce screen flickering.

danielmarcano:
> > Common/AvatarGroup/Avatar: The delayMs prop has been removed from the RedixAvatar.Fallback component, as it was creating a layout shift issue, which is not accessible, as it increases [Cumulative Layout Shift](https://web.dev/articles/cls), a Web Core Vital that is supposed to stay as low as possible. Here's a short video demonstrating this:
> 
> Could you let me know if you are sure about this? According to Radix documentation, `delayMs` exist exactly to reduce layout shift and reduce screen flickering.

Hi, @ovflowd! Thank you for your comment.

I checked the Radix UI docs before removing the `delayMs` prop, and noticed that they mentioned flickering, which tends to be associated with the "swap" of two elements, in this case the `RadixAvatar.Image` and the `RadixAvatar.Fallback` ones, but does not cause a layout shift.

Here's a video showing the result of removing `delayMs`. You can notice the flickering issue for `src` values that do resolve to images:

https://github.com/nodejs/nodejs.org/assets/93147810/feb7f331-1eef-4055-9179-74091f6f4b30

After reading your comment, I have come up with a better way of avoiding the layout shift. This way, we can keep the `delayMs` prop, and avoid both the flickering issue, and the layout shift issue.

I will push the changes now. This is how it looks like:

https://github.com/nodejs/nodejs.org/assets/93147810/79b83bf1-125e-433e-906b-f459a9ce1b84

components/Common/Card/index.module.css
```diff
@@ -0,0 +1,53 @@
+.container {
+  @apply max-w-full
+  bg-white
+  dark:bg-neutral-950;
```
canerakdas: Missing indentation 👀

components/Common/Card/index.module.css
```diff
@@ -0,0 +1,53 @@
...
+  @apply text-sm
+    font-semibold
+    text-neutral-900
+    dark:text-white;
+}
+
+.date {
+  @apply text-sm
+  text-neutral-800
+  dark:text-neutral-200;
```
canerakdas: Missing indentation

components/Common/Card/index.tsx
```diff
@@ -0,0 +1,67 @@
+import { useMemo, type ComponentProps, type FC } from 'react';
```
canerakdas: Type imports should be separated from regular imports

components/Common/Card/index.tsx
```diff
@@ -0,0 +1,67 @@
...
+import Preview from '@/components/Common/Preview';
+import { shortHumanReadableDate } from '@/util/shortHumanReadableDate';
+
+import styles from './index.module.css';
+
+type Author = {
+  firstName: string;
+  lastName: string;
+  src: ComponentProps<typeof AvatarGroup>['avatars'][number]['src'];
+};
```
canerakdas: Do we need to obtain `firstName` and `lastName` information separately? It might be better to use `Avatar` props directly

components/Common/Card/index.tsx
```diff
@@ -0,0 +1,67 @@
...
+
+export type CardProps = {
+  title: ComponentProps<typeof Preview>['title'];
+  type: ComponentProps<typeof Preview>['type'];
+  subtitle: string;
+  description: string;
+  author: Author;
```
canerakdas: As I mentioned above, I think we can directly use the props of the `Avatar` component. Additionally, as stated in the issue description, it must be possible to have more than one avatar

```suggestion
  authors: ComponentProps<typeof Avatar>[];
```

components/Common/Card/index.tsx
```diff
@@ -0,0 +1,67 @@
...
+  }, [author.firstName, author.lastName, author.src]);
+
+  return (
+    <article className={styles.container}>
+      <Preview
+        title={title}
+        type={type}
+        height="auto"
+        className={styles.preview}
+      />
```
canerakdas: @ovflowd It seems like a refactor needs to be made about the content of this component. I will examine it and try to resize the logo and text according to the container's width.

components/Common/Preview/index.tsx
```diff
@@ -37,7 +37,7 @@ const Preview: FC<PreviewProps> = ({
           src={`${basePath}/static/images/logos/js-white.svg`}
           alt="Node.js"
         />
-        <h1>{title}</h1>
+        <h2>{title}</h2>
```
canerakdas, line 40: When I examined the designs, I saw that this component(`Preview`) was used in more than one place. Marking this element directly as `h1`, `h2` or `h3` may give us a disadvantage in terms of SEO. Instead, I recommend using it directly. We can pass the heading elements from above

```suggestion
        {title}
```

components/Common/Card/index.tsx
```diff
@@ -0,0 +1,67 @@
...
+      <p className={styles.subtitle}>{subtitle}</p>
+      <h3 className={styles.title}>{title}</h3>
+      <p className={styles.description}>{description}</p>
+      <footer className={styles.footer}>
+        <AvatarGroup avatars={avatars} />
+        <div>
+          <p className={styles.author}>
+            {author.firstName} {author.lastName}
+          </p>
+          <p className={styles.date}>{shortHumanReadableDate(date)}</p>
```
canerakdas: We can use [the time element](https://www.w3.org/TR/2011/WD-html5-author-20110809/the-time-element.html) there

components/Common/Card/index.tsx
```diff
@@ -0,0 +1,67 @@
...
+        className={styles.preview}
+      />
+      <p className={styles.subtitle}>{subtitle}</p>
+      <h3 className={styles.title}>{title}</h3>
+      <p className={styles.description}>{description}</p>
+      <footer className={styles.footer}>
+        <AvatarGroup avatars={avatars} />
+        <div>
+          <p className={styles.author}>
+            {author.firstName} {author.lastName}
```
canerakdas: It may be a page written by more than one person, such as the example page below;
https://nodejs.org/en/blog/release/v18.14.0

components/Common/Card/index.stories.tsx
```diff
@@ -0,0 +1,47 @@
...
+    withThemeByDataAttribute({
+      themes: {
+        light: '',
+        dark: 'dark',
+      },
+      defaultTheme: 'dark',
+      attributeName: 'data-theme',
+    }),
+  ],
+};
```
canerakdas: Dont need to add a story for the Dark mode. Chromatic gives output for light and dark mode

components/Common/Card/index.tsx
```diff
@@ -0,0 +1,67 @@
...
+  const avatars = useMemo(() => {
+    const avatars = [
+      {
+        alt: `${author.firstName} ${author.lastName}`,
+        src: author.src,
+      },
+    ];
+
+    return avatars;
+  }, [author.firstName, author.lastName, author.src]);
```
canerakdas: With the solution I mentioned above, there will be no need to make these changes

canerakdas (commented):
Great work and the component layout exactly matches! I have added some reviews to help 😄

components/Common/Card/index.tsx
danielmarcano: Thank you for your suggestion.

I have added these props as they add more clarity. Using `alt` directly is way less explicit.

components/Common/Card/index.tsx
danielmarcano: Thank you for your suggestion. As mentioned [here](https://github.com/nodejs/nodejs.org/pull/6037/files#r1367783451), using the `alt` prop directly conveys less meaning/intention than asking directly for a `firstName` and a `lastName`.

As for using `authors` in plural, I have just left a doubt [here](https://github.com/nodejs/nodejs.org/pull/6037/files#r1367784256).

components/Common/Card/index.tsx
```diff
@@ -0,0 +1,67 @@
...
+  subtitle: string;
+  description: string;
+  author: Author;
+  date: Date;
+};
+
+const Card: FC<CardProps> = ({
+  title,
+  type,
+  subtitle,
```
canerakdas: Im not sure but instead of adding similar props, it might be better to just send the type value and use the localized version inside the component

components/Common/Card/index.tsx
danielmarcano: Nice catch. Will update these. Thanks!

components/Common/Card/index.stories.tsx
danielmarcano: You're right. I had kept two stories as the Common/Card component task described that it was necessary to have a story per each state, and mentioned dark mode as its own state.

However, what you just mentioned is true. There's no need to have two different stories as we have the theme dropdown.

components/Common/Card/index.module.css
danielmarcano: Thanks! Will update these

components/Common/Card/index.module.css
danielmarcano: Thanks!

components/Common/Card/index.tsx
danielmarcano: Thanks for your suggestion! As mentioned [here](https://github.com/nodejs/nodejs.org/pull/6037/files#r1367783451), however, I have added these properties for better clarity on what the intended usage of these values is.

components/Common/Card/index.tsx
danielmarcano: This is actually a doubt that I had, but forgot to mention in the description of this PR.

@ovflowd, how would the component look like if we had more than 1 author? There are no [designs](https://www.figma.com/file/pu1vZPqNIM7BePd6W8APA5/Node.js?type=design&node-id=497-10522&mode=dev) for this.

components/Common/Card/index.tsx
danielmarcano: Indeed. I thought the same, while trying to adapt this component to the `Common/Card` one, but that was out of the scope of this PR. Definitely worth looking into.

components/Common/Card/index.tsx
danielmarcano: Great, thank you! I was not sure about using it, as I tend to use it for dates that are partially shown which is the case for all the examples they have on MDN:

```html
<time datetime="2018-07-07">July 7</time>
<time datetime="20:00">20:00</time>
<time datetime="PT2H30M">2h 30m</time>
```

It will still be useful for screen readers, though, so I will definitely add it!

components/Common/Preview/index.tsx
danielmarcano: Thank you for your suggestion. I have the following concerns with this:

- The `Preview` component relies on the `title` property.
- The `Common/Card` components are designed to have consistent hierarchies.
- Allowing any HTML tag (e.g., `p` or `h1`) as the `title` feels like exposing internal logic, in a way, which may impact accessibility.

Therefore, I suggest that we keep the `title` as a string for now.

components/Common/Card/index.tsx
danielmarcano: Thank you!

This is another doubt I had, but was not really sure from the description of this `Common/Card` component, as it suggests including the `subtitle` prop.

There seems to be a relationship between the `type` and the `subtitle` but they're not exactly the same.

There's only three types: 'announcement' | 'release' | 'vulnerability', and in the designs, the type 'vulnerability' is shown as `Vulnerabilities`.

vulnerabilities-subtitle

@ovflowd knowing this, could you confirm if such translation strings should be 'Vulnerabilities', 'Releases', and 'Announcements'? This way, I could add them in this PR, and remove the `subtitle` property as suggested by @canerakdas

danielmarcano:
> Great work and the component layout exactly matches! I have added some reviews to help 😄

Thank you for the kind review! I have applied most of the changes you have suggested, and replied to all of them.

I have also left a couple of doubts for @ovflowd to solve, as neither of them are clear from the designs.

components/Common/Card/index.tsx
canerakdas: Nope, I'm thinking like:

Adding enums into the `/i18n/locales/en.json`

```json
"components.common.card.vulnerability": "Vulnerabilities",
"components.common.card.announcement": "Announcements",
"components.common.card.release": "Releases"
```

And use;

```jsx
<FormattedMessage id={`components.common.card.${type}`} />
```

bmuenzenmeyer: I wanted these to be props and the responsibility of the caller because I am not 100% certain that there won't be subtitles OTHER than the `type` values


edit: i had to go back something like 4 years to this post https://nodejs.org/en/blog/uncategorized/10-lts-to-12-lts to find something that wasn't - and it doesn't even render. I suppose it's fine to start with combining `subtitle` and `type` but yes I agree with @canerakdas then it needs to be internationalized

danielmarcano: @canerakdas I had understood you correctly, yes. That's what I meant by adding the translation strings.

Got it, @bmuenzenmeyer, thank you! I will leave the `subtitle` prop as-is then.

components/Common/Card/index.tsx
bmuenzenmeyer: I think it would the overlapped avatar group, and then names in order, like https://nodejs.org/en/blog/announcements/node-js-march-17-incident

danielmarcano: Thanks! The names without the "By", and the `AvatarGroup`, right? As @canerakdas had suggested.

What is the `limit` of avatars we should show per card?

danielmarcano: I have pushed these changes! @bmuenzenmeyer

danielmarcano:
Hi @bmuenzenmeyer, I have just pushed the latest changes, allowing the `Common/Card` component to receive more than one author.

ovflowd:
> I have also left a couple of doubts for @ovflowd to solve, as neither of them are clear from the designs.

Hey there, are there still open questions for me?

danielmarcano:
> > I have also left a couple of doubts for @ovflowd to solve, as neither of them are clear from the designs.
> 
> 
> 
> Hey there, are there still open questions for me?

Hi, @ovflowd! Thanks for asking.

Just a single doubt left: should we pass a `limit` to the `AvatarGroup`?

components/Common/Card/index.tsx
canerakdas: I got your point, makes sense. But afaik this data is created from files in markdown formatted. You can view the sample file here; https://raw.githubusercontent.com/nodejs/nodejs.org/main/pages/en/blog/announcements/adjusted-release-schedule-covid.md

IMHO It doesn't seem sense to parse the names in these files as name/surname via custom hooks and combine them again as full names within the component

ovflowd:
> > > I have also left a couple of doubts for @ovflowd to solve, as neither of them are clear from the designs.
> > 
> > 
> > Hey there, are there still open questions for me?
> 
> Hi, @ovflowd! Thanks for asking.
> 
> Just a single doubt left: should we pass a `limit` to the `AvatarGroup`?

Nah, I don't think it's needed for now.

danielmarcano:
> > > > I have also left a couple of doubts for @ovflowd to solve, as neither of them are clear from the designs.
> > > 
> > > 
> > > Hey there, are there still open questions for me?
> > 
> > 
> > Hi, @ovflowd! Thanks for asking.
> > Just a single doubt left: should we pass a `limit` to the `AvatarGroup`?
> 
> Nah, I don't think it's needed for now.

Great. Then I would say that this component is good to go!

components/Common/Card/index.module.css
```diff
@@ -0,0 +1,53 @@
+.container {
+  @apply max-w-full
+    bg-white
+    dark:bg-neutral-950;
+}
+
+.preview {
+  @apply mb-6
+    max-w-full
+    rounded p-4;
```
canerakdas: More than one Tailwind class should not be on the same line 👀

components/Common/Card/index.tsx
canerakdas: Based on this informations, it would be better to update as I suggested @danielmarcano 
> edit: i had to go back something like 4 years to this post https://nodejs.org/en/blog/uncategorized/10-lts-to-12-lts to find something that wasn't - and it doesn't even render. I suppose it's fine to start with combining `subtitle` and `type` but yes I agree with @canerakdas then it needs to be internationalized

components/Common/Card/index.tsx
canerakdas: I'm still not sure about this issue. What do you think about it @bmuenzenmeyer @ovflowd

components/Common/Card/index.module.css
danielmarcano: Thanks! Just pushed these changes

components/Common/Card/index.tsx
danielmarcano: Thanks for pointing this out, @canerakdas. I hadn't read the edited version of @bmuenzenmeyer's comment.

I will remove the `subtitle` prop, in that case.

components/Common/Card/index.tsx
danielmarcano: Hi, @canerakdas. I didn't know information came from markdown files. Thank you for letting me know.

However, seeing that the full name comes from an `author` "property", I would still keep a custom property.

Just a single one, called `fullName`, as still, using `alt` would not convey to people that:
- We expect people to pass the full author name as the `alt` value of each `Avatar`
- The `alt` value would also be used to display a list of all author names, comma-separated

What are your thoughts on this? CC @ovflowd @bmuenzenmeyer

junwen-k:
Should this component be renamed as `ArticleCard` / `BlogPostCard` or something? `Card` feels a bit too generic for the current use case IMO. I would expect card to just render the outline of a Card style instead of rendering Authors, Avatars etc. (In Figma it is called `Blog Post Card`) 👀

danielmarcano:
> Should this component be renamed as `ArticleCard` / `BlogPostCard` or something? `Card` feels a bit too generic for the current use case IMO. I would expect card to just render the outline of a Card style instead of rendering Authors, Avatars etc. (In Figma it is called `Blog Post Card`) 👀 

Thank you. Makes sense to rename it to `BlogPostCard` indeed.

What are your thoughts on this? @canerakdas @ovflowd

bmuenzenmeyer:
I named it a bit more generically to account for potential future reuse. A card is a window into other content, a preview. Material has made everyone think it's just an elevated container.

I'm fine with whatever (and went with the Figma names -most- of the time)... You know what they say about having things....😏

danielmarcano:
> I named it a bit more generically to account for potential future reuse. A card is a window into other content, a preview. Material has made everyone think it's just an elevated container.
> 
> I'm fine with whatever (and went with the Figma names -most- of the time)... You know what they say about having things....😏

Thank you for the feedback.

In that case, since the Figma designs call this component a "Blog Post Card" as pointed by @junwen-k, I have renamed the component to `BlogPostCard`.

I have also applied these last two changes:

- Removed `subtitle` prop, and used the `type` to determine the correct subtitle translation string
- Merged `firstName` and `lastName` into a single `fullName` prop 

With those changes, I would say that this component is now complete!

CC @canerakdas

canerakdas (approved):
LGTM :shipit: Thank you for your contribution :tada:

components/Common/BlogPostCard/index.stories.tsx
```diff
@@ -0,0 +1,45 @@
+import type { Meta as MetaObj, StoryObj } from '@storybook/react';
+
+import BlogPostCard from './index';
```
ovflowd:
```suggestion
import BlogPostCard from '@/components/Common/BlogPostCard';
```

components/Common/BlogPostCard/index.tsx
```diff
@@ -0,0 +1,71 @@
...
+  type,
+  description,
+  authors,
+  date,
+}) => {
+  const avatars = useMemo(() => {
+    const parsedAvatars = authors.map(({ fullName, src }) => ({
+      alt: fullName,
+      src,
+      toString() {
```
ovflowd: Interesting. But can it be on a arrow notation? `toString: () => fullName`?

components/Common/BlogPostCard/index.tsx
```diff
@@ -0,0 +1,71 @@
...
+        return fullName;
+      },
+    }));
+
+    return parsedAvatars;
+  }, [authors]);
+
+  const authorNames = avatars.join(', ');
```
ovflowd: Does this need to be a variable on its own?

util/shortHumanReadableDate.ts
```diff
@@ -0,0 +1,9 @@
+export function shortHumanReadableDate(date: Date, locale?: string) {
```
ovflowd: I would actually prefer a generic DateTimeFormat function that accepts the props of DateTimeFormat.

components/Common/BlogPostCard/index.stories.tsx
danielmarcano: What do you think about leaving it as `'./'` instead, to maintain consistency? The thing is that the other stories from this `components` directory all have either `'./'` or `'./index'`

components/Common/BlogPostCard/index.tsx
danielmarcano: Thanks. Yes, it can be.

components/Common/BlogPostCard/index.tsx
danielmarcano: I created it so that its intended usage was more clear, but if you think it's unnecessary, I can remove it.

util/shortHumanReadableDate.ts
danielmarcano: You mean renaming and updating the function to something like this?

```ts
export function dateTimeFormat(
  date: Date,
  options: Intl.DateTimeFormatOptions,
  locale?: string
) {
  const formattedDate = new Intl.DateTimeFormat(
    locale ?? navigator.language,
    options
  ).format(date);

  return formattedDate;
}

```

The way I see it, this diminishes the intended value of this `shortHumanReadableDate`, as it was meant as an abstraction.

By updating it and leaving it as broad as possible, people would need to check the `Intl.DateTimeFormat`'s `options` parameter API, and understand how to achieve what they want, so its usage would be a bit less straightforward, and not so consistent, as some might configure them a bit differently:

```ts
           {dateTimeFormat(date, {
              day: 'numeric',
              month: 'short',
              year: 'numeric',
            })}
```

What are your thoughts on this?

ovflowd: > The way I see it, this diminishes the intended value of this shortHumanReadableDate, as it was meant as an abstraction.

I don't see the value of having a helper used in one (see the rule of 3 https://blog.codinghorror.com/rule-of-three/) place for formatting dates, you instead want to simply invoke Intl.DateTimeFormat directly from your Component, then.

Also remember that you shouldn't instantiate `Intl.DateTimeFormat` within a component. Please simply crate the format you want on the root of your Component file and then invoke the `myTimeFormat.format(myDateObj)` within the component. Remember to memoize it if you don't want a new date/format to happen on every rerender.

> By updating it and leaving it as broad as possible, people would need to check the Intl.DateTimeFormat's options parameter API, and understand how to achieve what they want, so its usage would be a bit less straightforward, and not so consistent, as some might configure them a bit differently:

Exactly, that's the conclusion I wanted you to reach. It's a useless abstraction.

components/Common/BlogPostCard/index.tsx
ovflowd: Yup please remove it.

components/Common/BlogPostCard/index.stories.tsx
ovflowd: https://github.com/nodejs/nodejs.org/pull/6046 the consistency is to use `@/` for component imports.

util/shortHumanReadableDate.ts
danielmarcano: Thank you. Great read.

I also agree in avoiding early abstractions but had incorrectly assumed that this util function would be used in several places.

I will add these changes 🚀

components/Common/BlogPostCard/index.stories.tsx
danielmarcano: Great to know! Thank you. I will update this.

util/shortHumanReadableDate.ts
ovflowd: Right, so Intl.DateTimeFormat constructor returns an instance of DateTimeFormat that you can use against any date object. It is something static. You don't need to call the constructor inside a React Component as it will never change, meaning:

```tsx
const shortDateTimeFormat = new Intl.DateTimeFormat({ ... });

const MyComponent = ({ date }) => {
  const formattedDate = useMemo(() => 
    shortDateTimeFormat.format(date), [date]);

  ...
}
```

danielmarcano:
Hi, @ovflowd.

I have pushed the changes you have suggested, along with new unit tests for the `Common/BlogPostCard` component.

components/Common/BlogPostCard/index.tsx
```diff
@@ -0,0 +1,83 @@
...
+  month: 'short',
+  year: 'numeric',
+});
+
+type Author = {
+  fullName: string;
+  src: ComponentProps<typeof AvatarGroup>['avatars'][number]['src'];
+};
+
+export type BlogPostCardProps = {
```
ovflowd:
```suggestion
type BlogPostCardProps = {
```
We don't export ComponentProps as they in general can be inferred due `ComponentProps<typeof Component>`

components/Common/BlogPostCard/index.tsx
```diff
@@ -0,0 +1,83 @@
...
+
+const dateTimeFormat = new Intl.DateTimeFormat(navigator.language, {
+  day: 'numeric',
+  month: 'short',
+  year: 'numeric',
+});
+
+type Author = {
+  fullName: string;
+  src: ComponentProps<typeof AvatarGroup>['avatars'][number]['src'];
```
ovflowd: Isn't `src` just a string? This sounds a mouthful of a type 😅

components/Common/BlogPostCard/index.tsx
```diff
@@ -0,0 +1,83 @@
...
+
+const BlogPostCard: FC<BlogPostCardProps> = ({
+  title,
+  type,
+  description,
+  authors,
+  date,
+}) => {
+  const avatars = useMemo(() => {
+    const parsedAvatars = authors.map(({ fullName, src }) => ({
```
ovflowd: Is a variable assignment really needed? Can't this be just inline without return statement and variable assignment?

```tsx
const avatars = useMemo(() => authors.map(...), [authors]);
```

components/Common/BlogPostCard/index.tsx
```diff
@@ -0,0 +1,83 @@
...
+      <Preview
+        title={title}
+        type={type}
+        height="auto"
+        className={styles.preview}
+      />
+      <p className={styles.subtitle}>
+        <FormattedMessage id={`components.common.card.${type}`} />
+      </p>
+      {/* We hide it from screen-readers as Preview component already wraps the title in an h2 tag */}
```
ovflowd: I'm often opposed to inline comments within JSX. I would say these things are self-explanatory.

ovflowd (approved):
I know I've left a few last comments but after that I swear, this is good to go.

components/Common/BlogPostCard/index.tsx
danielmarcano: Thank you! I actually thought so, and that's why I ended up using `ComponentProps<typeof Component>`, but forgot to remove this `export`.

components/Common/BlogPostCard/index.tsx
danielmarcano: I'm with you on this, thanks for pointing it out. I will set `src` to a `string`.

components/Common/BlogPostCard/index.tsx
danielmarcano: I did not know if it was going to be self-explanatory, although it's true that there's a unit test covering this, so I will remove the comment.

components/Common/BlogPostCard/index.tsx
danielmarcano: Thank you! The variable is indeed not needed. We can definitely inline `authors.map`.

danielmarcano:
> I know I've left a few last comments but after that I swear, this is good to go.

No worries! Thanks for your suggestions and for your feedback. I have just pushed the changes 🚀

aymen94 (approved):

ovflowd:
Hey @danielmarcano just wanted to again appreciate the fantastic work you've been doing. We truly appreciate it!

danielmarcano:
> Hey @danielmarcano just wanted to again appreciate the fantastic work you've been doing. We truly appreciate it!

That's really kind of you, @ovflowd. Thank you for sharing these thoughts with me!

PR #6038, opened 2023-10-21
feat: add CodeBox component and code tabs plugin

### LINKED ISSUE #5819: Create `CodeBox` component

devjvao:
## Description

This PR introduces two new components related to the code block:

- `CodeBox`: The main component for rendering the code block;
- `CodeTabs`: The component that customizes the `Tabs` component to match the `CodeBox` style.

And introduces a plugin for resolving the desired format in markdown files into the code tabs structure:

- `codeTabsResolver`: The plugin transforms the format below into the code tabs structure.

  ```md
  ```mjs|cjs labels="MJS,CJS" link="[More options](https://github.com/nodejs/nodejs.org)"
  const { createHmac } = await import('node:crypto');
  --------------
  const { createHmac } = require('node:crypto');
  ``` <!-- prevent markdown from formatting this, ignore it -->
  ```

### Single file

image

### Multiple files

image

### Multiple files with link

image

## Considerations

Some things were not clear to me about how the implementation should go.

- Is there a minimum/maximum width?;
- Where should the copied notification appear? (top left, top right...);
- Is it okay to create the `CodeTabs` for dealing with the tabs structure? I couldn't manage to make it work with a single component;
- I also had to update the `Tabs` implementation to accept an `addons` prop so that we can render it correctly without harming the responsive layout;
- In the `CodeBox` component I'm using a `languageMap` variable for mapping the display language with the language that is used in the markdown. Where should this go? Should this have internationalization support?;
- In the `codeTabsResolver` I'm using the `normalizeLanguage` function to handle JavaScript file extensions like `mjs`. Is it ok? It seemed fragile to me, but I couldn't think of anything else;
- The `ArrowUpRightIcon` looks different from the one used in Figma;
- The green used in the `Tabs` component is different from the one used in Figma, so I had to overwrite it, but it seems wrong to do it. Should we update the `Tabs` green color or remove the style overriding it in the `CodeTabs`?

## Related Issues

Addresses #5819.

### Check List


components/Common/CodeBox/CodeTabs/index.tsx
```diff
@@ -0,0 +1,40 @@
...
+import type { TabsProps } from '@/components/Common/Tabs';
+import Tabs from '@/components/Common/Tabs';
+import LocalizedLink from '@/components/LocalizedLink';
+
+import styles from './index.module.css';
+
+type CodeTabsProps = Pick<
+  TabsProps,
+  'tabs' | 'onValueChange' | 'defaultValue'
+> & {
```
ovflowd: I'd rather than addons could be just `children`? So we can simply add what we want?

Then you can create simply update the `LocalizedLink` component to a have a new prop called `external` (boolean) and when present it renders the Arrow.

Then we simply pass on `<CodeTabsProps><LocalizedLink external={true}>....`

components/Common/CodeBox/CodeTabs/index.tsx
```diff
@@ -0,0 +1,40 @@
...
+  linkUrl?: string;
+  linkText?: string;
+};
+
+const CodeTabs: FC<PropsWithChildren<CodeTabsProps>> = props => {
```
ovflowd: We recommend destructuring props, instead of assigning to a variable. That's what we've been doing so far.

components/Common/CodeBox/CodeTabs/index.tsx
```diff
@@ -0,0 +1,40 @@
...
+  const { children, linkUrl, linkText, ...rest } = props;
+
+  return (
```
ovflowd: Then a return statement will not be needed

components/Common/CodeBox/__tests__/index.test.mjs
```diff
@@ -0,0 +1,102 @@
+import * as Toast from '@radix-ui/react-toast';
+import { render, screen, waitFor, within } from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
+import { IntlProvider } from 'react-intl';
+
+import CodeBox from '@/components/Common/CodeBox';
+
+jest.mock('@heroicons/react/24/outline', () => ({
```
ovflowd: Is this mock needed... Why?

components/Common/CodeBox/__tests__/index.test.mjs
```diff
@@ -0,0 +1,102 @@
...
+  const Sut = props => {
+    const {
+      language = 'javascript',
+      content = 'const value = 10;',
+      showLineNumbers,
+    } = props;
+
+    return (
+      <Toast.Provider>
+        <IntlProvider locale="en" messages={messages}>
```
ovflowd: We don't recommend passing IntlProvider on tests. We recommend passing our own `LocaleProvider`

components/Common/CodeBox/index.stories.tsx
```diff
@@ -0,0 +1,130 @@
...
+server.listen(port, hostname, () => {
+  console.log(\`Server running at http://\${hostname}:\${port}/\`);
+});
+\`\`\``,
+  },
+};
+
+export const MultipleFiles: Story = {
+  args: {
+    children: `\`\`\`mjs|cjs showLineNumbers labels="MJS,CJS"
```
ovflowd: So showLineNumbers is something from Shiki? Is this labels field also?

mdx/plugins/codeTabsResolver.mjs
```diff
@@ -0,0 +1,152 @@
...
+        Object.assign(parent, {
+          tagName: 'codeTabs',
+          properties: {
+            labels: metaLabels,
+            languages: languages.join('|'),
+            ...link,
+          },
+          children: languages.map((language, index) => ({
+            type: 'element',
+            // `rehype-pretty-code` only formats code elements as a code block
```
ovflowd: But that is correct. We should only transform `pre` elements, no?

components/Common/CodeBox/index.stories.tsx
```diff
@@ -0,0 +1,130 @@
...
+npm -v
+\`\`\``,
+  },
+};
+
+export default {
+  component: CodeBox,
+  render: (_, { loaded: { Content } }) => Content,
+  loaders: [
+    async ({ args }) => {
```
ovflowd: I would rather that we export the part of serialize of MDX of `next.dynamic.mjs` to a different function and then simply reimport it here.

Otherwise we can easily forget about this.

components/Common/CodeBox/index.tsx
```diff
@@ -0,0 +1,79 @@
...
+import { useRef } from 'react';
+import { FormattedMessage } from 'react-intl';
+
+import Button from '@/components/Common/Button';
+import Notification from '@/components/Common/Notification';
+import { useCopyToClipboard } from '@/hooks/useCopyToClipboard';
+
+import styles from './index.module.css';
+
+const languageMap: Record<string, string[]> = {
```
ovflowd: Move this to our next.shiki.mjs file

components/Common/CodeBox/index.tsx
```diff
@@ -0,0 +1,79 @@
...
+  JSON: ['json'],
+  XML: ['xml'],
+  YML: ['yml', 'yaml'],
+  Bash: ['bash', 'sh', 'shell', 'zsh', 'shellscript'],
+  Shell: ['console', 'shellsession'],
+  TypeScript: ['ts', 'typescript'],
+  JavaScript: ['cjs', 'mjs', 'js', 'javascript'],
+};
+
+const labelsMap = Object.fromEntries(
```
ovflowd: I don't like mappings being done (function calls) at root level of a file, it's not practical and pollutes global context. Transform this into a function if needed.

components/Common/CodeBox/index.tsx
```diff
@@ -0,0 +1,79 @@
...
+    languages.map(language => [language, label])
+  )
+);
+
+type CodeBoxProps = {
+  language: string;
+  showLineNumbers?: boolean;
+};
+
+const CodeBox: FC<PropsWithChildren<CodeBoxProps>> = props => {
```
ovflowd: Destructuring of props

components/Common/CodeBox/index.tsx
```diff
@@ -0,0 +1,79 @@
...
+  const { children, language, showLineNumbers = false } = props;
+
+  const ref = useRef<HTMLPreElement>(null);
+
+  const [copied, copyToClipboard] = useCopyToClipboard();
+
+  const onCopy = () => {
+    if (ref.current !== null && ref.current.textContent !== null) {
```
ovflowd: Hmmm, what are the cases whe it is null or textNull is null?

components/Common/CodeBox/index.tsx
```diff
@@ -0,0 +1,79 @@
...
+      copyToClipboard(ref.current.textContent);
+    }
+  };
+
+  return (
+    <div className={styles.root}>
+      <div
```
ovflowd: Is there a way to reduce one div?

components/Common/CodeBox/index.tsx
```diff
@@ -0,0 +1,79 @@
...
+      <div className={styles.footer}>
+        <span className={styles.language}>
+          {labelsMap[language] ?? language}
+        </span>
+        <Button type="button" className={styles.action} onClick={onCopy}>
+          <DocumentDuplicateIcon className={styles.icon} />
+          <FormattedMessage id="components.common.codebox.copy" />
+        </Button>
+      </div>
+      <Notification open={copied}>
```
ovflowd: FYI, we use a hook to dispatch notifications, we don't invoke it directly. Can you use the hook?

components/Common/Tabs/index.tsx
```diff
@@ -9,33 +9,42 @@ type Tab = {
...
-        >
-          {tab.label}
-        </TabsPrimitive.Trigger>
-      ))}
-    </TabsPrimitive.List>
-    {children}
-  </TabsPrimitive.Root>
-);
+}) => {
+  return (
```
ovflowd: No return statement please

components/Common/Tabs/index.tsx
```diff
@@ -9,33 +9,42 @@ type Tab = {
   label: string;
 };
 
-type TabsProps = {
+export type TabsProps = {
```
ovflowd: No export of Component Types, if you need them, simply do `ComponentProps<typeof Component>`

mdx/plugins/codeTabsResolver.mjs
ovflowd, line 1: I'm a bit in the fence about this. I usually don't like using custom syntax, even more as API Docs will get through similar tech, having an extra unified parser.... And custom syntax, feels kinda wrong.

We should IMO, always show the line numbers (so no need of this lineNumberParam) and we shouldn't have this param called "labels". I don't even think we should have a mapping that replaces "mjs/cjs" etc to "js".

Since we use Shiki, you can simply add them as alias here https://github.com/nodejs/nodejs.org/blob/main/shiki.config.mjs#L30

I would also say you can update this same variable and add a new field called "label". So label for any of the languages that match the id or aliases would have label of "JavaScript"

This reduce code complexity, code duplication, need of extra variables and non-standard Markdown syntax inside code blocks.

Unfortunately we still need to handle the `|` divisor and the `--------` **OR** since you are already making an Unified plugin, you could simply do the following:

The cases from MDX when we should have only one Codebox with two languages, is when one Codebox is followed immediately by another Codebox.

Hence you could try to handle this scenario, and merge the two Codeboxes in one element.

Then with this approach we **don't** need to change the Markdown sources. I would love if you could do that!

devjvao: Nice review!

I didn't notice the alias [here](https://github.com/nodejs/nodejs.org/blob/main/shiki.config.mjs#L30), it is much simpler. :sweat_smile: 

I'll take a look and try implementing it as you suggested, it'll result in a much more clean code.

components/Common/CodeBox/__tests__/index.test.mjs
devjvao: Because assigning a `data-testid` to the icon allow us to ensure the icon is rendered on the DOM. Otherwise, it wouldn't be possible to differentiate more than one `svg` element.

Still, if you believe it is not needed I can just remove it, no problem. This is just a pattern I'm used to following.

mdx/plugins/codeTabsResolver.mjs
devjvao: Yep. The comment is just to explicit the need for the `pre` element before the `code` element. Do you think we can just remove this comment?

ovflowd: The comment is quite unclear, imo. Ill wait for you with the new implementation for further feedback.

ovflowd (commented):
Just wanted to say that I truly appreciate the effort you're putting in here, because this component is definitely a truly complex case.

Awesome work so far.

components/Common/CodeBox/index.stories.tsx
devjvao: I was able to do it but I exported it in a different file, because `next.dynamic.mjs` has some node-related imports.

components/Common/CodeBox/index.stories.tsx
devjvao: `showLineNumbers` is handled by `rehype-pretty-code`. It just adds an attribute `data-line-numbers`, but since we're always showing them I've just removed it.

`labels` is a custom meta property resolved by the `codeTabsResolver`, so we can display custom tab titles.

mdx/plugins/codeTabsResolver.mjs
devjvao: Got it. Well, I believe we can remove the comment to avoid any confusion.

components/Common/CodeBox/index.stories.tsx
ovflowd: Right but lets please remove all of this. It's just that we want to avoid anything too deviating from commonmark standard markdown or something that becomes cryptic to other people...

mdx/plugins/codeTabsResolver.mjs
devjvao: I've managed to make it work with the `followed immediately` approach. 

This is working now:

image

But, there is a problem. Actually, a doubt.

If we do not use the meta parameters for the tab title and the link on the right, where should we specify them? Especially now with more than one code block in the markdown.

Do you have a suggestion? I'm not sure how to continue.

ovflowd: Hey there, as I mentioned. You can add a nee property on each object within next.shiki.mjs called label. For cjs,esm,js it should simply be JavaScript imo

devjvao: Hey again! I believe we're talking two different things.

1. We have the language being shown at the bottom of the code block;
2. And the tab title at the top.

In the image below, both tabs will show `Bash` as the language. I've updated the implementation by adding a `displayName` as suggested, and it works beautifully.

My doubt is about the tab title at the top. How should we define the title and the link option in the markdown?

image

ovflowd: Hey, I think you're mistaking a few things. This component is not meant to be **only used with MDX**.

Meaning that based on a discussion I had with @bmuenzenmeyer there should be two variations of this component, which is what I thought you were doing.

One, the base one `Codebox` has one prop (`snippets`) of type similarly of `{ label: string, content: React.ReactNode }[]`

The content is whatever already parsed content (either by MDX or raw JSX such as `<code>something</code>` that is simply rendered within the `Codebox`.

The `label` field is what is shown **below**, and for the MDX Variant of this Component you would be wiring up the Label from `next.shiki.mjs`.

So the `MDXCodebox` receives only a children which is the parsed MDX, and then it should infer the label of the language based on the className or something (and map towards the shiki.mjs file) (You don't need to do necessarily like this, but that's the idea)

Then you have similarly for the `CodeboxTabs`

The Base component which is exactly how you've done it, which contains the Surrounding Tab Component. And it has the children that has `Tabs.Content` and that `Tabs.Content` can be simply `Codebox` elements or `<code>` or anything.

Then you'd have the `MDXCodeboxTabs` this one also only uses MDX to infer. Not sure how you will do it, but the idea is that for the **top labels**  you will simply render what the language name is, ie cjs/esm/js/etc. No need for fancy mappings here.

And at the bottom you'll render whatever is mapped on shiki.mjs. It's important to mention that we don't care much about the top title, as in fact for all code snippets that have tabs, will either be manually JSX usages (where we manually pass whatever we want, like the home page one) or the ones from API docs, where the title is always the language code. Like on current API Docs.

Now, if you want, you can add a field called "displayName" but that **must** be optional (on the MDX corpus)

ovflowd: (Pretty much your Codebox component right now only supports MDX (children arg) and the reality is that the Base component should only support JSX. And the MDX one extends the "JSX" one.

(Hence two different stories) (You can if you want make the Component variations in the same file)

i.e. `import { MDXCodebox, Codebox } from '@/components/Common/Codebox';`

components/Common/CodeBox/index.stories.tsx
ovflowd: Gotcha

mdx/plugins/codeTabsResolver.mjs
ovflowd: Hey @devjvao I hope my explanation makes sense and I did not overwhelm you with all these details 🙈 let me know if you need any support!

devjvao: Hey again! Don't worry, I was waiting for some free time to look calmly. 

It makes sense, but that wasn't clear in the described issue. At least I did not understand it like that.

Still, I'm working on updating the implementation considering the last explanation.

By the way, I'm following this format:

devjvao: Hey again! Don't worry, I was waiting for some free time to look calmly.

It makes sense, but that wasn't clear in the described issue. At least I did not understand it like that.

Still, I'm working on updating the implementation considering the last explanation.

By the way, I'm following this format:

```ts
import { MDXCodeBox, CodeBox } from '@/components/Common/CodeBox';
import { MDXCodeTabs, CodeTabs } from '@/components/Common/CodeBox/CodeTabs';
```

And, regarding this part of the explanation:

> One, the base one Codebox has one prop (snippets) of type similarly of { label: string, content: React.ReactNode }[]...

The `CodeTabs` is the one that receives the `snippets`, and the `CodeBox` is the simplest one. I hope that is not a problem.

Anyway, I'll update the PR soon! Then you can review it again, and we can progress with this feature.

components/Common/CodeBox/CodeTabs/index.tsx
devjvao: If this is about the `addons` added in the `Tabs` component, then we can't just render the link as a child because the level is different. 

In order to respect this layout:

image

We need to render the link inside the `TabsPrimitive.List`. This way it is possible to use a flex layout and harmonize the tabs with the link.

image

Do you think of something else?

components/Common/CodeBox/index.tsx
devjvao: This is basically for type safety.

 `ref.current ` is null when the component is not mounted and `ref.current.textContent` is null for specific elements like the `document` itself.

mdx/plugins/codeTabsResolver.mjs
devjvao: Good to go for the next review. @ovflowd

components/Common/CodeBox/CodeTabs/index.tsx
ovflowd: I see... Well it's fine then.

components/Common/CodeBox/index.tsx
ovflowd: Right can't you just `if(ref.current?.textContent)`

components/Common/CodeBox/CodeTabs/index.stories.tsx
```diff
@@ -0,0 +1,86 @@
...
+
+type Story = StoryObj<typeof CodeTabs>;
+type Meta = MetaObj<typeof CodeTabs>;
+
+const TabsContent: FC = () => (
+  <>
+    <TabsPrimitive.Content key="mjs" value="mjs">
+      <CodeBox language="JavaScript (MJS)">
+        <code>
+          {`import * as http from 'http';
```
ovflowd: Can you extract this to a constant? So that we don't mix the text with the code?

components/Common/CodeBox/CodeTabs/index.stories.tsx
```diff
@@ -0,0 +1,86 @@
...
+  },
+};
+
+export default {
+  component: CodeTabs,
+  args: {
+    children: <TabsContent />,
+    defaultValue: 'mjs',
+    tabs: [
+      {
```
ovflowd:
```suggestion
      { key: 'mjs', label: 'MJS' }
```

We prefer inline objects when possible, when Prettier alows it, just to use less lines

components/Common/CodeBox/CodeTabs/index.stories.tsx
```diff
@@ -0,0 +1,86 @@
...
+        key: 'mjs',
+        label: 'MJS',
+      },
+      {
+        key: 'cjs',
+        label: 'CJS',
+      },
+    ],
+  },
+  parameters: {
```
ovflowd: What are these controls?

components/Common/CodeBox/CodeTabs/index.tsx
```diff
@@ -0,0 +1,87 @@
...
+  linkUrl,
+  linkText,
+  ...props
+}) => (
+  <Tabs
+    {...props}
+    className={styles.root}
+    headerClassName={styles.header}
+    addons={
+      linkUrl !== undefined && linkText !== undefined ? (
```
ovflowd:
```suggestion
      linkUrl && linkText && (
```

No need to check for undefined and there's no need for a ternary. As if any of these fail, they will be undefined and pass undefined.

components/Common/CodeBox/index.module.css
```diff
@@ -0,0 +1,92 @@
...
+          min-w-0
+          pl-8;
+
+        & > span {
+          @apply whitespace-break-spaces
+            break-words;
+        }
+
+        &::before {
+          /* Use an empty content with inline-block so the empty line renders
```
ovflowd: I don't think we need an inline comment here

components/Common/CodeBox/index.stories.tsx
```diff
@@ -0,0 +1,39 @@
...
+
+type Story = StoryObj<typeof CodeBox>;
+type Meta = MetaObj<typeof CodeBox>;
+
+export const Default: Story = {
+  args: {
+    language: 'JavaScript (CJS)',
+    children: (
+      <code>
+        {`const http = require('http');
```
ovflowd: Extract to a constant

components/Common/CodeBox/index.tsx
```diff
@@ -0,0 +1,107 @@
...
+export const MDXCodeBox: FC<PropsWithChildren> = ({ children: code }) => {
+  const props = isValidElement(code) ? code.props : {};
+  const { 'data-language': language } = props;
+
+  return (
+    <CodeBox language={getCodeLanguageDisplayName(language)}>{code}</CodeBox>
+  );
+};
+
+function transformCode(code: ReactNode): ReactNode {
```
ovflowd: Why exactly do we need this?

providers/mdxProvider.tsx
```diff
@@ -1,14 +1,20 @@
...
+import { MDXCodeBox } from '@/components/Common/CodeBox';
+import { MDXCodeTabs } from '@/components/Common/CodeBox/CodeTabs';
 import NodeApiVersionLinks from '@/components/Docs/NodeApiVersionLinks';
 
-const mdxComponents: MDXComponents = {
+export const mdxComponents: MDXComponents = {
   NodeApiVersionLinks: NodeApiVersionLinks,
   blockquote: ({ children }) => <div className="highlight-box">{children}</div>,
+  pre: ({ children }) => <MDXCodeBox>{children}</MDXCodeBox>,
+  codeTabs: ({ children, ...props }: ComponentProps<typeof MDXCodeTabs>) => (
```
ovflowd: Code Tabs is not a native Markdown element. What is this and why is it needed here?

If the point is to make it easy to invoke `<CodeTabs />` from any MDX document, then it should be `CodeTabs` not `codeTabs`

components/Common/CodeBox/mdx.stories.tsx
ovflowd, line 1: This should be in the same `index.stories.tsx` IMO. But I guess it's fine being on a separate file

components/Common/CodeBox/mdx.stories.tsx
```diff
@@ -0,0 +1,88 @@
+import type { Meta as MetaObj, StoryObj } from '@storybook/react';
+import { MDXRemote } from 'next-mdx-remote';
+import { serialize } from 'next-mdx-remote/serialize';
+import { setCDN } from 'shiki';
+
+import { nextRehypePlugins } from '@/next.mdx.mjs';
+import { mdxComponents } from '@/providers/mdxProvider';
+
+setCDN('https://unpkg.com/shiki');
```
ovflowd: Why? We should **never** need to do this. That's the reason `shiki.config.mjs` exists.

mdx/plugins/codeTabsResolver.mjs
ovflowd, line 1: I'm not sure this should be a in a new folder called `mdx/plugins`...

You could for now simply put it on root directory and call it `next.mdx.codebox.mjs`

ovflowd:
The implementation is looking fantastically good! I left a few minor comments, but besides that, this is great!

ovflowd:
FYI @devjvao Next.js build is failing (the full build) and storybook build also apparently crashed.

Can you run:

```
npm run deploy
npm run storybook:build
```

Just to debug these locally? Ty!

ovflowd:
It'd be great if we could uppercase these

image

ovflowd:
Also, the More Options text color looks wrong.

image

I'd recommend double-checking everything on Figma just to be ✅ that the designs are 👌

ovflowd:
Also, missing a border-radius here 👀...


image

It is also missing a border on the dark theme 👀 (Image below how it's supposed to look on dark and white)

image

image

components/Common/CodeBox/CodeTabs/index.stories.tsx
devjvao: The available props to control the story. In the case of CodeBox and CodeTabs there is nothing controllable, so by not including any prop it leaves us with an empty control tab.

image

components/Common/CodeBox/index.tsx
devjvao: It is transforming a raw code element by breaking it down and building the content using `spans`. This way it is possible to show the line numbers.

Example:

Given the following code content:

image

It is transformed to the following DOM structure:

image

devjvao: (only for non mdx content)

providers/mdxProvider.tsx
devjvao: That's the point. I'll rename it. 

This element is used in the `codeTabsResolver`, making it easy to build the tabs structure.

mdx/plugins/codeTabsResolver.mjs
devjvao: It sound good to me.

`next.mdx.codebox.mjs` or `next.mdx.codetabs.mjs`?

ovflowd: `next.mdx.codebox` imo.

components/Common/CodeBox/index.tsx
ovflowd: Right that makes sense, can you make it a arrow function instead of regular function. We avoid using regular functions on TypeScript files (just for consistency)

And it'd be great if you document inline what this function does.

components/Common/CodeBox/CodeTabs/index.stories.tsx
ovflowd: I don't think you even need to add this. It should still work normally (just by removing it) imo.

devjvao:
> It'd be great if we could uppercase these
> 
> image

This means that every tab with MDX will be in uppercase. Is it ok?

```
```javascript -> JAVASCRIPT
```
```
```mjs -> MJS
```

ovflowd:
> > It'd be great if we could uppercase these
> 
> > 
> 
> > image
> 
> 
> 
> This means that every tab with MDX will be in uppercase. Is it ok?
> 
> 
> 
> ```
> 
> ```javascript -> JAVASCRIPT
> 
> ```
> 
> ```
> 
> ```mjs -> MJS
> 
> ```

No. Only if no displayName is provided.

Ie:

````
```bash displayName="macOS"

some bash
```
````
Should render macOS

But in case I did the same snippet without displayName then it would render "BASH"

**Comment by devjvao** [comment]
I have a problem with rendering the MDX content in storybook...

I'm using `loaders` to compile the MDX content to pass it down to the `MDXRemote`.

image

This runs on the client side, and `shikiji` uses node-related imports, so it's breaking when entering any page.

image

It would be nice if storybook had support for SSR.

So... can you guide me on how to proceed? Should I just remove this story? Do you have any other ideas? @ovflowd

**Comment by ovflowd** [comment]
> This runs on the client side, and shikiji uses node-related imports, so it's breaking when entering any page.

Shikiji doesn't use Node.js imports. It's actually the opposite of what Shiki does. Apparently `shikiji-compat` does use `node:fs`

What if we use `rehype-shikiji` instead of `rehype-pretty-code`? https://github.com/antfu/shikiji/blob/main/packages/rehype-shikiji/README.md let me give an 👀 on that and work on a fix 🙇

**Comment by ovflowd** [comment]
Hey @devjvao would you mind joining the OpenJS Slack -> #nodejs-website channel? Would love to chat with you there :)

File: components/Common/CodeBox/CodeTabs/index.tsx [hunk 26]
```diff
@@ -0,0 +1,101 @@
...
+import type {
+  ComponentProps,
+  FC,
+  PropsWithChildren,
+  ReactElement,
+} from 'react';
+import { useMemo, useState } from 'react';
+
+import Tabs from '@/components/Common/Tabs';
+import { Link } from '@/navigation.mjs';
```
ovflowd: With the latest version of main the right Link component is `components/Link`

components/Common/CodeBox/CodeTabs/index.tsx
```diff
@@ -0,0 +1,101 @@
...
+  </Tabs>
+);
+
+type MDXCodeTabsProps = {
+  children: Array<ReactElement>;
+  languages: string;
+  displayNames?: string;
+};
+
+export const MDXCodeTabs: FC<MDXCodeTabsProps> = props => {
```
ovflowd: If MDXCodeTabs, and MDXCodeBox were inside `components/MDX` instead of `components/Common` :) as they are MDX specific.

ovflowd: Also, the MDXCodeTabs component should be a client-side component by using `"use client";` the other ones don't need to.

components/Common/CodeBox/index.tsx
```diff
@@ -0,0 +1,119 @@
...
+      </div>
+    </div>
+  );
+};
+
+type MDXCodeBoxProps = {
+  className?: string;
+};
+
+export const MDXCodeBox: FC<PropsWithChildren<MDXCodeBoxProps>> = ({
```
ovflowd: This should be inside `components/MDX`

components/Common/CodeBox/index.tsx
```diff
@@ -0,0 +1,119 @@
...
+          <div className={styles.notification}>
+            <CodeBracketIcon className={styles.icon} />
+            {t('components.common.codebox.copied')}
+          </div>
+        ),
+      });
+    }
+  };
+
+  const code = useMemo(() => transformCode(children), [children]);
```
ovflowd: Is memoization needed here? I wonder how often this CodeBox would get re-rendered? Afaik the children and language props would never change.

Hence, no memorization is needed here

components/Common/CodeBox/index.tsx
```diff
@@ -0,0 +1,119 @@
...
+  language: string;
+};
+
+export const CodeBox: FC<PropsWithChildren<CodeBoxProps>> = ({
+  children,
+  language,
+}) => {
+  const ref = useRef<HTMLPreElement>(null);
+
+  const notify = useNotification();
```
ovflowd: This is a client-side, hook. Which makes this whole component dependent on a client-side effect.

Hence, this component needs to be a `"use client";`

ovflowd: I wonder how that would affect us, as pretty much, CodeBox is something so important, that making it client-only feels like a limitation. But I don't see an easy alternative at the moment, so let it be.

next.mdx.shiki.mjs
```diff
@@ -19,8 +19,150 @@ const memoizedShikiji = await getHighlighterCore({
...
+        const compensatedRange = {
+          start: range.start - lengthOffset,
+          end: range.end - lengthOffset,
+        };
+        const deleteCount = compensatedRange.end - compensatedRange.start + 1;
+
+        // Replace the sequential code boxes with a code tabs element
+        children.splice(compensatedRange.start, deleteCount, {
+          type: 'element',
+          tagName: 'CodeTabs',
```
ovflowd, line 155: You're making an implicit convertion to `CodeTabs`? Is this the best path here?

This pretty much creates an invisible requirement of always providing the CodeTabs component. I'm fine with it, but I wonder how... performatic this is

next.mdx.shiki.mjs
```diff
@@ -19,8 +19,150 @@ const memoizedShikiji = await getHighlighterCore({
...
+  return Object.entries(rangeMap).map(([start, end]) => ({
+    start: Number(start),
+    end: Number(end),
+  }));
+}
+
 export default function rehypeShikiji() {
   return async function (tree) {
+    // Retrieve all sequential code boxes to transform
+    const ranges = getCodeTabsRange(tree);
```
ovflowd, line 101: You're aware this change will have an immediate effect on the legacy website right? I wonder what the penalty in performance will be and what changes this might imply.

components/Common/CodeBox/index.tsx
devjvao: Should the component be renamed to `CodeBox`? Since it's going to be inside the mdx dir.

next.mdx.shiki.mjs
```diff
@@ -19,8 +19,150 @@ const memoizedShikiji = await getHighlighterCore({
...
+ */
+function getCodeTabsRange(tree) {
+  const rangeMap = {};
+  let start = null;
+
+  visit(tree, 'element', (node, index, parent) => {
+    // Adding 2 since there is one text node between every element
+    const next = index + 2;
+
+    if (
```
ovflowd: Can you separate this big if into maybe smaller if statements as I've done in the same file? And explain why these if statements exist and why we check for such things?

next.mdx.use.mjs
```diff
@@ -28,4 +31,7 @@ export const mdxComponents = {
 export const htmlComponents = {
   a: Link,
   blockquote: ({ children }) => <div className="highlight-box">{children}</div>,
+  pre: ({ children, ...props }) => (
```
ovflowd, line 38: Is this needed? I think just `pre: MDXCodeBox` should work, imo.

components/Common/CodeBox/mdx.stories.tsx
```diff
@@ -0,0 +1,76 @@
+import type { Meta as MetaObj, StoryObj } from '@storybook/react';
+import { VFile } from 'vfile';
+
+import { MDXRenderer } from '@/components/mdxRenderer';
+import { compileMDX } from '@/next.mdx.compiler.mjs';
+
+type Props = {
```
ovflowd: Instead of this, just make `type Story = StoryObj<PropsWithChildren>`

components/Common/CodeBox/index.tsx
```diff
@@ -0,0 +1,119 @@
...
+const transformCode = (code: ReactNode): ReactNode => {
+  if (!isValidElement(code)) {
+    // Early return when the `CodeBox` child is not a valid element since the
+    // type is a ReactNode, and can assume any value
+    return code;
+  }
+
+  const content = code.props?.children;
+
+  if (code.type !== 'code' || typeof content !== 'string') {
```
ovflowd, line 28: And in which situations would that happen?

components/Common/CodeBox/index.tsx
```diff
@@ -0,0 +1,119 @@
...
+
+  return (
+    <CodeBox language={getCodeLanguageDisplayName(language)}>{code}</CodeBox>
+  );
+};
+
+// Transforms a code element with plain text content into a more structured
+// format for rendering with line numbers
+const transformCode = (code: ReactNode): ReactNode => {
+  if (!isValidElement(code)) {
```
ovflowd, line 20: When would that happen? Is this just a security check or can that actually happen?

components/Common/CodeBox/index.tsx
```diff
@@ -0,0 +1,119 @@
...
+  children,
+  language,
+}) => {
+  const ref = useRef<HTMLPreElement>(null);
+
+  const notify = useNotification();
+  const [, copyToClipboard] = useCopyToClipboard();
+  const t = useTranslations();
+
+  const onCopy = () => {
```
ovflowd: I'm almost thinking that we should make the whole onCopy just a `prop` (callback) and let it be defined somewhere else. It is unpractical to make this a client-side component just because of the copy part of the codebox.

Then we can make MDXCodeBox the actual component that implements the `onCopy` (provides a valid callback) and the actual one to be "client-side". Since transformCode is done on the base CodeBox component, that would reduce the burden of the client-side of "parsing" and mapping the codebox for every CodeBox.

ovflowd:
@devjvao can you rebase your branch locally?

Apparently it is failing due to this:

```
Failed to compile.
--
19:15:50.099 |  
19:15:50.100 | ./components/Common/CodeBox/CodeTabs/index.tsx
19:15:50.100 | ReactServerComponentsError:
19:15:50.100 |  
19:15:50.100 | You're importing a component that needs useState. It only works in a Client Component but none of its parents are marked with "use client", so they're Server Components by default.
19:15:50.100 | Learn more: https://nextjs.org/docs/getting-started/react-essentials
19:15:50.100 |  
19:15:50.100 | ,-[/vercel/path0/components/Common/CodeBox/CodeTabs/index.tsx:6:1]
19:15:50.100 | 6 \|   PropsWithChildren,
19:15:50.100 | 7 \|   ReactElement,
19:15:50.102 | 8 \| } from 'react';
19:15:50.102 | 9 \| import { useMemo, useState } from 'react';
19:15:50.102 | :                   ^^^^^^^^
19:15:50.102 | 10 \|
19:15:50.102 | 11 \| import Tabs from '@/components/Common/Tabs';
19:15:50.103 | 12 \| import { Link } from '@/navigation.mjs';
19:15:50.103 | `----
19:15:50.103 |  
19:15:50.103 | Maybe one of these should be marked as a client entry with "use client":
19:15:50.103 | ./components/Common/CodeBox/CodeTabs/index.tsx
19:15:50.103 | ./next.mdx.use.mjs
19:15:50.103 | ./components/mdxRenderer.tsx
19:15:50.103 | ./app/[locale]/[[...path]]/page.tsx
19:15:50.103 |  
19:15:50.103 | ./components/Common/CodeBox/index.tsx
19:15:50.103 | ReactServerComponentsError:
19:15:50.103 |  
19:15:50.103 | You're importing a component that needs useRef. It only works in a Client Component but none of its parents are marked with "use client", so they're Server Components by default.
19:15:50.103 | Learn more: https://nextjs.org/docs/getting-started/react-essentials
19:15:50.103 |  
19:15:50.103 | ,-[/vercel/path0/components/Common/CodeBox/index.tsx:4:1]
19:15:50.103 | 4 \| } from '@heroicons/react/24/outline';
19:15:50.103 | 5 \| import { useTranslations } from 'next-intl';
19:15:50.104 | 6 \| import type { FC, PropsWithChildren, ReactNode } from 'react';
19:15:50.104 | 7 \| import { Fragment, isValidElement, useMemo, useRef } from 'react';
19:15:50.104 | :                                             ^^^^^^
19:15:50.104 | 8 \|
19:15:50.104 | 9 \| import Button from '@/components/Common/Button';
19:15:50.104 | 10 \| import { useCopyToClipboard, useNotification } from '@/hooks';
19:15:50.104 | `----
19:15:50.104 |  
19:15:50.104 | Maybe one of these should be marked as a client entry with "use client":
19:15:50.104 | ./components/Common/CodeBox/index.tsx
19:15:50.104 | ./next.mdx.use.mjs
19:15:50.104 | ./components/mdxRenderer.tsx
19:15:50.104 | ./app/[locale]/[[...path]]/page.tsx
``` 

Which IMO means your component depends on client-side artifacts. `useState` is client-side only.

components/Common/CodeBox/index.tsx
ovflowd: Sure

ovflowd: `components/MDX/CodeBox/index.tsx`

next.mdx.shiki.mjs
ovflowd: Anyhow, what worries me the most here, is that with this change, we will "visit" the whole tree at least N times.

I understand why this is being done, but it feels **fragile** My worry here is that we might be overcomplicating everything here.

I wonder if instead of having all of this complexity, we could visit the root tree and simply go down:

```
-> root tree
  -> detects pre element
     -> goes down that pre element
       -> transforms into shiki
       -> add the resulting element to a temporary array
     -> seeks if there is another pre element
       -> if yes, same logic as below
  -> once goes all close codeboxes
     -> replaces the elements with the contents of the temporary array (the whole range)
  -> keeps going down the tree, identifying pre elements
```

With this IMO, we only need one `visit` call. Remember that `visit` goes by every element of the tree by itself, we don't need to call `visit` on deeper items of the tree... I think.

I just believe we can optimise this further, as I'm truly fearful of issues.

components/Common/CodeBox/index.tsx
devjvao: Makes sense.

components/Common/CodeBox/index.tsx
devjvao: It's more of a security check since it is a react node, and considering we are transforming react elements.

ovflowd: But it should not happen ever. I would rather not have this check and having the app crash if any new page that introduced something ungodly, than "ignoring"

Sometimes the best thing is to scream the error, let it crash, so we know that something is wrong

components/Common/CodeBox/index.tsx
devjvao: The code box component accepts any children, so it could be rendered with a div or whatever (ik it's not supposed to be used like that, but it does accept).

And the typeof of the content is to make sure we will transform text content, because of the split right above.

next.mdx.shiki.mjs
devjvao: I thought it was the only path, but I've managed it to work with the `pre` element. So the CodeTabs won't be needed in the mdx components map.

components/Common/CodeBox/index.tsx
ovflowd: I understand that, but I doubt anyone here would make misuse of this component.

Hence these checks add only extra logic that imo is not needed here. Since we know that during the runtime this shouldn't be possible, and if it is, we rather have this crashing and seek the source of the error

devjvao: Got it. It makes sense, I'll remove what is not needed.

The `typeof content !== 'string` condition is required, I'll just update the comment to explain it.

ovflowd:
Hey @devjvao just checking-in if you need support here or how's the progress going on here? 😄

devjvao:
> Hey @devjvao just checking-in if you need support here or how's the progress going on here? 😄

Hey! I'm sorry for taking it so long, I wasn't fully focused on this PR.

Regarding the last review: I've adjusted everything and rebased the branch locally, but I'm still having some trouble with the plugin logic. Doing all operations using a single visit is really tricky, but I'm progressing.

ovflowd:
> > Hey @devjvao just checking-in if you need support here or how's the progress going on here? 😄
> 
> Hey! I'm sorry for taking it so long, I wasn't fully focused on this PR.
> 
> Regarding the last review: I've adjusted everything and rebased the branch locally, but I'm still having some trouble with the plugin logic. Doing all operations using a single visit is really tricky, but I'm progressing.

No worries! If you need support from our side, just let us know!

ovflowd:
Hey @devjvao just checking in if you have had time to make any changes? And if you could update this PR with them? If you don't have time to finish this PR, I can takeover if you're OK with that. But I recall that you had a few changes already done locally and I wondered if you'd like to push them.

Thank you!

ovflowd (approved):
I've rebased and applied some bug fixes and the code works now 100% (It was already working just made some fixes).

Although I wish we could iterate this PR further and simplify the Shiki Plugin and other related aspects, I'm fine with the code as it is.

cc @nodejs/nodejs-website for anyone else interested in reviewing this PR.

araujogui (approved):

AugustinMauroy (approved):

ovflowd (changes requested):
@canerakdas your commit broke a few things:

- The code lines are not aligned anymore with the line numbers of the CodeBox
- There's a horizontal scrollbar on CodeBoxes... The line break was intentional as it's cleaner... Could you add it back? Thanks!

canerakdas:
> * The code lines are not aligned anymore with the line numbers of the CodeBox

It looks aligned but I added back the leadings 🤔 (Since the font family has been changed, it looks a little confusing in the Chromatic)

> * There's a horizontal scrollbar on CodeBoxes... The line break was intentional as it's cleaner... Could you add it back? 

I have reverted this code. Actually, the reason why I added this is that I think it makes it difficult to read the code since the code is not prettified in any way in mobile resolutions;

image

mikeesto (commented):
I agree with @canerakdas , the CodeTabs story looks awkward on a mobile resolution. If the code was formatted with prettier, we could set the printWidth to something that would be a better fit for a smaller screen?

ovflowd:
Im not sure applying prettier on the code snippets during runtime would be a good idea. Prettier is extremely costly.

We might be able to fix tabulation for mobile with pure CSS. But also, to be honest, this website is not something that I see people accessing from mobile. Node.js website is mostly accessed for downloading Node.js; and sure we have blog and learn content but I'd discourage people actually reading these from mobile anyways

mikeesto:
For me it's not a blocker to this PR, I'm in favour of seeing it land and overall it looks good. We can always tweak the styling in the future. I agree that many people probably visit the site on a desktop to download Node, but I'm sure we have some users on mobile too

ovflowd (approved):

PR #6072, opened 2023-10-31
feat: Added Navbar component

### LINKED ISSUE #6060: Create `Navbar` Component
Create a `Navbar` component using the new [figma design (direct link)](https://www.figma.com/file/pu1vZPqNIM7BePd6W8APA5/Node.js?type=design&node-id=415-8491&mode=dev). 

This component starts the responsive header for the new site.

---

### Before You Start...

---

### Details

image

- Create a new directory called `Navbar` within the [`sections` directory](https://github.com/nodejs/nodejs.org/tree/main/components/sections) that contains all the new code.
- Omit search fpr now, that is the scope of another story not yet started.
- Use the new `NavItem` component introduced within https://github.com/nodejs/nodejs.org/pull/5972
- Use the new `ThemeToggle` component introduced within https://github.com/nodejs/nodejs.org/pull/6007
- Use the new `LanguageDropdown` component introduced within https://github.com/nodejs/nodejs.org/pull/5982
- Use the new [social-github.svg](https://github.com/nodejs/nodejs.org/blob/main/public/static/images/logos/social-github.svg) along with [AccessibleIcon](https://www.radix-ui.com/primitives/docs/utilities/accessible-icon) to link to https://github.com/nodejs/node
- The figma puts the mobile breakpoint at 375px. Explore if we can use that directly or an [new/existing tailwind token (if any)](https://github.com/nodejs/nodejs.org/blob/main/tailwind.config.ts)
- The links must be internationalized. 
   - Add the following new keys to the `i18n/locales/en.json` file. 
      - `components.headers.links.learn` 
   - Use the existing `components.headers.links.[about,docs,download,blog,certification] keys 
- FormattedMessage from react-intl should be used to reference the keys. Look for examples elsewhere in the codebase.

### Storybook

Add a storybook story file `components/sections/Navbar/index.stories.tsx` which exercises each of the component's states.

There are 4 Navbar states to capture within styles and stories:

- Light
- Light, mobile
- Dark
- Dark, mobile

Light and dark modes toggle between https://github.com/nodejs/nodejs.org/blob/main/public/static/images/logos/horizontal-light.svg and https://github.com/nodejs/nodejs.org/blob/main/public/static/images/logos/horizontal-dark.svg. I imagine there is a way to do this with just a fill, but I didn't confirm.

### Open Questions

- Should the links be defined within `site.json`? Today many are in `navigation.json` but that's for the live site.

Rounak-stha:
## Description

This PR creates `Navbar`component. The figma design is followed and the component is tested visually.

## Validation

image

image

## Related Issues

Addresses #6060

### Check List


Rounak-stha:
Here are some of my concerns
- It is mentioned in the issue to add the key `components.headers.links.learn` in the `en.json` file has a key `components.header.links.learn`. Notice the header(s) in the key mentioned in the issue
- Currently, the navbar has seven items, but in the new design, only six are used; the `Get Involved` nav item is removed. The current `navigation.json` file has all seven items and like in the current implementation of `Header`, I've used the `useNavigation` hook but exported new items from it to only get items listed in the new design.
- The navigation item text for `blog` is `blog` in design file but the formattedMessage is `news` 
- I didn't see the nav items design for mobile. The design hides the nav items and has added a Hambuger icon; how should the items be shown on open state on mobile.

components/sections/Navbar/index.tsx
```diff
@@ -0,0 +1,72 @@
...
+      <div className="flex items-center gap-6">
+        <ThemeToggle
+          onClick={() => setTheme(theme === 'dark' ? 'light' : 'dark')}
+        />
+        <LanguageDropdown
+          availableLanguages={availableLanguages}
+          currentLanguage={currentLanguage}
+        />
+        <AccessibleIcon.Root label={'Nodejs.org Github'}>
+          <a href="https://github.com/nodejs/node">
```
AugustinMauroy: use `Link` on a next project Link is more optimized

.storybook/preview.tsx
```diff
@@ -32,9 +33,11 @@ const preview: Preview = {
       <SiteProvider>
         <LocaleProvider>
           <NotificationProvider viewportClassName="absolute top-0 left-0 list-none">
-            <div className={rootClasses}>
-              <Story />
-            </div>
+            <ThemeProvider>
```
AugustinMauroy: In my opinion this should be put on `decorators` of navbar stories

components/sections/Navbar/index.tsx
```diff
@@ -0,0 +1,72 @@
...
+  const currentLanguage = useMemo(
+    () => ({
+      name: currentLocale.name,
+      code: currentLocale.code,
+    }),
+    [currentLocale]
+  );
+
+  return (
+    <nav className="flex h-16 justify-between border-b-2 border-neutral-200 px-8 dark:border-0">
```
AugustinMauroy: Read [How to style a component](https://github.com/nodejs/nodejs.org/blob/main/COLLABORATOR_GUIDE.md#styling-a-component) on our contribution guideline

components/sections/Navbar/index.tsx
Rounak-stha: I didn't use the `Link` component here because it's linking to an external site. Yes, the `Link` would still work but aren't all the optimizations for client-side navigation within the SPA?

components/sections/Navbar/index.tsx
Rounak-stha: hey, I missed a part here. I meant to draft this PR but still wanted reviews as I had some concerns. Since adding a class on the markup itself is convenient during development, I decided to create css modules after all styles are fixed. Thank you for mentioning this.

Rounak-stha: Hey, I decided to mark this PR ready for review, I'll create modules css right away.

components/sections/Navbar/index.tsx
ovflowd: I agree that we should only use an `a` here.

.storybook/preview.tsx
ovflowd: We shouldn't use `ThemeProvider` on Storybook. We intentionally want the Theme Change to be done on Storybook Controls (as you can see on the "Dark / Light" theme chooser.

Hence, to make the Component happy because you're using `useTheme` you could probably either mock the ThemeProvider or something else, since by using the ThemeProvider the component/story becomes flaky (as ThemeProvider (next-themes)) injects some stuff on the HTML element.

components/sections/Navbar/index.tsx
```diff
@@ -0,0 +1,76 @@
...
+import LanguageDropdown from '@/components/Common/LanguageDropDown';
+import ThemeToggle from '@/components/Common/ThemeToggle';
+import NavItem from '@/components/sections/NavItem';
+import { useLocale } from '@/hooks/useLocale';
+import { useNavigation } from '@/hooks/useNavigation';
+
+import style from './index.module.css';
+
+const Navbar: React.FC = () => {
+  const { newNavigationItems } = useNavigation();
```
ovflowd: Instead of using Hooks can we pass these all as component props?

The idea behind is that these hooks will be invoked on the Parent Component. So that we don't need to provide Providers (such as the `ThemeProvider` on the NavBar.

This is the approach we've been doing on the components here.

components/sections/Navbar/index.tsx
```diff
@@ -0,0 +1,76 @@
...
+
+import LanguageDropdown from '@/components/Common/LanguageDropDown';
+import ThemeToggle from '@/components/Common/ThemeToggle';
+import NavItem from '@/components/sections/NavItem';
+import { useLocale } from '@/hooks/useLocale';
+import { useNavigation } from '@/hooks/useNavigation';
+
+import style from './index.module.css';
+
+const Navbar: React.FC = () => {
```
ovflowd:
```suggestion
const NavBar: FC = () => {
```

^ import the `FC` type, don't rely on React globals.

components/sections/Navbar/index.tsx
```diff
@@ -0,0 +1,76 @@
...
+      </div>
+      <div className={style.rightItems}>
+        <ThemeToggle
+          onClick={() => setTheme(theme === 'dark' ? 'light' : 'dark')}
+        />
+        <LanguageDropdown
+          availableLanguages={availableLanguages}
+          currentLanguage={currentLanguage}
+        />
+        <AccessibleIcon.Root label={'Nodejs.org Github'}>
```
ovflowd: I don't think we need to use AccessibleIcon here, as it is a button we can simply add the `aria-label` to the Button itself. And the label could be `aria-label="Node.js GitHub"`

components/sections/Navbar/index.tsx
```diff
@@ -0,0 +1,76 @@
...
+          <button className={style.iconWrapper}>
+            <a href="https://github.com/nodejs/node">
+              <Image alt="Github Icon" width={20} height={20} src={GHLogoSrc} />
```
ovflowd: No need for alt here imho

hooks/useNavigation.tsx
```diff
@@ -63,5 +63,10 @@ export const useNavigation = () => {
         context
       );
     },
+    newNavigationItems: (() => {
```
ovflowd: Why removing Get Involved from the Nav Bar? I know it's not on the Figma but we should render the Navigation as it is.

We will in the future rearrange the items and rename them, so this method is not needed.

components/sections/Navbar/index.tsx
Rounak-stha: We actually get an eslint warning warning if we miss to pass the alt.

image

components/sections/Navbar/index.tsx
```diff
@@ -0,0 +1,76 @@
...
+import { useNavigation } from '@/hooks/useNavigation';
+
+import style from './index.module.css';
+
+const Navbar: React.FC = () => {
+  const { newNavigationItems } = useNavigation();
+  const { availableLocales, currentLocale } = useLocale();
+  const { theme, setTheme } = useTheme();
+
+  const GHLogoSrc = `/static/images/logos/social-github${
```
ovflowd: Also you shouldn't use this to decide which image to render.

I recommend doing what we did here: https://github.com/nodejs/nodejs.org/blob/main/components/Common/ThemeToggle/index.tsx#L25-L26

Rendering the two images, but using Tailwind Classes to hide one or another depending of the current Theme :)

components/sections/Navbar/index.tsx
ovflowd: Sorry, not sure what you're talking about. You should pass `alt` on the `Image` element.

Rounak-stha: You suggested that there's no need for an alt but the alt must must be passed to Image element as you just mentioned.

components/sections/Navbar/index.tsx
Rounak-stha: Thank you for this suggestion. This really simplifies a lot of things.

.storybook/preview.tsx
Rounak-stha: I could only find a solution of decorating the story at some level with the provider. In the storybook, the component does not yell when it's not wrapped with the `ThemeProvider` except that the theme-toggle-button can't do its job without a provider. Using the Storybook theme control is working for now using your suggestion here
> I recommend doing what we did here: https://github.com/nodejs/nodejs.org/blob/main/components/Common/ThemeToggle/index.tsx#L25-L26

components/sections/Navbar/index.tsx
```diff
@@ -0,0 +1,93 @@
...
+import { useMemo } from 'react';
+
+import LanguageDropdown from '@/components/Common/LanguageDropDown';
+import ThemeToggle from '@/components/Common/ThemeToggle';
+import NavItem from '@/components/sections/NavItem';
+import { useLocale } from '@/hooks/useLocale';
+
+import style from './index.module.css';
+
+type NavItem = { key: string; link: string };
```
Rounak-stha: The name for this type is used by a component imported here. Although, this does not cause any errors should we use a different name for the type or should we just merge the type where it's used.
If we have to rename, I could really use a suggestion here.

ovflowd: This should be avoided. You should do:

```ts
type NavbarProps = { navItems: ComponentProps<NavItem>[] };
```

components/sections/Navbar/index.tsx
```diff
@@ -0,0 +1,93 @@
...
+type NavbarProps = { navItems: NavItem[] };
+
+const Navbar: FC<NavbarProps> = ({ navItems }) => {
+  const { availableLocales, currentLocale } = useLocale();
```
ovflowd: As I mentioned locale items should also be a prop.

And you can simply make it:

```tsx
type NavbarProps = { 
  navItems: ComponentProps<NavItem>[];
  languages: Pick<LocaleConfig, 'code' | 'name'>[];
  currentLanguage: string; // just the language code
  currentTheme: 'dark' | 'light';
  onThemeTogglerClick: () => void; 
};
```

components/sections/Navbar/index.tsx
Rounak-stha: The `NavItem` component accepts `children` as a prop which is displayed as text. 
so, I should do something like this.

```TS
type NavbarProps = {
  navItems: (Omit<ComponentProps<typeof NavItem>, 'children'> & {
    text: string;
  })[];
  localeConfig: ComponentProps<typeof LanguageDropdown>;
};
```

components/sections/Navbar/index.tsx
Rounak-stha: The `NavItem` component accepts `children` as a prop which is displayed as text. 
so, I should do something like this.

```TS
type NavbarProps = {
  navItems: (Omit<ComponentProps<typeof NavItem>, 'children'> & {
    text: string;
  })[];
  languages: Pick<LocaleConfig, 'code' | 'name'>[];
  currentLanguage: string; // just the language code
  currentTheme: 'dark' | 'light';
  onThemeTogglerClick: () => void; 
```

ovflowd:
```
navItems: (Omit<ComponentProps<typeof NavItem>, 'children'> & {
    text: string;
  })[];
```

Uh very interesting. NavItem uses children? I wonder if that was intentional per design...

So no, you're right here, we should not do `ComponentProps` in this case, but do the type you've made (`type NavItem = { text, href }`)

components/sections/Navbar/index.tsx
ovflowd: Right? :)

components/sections/Navbar/index.tsx
ovflowd: Oh no, sorry, yes we need an alt!

.storybook/preview.tsx
ovflowd: Right, let's not use the `useTheme` then for this Component as mentioned on below discussions. So that the Component is pretty much agnostic (as the LanguageDropdown and ThemeToggle) components are.

components/sections/Navbar/index.tsx
Rounak-stha: Here's what the latest NavBarProp looks like:
```TS
type NavbarProps = {
  navItems: NavItem[];
  languages: ComponentProps<typeof LanguageDropdown>;
  onThemeTogglerClick: () => void;
};
```

- The `LanguageDropdown` component requires `currentLanguage` to be of type `{ name, code }`, though it only uses code. So, I used `ComponentProps`.
- `currentTheme` is not required now as it was before used to change theme on the `themeTogglerCallback` but now it is passed as prop.

Also, there's this
> I didn't see the nav items design for mobile. The design hides the nav items and has added a Hambuger icon; how should the items be shown on open state on mobile.

ovflowd: > The LanguageDropdown component requires currentLanguage to be of type { name, code }, though it only uses code. So, I used ComponentProps.

I guess you can change the type of LanguageDropdown and make `currentLanguage` to be a string :) (Just to simplify the prop, if you want!)

> currentTheme is not required now as it was before used to change theme on the themeTogglerCallback but now it is passed as prop.

Right, afaik, we don't really need to pass currentTheme, we can simply use Tailwind to render icons with dark/light mode only if needed.

> I didn't see the nav items design for mobile. The design hides the nav items and has added a Hambuger icon; how should the items be shown on open state on mobile.

The hamburger designs are in this page: https://www.figma.com/file/pu1vZPqNIM7BePd6W8APA5/Node.js?node-id=696%3A19655&mode=dev

components/sections/Navbar/index.tsx
```diff
@@ -0,0 +1,77 @@
...
+  languages: ComponentProps<typeof LanguageDropdown>;
+  onThemeTogglerClick: () => void;
+};
+
+const Navbar: FC<NavbarProps> = ({
+  navItems,
+  languages,
+  onThemeTogglerClick,
+}) => {
+  return (
```
ovflowd: No need of a return statement and `{}` surrounding the component, as its just component here

components/sections/Navbar/index.tsx
```diff
@@ -0,0 +1,77 @@
...
+
+type NavItem = { text: string; href: string };
+
+type NavbarProps = {
+  navItems: NavItem[];
+  languages: ComponentProps<typeof LanguageDropdown>;
+  onThemeTogglerClick: () => void;
+};
+
+const Navbar: FC<NavbarProps> = ({
```
ovflowd:
```suggestion
const NavigationBar: FC<NavbarProps> = ({
```

I feel if wename this NavigationBar it feels more meaningful (not to mention we should fix the casing)

components/sections/Navbar/index.tsx
```diff
@@ -0,0 +1,77 @@
...
+          height={24}
+          width={80}
+          src="/static/images/logos/horizontal-dark.svg"
+          alt="Node.js"
+        />
+        <Image
+          className="hidden dark:block"
+          height={24}
+          width={80}
+          src="/static/images/logos/horizontal-light.svg"
```
ovflowd:
```suggestion
          src={`${BASE_PATH}/static/images/logos/horizontal-light.svg`}
```

For all these links, important to add the `BASE_PATH` variable from `next.constants.mjs` as sometimes deployments happen in subfolders, not the / of domains.

ovflowd: (Applies for all the calls of `/static` here

ovflowd (dismissed):
We're getting pretty close of the final result. Just some changes as I mentioned here, and the comments and (yes the mobile designs) and then we're here :D

ovflowd:
BTW a few style issues:

- The bottom border of the NavBar should be 1px
- The Node.js logo is unaligned with the remaining of the items
- The GitHub Button/Icon colour is wrong (you can use color attribute) and hovering should create same effect as on the ThemeToggler/LanguageDropdown

components/sections/Navbar/index.stories.tsx
```diff
@@ -0,0 +1,44 @@
...
+import Navbar from './index';
+
+type Story = StoryObj<typeof Navbar>;
+type Meta = MetaObj<typeof Navbar>;
+
+export const Default: Story = {
+  args: {
+    navItems: [
+      {
+        text: 'about',
```
ovflowd: Could you write these entries as on Figma?

Learn, About, Docs, Download, Blog, Certification (in this order and this casing) (You also forgot to add the Download entry)

Rounak-stha:
>     * The GitHub Button/Icon colour is wrong (you can use color attribute) and hovering should create same effect as on the ThemeToggler/LanguageDropdown

For the GithHub icon, I am using the svg in the `public/static/..` folder.  Should I change the fill here?

ovflowd:
> >     * The GitHub Button/Icon colour is wrong (you can use color attribute) and hovering should create same effect as on the ThemeToggler/LanguageDropdown
> 
> 
> 
> For the GithHub icon, I am using the svg in the `public/static/..` folder.  Should I change the fill here?
> 
> 
> 
> 

No. Simply have a className in place that changes the colour of the fill/svg. If the colour is statically defined on the SVG, let me know and I can make a fix that allows the colour from being defined by css.

ovflowd:
cc @bmuenzenmeyer 👆 another reason why I believe our svgs should be react components and not img elements.

components/sections/Navbar/index.tsx
Rounak-stha: Should rename the filename roo, right?

Rounak-stha:
> If the colour is statically defined on the SVG, let me know

Yes, it's statically defined. I think changing it to `currentColor` would do if we're changing fill color with classname. If updating tht is within the scope of this PR, I could do that.

components/sections/Navbar/index.tsx
Rounak-stha: Should I use `@radix-ui/react-dropdown-menu` as in the `LanguageDropDown` component for the mobile nav menu? though that would result in a few duplications of markups?

ovflowd: > Should I use `@radix-ui/react-dropdown-menu` as in the `LanguageDropDown` component for the mobile nav menu? though that would result in a few duplications of markups?

No. The mobile NavBar is not a dropdown. Let me think about how we should approach this for a little while.

Rounak-stha: Hey, I have added the mobile view. I used CSS sibling selector to show and hide the items but to change the icon from Ham to XMark, I had to use state.

ovflowd: Let me review that. I was actually thinking to use the `useMediaQuery` hook to observe on media query changes.

Or simply tailwindcss to mutate the styles, but let me see what you did :)

components/sections/Navbar/index.tsx
ovflowd: Right

components/sections/NavigationBar/index.tsx
```diff
@@ -0,0 +1,105 @@
...
+            src={`${BASE_PATH}/static/images/logos/horizontal-light.svg`}
+            alt="Node.js"
+          />
+        </div>
+        <label
+          onClick={() => setShowHamIcon(prev => !prev)}
+          className={style.sidebarItemTogglerLabel}
+          htmlFor="sidebarItemToggler"
+        >
+          {showHamIcon ? (
```
ovflowd: What if we have a simply silly map:

```tsx
// outside the component
const navInteractionIcons = {
  'show': Hamburger,
  'close': XMark,
}

// in the component
{navInteractionIcons[isOpen ? 'close' : 'show']}
```

And let's rename the state from `showHamIcon` to `isOpen` or `isMenuOpen`

ovflowd:
> > If the colour is statically defined on the SVG, let me know
> 
> Yes, it's statically defined. I think changing it to `currentColor` would do if we're changing fill color with classname. If updating tht is within the scope of this PR, I could do that.

Feel free to do it :)

components/sections/NavigationBar/index.tsx
```diff
@@ -0,0 +1,105 @@
...
+            className="hidden dark:block"
+            height={24}
+            width={80}
+            src={`${BASE_PATH}/static/images/logos/horizontal-light.svg`}
+            alt="Node.js"
+          />
+        </div>
+        <label
+          onClick={() => setShowHamIcon(prev => !prev)}
+          className={style.sidebarItemTogglerLabel}
```
ovflowd: Why are we using a label element instead of button here? 🤔

ovflowd: Because of the nature of making it show/hide without any javascript interaction?

If yes, that's great. But can we use this https://www.radix-ui.com/primitives/docs/components/label then? and also make the cursor a pointer?

ovflowd (commented):
This is looking really good, thank you so much for the work so far!

Rounak-stha:
>I think changing it to `currentColor` would do if we're changing the fill color with class name. 

Turns out, when `img` tag is rendering the `svg`, we can't change the color using classname. The `svg` has to be the child of the tag where we're setting the colors. So, now we either have to hardcode the color in the `svg` file or create `svg component`.

And if we're changing the fill with class name, we don't need to 2 `svg` file for each `light` and `dark` mode.

What do you think we should do?

Rounak-stha:
> This is looking really good, thank you so much for the work so far!

Really happy to be contributing to the project. Thank you for your help and patience.

ovflowd:
> Turns out, when img tag is rendering the svg, we can't change the color using classname. The svg has to be the child of the tag where we're setting the colors. So, now we either have to hardcode the color in the svg file or create svg component.

Feel free to make this SVG (GitHub one) a Component, and add it under `components/Icons`.  (I am going anyways to open an issue soon or @bmuenzenmeyer to make all current Icon SVGS (Social Logos and Node.js logos) React Components under `components/Icons`

Rounak-stha:
> Feel free to make this SVG (GitHub one) a Component,

The component expects just one prop (size) for now. There might be some design decisions on structuring the component especially it's props when a dedicated issue is created. But please let me know if I should do any updates.

ovflowd:
> > Feel free to make this SVG (GitHub one) a Component,
> 
> 
> 
> The component expects just one prop (size) for now. There might be some design decisions on structuring the component especially it's props when a dedicated issue is created. But please let me know if I should do any updates.
> 
> 

You mean the SVG? Why? 🤔

SVGs even as a react component should support whatever size you enforce with CSS. Afaik we should only define on a SVG their default sizes and viewboxes 👀

Rounak-stha:
> SVGs even as a react component should support whatever size you enforce with CSS. Afaik we should only define on a SVG their default sizes and viewboxes 👀

So, the only prop the SVG component should expect is `className`?

ovflowd:
> > SVGs even as a react component should support whatever size you enforce with CSS. Afaik we should only define on a SVG their default sizes and viewboxes 👀
> 
> So, the only prop the SVG component should expect is `className`?

Nope. You still the SVG by simply styling the SVG tag:

```css

svg {
  ... 
}
```

I think in your case it'd be:

```css

.nodeIconWrapper {
  svg {

  }
}
```

Note, that IMO you can make the component be:

```tsx
const HorizontalDarkLogo: FC<HTMLAttributes<HTMLSVGSVGElement>> = props => <SVG regularProps.. {...props}> ... </SVG>
```

So that we can easily pass props to the element without the need of declaring them

components/Icons/Github.tsx
ovflowd, line 1: Can this be named `GitHubLogo`? (With this exact casing)

components/Icons/Github.tsx
```diff
@@ -0,0 +1,18 @@
+import type { FC, HTMLAttributes } from 'react';
+
+const Github: FC<HTMLAttributes<SVGElement>> = props => {
+  return (
```
ovflowd: FYI, no need of `return` and `{}`

components/sections/NavigationBar/index.tsx
```diff
@@ -0,0 +1,95 @@
...
+  currentNavItem,
+  languages,
+  onThemeTogglerClick,
+}) => {
+  const [isMenuOpen, setIsMenuOpen] = useState(false);
+  return (
+    <nav className={style.container}>
+      <div className={style.leftItems}>
+        <div className={style.nodeIconWrapper}>
+          <Image
```
ovflowd: Since we're making other logos SVG components, would you mind doing the same with these two?

components/sections/NavigationBar/index.tsx
```diff
@@ -0,0 +1,95 @@
...
+  close: <XMark className="h-6 w-6" />,
+};
+
+const NavigationBar: FC<NavbarProps> = ({
+  navItems,
+  currentNavItem,
+  languages,
+  onThemeTogglerClick,
+}) => {
+  const [isMenuOpen, setIsMenuOpen] = useState(false);
```
ovflowd: Can we avoid using states here? Like I wonder if we can just with pure CSS keep track if the label was clicked or not. I assume maybe we can use the `:active` state? IDK

The reason why I am "cringing" over the "useState" statement, is that this component would need to become a Client-side component (aka you need to add `"use client";` at the beginning of this component, once you rebase this branch.

ovflowd (approved):
Cool! I left some final comments over here, mostly about making the Node.js logo also SVG Components and if we can find a way of removing "useState" to prevent this component from being a client component.

Thank you so much for all this work! For now approving as these comments are nits. (But don't forget to rebase, and add `"use client";` if you intend to keep the component as it is.

components/sections/NavigationBar/index.module.css
```diff
@@ -0,0 +1,86 @@
...
+    md:px-0;
+}
+
+.nodeIconWrapper {
+  @apply h-[30px]
+    flex-1;
+}
+
+.sidebarItemTogglerLabel {
+  @apply block md:hidden;
```
canerakdas: Each Tailwind class must be specified on a new line 👀

components/sections/NavigationBar/index.module.css
```diff
@@ -0,0 +1,86 @@
...
+
+.ghIconWrapper {
+  @apply h-9
+    w-9
+    rounded-md
+    p-2;
+
+  svg {
+    @apply fill-neutral-700
+    dark:fill-neutral-300;
```
canerakdas: Missing indentation

components/sections/NavigationBar/index.tsx
```diff
@@ -0,0 +1,95 @@
+import Hamburger from '@heroicons/react/24/solid/Bars3Icon';
+import XMark from '@heroicons/react/24/solid/XMarkIcon';
+import * as Label from '@radix-ui/react-label';
+import classNames from 'classnames';
+import Image from 'next/image';
+import { type FC, type ComponentProps, useState } from 'react';
```
canerakdas: Type imports should be separated from other imports

components/sections/NavigationBar/index.tsx
```diff
@@ -0,0 +1,95 @@
...
+type NavbarProps = {
+  navItems: NavItem[];
+  currentNavItem: string;
+  languages: ComponentProps<typeof LanguageDropdown>;
+  onThemeTogglerClick: () => void;
+};
+
+const navInteractionIcons = {
+  show: <Hamburger className="h-6 w-6" />,
+  close: <XMark className="h-6 w-6" />,
```
canerakdas: IMO It might be better to add these Tailwind classes to the style file to manage these icon styles from a single place

components/sections/NavigationBar/index.tsx
```diff
@@ -0,0 +1,95 @@
...
+        </Label.Root>
+      </div>
+      <input className="peer hidden" id="sidebarItemToggler" type="checkbox" />
+      <div className={`${style.main} peer-checked:block`}>
+        <div className={style.navItems}>
+          {navItems.map(({ text, href }) => (
+            <NavItem
+              className={classNames('md:bg-transparent', {
+                'bg-green-600 !text-white': currentNavItem === text,
+              })}
```
canerakdas: IMO moving these Tailwind classes to the section's style file will make the code more readable. Additionally, it would be great if we can do it without using important 👀

components/sections/NavigationBar/index.tsx
```diff
@@ -0,0 +1,95 @@
...
+              key={text}
+              href={href}
+            >
+              {text}
+            </NavItem>
+          ))}
+        </div>
+        <div className={style.rightItems}>
```
canerakdas: Nit: IMHO we should name the elements according to the elements they contain, not their positions. These positions may change in the future/device, or it may be weird for a contributor who uses a right-to-left script

components/sections/NavigationBar/index.tsx
```diff
@@ -0,0 +1,95 @@
...
+          <ThemeToggle onClick={onThemeTogglerClick} />
+          <LanguageDropdown
+            availableLanguages={languages.availableLanguages}
+            currentLanguage={languages.currentLanguage}
+          />
+          <a href="https://github.com/nodejs/node">
+            <button className={style.ghIconWrapper} aria-label="Node.js Github">
+              <Github />
+            </button>
+          </a>
```
canerakdas: Invalid HTML;

> Transparent, but there must be no interactive content descendant, `a` element descendant, or descendant with the `tabindex` attribute specified.


https://html.spec.whatwg.org/multipage/text-level-semantics.html#the-a-element

Can we use only the `Link` component there instead of `a > button` 🤔

components/sections/NavigationBar/index.stories.tsx
```diff
@@ -0,0 +1,49 @@
+import type { Meta as MetaObj, StoryObj } from '@storybook/react';
+
+import NavigationBar from './index';
```
canerakdas:
```suggestion
import NavigationBar from '@/components/sections/NavigationBar';
```

Need to use absolute path.

@ovflowd I think the location of the sections has been updated to `@/components/Containers`, shouldn't we move this section there as well?

canerakdas (commented):
Great work! @Rounak-stha I have some minor comments regarding code styling and inline Tailwind classes, other than that it looks like it's ready to go 🚀

components/sections/NavigationBar/index.module.css
ovflowd: (still missing fyi)

components/sections/NavigationBar/index.stories.tsx
ovflowd: Yes, NavigationBar should be moved to the "Containers" folder

components/sections/NavigationBar/index.tsx
```diff
@@ -0,0 +1,84 @@
...
+  languages,
+  onThemeTogglerClick,
+}) => {
+  const [isMenuOpen, setIsMenuOpen] = useState(false);
+  return (
+    <nav className={style.container}>
+      <div className={style.nodeIconAndMobileItemsToggler}>
+        <div className={style.nodeIconWrapper}>
+          <NodejsLogoDark className="h-6 w-20 dark:hidden" />
+          <NodejsLogoLight className="hidden h-6 w-20 dark:block" />
```
ovflowd: Nice!

components/sections/NavigationBar/index.tsx
```diff
@@ -0,0 +1,84 @@
...
+          ))}
+        </div>
+        <div className={style.actionsWrapper}>
+          <ThemeToggle onClick={onThemeTogglerClick} />
+          <LanguageDropdown
+            availableLanguages={languages.availableLanguages}
+            currentLanguage={languages.currentLanguage}
+          />
+          <a href="https://github.com/nodejs/node">
+            <button className={style.ghIconWrapper} aria-label="Node.js Github">
```
ovflowd: Reminder that @canerakdas mentioned that having a button inside an a element is sorta invalid. We don't need the button element if we have the `a` element, afaik.

components/Containers/NavigationBar/index.tsx
```diff
@@ -0,0 +1,88 @@
...
+        </div>
+        <div className={style.actionsWrapper}>
+          <ThemeToggle onClick={onThemeTogglerClick} />
+          <LanguageDropdown
+            availableLanguages={languages.availableLanguages}
+            currentLanguage={languages.currentLanguage}
+          />
+          <a
+            className={style.ghIconWrapper}
+            href="https://github.com/nodejs/node"
```
ovflowd, line 69: Should this still have an aria-label?

ovflowd (approved):
Looks perfect. @canerakdas any last input here?

Otherwise, we're good to merge. Gosh, thank you so much for bearing with us here. 🙇

canerakdas (approved):
LGTM! Thank you for your effort @Rounak-stha

ovflowd:
Hmmm @Rounak-stha I'm not seeing the styles of active navigation item here 🤔 

image

ovflowd:
I can see them working on mobile (@Rounak-stha) but with wrong text colour on light theme 👀 


image

ovflowd:
And we still need a cursor pointer for the label toggle on mobile xD

Rounak-stha:
> I can see them working on mobile (@Rounak-stha) but with wrong text colour on light theme 👀 

Yes, I've mentioned this
The NavItem component has set its own styles for the text color and we can't change that frok the nav component. May be we need to expose a additional prop on the NabItem component to control the text color of the active item.

Rounak-stha:
> Hmmm @Rounak-stha I'm not seeing the styles of active navigation item here 🤔 

TBH, this was bugging me too. As I didn't see the style for the active item in the design file like in the mobile view, I skipped to add it.
I'll be adding this and changing the cursor to pointer on Label too.

ovflowd:
> > I can see them working on mobile (@Rounak-stha) but with wrong text colour on light theme 👀 
> 
> Yes, I've mentioned this
> The NavItem component has set its own styles for the text color and we can't change that frok the nav component. May be we need to expose a additional prop on the NabItem component to control the text color of the active item.
> 
> 

I'd say the NavItem styles are wrong then, as active item with light theme should have white text. I think you can fix these styles directly on NavItem component :)

Rounak-stha:
> I'd say the NavItem styles are wrong then, as active item with light theme should have white text. I think you can fix these styles directly on NavItem component :)

I went through the `NavItem` component in detail and found that it does have styles for active navItem which is passed to the `ActiveLocalizedLink` component which uses the `usePathname` hook. Because we're viewing in storyboook, the `pathname` does not actually match. The pathname is `null` as the url looks like this: `http://localhost:6006/?path=/story/containers-navigationbar--default&globals=theme:dark`.

ovflowd:
> > I'd say the NavItem styles are wrong then, as active item with light theme should have white text. I think you can fix these styles directly on NavItem component :)
> 
> I went through the `NavItem` component in detail and found that it does have styles for active navItem which is passed to the `ActiveLocalizedLink` component which uses the `usePathname` hook. Because we're viewing in storyboook, the `pathname` does not actually match. The pathname is `null` as the url looks like this: `http://localhost:6006/?path=/story/containers-navigationbar--default&globals=theme:dark`.

But **why** on the Mobile View Port it works? That's what bugs me k

ovflowd:
Also another bug, if I change viewport while the mobile menu was open... This happens:

image

I think this is edge case, as well viewports doesn't change that drastically, but probably something we should 👀

ovflowd:
Also @Rounak-stha can you check why ActiveLocalizedLink is not working? I think there might be a bug there (I had to change the logic there... The unit tests are passing but they are weird and I don't even know **why** they are passing 😅

Rounak-stha:
> But **why** on the Mobile View Port it works? That's what bugs me k

Because I am adding those styles from the NavigationBar component (as in design file; right now looks like I kinda forced it)

ovflowd:
> > But **why** on the Mobile View Port it works? That's what bugs me k
> 
> Because I am adding those styles from the NavigationBar component (as in design file; right now looks like I kinda forced it)

Hmm, I think we shouldn't force those styles (active styles) IMO, are they really needed? The pathname checks for ActiveLocalizedLink are fine, and I feel that's how we should also check.

I think we can remove the `currentNavItem` prop, and just let ActiveLocalizedLink do its job. Otherwise we would need to invoke the same logic anyways to the component rendering NavigationBar

aymen94 (approved):

Rounak-stha:
As the `ActiveLocalizedLink` component uses the `usePathName` hook to check if the NavItem is the current item, this does not play well with storybook. I used the `NavigationBar` component on the actual web page and the result looks like this:

image

ovflowd:
> As the `ActiveLocalizedLink` component uses the `usePathName` hook to check if the NavItem is the current item, this does not play well with storybook. I used the `NavigationBar` component on the actual web page and the result looks like this:
> 
> 
> 
> image
> 
> 

So on actual usages it works well?

I wonder if we can mock next router on storybook so that it returns the path that we want.

If you feel like spending tile investigating on that, that's an extra, as it'd be cool to test the active item happy path on storybook (mostly to see if the designs look well)

Otherwise, let it be

Rounak-stha:
> If you feel like spending tile investigating on that, that's an extra, as it'd be cool to test the active item happy path on storybook (mostly to see if the designs look well)

Sure. I'll take a look at mocking that data

ovflowd:
Just don't forget about this (@Rounak-stha) > https://github.com/nodejs/nodejs.org/pull/6072#issuecomment-1810991397 (didn't see you giving it a 👍 so maybe you missed that one)

ovflowd:
@Rounak-stha I was able to find the bug with ActiveLocalizedLink on storybook and pushed a fix on `main`. `usePathname` was returning "null"...

So no need for more investigations on your side.

Rounak-stha:
> @Rounak-stha I was able to find the bug with ActiveLocalizedLink on storybook and pushed a fix on `main`. `usePathname` was returning "null"...

Yeap, I checked. In the edfault SB state, the `usePathname` returns `/` which is not on the list of our navItems. I changed the href of oneof the items to get this:

image

The SB nextjs addon should handle the `Link` component navigation, shouldn't it?

ovflowd:
> The SB nextjs addon should handle the Link component navigation, shouldn't it?

wdym?

> Yeap, I checked. In the edfault SB state, the usePathname returns / which is not on the list of our navItems. I changed the href of oneof the items to get this:

Yeah, that's how I tested. I had to update the Storybook settings to "enable" the App Router functionality. So working fine now :)

ovflowd:
Honestly speaking, I think there's nothing else blocking us and everything seems to be working!

So let's get this merged 🎉 I don't think I've missed anything else and all styles seem correct :)

We can leave the "mock" of usePathname for another PR.

PR #6090, opened 2023-11-02
Add shas to workflow images

bmuenzenmeyer:
## Description

Completing the work started during hacktoberfest - the dependency update work always highlights that we are inconsistent in our approach here.

The work also revealed some risk in chromaui/action in that there is no discernable tag or url to view the sha, only a moment in time that is `latest` on their repositories. I have added on to an issue report that is is unusual and make for a difficult time maintaining intentionality as a customer.

We could look into something like https://github.com/zgosalvez/github-actions-ensure-sha-pinned-actions to enforce this - but I don't yet know if that is overkill.

## Validation

All images the same, but more readable.

### Check List

### DIFF
--- a/.github/workflows/lint-and-tests.yml
+++ b/.github/workflows/lint-and-tests.yml
@@ -75,7 +75,7 @@ jobs:
           ref: ${{ github.event.pull_request.head.ref || github.ref }}
 
       - name: Restore Lint Cache
-        uses: actions/cache/restore@704facf57e6136b1bc63b828d79edcd491f0ee84
+        uses: actions/cache/restore@704facf57e6136b1bc63b828d79edcd491f0ee84 # v3.3.2
         with:
           path: |
             .turbo/cache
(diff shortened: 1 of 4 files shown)


bmuenzenmeyer:
requesting fast-track because this change is trivial

.github/workflows/lint-and-tests.yml
```diff
@@ -216,6 +216,7 @@ jobs:
           (github.event_name == 'pull_request_target' &&
             startsWith(github.event.pull_request.head.ref, 'dependabot/') == false &&
             github.event.pull_request.head.ref != 'chore/crowdin')
+        # sha reference has no stable git tag reference or URL. see https://github.com/chromaui/chromatic-cli/issues/797
```
bmuenzenmeyer, line 219: :yikes:

MattIPv4 (approved):

ovflowd (approved):

PR #6174, opened 2023-12-02
fix: Layout max width

canerakdas:
## Description

The maximum width given to the content area has been moved to the container area, ensuring that the layout does not shrink in cases where the width of the content width is low

## Validation

image

image

In cases where the content is long or short, the layout must comply with the maximum width(`90rem`/ `1440px`) we have given and appear correct

### Check List


tailwind.config.ts
```diff
@@ -130,8 +130,9 @@ export default {
         xs: '0px 1px 2px 0px theme(colors.shadow / 5%)',
         lg: '0px 4px 6px -2px theme(colors.shadow / 3%), 0px 12px 16px -4px theme(colors.shadow / 8%)',
       },
-      spacing: { '4.5': '1.125rem' },
+      spacing: { '4.5': '1.125rem', '18': '4.5rem' },
```
ovflowd, line 133: We don't necessarily have to have 4.5rem, right? What happens if we use the closest `px` for this value? How different it looks?

ovflowd (approved):
LGTM! Thanks for the changes!

tailwind.config.ts
canerakdas: The required spacing is 72px(4.5rem), I can add a spacing of 64px(4rem) or 80px(5rem) instead

64px / 4rem;
image

80px / 5rem;
image

IMO 64px one looks good, and hard to tell difference 👀

AugustinMauroy (approved):

PR #6246, opened 2024-01-14
fix: de-emphasize copy to clipboard button

### LINKED ISSUE #6244: `Copy to Clipboard` button is too prominent
A tiny design nit is that the `Copy to Clipboard` button is very prominent in the smaller coding snippets, which I find distracting from the content. This should be a de-emphasized control in my opinion.

image

Compare to the homepage which uses a de-emphasized version

image

_Originally posted by @bmuenzenmeyer in https://github.com/nodejs/nodejs.org/issues/6215#issuecomment-1890272029_

bmuenzenmeyer:
## Description

De-emphasizes the copy to clipboard button. also removed the homepage customization. yes, this is a deviation of the figma- but now met with real content, especially repeated examples are distracting. this is a tertiary act for the site, it should not have primary/action color

## Validation

same homepage

image

less busy examples within content

| Old | New |
| - | - |
| image  |image |

## Related Issues

closes #6244

### Check List

### DIFF
--- a/components/Common/CodeBox/index.module.css
+++ b/components/Common/CodeBox/index.module.css
@@ -67,18 +67,11 @@
 
     & > .action {
       @apply flex
-        cursor-pointer
         items-center
         gap-2
-        border-none
-        bg-neutral-900
         px-3
         py-1.5
         font-medium;
-
-      &:hover {
-        @apply bg-neutral-800;
-      }
     }
   }
 }
--- a/components/Common/CodeBox/index.module.css
+++ b/components/Common/CodeBox/index.module.css
@@ -70,8 +70,15 @@
         cursor-pointer
         items-center
         gap-2
+        border-none
+        bg-neutral-900
         px-3
-        py-1.5;
+        py-1.5
+        font-medium;
+
+      &:hover {
+        @apply bg-neutral-800;
+      }
     }
   }
 }
--- a/components/Common/Button/index.module.css
+++ b/components/Common/Button/index.module.css
@@ -9,6 +9,26 @@
     @apply cursor-not-allowed;
   }
 
+  &.neutral {
+    @apply rounded
+      bg-neutral-900
+      text-white
+      dark:text-neutral-200;
+
+    &:hover {
+      @apply bg-neutral-800;
+    }
+
+    &[aria-disabled='true'] {
+      @apply bg-neutral-900
+        opacity-50;
+    }
+
+    &:focus {
+      @apply bg-neutral-800;
+    }
+  }
+
   &.primary {
     @apply rounded
       border
--- a/components/Common/Button/index.stories.tsx
+++ b/components/Common/Button/index.stories.tsx
@@ -5,6 +5,14 @@ import Button from '@/components/Common/Button';
 type Story = StoryObj<typeof Button>;
 type Meta = MetaObj<typeof Button>;
 
+export const Neutral: Story = {
+  args: {
+    kind: 'neutral',
+    children: 'Download Node (LTS)',
+    disabled: false,
+  },
+};
+
 export const Primary: Story = {
   args: {
     kind: 'primary',
--- a/components/Common/Button/index.tsx
+++ b/components/Common/Button/index.tsx
@@ -6,7 +6,7 @@ import Link from '@/components/Link';
 import styles from './index.module.css';
 
 type ButtonProps = AnchorHTMLAttributes<HTMLAnchorElement> & {
-  kind?: 'primary' | 'secondary' | 'special';
+  kind?: 'neutral' | 'primary' | 'secondary' | 'special';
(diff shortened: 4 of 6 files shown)


ovflowd:
@bmuenzenmeyer does it make sense to keep the green colour on light theme? Like keeping the gray on dark theme and green on light theme? I have the feeling that within the light theme the gray button is more emphasized than the green one.

ovflowd (approved):

bmuenzenmeyer:
I thought about that but the codebox is always dark, so this still seems appropriate to me.

Open to more thoughts, or iterating on this for a bit.

AugustinMauroy (approved):
LGTM

I think that on `en/learn/getting-started/introduction-to-nodejs` the code snippet that displays functions should be inline code.

shanpriyan (approved):
LGTM!

aymen94 (approved):

components/Common/CodeBox/index.module.css
```diff
@@ -70,8 +70,15 @@
...
+        border-none
+        bg-neutral-900
         px-3
-        py-1.5;
+        py-1.5
+        font-medium;
+
+      &:hover {
+        @apply bg-neutral-800;
+      }
```
canerakdas: Wouldn't it be reusable to add a`neutral` kind to the`Button` component instead of overwriting the styles here?

canerakdas (approved):

components/Common/CodeBox/index.module.css
ovflowd: It does make sense indeed of creating another Button variant, this one could be useful in some places!

ovflowd: cc @bmuenzenmeyer

components/Common/CodeBox/index.module.css
bmuenzenmeyer, line 1: this change and the removal in layouts.module.css removes the customization that gives use the neutral color, but retains the customization that allows for both a smaller profile AND also one that accomodates an icon

we could explore adding both to the button component, but it's starting to scope creep IMO

bmuenzenmeyer (commented):
added a neutral button `kind` - alphabetically first across all diffs

components/Common/CodeBox/index.module.css
bmuenzenmeyer: addressed via https://github.com/nodejs/nodejs.org/pull/6246/commits/1a8280e1c6468c413623d3256d5c3a2251239175

components/Common/CodeBox/index.module.css
ovflowd: > this change and the removal in layouts.module.css removes the customization that gives use the neutral color, but retains the customization that allows for both a smaller profile AND also one that accomodates an icon

wdym?

ovflowd (approved):

components/Common/CodeBox/index.module.css
bmuenzenmeyer: i mean that codebox is the only button that has an icon, and as such we introduced a customization to the button to make it look nice. without these values here the icon would break onto a new line vs the text

image

my comment was that in some design systems, a button and icon is a standard use case that would promote these flexbox styles into the `Button` itself - but I don't think we should do that until we have to

ovflowd: > i mean that codebox is the only button that has an icon, and as such we introduced a customization to the button to make it look nice. without these values here the icon would break onto a new line vs the text

Right I think @canerakdas is fixing that on another PR, that the Button should be flex by default. I wonder if he fixed that already?

bmuenzenmeyer: @canerakdas can you review the above? I don't want us to head into a conflict - but glancing at the PRs I didn't notice anything. I'd like to merge this if we can.

canerakdas: @ovflowd Not yet, I'm planning to create another PR for this since many places are affected

ovflowd:
@bmuenzenmeyer apparently the merge queue is failing. You might want to check why 👀

ovflowd:
@bmuenzenmeyer I guess rebase is needed!

bmuenzenmeyer:
Yes I saw that, just too busy today. 4 sick people here and lice, lol.

Standby

bmuenzenmeyer:
@ovflowd this lint error above is directly conflicting with https://github.com/nodejs/nodejs.org/commit/31a50a692ee118350d3bc89193cdf2a3f7692539 - as if we have dueling format output

image

i re-generated my dependencies atop the lockfile - can you confirm you are using the latest too?

ovflowd:
I mean yea, we updated the tailwind prettier plugin, so main branch has new rules 👀

bmuenzenmeyer:
cursed pr

PR #6473, opened 2024-03-12
chore: update incorrect code comments regarding dynamicParams

dario-piotrowicz:
## Description

Very minor, this PR is simply updating a couple of code comments around `dynamicParams` which I am pretty sure are wrong and just [copy-pastas](https://github.com/search?q=repo%3Anodejs%2Fnodejs.org%20%22Forces%20that%20only%20the%20paths%20from%20%60generateStaticParams%60%20are%20allowed%2C%20giving%20404%20on%20the%20contrary%22&type=code) (and indeed they confused me at first)

### Check List

### DIFF
--- a/app/[locale]/feed/[feed]/route.ts
+++ b/app/[locale]/feed/[feed]/route.ts
@@ -29,7 +29,9 @@ export const GET = async (_: Request, { params }: StaticParams) => {
 export const generateStaticParams = async () =>
   siteConfig.rssFeeds.map(feed => ({ feed: feed.file, locale }));
 
-// Forces that only the paths from `generateStaticParams` are allowed, giving 404 on the contrary
+// In this case we want to catch-all possible requests. This is so that if a non defined feed is
+// requested we can manually return a 404 response for it instead of having Next.js handle it
+// and return our top level custom 404 html page instead
 // @see https://nextjs.org/docs/app/api-reference/file-conventions/route-segment-config#dynamicparams
 export const dynamicParams = true;
 
--- a/app/[locale]/next-data/og/route.tsx
+++ b/app/[locale]/next-data/og/route.tsx
@@ -64,7 +64,8 @@ export const generateStaticParams = async () => [
 // @see https://nextjs.org/docs/app/api-reference/file-conventions/route-segment-config#runtime
 export const runtime = VERCEL_ENV ? 'edge' : 'nodejs';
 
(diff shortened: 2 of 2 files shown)


ovflowd (approved):
Oh yes, these are definitely wrong. With dynamicParams = true the behaviour is definitely not the one commented there.

Nice catch!

app/[locale]/feed/[feed]/route.ts
```diff
@@ -29,7 +29,9 @@ export const GET = async (_: Request, { params }: StaticParams) => {
 export const generateStaticParams = async () =>
   siteConfig.rssFeeds.map(feed => ({ feed: feed.file, locale }));
 
-// Forces that only the paths from `generateStaticParams` are allowed, giving 404 on the contrary
+// In this case we want to catch-all possible requests. This is so that if a non defined feed is
+// requested we can manually return a 404 response for it instead of having Next.js handle it
+// and return our top level custom 404 html page instead
 // @see https://nextjs.org/docs/app/api-reference/file-conventions/route-segment-config#dynamicparams
 export const dynamicParams = true;
```
Comment (ovflowd) [inline app/[locale]/feed/[feed]/route.ts:36]: On a side note, I'm quite sure dynamicParams should not be active for blog feeds, but I don't recall now why this is set to true or if it was an accident

Comment (dario-piotrowicz) [reply app/[locale]/feed/[feed]/route.ts:36]: as I mentioned in the comment, with it set to `true` you don't get the custom 404 html page, and instead you get a 404 with an empty body: 
Screenshot 2024-03-12 at 22 36 07

my guess is that you didn't want to return some html as this could cause issues with RSS apps? ¯\\_(ツ)_/¯

Comment (ovflowd) [reply app/[locale]/feed/[feed]/route.ts:36]: Oh true, I recall now. Indeed because in some endpoints we are self consuming the data, if the 404 returned a HTML page it would break the JSON.parse method 👀

Comment (ovflowd) [reply app/[locale]/feed/[feed]/route.ts:36]: I sorta question the decision that it renders a HTML page when dynamicParams is disabled. I understand that for pages that makes sense... But for API routes 🤔

Comment (dario-piotrowicz) [reply app/[locale]/feed/[feed]/route.ts:36]: yeah.... weird that Pages routing seems to catch also the api requests 😅 🤷

AugustinMauroy (approved):
LGTM ! Thanks for you first contribution 🎉

aymen94 (approved):

bmuenzenmeyer:
fast tracking - simple (but important!) change

PR #6481, opened 2024-03-19
add mascot to 404

### LINKED ISSUE #6480: The new website is missing prominent display of the new project mascot
### URL:

https://nodejs.org/en

### Browser Name:

All Browsers

### Browser Version:

All Versions

### Operating System:

All OSs

https://nodejs.org/en/about/branding is the only place I could find it.

We need more rocket turtle!

bmuenzenmeyer:
pivoted to 404 as part of the #6480 discussion

## Description

Adds the new mascot to the 404 page
@AugustinMauroy suggested adding it to the 404, which I like

## Validation

image
image

## Related Issues

closes #6480

### Check List

### DIFF
--- a/app/[locale]/not-found.tsx
+++ b/app/[locale]/not-found.tsx
@@ -22,9 +22,8 @@ const NotFoundPage: FC = () => {
           <Image
             src="/static/images/node-mascot.svg"
             alt="The Node.js mascot"
-            className="w-[100px]"
-            height={200}
-            width={174.375}
+            height={114.69}
+            width={100}
           />
         </div>
         <p className="-mt-4 max-w-sm text-center text-lg">
--- a/app/[locale]/not-found.tsx
+++ b/app/[locale]/not-found.tsx
@@ -1,6 +1,7 @@
 'use client';
 
 import { ArrowRightIcon } from '@heroicons/react/24/solid';
+import Image from 'next/image';
 import { useTranslations } from 'next-intl';
 import type { FC } from 'react';
 
@@ -17,6 +18,15 @@ const NotFoundPage: FC = () => {
       <main>
         404
         <h1 className="special -mt-4">{t('layouts.error.notFound.title')}</h1>
+        <div className="my-4 flex items-center justify-center">
+          <Image
+            src="/static/images/node-mascot.svg"
+            alt="The Node.js mascot"
+            className="w-[100px]"
+            height={200}
+            width={174.375}
+          />
+        </div>
         <p className="-mt-4 max-w-sm text-center text-lg">
           {t('layouts.error.notFound.description')}
         </p>
--- a/pages/en/about/index.mdx
+++ b/pages/en/about/index.mdx
@@ -3,12 +3,6 @@ title: About Node.js®
 layout: about
 ---
 
-<div className="my-4 flex items-center justify-center">
-  <img src="/static/images/node-mascot.svg" className="w-[100px]" />
-</div>
-
----
-
 # About Node.js®
 
 As an asynchronous event-driven JavaScript runtime, Node.js is designed to build
--- a/pages/en/about/index.mdx
+++ b/pages/en/about/index.mdx
@@ -3,6 +3,12 @@ title: About Node.js®
 layout: about
 ---
 
+<div className="my-4 flex items-center justify-center">
+  <img src="/static/images/node-mascot.svg" className="w-[100px]" />
+</div>
+
+---
+
 # About Node.js®
 
 As an asynchronous event-driven JavaScript runtime, Node.js is designed to build


ovflowd (dismissed):
The removal was intentional as we got feedback from the TSC that it didn't fit on the About page.

I'm fine if the TSC is fine @nodejs/tsc

AugustinMauroy:
-1 for about page, because if a user goes to this page it's to read the text, whereas the first thing they see is the mascot, and this is even more pronounced on phones where the mascot takes up half the screen.

bmuenzenmeyer:
to me it's an injection of character into an otherwise serious website.

ovflowd:
> to me it's an injection of character into an otherwise serious website.

Right, that was one of the issues.

bmuenzenmeyer:
Unsure how this seems inappropriate. The "fold" is still there providing an affordance to users. Without the mascot it's not like they can see all the content.

The mascot is official. We should be proud of it.

ovflowd:
Right, and I certainly agree. But it's not my decision. It's the @nodejs/tsc that should approve this or not 🙈

GeoffreyBooth:
My only concern with the screenshots on this thread is that the mascot takes up a lot of space and pushes down the text, so you see less text content in the first screenful. Could we perhaps put it on the side so that the text can wrap around it?

targos:
Why is it up to the TSC to approve this specific PR? What does it mean? A vote? A majority by counting the number of approvals?

bmuenzenmeyer:
> Why is it up to the TSC to approve this specific PR? What does it mean? A vote? A majority by counting the number of approvals?

This is hard to answer without understanding the (unknown to me) circumstances that led to its addition and removal. 

- Added February 23rd [as a direct commit on `main`](https://github.com/nodejs/nodejs.org/commit/e4b81a35a91879f4f82a1159565c522cb9c4a183)
- Removed March 4th without mention as part of https://github.com/nodejs/nodejs.org/pull/6435

Per [CONTENT VS CODE](https://github.com/nodejs/nodejs.org/blob/main/CONTENT_VS_CODE.md), the website team defers to core collaborators, not specifically the TSC, on all matters of content. Without input, the website team has control.

IMO, the adoption of the mascot by the project is tacit approval to add it to the website.

app/[locale]/not-found.tsx
Comment (bmuenzenmeyer) [inline app/[locale]/not-found.tsx:1]: I noted this entire file is bleeding tailwind classnames into the markup. that seems to predate this PR but we can fix here or as another item. we still care right?

Comment (AugustinMauroy) [reply app/[locale]/not-found.tsx:1]: I think we'll have to do something to clean up this kind of inline markup tailwindcss. So let this pr as it

AugustinMauroy (approved):
WOOOOOWWW I love this 
🐢🚀

mcollina (approved):
lgtm

UlisesGascon (approved):
I love it! This is quite cool :)

tniessen (approved):

app/[locale]/not-found.tsx
```diff
@@ -17,6 +18,15 @@ const NotFoundPage: FC = () => {
       <main>
         404
         <h1 className="special -mt-4">{t('layouts.error.notFound.title')}</h1>
+        <div className="my-4 flex items-center justify-center">
+          <Image
+            src="/static/images/node-mascot.svg"
+            alt="The Node.js mascot"
+            className="w-[100px]"
+            height={200}
+            width={174.375}
```
Comment (canerakdas) [inline app/[locale]/not-found.tsx]:
```suggestion
            height={114.69}
            width={100}
```

We can directly give the width we want according to the aspect ratio

Comment (bmuenzenmeyer) [reply app/[locale]/not-found.tsx]: curious, why is this better than giving it a height based on the aspect ratio? because of the TW class?

Comment (canerakdas) [reply app/[locale]/not-found.tsx]: AFAIK When we add the `w-[100px]` class, the width and height values we give to the image element no longer matter. The width of the rendered image with becomes `100px` and the height becomes `114.69px`.

image

In addition to seeing the given input directly, it also helps the `next/image` component to prepare a more optimized `srcset` for us. In this scenario, it dont looks make any sense since we are using svg, but it is quite useful under normal conditions.

Example srcset for `100x114.69`;
`srcset="/_next/image?url=%2Fstatic%2Fimages%2Fnode-mascot.svg&w=128&q=75 1x, /_next/image?url=%2Fstatic%2Fimages%2Fnode-mascot.svg&w=256&q=75 2x"`

Example srcset for `174.375x200`;
`srcset="/_next/image?url=%2Fstatic%2Fimages%2Fnode-mascot.svg&w=256&q=75 1x, /_next/image?url=%2Fstatic%2Fimages%2Fnode-mascot.svg&w=384&q=75 2x"`

kyliewd (approved):

canerakdas (approved):

PR #6487, opened 2024-03-19
fix: open external link in new tab for navbar items

hisamafahri:
## Description

Open external navigation bar's link in new tab

## Validation

If clicked in the `Certification` and `Docs` in the navbar, should open it in the new tab.

## Related Issues

None

### Check List

### DIFF
--- a/components/Containers/NavBar/NavItem/index.tsx
+++ b/components/Containers/NavBar/NavItem/index.tsx
@@ -1,6 +1,6 @@
 import { ArrowUpRightIcon } from '@heroicons/react/24/solid';
 import classNames from 'classnames';
-import type { FC, PropsWithChildren } from 'react';
+import type { FC, HTMLAttributeAnchorTarget, PropsWithChildren } from 'react';
 
 import ActiveLink from '@/components/Common/ActiveLink';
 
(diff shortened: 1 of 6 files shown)


AugustinMauroy:
Honestly, I'm not in favour of this change, because if a user wants to go to the docs and goes through nodejs.org, they'll end up with two tabs open, which will "take time" to close. I think that opening a new tab should only be done when the user needs to keep the source page open, which isn't the case here.

hisamafahri:
@AugustinMauroy That's a good point. However, I believe the expectations for most users are to open the clicked link in a new tab whenever they see a top-right arrow (↗) since it is mostly used as an indicator for a link from another site.

Either way, whatever decision will be made (open in a new tab, remove the arrow, or even keep the current behaviour), I am all for it 🙆‍♂️ .

bmuenzenmeyer (approved):
I'm 👍 on this UX behavior

aymen94 (approved):

bmuenzenmeyer:
@AugustinMauroy are you okay with moving this forward?

AugustinMauroy:
> @AugustinMauroy are you okay with moving this forward?

I'm in favor of keeping this behavior. But modifying api docs link to not have icons and blank page.

bmuenzenmeyer:
@AugustinMauroy i appreciate your thoughts but am going to merge this. it's generally common consensus that 

- the arrow means you are leaving the site
- its a courteous gesture in today's tab based browsing to open blank

if you feel strongly, please block it in future discussions

bmuenzenmeyer:
that being said, some of the usage here is probably wrong (credit @ovflowd for mentioning this)
image

PR #6489, opened 2024-03-19
fix: update progression item spacing

Jhinger:
## Description
Increase spacing between progression items to make it easier to read.

Before: 
before
After:
after

### Check List

### DIFF
--- a/components/Common/ProgressionSidebar/ProgressionSidebarItem/index.module.css
+++ b/components/Common/ProgressionSidebar/ProgressionSidebarItem/index.module.css
@@ -1,6 +1,8 @@
 a.item {
   @apply relative
     z-20
+    mx-0
+    my-2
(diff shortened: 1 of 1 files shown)


AugustinMauroy:
from one point of view I agree with you that adding spacing helps reading. On the other hand, with this large spacing, navigation is super long, so searching for a resource is a bit tedious.

Another thing to note is that a professional designer worked on the system design, so he had to think about the spacing between each element to make it easy to read.

ovflowd (changes requested):
Unfortunately this spacing seems rather too much. It breaks the design of the Website.

I'm against this change, sorry.

bmuenzenmeyer:
closing - its against the design intent of the site. we can monitor sentry data and watch for other comments about this.

PR #6537, opened 2024-03-24
feat: Improve keyboard interactions in searchbox

Baccega:
## Description

This PR is improving the UX of the searchbox by adding interactivity with the up/down arrows and enter keys, to mimic the behaviour of similar searchboxes.
While developing the feature I've also found a bug in the `useKeyboardCommands` (the eventListener wasn't properly cleaned) that was causing issues to my addition.

## Validation

I've validated it locally by setting some random search results from prod into the react state, a video of the feature in action is available [here](https://www.loom.com/share/d21720da085443419512fa5101253739)

### Check List


components/Common/Search/utils.ts
```diff
@@ -1,7 +1,18 @@
+import type { Result } from '@orama/orama';
```
AugustinMauroy: IMHO this should be put on `util/search`.

components/Common/Search/States/WithSearchBox.tsx
```diff
@@ -16,7 +16,8 @@ import { WithError } from '@/components/Common/Search/States/WithError';
 import { WithNoResults } from '@/components/Common/Search/States/WithNoResults';
```
AugustinMauroy, line 14: please remove `console.log()`

AugustinMauroy (commented):
Cool improvement

components/Common/Search/utils.ts
Baccega: Moved it

components/Common/Search/States/WithSearchBox.tsx
Baccega: Removed (Whoops)

Baccega:
While implementing the feedbacks I've improved the feature a little bit further by bounding the selection (no overflowing now) and adding a scrollIntoView effect

AugustinMauroy (approved):
LGTM ! Super cool first contribution 🎉

components/Common/Search/States/WithSearchBox.tsx
```diff
@@ -78,15 +80,55 @@ export const WithSearchBox: FC<SearchBoxProps> = ({ onClose }) => {
     [searchTerm, selectedFacet]
   );
 
+  useKeyboardCommands(cmd => {
+    if (!searchResults) return;
+
+    switch (cmd) {
+      case 'down':
+        setSelectedResult(prev =>
+          Math.min(
```
ovflowd: Sorry, this logic is convoluted as hell and not needed at all.

ovflowd (changes requested):
We shouldn't use JavaScript to achieve this. This can easily be done with pure HTML and the proper aria-* tags (and other attributes such as tabindex, role, etc)

I appreciate the invested time in this, but unfortunately we shouldn't try to solve accessibility with JavaScript.

Baccega:
> We shouldn't use JavaScript to achieve this. This can easily be done with pure HTML and the proper aria-* tags (and other attributes such as tabindex, role, etc)
> 
> I appreciate the invested time in this, but unfortunately we shouldn't try to solve accessibility with JavaScript.

Do you mean that we shouldn't have this behaviour at all? [MDN Web Docs](https://developer.mozilla.org/en-US/) has this feature (With all the proper accessibility tags) and it does not work without js

In that case I can revert the feature and just leave the fix of the `useKeyboardCommands` hook, otherwise I can imitate the MDN searchbox accessibility features (with aria-selected, role and aria-activedescendant)

ovflowd:
> > We shouldn't use JavaScript to achieve this. This can easily be done with pure HTML and the proper aria-* tags (and other attributes such as tabindex, role, etc)
> 
> > 
> 
> > I appreciate the invested time in this, but unfortunately we shouldn't try to solve accessibility with JavaScript.
> 
> 
> 
> Do you mean that we shouldn't have this behaviour at all? [MDN Web Docs](https://developer.mozilla.org/en-US/) has this feature (With all the proper accessibility tags) and it does not work without js
> 
> 
> 
> In that case I can revert the feature and just leave the fix of the `useKeyboardCommands` hook, otherwise I can imitate the MDN searchbox accessibility features (with aria-selected, role and aria-activedescendant)

Interesting. My understanding is that navigation via tab key is sufficient for the a11y standard.

That's because, to my understanding, there's no sense of what is left and right, up and down.

I believe if we make navigation through tab keys work (which is possible with plain html) we could content with that.

Navigation with arrow keys indeed requires JavaScript, but I wonder if we should even address that. If that is indeed the goal, I still believe the proposed patch can be simplified and no Math calculations are needed.

You can do it with a simple state that stores the current index (default 0)

And have an if statement (if the next index < length of results) then update the state, otherwise do nothing.

Baccega:
> > > We shouldn't use JavaScript to achieve this. This can easily be done with pure HTML and the proper aria-* tags (and other attributes such as tabindex, role, etc)
> > 
> > 
> > > 
> > 
> > 
> > > I appreciate the invested time in this, but unfortunately we shouldn't try to solve accessibility with JavaScript.
> > 
> > 
> > Do you mean that we shouldn't have this behaviour at all? [MDN Web Docs](https://developer.mozilla.org/en-US/) has this feature (With all the proper accessibility tags) and it does not work without js
> > In that case I can revert the feature and just leave the fix of the `useKeyboardCommands` hook, otherwise I can imitate the MDN searchbox accessibility features (with aria-selected, role and aria-activedescendant)
> 
> Interesting. My understanding is that navigation via tab key is sufficient for the a11y standard.
> 
> That's because, to my understanding, there's no sense of what is left and right, up and down.
> 
> I believe if we make navigation through tab keys work (which is possible with plain html) we could content with that.
> 
> Navigation with arrow keys indeed requires JavaScript, but I wonder if we should even address that. If that is indeed the goal, I still believe the proposed patch can be simplified and no Math calculations are needed.
> 
> You can do it with a simple state that stores the current index (default 0)
> 
> And have an if statement (if the next index < length of results) then update the state, otherwise do nothing.

I've updated the PR with the simplified logic you suggested and imitated the searchbox behaviour of MDN (which I found out that implements the [combobox pattern](https://www.w3.org/WAI/ARIA/apg/patterns/combobox/)) 👍🏻

ovflowd:
> > > > We shouldn't use JavaScript to achieve this. This can easily be done with pure HTML and the proper aria-* tags (and other attributes such as tabindex, role, etc)
> > > 
> > > 
> > > > 
> > > 
> > > 
> > > > I appreciate the invested time in this, but unfortunately we shouldn't try to solve accessibility with JavaScript.
> > > 
> > > 
> > > Do you mean that we shouldn't have this behaviour at all? [MDN Web Docs](https://developer.mozilla.org/en-US/) has this feature (With all the proper accessibility tags) and it does not work without js
> > > In that case I can revert the feature and just leave the fix of the `useKeyboardCommands` hook, otherwise I can imitate the MDN searchbox accessibility features (with aria-selected, role and aria-activedescendant)
> > 
> > 
> > Interesting. My understanding is that navigation via tab key is sufficient for the a11y standard.
> > That's because, to my understanding, there's no sense of what is left and right, up and down.
> > I believe if we make navigation through tab keys work (which is possible with plain html) we could content with that.
> > Navigation with arrow keys indeed requires JavaScript, but I wonder if we should even address that. If that is indeed the goal, I still believe the proposed patch can be simplified and no Math calculations are needed.
> > You can do it with a simple state that stores the current index (default 0)
> > And have an if statement (if the next index < length of results) then update the state, otherwise do nothing.
> 
> I've updated the PR with the simplified logic you suggested and imitated the searchbox behaviour of MDN (which I found out that implements the [combobox pattern](https://www.w3.org/WAI/ARIA/apg/patterns/combobox/)) 👍🏻

Appreciate it! Reviewing now 👀

components/Common/Search/States/WithSearchBox.tsx
```diff
@@ -78,15 +80,55 @@ export const WithSearchBox: FC<SearchBoxProps> = ({ onClose }) => {
     [searchTerm, selectedFacet]
   );
 
+  useKeyboardCommands(cmd => {
+    if (searchError || !searchResults || searchResults.count <= 0) return;
+
+    switch (true) {
+      case cmd === 'down' && selectedResult == null:
```
ovflowd: I've never seen boolean algebra done this way with a switch statement. Kinda odd, but neat,

components/Common/Search/States/WithSearchBox.tsx
```diff
@@ -78,15 +80,55 @@ export const WithSearchBox: FC<SearchBoxProps> = ({ onClose }) => {
...
+
   const reset = () => {
     setSearchTerm('');
     setSearchResults(null);
+    setSelectedResult(null);
     setSelectedFacet(0);
   };
 
+  const handleEnter = () => {
+    if (selectedResult == null || (searchResults && searchResults?.count <= 0))
```
ovflowd: On this project we always use `{}` for if statements. could you add the missing brackets? :) (to all the if statements below?)

components/Common/Search/States/WithSearchBox.tsx
```diff
@@ -78,15 +80,55 @@ export const WithSearchBox: FC<SearchBoxProps> = ({ onClose }) => {
...
+      return;
+    const selectedHit = searchResults?.hits[selectedResult ?? 0];
+    if (!selectedHit) return;
+
+    router.push(searchHitToLinkPath(selectedHit));
+    reset();
```
ovflowd: Shouldn't the searchbox be reset/closed before the route change?

Also, could you add the "reset" to be part directly of the onClose? Makes sense IMO for search results to always reset during onClose

components/Common/Search/States/WithSearchBox.tsx
```diff
@@ -131,6 +173,21 @@ export const WithSearchBox: FC<SearchBoxProps> = ({ onClose }) => {
             <form onSubmit={onSubmit}>
               <input
                 ref={searchInputRef}
+                aria-activedescendant={
```
ovflowd, line 179: Nice!

util/search.ts
ovflowd, line 1: Rename the file to `searchUtils`

components/Common/Search/States/WithSearchBox.tsx
```diff
@@ -156,7 +213,11 @@ export const WithSearchBox: FC<SearchBoxProps> = ({ onClose }) => {
             ))}
           </div>
 
-          <div className={styles.fulltextResultsContainer}>
+          <div
+            id={`fulltext-results-container`}
```
ovflowd:
```suggestion
            id="fulltext-results-container"
```

components/Common/Search/States/WithSearchResult.tsx
```diff
@@ -1,32 +1,43 @@
...
 export const WithSearchResult: FC<SearchResultProps> = props => {
-  const isAPIResult = props.hit.document.siteSection.toLowerCase() === 'docs';
-  const basePath = isAPIResult ? BASE_URL : '';
-  const path = `${basePath}/${props.hit.document.path}`;
+  const divRef = useRef<HTMLDivElement>(null);
+  const path = searchHitToLinkPath(props.hit);
+
+  useEffect(() => {
+    if (props.selected && divRef.current) {
+      divRef.current?.scrollIntoView({ block: 'center' });
```
ovflowd:
```suggestion
      divRef.current.scrollIntoView({ block: 'center' });
```

You're already checking if `divRef.current` is defined on the if statement

hooks/react-client/useKeyboardCommands.ts
```diff
@@ -27,9 +27,11 @@ const useKeyboardCommands = (fn: KeyboardCommandCallback) => {
           fn('up');
           break;
       }
-    });
+    }
 
-    return () => document.removeEventListener('keydown', () => {});
+    document.addEventListener('keydown', handleKeyDown);
+
+    return () => document.removeEventListener('keydown', handleKeyDown);
```
ovflowd, line 34: Nice cleanup! That removeEventListener was doing nothing. Not sure how this got passed on code reviews 🤦 

But instead of using regular functions. Could you use an arrow function?

hooks/react-client/useKeyboardCommands.ts
```diff
@@ -6,7 +6,7 @@ type KeyboardCommandCallback = (key: KeyboardCommand) => void;
 
 const useKeyboardCommands = (fn: KeyboardCommandCallback) => {
   useEffect(() => {
-    document.addEventListener('keydown', event => {
+    function handleKeyDown(event: KeyboardEvent) {
```
ovflowd:
```suggestion
    const handleKeyDown = (event: KeyboardEvent) => {
```

ovflowd (commented):
Really good stuff! This is a tenfold enhancement to what we have at the moment!

components/Common/Search/States/WithSearchBox.tsx
```diff
@@ -78,15 +80,55 @@ export const WithSearchBox: FC<SearchBoxProps> = ({ onClose }) => {
     [searchTerm, selectedFacet]
   );
 
+  useKeyboardCommands(cmd => {
+    if (searchError || !searchResults || searchResults.count <= 0) return;
```
ovflowd: Also brackets here

components/Common/Search/States/WithSearchBox.tsx
```diff
@@ -31,6 +32,7 @@ type SearchBoxProps = { onClose: () => void };
 export const WithSearchBox: FC<SearchBoxProps> = ({ onClose }) => {
   const [searchTerm, setSearchTerm] = useState('');
   const [searchResults, setSearchResults] = useState<SearchResults>(null);
+  const [selectedResult, setSelectedResult] = useState<Nullable<number>>();
```
ovflowd: This doesn't need to be a nullable, if you set "null" as the default value. TypeScript will automatically infer that. At the moment this type is "null | undefined | number" if you do `useState<number>(null)` it'll be automatically `null | number`

Baccega: I tried to implement the simplification but the inferred type is actually just `number` for some reason (and it complains that I'm setting `null` as the initial value) 🤷🏻 :

image

But if I set it to `usestate<number>()` the resulting inference is similar to what you described, but with `number | undefined` as the resulting type:

image
 
I've now set it to `number | null` explicitly to shrink the type so that `undefined` is not a valid value for the state 👍🏻

components/Common/Search/States/WithSearchBox.tsx
Baccega: Added the missing brackets 👍🏻

components/Common/Search/States/WithSearchBox.tsx
Baccega: Added the missing brackets 👍🏻

components/Common/Search/States/WithSearchBox.tsx
Baccega: Yeah, it now uses a single `handleClose` function

components/Common/Search/States/WithSearchBox.tsx
Baccega: Changed it

components/Common/Search/States/WithSearchResult.tsx
Baccega: True, changed it

hooks/react-client/useKeyboardCommands.ts
Baccega: Changed it

hooks/react-client/useKeyboardCommands.ts
Baccega: Yeah, It was causing problems for my new keybindings 😄 Now it's using an arrow function

util/search.ts
Baccega: Renamed it

aymen94 (approved):

components/Common/Search/States/WithSearchBox.tsx
ovflowd: Does it need to be null? Just out of curiosity... Why not undefined?

Baccega: No reason actually, changed it to `number|undefined` 👍🏻

components/Common/Search/States/WithSearchBox.tsx
```diff
@@ -78,16 +89,57 @@ export const WithSearchBox: FC<SearchBoxProps> = ({ onClose }) => {
...
+        handleEnter();
+        break;
+      default:
+    }
+  });
+
+  const handleEnter = () => {
+    if (
+      selectedResult === undefined ||
+      (searchResults && searchResults?.count <= 0)
```
ovflowd:
```suggestion
      (searchResults?.count <= 0)
```

components/Common/Search/States/WithSearchBox.tsx
```diff
@@ -78,16 +89,57 @@ export const WithSearchBox: FC<SearchBoxProps> = ({ onClose }) => {
...
+    ) {
+      return;
+    }
+    const selectedHit = searchResults?.hits[selectedResult ?? 0];
```
ovflowd:
```suggestion
    const selectedHit = searchResults.hits[selectedResult];
```

You are already checking for undefined on the other if statement

components/Common/Search/States/WithSearchBox.tsx
```diff
@@ -78,16 +89,57 @@ export const WithSearchBox: FC<SearchBoxProps> = ({ onClose }) => {
...
+    if (!selectedHit) {
```
ovflowd, line 123: This should never happen, this if statement is unnecessary

components/Common/Search/States/WithSearchBox.tsx
```diff
@@ -78,16 +89,57 @@ export const WithSearchBox: FC<SearchBoxProps> = ({ onClose }) => {
...
+      return;
+    }
+
+    router.push(searchHitToLinkPath(selectedHit));
+    handleClose();
```
ovflowd, line 127: handleClose should happen probably before the router push

components/Common/Search/States/WithSearchBox.tsx
```diff
@@ -78,16 +89,57 @@ export const WithSearchBox: FC<SearchBoxProps> = ({ onClose }) => {
...
   };
 
   const onSubmit = (e: React.FormEvent<HTMLFormElement>) => {
     e.preventDefault();
+
+    // Prevent searching if a result is selected
```
ovflowd: In which situation could happen that a result is selected and that you hit enter on the input? Seems pretty impossible no?

components/Common/Search/States/WithSearchBox.tsx
```diff
@@ -131,6 +183,21 @@ export const WithSearchBox: FC<SearchBoxProps> = ({ onClose }) => {
             <form onSubmit={onSubmit}>
               <input
                 ref={searchInputRef}
+                aria-activedescendant={
+                  selectedResult != undefined
```
ovflowd:
```suggestion
                  selectedResult !== undefined
```

components/Common/Search/States/WithSearchBox.tsx
```diff
@@ -131,6 +183,21 @@ export const WithSearchBox: FC<SearchBoxProps> = ({ onClose }) => {
...
+                    ? `search-hit-${selectedResult}`
+                    : undefined
+                }
+                aria-autocomplete="list"
+                aria-controls="fulltext-results-container"
+                aria-expanded={
```
ovflowd:
```tsx
aria-expanded={Boolean(
  !searchError &&
  searchTerm &&
  searchResults && 
  searchResults.count
)}
```

components/Common/Search/States/WithSearchBox.tsx
Baccega: Unfortunately we need the null check, otherwise typescript complains, but I've removed the optional chaining since it's unnecessary with the null check

image

components/Common/Search/States/WithSearchBox.tsx
Baccega: True, changed

components/Common/Search/States/WithSearchBox.tsx
Baccega: True, removed

components/Common/Search/States/WithSearchBox.tsx
Baccega: Yeah, this sounds like a safer flow, implemented

components/Common/Search/States/WithSearchBox.tsx
Baccega: Changed it.

This is just to ensure the output right? 🤔

components/Common/Search/States/WithSearchBox.tsx
Baccega: Changed it

components/Common/Search/States/WithSearchBox.tsx
Baccega: Yeah, I've just tested it to make sure of it, after selecting a result this submit does not fire, removed it

canerakdas:
I saw this PR a little late. First of all, I apologize for my late comment 🙈 In my opinion, we should not bother with implementing such accessibility features manually, if there is a component for this in Radix primitives, let's use it. As an alternative, I know that [`cmdk`](https://cmdk.paco.me/) provides an accessible, unstyled command menu.

Why am I a fan of doing it externally?

   - Accessibility features work in harmony with each other, a feature we add manually may conflict with another feature we use.
   - Ensuring accessibility may seem easy at first, but implementing it exactly according to standards is much more difficult than it seems. (For example, the modal created here)
   - I think we should consider a general solution instead of providing accessibility only to a certain group. IMO correcting this situation manually will complicate possible accessibility changes to be made in the future. When I first looked, there were also these accessibility problems here;
     - Modal focus is not trapped
     - Facets not being controllable with arrow keys

We can use Radix's [primitive dialog](https://www.radix-ui.com/primitives/docs/components/dialog) and [tabs](https://www.radix-ui.com/primitives/docs/components/tabs)(for facets), but I'm not sure if there is a primitive component to provide this combobox pattern on the Radix side 🤔

ovflowd:
Hey, @Baccega, would you be willing to attempt to implement the suggestions @canerakdas mentioned? If not, we can merge this as it is (if @canerakdas is fine) and iterate to achieve what he mentioned.

Your contribution and time spent on this so far are sincerely appreciated!

canerakdas (approved):
LGTM is a great first PR and I learned something new from the comments here. (Any PR that takes us further from where we are looks good to me 🚀 🐢)

Baccega:
Using the tabs from our components it would look like this:
image

But there are a couple of things concerning me before going forward with this:
1. The [figma design](https://www.figma.com/file/pu1vZPqNIM7BePd6W8APA5/Node.js?type=design&node-id=464-8657&mode=design&t=Q6CzQE0kLeZRyGSG-4) looks different from the current implementation, so I'm not sure if I should work to keep the current tab style or if I can use the tab default style from our component (that IMO would be more in line with the rest of the app, but maybe you have access/knowledge of more wireframes, so I wanted to ask)
2.  We are currently not using tabs conventionally in this component because we don't really have "tabs", they feel like it, but in reality the component is fetching and rendering results in the same spot (there is no show-hide logic), and I don't know if this is a issue accessibility-wise 🤔

canerakdas:
> 1. The [figma design](https://www.figma.com/file/pu1vZPqNIM7BePd6W8APA5/Node.js?type=design&node-id=464-8657&mode=design&t=Q6CzQE0kLeZRyGSG-4) looks different from the current implementation, so I'm not sure if I should work to keep the current tab style or if I can use the tab default style from our component (that IMO would be more in line with the rest of the app, but maybe you have access/knowledge of more wireframes, so I wanted to ask)

I think it looks better, I am a fan of using components included in the design system whenever possible.
In this area, we can use the already existing `Button` component and Radix primitive Tabs or the existing `Tabs` component as you shared as an example.

In my opinion, both methods would be better than a component that is not available in the current design system. (Of course, these are my opinions, it is valuable for the team to comment, cc @nodejs/nodejs-website)

> 2. We are currently not using tabs conventionally in this component because we don't really have "tabs", they feel like it, but in reality the component is fetching and rendering results in the same spot (there is no show-hide logic), and I don't know if this is a issue accessibility-wise 🤔

Good point, as far as [I've read](https://www.w3.org/WAI/ARIA/apg/practices/keyboard-interface/#kbd_selection_follows_focus), it would be better to pass it manually with the "Enter" key instead of automatically focusing it

It seems like we can make this adjustment with the `activationMode` prop in the Root element of [Radix primitive Tabs](https://www.radix-ui.com/primitives/docs/components/tabs)

Baccega:
I've updated the PR to use the tabs component for the facets 👍🏻 
I feel that introducing cmdk in this PR is a bit out of scope, as it would require a total overall of the searchbox.

ovflowd:
Going to review this once Ive recovered from my surgery 👀

components/Common/Search/States/WithSearchBox.tsx
```diff
@@ -78,19 +89,53 @@ export const WithSearchBox: FC<SearchBoxProps> = ({ onClose }) => {
...
+        setSelectedResult(selectedResult - 1);
+        break;
+      case cmd === 'enter':
+        handleEnter();
+        break;
+      default:
+    }
+  });
+
+  const handleEnter = () => {
```
ovflowd, line 116:
```tsx
if (!searchResults) {
  return;
}

const selectedHit = searchResults.hits[selectedResult];

if (!selectedHit) {
  return;
}

handleClose();
router.push(searchHitToLinkPath(selectedHit));
```

This seems cleaner to me 👀

ovflowd (approved):
Besides my final comment, LGTM! Awesome stuff. Thank you so much for bearing with us and doing all these changes ❤️ 

If there's only two changes I could ask for is:

- Override the Tabs styles to look like the previous styles 
  image
- When there's no search query, show all results from the selected tab (or at least follow the logic of showing the fist X results)

I think that'd be great probably?

components/Common/Search/States/WithSearchBox.tsx
Baccega: True, changed 👍🏻

Baccega:
> Besides my final comment, LGTM! Awesome stuff. Thank you so much for bearing with us and doing all these changes ❤️

Happy to contribute 😄 

> 
> If there's only two changes I could ask for is:
> 
> * Override the Tabs styles to look like the previous styles
>   image

I've now update the tabs to include a "secondaryLabel" option, similar to the previous style

Screenshot 2024-04-10 at 13 52 19

Screenshot 2024-04-10 at 13 52 27

> * When there's no search query, show all results from the selected tab (or at least follow the logic of showing the fist X results)

I've removed the empty ("Search something...") state, but I'm not sure if the results will actually show on production because they depend on how the search with an empty term will behave (locally I've use mock data to develop the feature).
Do you have any suggestion/do you know if the search returns something with an empty search term?

> 
> I think that'd be great probably?

ovflowd:
> I've removed the empty ("Search something...") state, but I'm not sure if the results will actually show on production because they depend on how the search with an empty term will behave (locally I've use mock data to develop the feature).
Do you have any suggestion/do you know if the search returns something with an empty search term?

@micheleriva is there a way folks can have a test database locally? 🤔

ovflowd:
I also guess you can always wait for the Vercel deployment to test things out :pokerface:

micheleriva:
@ovflowd you can set the environment variables for local development!

ovflowd:
> @ovflowd you can set the environment variables for local development!

That is not the issue. I don't want to share our api keys publicly. Is there a way folks can test things locally without using our keys?

micheleriva:
@ovflowd I understand, but keep in mind that `ORAMA_CLOUD_ENDPOINT` and `ORAMA_CLOUD_API_KEY` are commit-safe and can be shared publicly. These are the **only** two variables you could share.

PR #6543, opened 2024-03-25
feat(tabs): fix contrast ratio

AugustinMauroy:
## Description

To solve issue I had follow Figma design. And allow forcing dark theme to allow codeTabs to force theme.

## Related Issues

Fixe #6527 

### Check List

- *NA* I've covered new added functionality with unit tests if necessary.

### DIFF
--- a/components/Common/CodeTabs/index.tsx
+++ b/components/Common/CodeTabs/index.tsx
@@ -23,6 +23,7 @@ const CodeTabs: FC<PropsWithChildren<CodeTabsProps>> = ({
   <Tabs
     {...props}
     className={styles.root}
+    forceDarkTheme={true}
     addons={
       linkUrl &&
       linkText && (
--- a/components/Common/CodeTabs/index.tsx
+++ b/components/Common/CodeTabs/index.tsx
@@ -23,7 +23,7 @@ const CodeTabs: FC<PropsWithChildren<CodeTabsProps>> = ({
   <Tabs
     {...props}
     className={styles.root}
-    forceDarkTheme={true}
+    triggerClassName={styles.trigger}
     addons={
       linkUrl &&
       linkText && (
--- a/components/Common/CodeTabs/index.module.css
+++ b/components/Common/CodeTabs/index.module.css
@@ -14,7 +14,7 @@
       pt-3
       xs:px-2;
 
-    > button {
+    .trigger {
       @apply border-b
         border-b-transparent
         px-1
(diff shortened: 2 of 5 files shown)


components/Common/CodeTabs/index.tsx
```diff
@@ -23,6 +23,7 @@ const CodeTabs: FC<PropsWithChildren<CodeTabsProps>> = ({
   <Tabs
     {...props}
     className={styles.root}
+    forceDarkTheme={true}
```
ovflowd: Why? Just apply the styles by default. Our Codebox is always dark

AugustinMauroy: because the tabs component could be used in a location other than the codeBox

ovflowd: This fix should not be in the Tabs component. But directly within the CodeTabs component styles. Please remove these changes from the Tabs component

AugustinMauroy: Ok, I see what you mean. I allow with you it's will more clear.

ovflowd:
Screenshots of before/after pls?

AugustinMauroy:
> Screenshots of before/after pls?

yep _**note** there are no change on dark mode and on tabs comment _

# before

Capture d’écran 2024-03-28 à 12 43 51

# after

Capture d’écran 2024-03-28 à 12 45 08

canerakdas (approved):
LGTM

components/Common/CodeTabs/index.tsx
```diff
@@ -23,6 +23,7 @@ const CodeTabs: FC<PropsWithChildren<CodeTabsProps>> = ({
   <Tabs
     {...props}
     className={styles.root}
+    triggerClassName={styles.trigger}
```
ovflowd, line 26: I don't think we need to add a new prop called "triggerClassName" you can achieve the same by overriding the Tab styles with !important (`!`) I know in general that's not the best thing, but in this case it makes sense.

AugustinMauroy: The `!important` directive also works. But I've chosen to use the className so that the css is explicit (it says the class trigger to this style).

ovflowd: It's more like, this feels very specific. But I'm fine either ways.

AugustinMauroy:
@ovflowd 👋 can we merge ?

ovflowd:
Lol, sure, go ahead ;)

PR #6545, opened 2024-03-25
fix(og): use the resolved page title

### LINKED ISSUE #6483: odd og image
is the following OG image correct? I think we should do something better than undefined:

IMG_0643

bmuenzenmeyer:
## Description

OG image generation was using the `title` - but displaying `undefined` for some pages. This was because the previous logic was only taking account `title` from frontmatter in markdown pages. The new logic now directly mirrors the implementation of the `og:twitter` title

https://github.com/nodejs/nodejs.org/blob/main/next.dynamic.mjs#L221:

```js
    const { data } = matter(source);

    pageMetadata.title = data.title
      ? `${siteConfig.title} — ${data.title}`
      : siteConfig.title;

    pageMetadata.twitter.title = pageMetadata.title;
```

## Validation

Before

image

After

image

## Related Issues

Fixes #6483

### Check List

### DIFF
--- a/next.dynamic.mjs
+++ b/next.dynamic.mjs
@@ -218,7 +218,7 @@ const getDynamicRouter = async () => {
         path
       );
       pageMetadata.openGraph.images = [
-        `${currentLocale}/next-data/og?title=${data.title}&type=${data.category ?? 'announcement'}`,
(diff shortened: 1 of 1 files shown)


bmuenzenmeyer:
requesting fast-track, as this is a small but visible bug

AugustinMauroy (approved):
LGTM !!!

PR #6547, opened 2024-03-25
Fixes search page with no query should not throw

SaraVieira:
## Description

Fixes the 500 error that happened when a user landed on the search page with no query

When this happens the page does not break and also shows a Searchbox

CleanShot 2024-03-25 at 16 00 02

## Validation

Local testing

### Check List

- [x ] I have read the [Contributing Guidelines](https://github.com/nodejs/nodejs.org/blob/main/CONTRIBUTING.md) and made commit messages that follow the guideline.


AugustinMauroy (approved):
LGTM ! Thanks for you first contribution 🎉

AugustinMauroy:
@nodejs/nodejs-website I'm requiring fast-track. you can 👍🏼 if you are ok

PR #6554, opened 2024-03-26
feat(search): Enhance shortcut key transition for OS detection

edereagzi:
## Description

This pull request introduces a smoother transition for the shortcut key display within the search component. Previously, the shortcut key would abruptly change from "Ctrl K" to "Cmd K" (or vice versa) based on the operating system detection. This behavior caused a visual flash that was jarring to the user.

## Validation

Before:

https://github.com/nodejs/nodejs.org/assets/81780268/3dbf1ebd-46eb-4f44-a135-050d819a3507

After:

https://github.com/nodejs/nodejs.org/assets/81780268/f4de5878-21c2-4b8e-bc41-4151efc53824

### Check List

### DIFF
--- a/components/Common/Search/index.module.css
+++ b/components/Common/Search/index.module.css
@@ -29,8 +29,8 @@
     absolute
     right-2
     ml-2
-    transition-opacity
-    duration-500
-    md:visible
-    [&>kbd]:font-ibm-plex-mono;
+    font-ibm-plex-mono
+    motion-safe:transition-opacity
+    motion-safe:duration-100
+    md:visible;
 }
--- a/components/Common/Search/index.module.css
+++ b/components/Common/Search/index.module.css
@@ -29,6 +29,8 @@
     absolute
     right-2
     ml-2
+    transition-opacity
+    duration-500
     md:visible
     [&>kbd]:font-ibm-plex-mono;
 }
--- a/components/Common/Search/index.tsx
+++ b/components/Common/Search/index.tsx
@@ -1,6 +1,7 @@
 'use client';
 
 import { MagnifyingGlassIcon } from '@heroicons/react/24/outline';
+import classNames from 'classnames';
 import { useTranslations } from 'next-intl';
 import { useState, type FC } from 'react';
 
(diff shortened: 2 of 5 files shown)


components/Common/Search/index.module.css
```diff
@@ -29,6 +29,8 @@
     absolute
     right-2
     ml-2
+    transition-opacity
+    duration-500
```
canerakdas:
```suggestion
    motion-safe:transition-opacity
    motion-safe:duration-200
```

It may be better to use `motion-safe` for users who are sensitive to this type of interaction. IMO 500 ms is a little bit long 😄 

https://drafts.csswg.org/mediaqueries-5/#prefers-reduced-motion

components/Common/Search/index.module.css
```diff
@@ -29,6 +29,8 @@
...
     md:visible
     [&>kbd]:font-ibm-plex-mono;
```
canerakdas: Since we use the `kbd` element instead of the `span` element, you can make this definition directly to the element itself

components/Common/Search/index.module.css
edereagzi: Thank you for your feedback! I agree with you on the importance of accommodating users who are sensitive to motion. That's why I've implemented the motion-safe utility to ensure that transitions only occur if the user has not specified a preference for reduced motion. Regarding the duration, I've adjusted it to 100ms to make the transition quicker while still providing a noticeable effect. 😊

components/Common/Search/index.module.css
edereagzi: Great point! I've moved the style definition directly to the kbd element.

canerakdas (approved):
LGTM! Thank you for the contribution 🚀

AugustinMauroy (approved):
LGTM ! thanks for this first contribution

shanpriyan (approved):
LGTM!

aymen94 (approved):

mikeesto (approved):

PR #6562, opened 2024-03-26
fix: RSS alternate link tag does not expose release/vuln feeds

### LINKED ISSUE #6559: RSS alternate link tag does not expose release/vuln feeds
### URL:

https://nodejs.org/en/blog/release

### Browser Name:

Firefox

### Browser Version:

124.0b9 (64-bit)

### Operating System:

macOS 14.3.1 (23D60)

Observe `application/rss+xml` alternate link tag in `head`. Should be `https://nodejs.org/en/feed/releases.xml`, but is `https://nodejs.org/en/feed/blog.xml`. Same applies to https://nodejs.org/en/blog/vulnerability

thisalihassan:
## Description

This PR addresses an issue where the RSS alternate link tags in the website's head were incorrectly pointing to the general blog feed (blog.xml) for both release and vulnerability pages. Instead, these pages should specifically reference their respective RSS feeds (releases.xml and vulnerabilities.xml). However for all other pages, the general blog RSS feed (blog.xml) is used as it is being done already (current behaviour)

## Validation

image
image

## Related Issues

closes #6559

### Check List

### DIFF
--- a/next.dynamic.mjs
+++ b/next.dynamic.mjs
@@ -212,6 +212,21 @@ const getDynamicRouter = async () => {
       path
     );
 
+    const match = path.match(/blog\/(release|vulnerability)/);
+
+    if (match) {
+      const category = match[1];
+      const currentFile = siteConfig.rssFeeds.find(
+        item => item.category === category
+      )?.file;
+      // Dynamically construct the XML path for blog/release and blog/vulnerability
+      pageMetadata.alternates.types['application/rss+xml'] =
+        `${baseUrlAndPath}/en/feed/${currentFile}`;
+    } else {
+      pageMetadata.alternates.types['application/rss+xml'] =
+        `${BASE_URL}${BASE_PATH}/en/feed/blog.xml`;
+    }
+
     availableLocaleCodes.forEach(currentLocale => {
       pageMetadata.alternates.languages[currentLocale] = getUrlForPathname(
         currentLocale,
--- a/next.dynamic.mjs
+++ b/next.dynamic.mjs
@@ -224,7 +224,7 @@ const getDynamicRouter = async () => {
         `${baseUrlAndPath}/en/feed/${currentFile}`;
     } else {
       pageMetadata.alternates.types['application/rss+xml'] =
-        `${BASE_URL}${BASE_PATH}/en/feed/blog.xml`;
+        `${baseUrlAndPath}/en/feed/blog.xml`;
     }
 
     availableLocaleCodes.forEach(currentLocale => {
--- a/next.dynamic.mjs
+++ b/next.dynamic.mjs
@@ -212,19 +212,23 @@ const getDynamicRouter = async () => {
       path
     );
 
-    const match = path.match(/blog\/(release|vulnerability)/);
-
-    if (match) {
-      const category = match[1];
+    const blogMatch = path.match(/^blog\/(release|vulnerability)(\/|$)/);
+    if (blogMatch) {
+      const category = blogMatch[1];
       const currentFile = siteConfig.rssFeeds.find(
         item => item.category === category
       )?.file;
-      // Dynamically construct the XML path for blog/release and blog/vulnerability
-      pageMetadata.alternates.types['application/rss+xml'] =
-        `${baseUrlAndPath}/en/feed/${currentFile}`;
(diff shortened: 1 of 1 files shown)


next.dynamic.mjs
```diff
@@ -212,6 +212,21 @@ const getDynamicRouter = async () => {
...
+      const currentFile = siteConfig.rssFeeds.find(
+        item => item.category === category
+      )?.file;
+      // Dynamically construct the XML path for blog/release and blog/vulnerability
+      pageMetadata.alternates.types['application/rss+xml'] =
+        `${baseUrlAndPath}/en/feed/${currentFile}`;
+    } else {
+      pageMetadata.alternates.types['application/rss+xml'] =
+        `${baseUrlAndPath}/en/feed/blog.xml`;
+    }
```
MattIPv4:
```suggestion
    pageMetadata.alternates.types['application/rss+xml'] =
      `${baseUrlAndPath}/en/feed/blog.xml`;
     
    // Dynamically construct the XML path for blog/release and blog/vulnerability   
    const blogMatch = path.match(/blog\/(release|vulnerability)/);
    if (blogMatch) {
      const category = blogMatch[1];
      const currentFile = siteConfig.rssFeeds.find(
        item => item.category === category
      )?.file;
      if (currentFile) {
        pageMetadata.alternates.types['application/rss+xml'] =
          `${baseUrlAndPath}/en/feed/${currentFile}`;
      }
    }
```
I'm curious what @ovflowd thinks here, but I wonder if we'd be better off defaulting to always setting the RSS feed to the blog, and then overwriting it if we find something "special"? With the current logic, it seems there is a risk that `currentFile` might be undefined (though in reality it won't as we know that `siteConfig.rssFeeds` has what we need).

Also, very much a nit, but `match` => `blogMatch` as a better variable name to make it clearer what it is.

next.dynamic.mjs
```diff
@@ -212,6 +212,21 @@ const getDynamicRouter = async () => {
       path
     );
 
+    const match = path.match(/blog\/(release|vulnerability)/);
```
MattIPv4: Should this have a `^` at the start to ensure some weird nested path on the site elsewhere never matches it, and perhaps a `(\/|$)` at the end, so that if a path like `/blog/release-something` ever existed it wouldn't accidentally match?

next.dynamic.mjs
```diff
@@ -212,6 +212,21 @@ const getDynamicRouter = async () => {
...
+      const category = match[1];
+      const currentFile = siteConfig.rssFeeds.find(
+        item => item.category === category
+      )?.file;
+      // Dynamically construct the XML path for blog/release and blog/vulnerability
+      pageMetadata.alternates.types['application/rss+xml'] =
+        `${baseUrlAndPath}/en/feed/${currentFile}`;
+    } else {
+      pageMetadata.alternates.types['application/rss+xml'] =
+        `${baseUrlAndPath}/en/feed/blog.xml`;
```
MattIPv4: Should these be constructed using `getUrlForPathname`, so the locale changes correctly?

next.dynamic.mjs
thisalihassan: that's a good idea I believe just to make sure, I will update the regex

next.dynamic.mjs
thisalihassan: Yeah I don't see why it will be undefined, after we update the regex we can make sure that it matches with the path properly

next.dynamic.mjs
thisalihassan: I believe currently we are always defaulting to en/feed/blog.xml this have to change for future as well

thisalihassan: Done

next.dynamic.mjs
thisalihassan: Done

next.dynamic.mjs
thisalihassan: Done

MattIPv4 (approved):

AugustinMauroy (approved):

aymen94 (approved):

thisalihassan:
Hi @ovflowd @MattIPv4 can we merge this? or do you have any comments for this PR?

PR #6619, opened 2024-04-04
fix: Search box button accessibility

canerakdas:
## Description

The contrast ratio of the background and text in the search box button does not meet the minimum contrast requirements (`4.5:1`). With this PR was updated to meet the minimum contrast ratio

https://dequeuniversity.com/rules/axe/4.8/color-contrast

## Validation

Old `3.69` / New `5.48`

image

image

### Check List

### DIFF
--- a/components/Common/Search/index.module.css
+++ b/components/Common/Search/index.module.css
@@ -8,7 +8,7 @@
     pr-4
     text-left
     text-sm
-    text-neutral-700
+    text-neutral-800
     hover:bg-neutral-300
     hover:text-neutral-900
     dark:bg-neutral-900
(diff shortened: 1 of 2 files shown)


components/Common/Search/index.module.css
```diff
@@ -28,8 +28,10 @@
   @apply invisible
```
AugustinMauroy, line 23: we should make this element as flex an item-center

canerakdas: Some stylings for this button bother me too, I'll take a look 👍

AugustinMauroy (approved):
LGTM !!! Clean 💯

bmuenzenmeyer (approved):
looks good to me - for some reason i cannot get the contrast overlay to show in the preview site for the search box

canerakdas:
> looks good to me - for some reason i cannot get the contrast overlay to show in the preview site for the search box

If there is another element other than text inside an element, it does not show contrast overlay interestingly. While checking, I manually deleted the elements(`svg`, `kbd`) and checked like that 😄

PR #6628, opened 2024-04-09
fix: redirect to default locale page when locale is not supported

yosuke-furukawa:
## Description

https://nodejs.org/ja/ pages are 404. this issue link reported about the problem.
https://github.com/nodejs/nodejs.org/issues/6501

When some locales are not supported in this page, this patch redirects the url to default locale `/en` url.

## Validation

- launch local dev environment
- go to https://localhost:3000/ja
- see the url from /ja to /en

## Related Issues

Related to [#6501](https://github.com/nodejs/nodejs.org/issues/6501)

### Check List

### DIFF
--- a/app/[locale]/[[...path]]/page.tsx
+++ b/app/[locale]/[[...path]]/page.tsx
@@ -1,5 +1,5 @@
 import { setContext, setTags } from '@sentry/nextjs';
-import { notFound } from 'next/navigation';
+import { notFound, redirect } from 'next/navigation';
 import { unstable_setRequestLocale } from 'next-intl/server';
 import type { FC } from 'react';
 
@@ -9,7 +9,11 @@ import WithLayout from '@/components/withLayout';
 import { ENABLE_STATIC_EXPORT, VERCEL_REVALIDATE } from '@/next.constants.mjs';
 import { PAGE_VIEWPORT, DYNAMIC_ROUTES } from '@/next.dynamic.constants.mjs';
 import { dynamicRouter } from '@/next.dynamic.mjs';
-import { availableLocaleCodes, defaultLocale } from '@/next.locales.mjs';
+import {
+  allLocaleCodes,
+  availableLocaleCodes,
+  defaultLocale,
+} from '@/next.locales.mjs';
 import { MatterProvider } from '@/providers/matterProvider';
 
 type DynamicStaticPaths = { path: Array<string>; locale: string };
@@ -67,7 +71,14 @@ const getPage: FC<DynamicParams> = async ({ params }) => {
     // Forces the current locale to be the Default Locale
     unstable_setRequestLocale(defaultLocale.code);
 
-    return notFound();
+    if (!allLocaleCodes.includes(locale)) {
+      // when the locale is not listed in the locales, return NotFound
+      return notFound();
+    }
+
+    // Redirect to the default locale path
+    const pathname = dynamicRouter.getPathname(path);
+    return redirect(`/${defaultLocale.code}/${pathname}`);
   }
 
   // Configures the current Locale to be the given Locale of the Request
--- a/next.locales.mjs
+++ b/next.locales.mjs
@@ -20,7 +20,11 @@ const availableLocalesMap = Object.fromEntries(
   localeConfig.map(locale => [locale.code, locale])
 );
 
+// Creates all supported locales
+const allLocaleCodes = localeConfig.map(locale => locale.code);
+
 export {
+  allLocaleCodes,
   availableLocales,
   availableLocaleCodes,
   availableLocalesMap,


app/[locale]/[[...path]]/page.tsx
```diff
@@ -67,7 +67,7 @@ const getPage: FC<DynamicParams> = async ({ params }) => {
     // Forces the current locale to be the Default Locale
     unstable_setRequestLocale(defaultLocale.code);
 
-    return notFound();
+    return redirect(`/${defaultLocale.code}`);
```
Comment (ovflowd) [inline app/[locale]/[[...path]]/page.tsx]: But wont that redirect all routes to default locale? I think the intent is to redirect to a page that exists in /en whenever it exists.

ovflowd (changes requested):

yosuke-furukawa:
updated!

app/[locale]/[[...path]]/page.tsx
```diff
@@ -67,7 +67,9 @@ const getPage: FC<DynamicParams> = async ({ params }) => {
     // Forces the current locale to be the Default Locale
     unstable_setRequestLocale(defaultLocale.code);
 
-    return notFound();
+    // Redirect to the default locale path
+    const pathname = dynamicRouter.getPathname(path);
+    return redirect(`/${defaultLocale.code}/${pathname}`);
```
Comment (ovflowd) [inline app/[locale]/[[...path]]/page.tsx:81]: Would this redirect work if the website is deployed on subpaths?

What would happen if you simply remove the redirect and do something like:

```tsx
const selectedLocale = availableLocaleCodes.includes(locale) ? locale : defaultLocale.code;

unstable_setRequestLocale(selectedLocale );
```

(And replace all other usages of `locale` to `selectedLocale`?)

without any redirect or return statement. Would an English version of said page be rendered with the links working? And if the page also doesn't exist on English, would it return a 404 then?

I think if the conditions above work the way I mentioned, then we have a good contender for a solution :)

Comment (mAmineChniti) [reply app/[locale]/[[...path]]/page.tsx:81]: i've tried out your code idea, all not found pages now are redirected to /en counterparts but that doesn't fix the navBar `Download` link

Comment (yosuke-furukawa) [reply app/[locale]/[[...path]]/page.tsx:81]: > Would this redirect work if the website is deployed on subpaths?

subpaths mean `https://nodejs.org/en/download`, `https://nodejs.org/en/blog` or something? I checked the following links.

- https://nodejs.org/en/download => 200 OK
- https://nodejs.org/en/blog => 200 OK
- https://nodejs.org/ja/download => Redirect to https://nodejs.org/en/download
- https://nodejs.org/ja/blog => Redirect to https://nodejs.org/en/blog
- https://nodejs.org/ja/foobarbaz => Redirect to https://nodejs.org/en/foobarbaz and 404
- https://nodejs.org/ja => Redirect to https://nodejs.org/en/

I believe it works well.

> What would happen if you simply remove the redirect and do something like ~ (And replace all other usages of locale to selectedLocale?) And if the page also doesn't exist on English, would it return a 404 then?

- https://nodejs.org/en/download => 200 OK
- https://nodejs.org/en/blog => 200 OK
- https://nodejs.org/ja/download => 200 OK but en page is shown.
- https://nodejs.org/ja/blog => 200 OK but en page is shown.
- https://nodejs.org/ja/foobarbaz => 404
- https://nodejs.org/ja/ => 500 unsupported locale error.

I don't feel this solution is good idea.

 if `/ja` pages could be treated as correct paths, it would be better to change i18n/config.json  like:

```
  {
    "code": "ja",
    "localName": "日本語",
    "name": "Japanese",
    "langDir": "ltr",
    "dateFormat": "YYYY.MM.DD",
    "hrefLang": "ja",
    "enabled": true, // previous false
    "default": false
  },
```

This change will enable Japanese page.  And links work fine.

- https://nodejs.org/en/download => 200 OK
- https://nodejs.org/en/blog => 200 OK
- https://nodejs.org/ja/download => 200 OK but en page is shown.
- https://nodejs.org/ja/blog => 200 OK but en page is shown.
- https://nodejs.org/ja/foobarbaz => 404
- https://nodejs.org/ja/ => 200 OK

IMHO, I feel redirect solution is better than locale enable solution. 

1. /ja pages does not have any translated pages. navigation bars, indexes etc...
2. other locales would be better to handle as correct paths? In that case, we need to update all locales to enable.

bmuenzenmeyer (approved):
Lgtm

AugustinMauroy (changes requested):
if you go on `/hhjjgkgjkgk` it's redirect to `/en` it's should display 404

yosuke-furukawa:
> if you go on `/hhjjgkgjkgk` it's redirect to `/en` it's should display 404

fixed.
if the locale is not listed in config.json, it shows 404 page.

AugustinMauroy (approved):
LGTM ! Thanks for your contribution

app/[locale]/[[...path]]/page.tsx
```diff
@@ -9,7 +9,11 @@ import WithLayout from '@/components/withLayout';
 import { ENABLE_STATIC_EXPORT, VERCEL_REVALIDATE } from '@/next.constants.mjs';
 import { PAGE_VIEWPORT, DYNAMIC_ROUTES } from '@/next.dynamic.constants.mjs';
 import { dynamicRouter } from '@/next.dynamic.mjs';
-import { availableLocaleCodes, defaultLocale } from '@/next.locales.mjs';
+import {
+  allLocales,
```
Comment (ovflowd) [inline app/[locale]/[[...path]]/page.tsx]: Do we really need to import "allLocales"? Won't availableLocaleCodes be enough?

app/[locale]/[[...path]]/page.tsx
Comment (ovflowd) [reply app/[locale]/[[...path]]/page.tsx:81]: So the reasoning behind not using redirect is because we should show 404 pages if a page doesn't actually exist instead of redirecting.

With the proposed approach:

- /ja/download would render the en download page, but still keep the /ja in the URL
- /ja/fsdjhfkisdhf would render a Not Found

But I'm fine with redirect tho.

app/[locale]/[[...path]]/page.tsx
```diff
@@ -67,7 +71,14 @@ const getPage: FC<DynamicParams> = async ({ params }) => {
     // Forces the current locale to be the Default Locale
     unstable_setRequestLocale(defaultLocale.code);
 
-    return notFound();
+    if (!allLocales.includes(locale)) {
```
Comment (ovflowd) [inline app/[locale]/[[...path]]/page.tsx]: I see, OK, that's fine I believe.

app/[locale]/[[...path]]/page.tsx
Comment (ovflowd) [inline app/[locale]/[[...path]]/page.tsx]: This variable should probably be named `allLocaleCodes`

Comment (ovflowd) [reply app/[locale]/[[...path]]/page.tsx]: But I'm also afraid of people mistaking or not understanding the difference between availableLocaleCodes and allLocaleCodes, which means probably availableLocaleCodes should be renamed to activeLocaleCodes

ovflowd (approved):
LGTM! I've left a few comments that you might or not want to address :)

app/[locale]/[[...path]]/page.tsx
Comment (ovflowd) [reply app/[locale]/[[...path]]/page.tsx:81]: (There's also the point that the redirect would cause one extra requests to our services, which is extra work and processing...)

app/[locale]/[[...path]]/page.tsx
Comment (yosuke-furukawa) [reply app/[locale]/[[...path]]/page.tsx]: Thank you. I just changed allLocales to allLocaleCodes. I feel a bit afraid that I change existing variable name. So I didn't change the availableLocaleCodes name.

app/[locale]/[[...path]]/page.tsx
Comment (ovflowd) [reply app/[locale]/[[...path]]/page.tsx:81]: cc @yosuke-furukawa if you have any thoughts.

Comment (yosuke-furukawa) [reply app/[locale]/[[...path]]/page.tsx:81]: I feel redirect solution is just a temporary solution. But currently, this is the best solution. if we would support `/ja` path, we might find the other solution when translation project will begin.

And those redirect requests are not so critical issue. If we could support i18n pages properly, those redirects are not so many. 

I believe translation crowdin will be progress in the near future.

Comment (ovflowd) [reply app/[locale]/[[...path]]/page.tsx:81]: I don't think you have the data to back you up. Please avoid speculating that these 404s requests wouldn't be a huge traffic.

The actual data, from Vercel logs, show that a lot of requests are 404. Yes, if actual languages are enabled that would reduce it, but you gotta understand every extra redirect is another request we need to process. For the user on a browser it might be transparent but 🤷

Im not really convinced with this approach. I understand it is a simple fire and go approach, but within this repository we did a lot of work to ensure we have a good performance tradeoff.

I would really appreciate if you could find an alternative. At this point I believe I'd only merge this if @bmuenzenmeyer is also fine with the change.

Comment (yosuke-furukawa) [reply app/[locale]/[[...path]]/page.tsx:81]: I'm sorry...
I don't have the knowledge that redirects have much impact on performance.
If some supported locales are enabled, we can see English pages on other locale paths without any redirects. I believe it works well for performance. If this project thinks performance is more important than translations, the solution is better.

Current translations are nothing, in that case, I have a concern with enabled flag is `true`. People can see locale switch menu, but it does not work well. 

So IMHO, I believe this PR solution is better than others.

Comment (ovflowd) [reply app/[locale]/[[...path]]/page.tsx:81]: I don't think we prioritise performance over translations, that's an unfair statement and quite provocative 😅

What Im trying to imply here is to propel a way that'd allowed us yo have both features working without a redirect. I'm sure that should be possible.

To be honest, Im not sure if the redirects depending on how many different distinct 404s we would have would affect our Vercel infrastructure. I do believe the effect is negligible, but I do believe that there might be a solution that won't require redirects. I might as well be wrong, but I want to have a 2nd opinion from Brian, before we merge this, as this PR has a great deal of an impact after all.

Comment (yosuke-furukawa) [reply app/[locale]/[[...path]]/page.tsx:81]: Thank you. I hope I could give one step forward to tackle this problem. Lots of pages are linked from blogs, books, tutorials in Japan. I believe our solution will solve those dead links.

PR #6634, opened 2024-04-09
FIx sentry integrations lazy loading

### LINKED ISSUE #6621: Sentry lazy loading in browser is incorrect

Related https://github.com/getsentry/sentry-javascript/issues/11460

r34son:
## Description

Using official guide to lazy load integrations https://docs.sentry.io/platforms/javascript/guides/nextjs/session-replay/#lazy-loading-replay

## Related Issues

Fixes https://github.com/nodejs/nodejs.org/issues/6621
Related https://github.com/getsentry/sentry-javascript/issues/11460

### Check List

### DIFF
--- a/sentry.client.config.ts
+++ b/sentry.client.config.ts
@@ -1,12 +1,12 @@
 import {
-  Dedupe,
-  Breadcrumbs,
-  HttpContext,
-  LinkedErrors,
   BrowserClient,
-  getCurrentHub,
   defaultStackParser,
   makeFetchTransport,
+  setCurrentClient,
+  breadcrumbsIntegration,
+  dedupeIntegration,
+  httpContextIntegration,
+  linkedErrorsIntegration,
 } from '@sentry/nextjs';
 
 import {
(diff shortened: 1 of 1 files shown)


r34son:
Do you have any workflow that checks bundle size?
For example i am using https://github.com/hashicorp/nextjs-bundle-analysis
Similar pr on mine repo gives -40kb https://github.com/r34son/profile/pull/346#issuecomment-2041414911

AugustinMauroy:
> Do you have any workflow that checks bundle size?
> For example i am using https://github.com/hashicorp/nextjs-bundle-analysis

used in the past. But since we catch all the roads, it doesn't give us much.

ovflowd:
It is important to mention that our Sentry integration only gets actually properly eval'd into the bundle on very specific conditions.

Which are not fulfilled by the builds within GitHub Actions.

I would appreciate it if you could run a local build with the sentry integration enabled (you can check next.config.mjs) to check the condition and our sentry config files to check when we will allow it. Generally, it requires an API token from Sentry, so you must use your own Sentry project to test this.

I'm probably not merging this until a screenshot shows the difference between the next.js build output and the webpack bundle analyzer, which shows the differences between the current `main` branch and this PR.

But Ill check the links youve attached regardless afterwards to see if your claims are valid, Sentry takes a lot of bundle size. But far, the duplication on the bundle is a bug on the webpack bundle analyzer since Sentry gets added both on the Server Side bundle and the Client Side bundle, which can at times show as duplication... This is why React looks like it is being bundled twice on the next bundle analyzer.

Note that I'm more than happy to merge this if actual data of before/after is attached :)

ovflowd (approved):
Codewise looks good, but please attach before/after so we can actually see if there are improvements 👀

r34son:
@ovflowd 
Using @next/bundle-analyzer, also set SENTRY_ENABLE to true to force using sentry locally
Selected only chunks that related to sentry
Before:
image
After:
image

ovflowd (approved):
Awesome! Looks good to me :)

Thank you for putting time to investigate this and make this happen!

PR #6635, opened 2024-04-09
1.Added a regex filter for the AIX OS. 2. Added constants for the STORYBOOK_MODE 3.Removed unecessary closing of with or '||' operator

mortale2004:
1. ### Added a regex filter for the AIX OS.
- This os was missing in the given regex for finding the current os of the user. 
- This regex will help to find out the current os by using switch case .

2. ### Added constants for the STORYBOOK_MODE 
- The constants will prevent from typos and we can change the value of it from single place.

3. ### Removed unecessary closing of <div> with or '||' operator.
- This can give error in the css styling and unecessary use of or '||' operator.
-

### DIFF
--- a/components/withNavBar.tsx
+++ b/components/withNavBar.tsx
@@ -4,6 +4,7 @@ import { useLocale } from 'next-intl';
 import { useTheme } from 'next-themes';
 import type { FC } from 'react';
 
+import { STORYBOOK_MODE_THEME } from '@/.storybook/constants';
 import NavBar from '@/components/Containers/NavBar';
 import WithBanner from '@/components/withBanner';
 import { useClientContext, useSiteNavigation } from '@/hooks';
@@ -19,7 +20,11 @@ const WithNavBar: FC = () => {
   const locale = useLocale();
 
   const toggleCurrentTheme = () =>
-    setTheme(resolvedTheme === 'dark' ? 'light' : 'dark');
+    setTheme(
+      resolvedTheme === STORYBOOK_MODE_THEME.dark
+        ? STORYBOOK_MODE_THEME.light
+        : STORYBOOK_MODE_THEME.dark
+    );
 
   return (
     <div>
--- a/components/withSidebarCrossLinks.tsx
+++ b/components/withSidebarCrossLinks.tsx
@@ -25,13 +25,13 @@ const WithSidebarCrossLinks: FC<WithCrossLinksProps> = ({ navKey }) => {
 
   return (
     <div className="mt-4 grid w-full grid-cols-2 gap-4 xs:grid-cols-1">
-      {(previousCrossLink && (
+      {previousCrossLink && (
         <CrossLink
           type="previous"
           text={previousCrossLink.label}
           link={previousCrossLink.link}
         />
-      )) || <div />}
+      )}
 
       {nextCrossLink && (
         <CrossLink
--- a/util/detectOS.ts
+++ b/util/detectOS.ts
@@ -1,7 +1,7 @@
 import type { UserOS } from '@/types/userOS';
 
 export const detectOsInUserAgent = (userAgent: string | undefined): UserOS => {
-  const osMatch = userAgent?.match(/(Win|Mac|Linux)/);
+  const osMatch = userAgent?.match(/(Win|Mac|Linux|AIX)/);
   switch (osMatch && osMatch[1]) {
     case 'Win':
       return 'WIN';
--- a/.storybook/constants.ts
+++ b/.storybook/constants.ts
@@ -1,21 +1,27 @@
+// This is a constant object which is used throughout the app for changing the theme type of the storybook
+// This object is read only due to the Object.freez so we can go with it because we are not changing its value anymore
+export const STORYBOOK_MODE_THEME = Object.freeze({
+  dark: 'dark',
+  light: 'light',
+});
 // This defines "execution" modes that Chromatic will run on the each Storybook Story
 // This allows us to test each Story with different parameters
 // @see https://www.chromatic.com/blog/introducing-story-modes/
 export const STORYBOOK_MODES = {
   'dark mobile': {
-    theme: 'dark',
+    theme: STORYBOOK_MODE_THEME.dark,
     viewport: 'small',
   },
   'dark desktop': {
-    theme: 'dark',
+    theme: STORYBOOK_MODE_THEME.dark,
     viewport: 'large',
   },
   'light mobile': {
-    theme: 'light',
+    theme: STORYBOOK_MODE_THEME.light,
     viewport: 'small',
   },
   'light desktop': {
-    theme: 'light',
+    theme: STORYBOOK_MODE_THEME.light,
     viewport: 'large',
   },
 };
--- a/.storybook/preview.tsx
+++ b/.storybook/preview.tsx
@@ -2,7 +2,11 @@ import { NextIntlClientProvider } from 'next-intl';
 
 import { withThemeByDataAttribute } from '@storybook/addon-themes';
 import { NotificationProvider } from '@/providers/notificationProvider';
-import { STORYBOOK_MODES, STORYBOOK_SIZES } from '@/.storybook/constants';
+import {
+  STORYBOOK_MODES,
+  STORYBOOK_SIZES,
+  STORYBOOK_MODE_THEME,
+} from '@/.storybook/constants';
 import type { Preview, ReactRenderer } from '@storybook/react';
 
 import englishLocale from '@/i18n/locales/en.json';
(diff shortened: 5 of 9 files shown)


ovflowd:
I don't think people download Node for AIX within AIX. It is the sort of enterprise OS that system administrators would connect remotely no?

cc @mhdawson

mortale2004:
> I don't think people download Node for AIX within AIX. It is the sort of enterprise OS that system administrators would connect remotely no?
> 
> cc @mhdawson

But there can be possibility of downloading node js on the AIX OS. If this case is not going to occur then we should remove the AIX os from the switch case also isn't it?

ovflowd:
> > I don't think people download Node for AIX within AIX. It is the sort of enterprise OS that system administrators would connect remotely no?
> 
> > 
> 
> > cc @mhdawson
> 
> 
> 
> But there can be possibility of downloading node js on the AIX OS. If this case is not going to occur then we should remove the AIX os from the switch case also isn't it? 

Are you sure AIX is part of the user agent for browsers running on AIX? Is there even any modern browser running on AIX?

Do you have example user agents?

ovflowd:
> > I don't think people download Node for AIX within AIX. It is the sort of enterprise OS that system administrators would connect remotely no?
> 
> > 
> 
> > cc @mhdawson
> 
> 
> 
> But there can be possibility of downloading node js on the AIX OS. If this case is not going to occur then we should remove the AIX os from the switch case also isn't it? 

The reason why AIX is on the switch is so people can download Node for AIX. Auto detection is just a sugar on the top, that we do for OSs that we know people open on the browser to download Node.

mortale2004:
> > > I don't think people download Node for AIX within AIX. It is the sort of enterprise OS that system administrators would connect remotely no?
> > 
> > 
> > > 
> > 
> > 
> > > cc @mhdawson
> > 
> > 
> > But there can be possibility of downloading node js on the AIX OS. If this case is not going to occur then we should remove the AIX os from the switch case also isn't it?
> 
> The reason why AIX is on the switch is so people can download Node for AIX. Auto detection is just a sugar on the top, that we do for OSs that we know people open on the browser to download Node.

In the code that i have made changes. We are just getting the constant name for the Operating Systems from the user agent of the browser and this can be used to give the Node js according to the OS of its system. This is just for the easy accessibility to download Node js.

ovflowd:
> > > > I don't think people download Node for AIX within AIX. It is the sort of enterprise OS that system administrators would connect remotely no?
> 
> > > 
> 
> > > 
> 
> > > > 
> 
> > > 
> 
> > > 
> 
> > > > cc @mhdawson
> 
> > > 
> 
> > > 
> 
> > > But there can be possibility of downloading node js on the AIX OS. If this case is not going to occur then we should remove the AIX os from the switch case also isn't it?
> 
> > 
> 
> > The reason why AIX is on the switch is so people can download Node for AIX. Auto detection is just a sugar on the top, that we do for OSs that we know people open on the browser to download Node.
> 
> 
> 
> In the code that i have made changes. We are just getting the constant name for the Operating Systems from the user agent of the browser and this can be used to give the Node js according to the OS of its system. This is just for the easy accessibility to download Node js.

You haven't answered my question. Is there any real user agent out there that contains "AIX"?

What is the point of adding AIX to the user agent query if there aren't any browser user agents with AIX within the user agent string?

ovflowd (approved):
But anyhow, the changes make sense at least in order to make the switch statement work. Although I dont think we should have AIX there to begin with. (In the OS detection, but I might be wrong tho!)

Thanks for the quick fix, appreciate the time you put into this. Ill check with @mhdawson regarding if AIX os detection is intended or not 🤔

mortale2004:
> > > > > I don't think people download Node for AIX within AIX. It is the sort of enterprise OS that system administrators would connect remotely no?
> > 
> > 
> > > >
> > 
> > 
> > > >
> > 
> > 
> > > > >
> > 
> > 
> > > >
> > 
> > 
> > > >
> > 
> > 
> > > > > cc @mhdawson
> > 
> > 
> > > >
> > 
> > 
> > > >
> > 
> > 
> > > > But there can be possibility of downloading node js on the AIX OS. If this case is not going to occur then we should remove the AIX os from the switch case also isn't it?
> > 
> > 
> > > 
> > 
> > 
> > > The reason why AIX is on the switch is so people can download Node for AIX. Auto detection is just a sugar on the top, that we do for OSs that we know people open on the browser to download Node.
> > 
> > 
> > In the code that i have made changes. We are just getting the constant name for the Operating Systems from the user agent of the browser and this can be used to give the Node js according to the OS of its system. This is just for the easy accessibility to download Node js.
> 
> You haven't answered my question. Is there any real user agent out there that contains "AIX"?
> 
> What is the point of adding AIX to the user agent query if there aren't any browser user agents with AIX within the user agent string?

Yes, you are right currently there is no any user agent specifiec to the AIX OS but in future it can come. Because AIX is also a popular OS.

mortale2004:
Thank You for your appreciation to this small issue 😊

mortale2004 (commented):
1. Added Filter for AIX OS
2. Used Constants for theme of STORY_BOOK
3. Removed Extra closing of </div>

ovflowd:
I see unrelated changes on this PR from the last time Ive approved the PR 🤔

Would you mind putting those changes on another PR as they are unrelated?

mortale2004:
> I see unrelated changes on this PR from the last time Ive approved the PR 🤔
> 
> Would you mind putting those changes on another PR as they are unrelated?

Okay I will do that just tell me when this PR will be merged?

ovflowd:
It is written on the CONTRIBUTING.md file, but in general changes require 48h to get merged, but since the original change is simple I can "fast track" this PR which would allow me to merge it immediately

mortale2004:
> It is written on the CONTRIBUTING.md file, but in general changes require 48h to get merged, but since the original change is simple I can "fast track" this PR which would allow me to merge it immediately

Is there need to create different PR for the other commit becuase I have added that all issues commited in the Description of this PR.

ovflowd:
> > It is written on the CONTRIBUTING.md file, but in general changes require 48h to get merged, but since the original change is simple I can "fast track" this PR which would allow me to merge it immediately
> 
> 
> 
> Is there need to create different PR for the other commit becuase I have added that all issues commited in the Description of this PR.

Unfortunately, yes. I can re review this PR with the new changes but with the new changes this PR is not eligible anymore to be fast tracked and it might take time to be reviewed.

mortale2004:
Ok no problem you can take your time

mhdawson:
I read this as indicating its unlikely people will be using a brower on AIX to directly access the site- https://www.ibm.com/support/pages/ibm-aix-firefox-web-browser-faq.  Instead most likely they will download from another system and transfer to the AIX server.

I'm ok with the autodetection but agree we are likely ok without it.

mortale2004:
> I read this as indicating its unlikely people will be using a brower on AIX to directly access the site- https://www.ibm.com/support/pages/ibm-aix-firefox-web-browser-faq. Instead most likely they will download from another system and transfer to the AIX server.
> 
> I'm ok with the autodetection but agree we are likely ok without it.

If you are Ok without autodetection then there is no need to keep AIX in the switch case for the detection of OS. It will won't work in any case because there is no any browser agent for the AIX OS.

ovflowd:
@mortale2004 @mhdawson how y'all want to proceed?

mhdawson:
I'd just leave out the autodection for AIX.

mortale2004:
Okay no problem

components/withSidebarCrossLinks.tsx
```diff
@@ -25,13 +25,13 @@ const WithSidebarCrossLinks: FC<WithCrossLinksProps> = ({ navKey }) => {
 
   return (
     <div className="mt-4 grid w-full grid-cols-2 gap-4 xs:grid-cols-1">
-      {(previousCrossLink && (
+      {previousCrossLink && (
```
ovflowd, line 28: Afaik the `div` element is intentional, otherwise the `nextCrossLink` goes to the left, instead of staying in the right 😅

util/detectOS.ts
```diff
@@ -1,7 +1,7 @@
 import type { UserOS } from '@/types/userOS';
 
 export const detectOsInUserAgent = (userAgent: string | undefined): UserOS => {
-  const osMatch = userAgent?.match(/(Win|Mac|Linux)/);
+  const osMatch = userAgent?.match(/(Win|Mac|Linux|AIX)/);
```
ovflowd, line 4: I think as mentioned before, instead of adding `AIX` to this match, we should remove the "case AIX"

components/withNavBar.tsx
```diff
@@ -4,6 +4,7 @@ import { useLocale } from 'next-intl';
 import { useTheme } from 'next-themes';
 import type { FC } from 'react';
 
+import { STORYBOOK_MODE_THEME } from '@/.storybook/constants';
```
ovflowd, line 7: As mentioned before, we shouldn't have Sotrybook constants being imported in Application code, you can create a new constant similar to the Storybook one, on an App constants file, like `next.constants.mjs`

PR #6643, opened 2024-04-10
fix: move update-v8-5.4.md to announcements folder

### LINKED ISSUE #6641: [Broken Link]: Node.js v7 has updated V8 to 5.4
### URL:

https://nodejs.org/en/blog/announcements/update-v8-5.4

### Browser Name:

Chrome

### Browser Version:

123.0.6312.87

### Operating System:

MacOS

1. Visit: https://nodejs.org/en/blog/announcements/nodejs-certified-developer-program
2. Click the next button: image
3. This leads to a 404 page to: https://nodejs.org/en/blog/announcements/update-v8-5.4
image

reeveng:
## Description

This approach assumes the category was right and it was placed in the wrong folder. The other approach that could fix the issue is switching out the category in the frontmatter, but think this is the right approach.

## Validation

Assumed and validated that the path is depended on the location within the folder structure (blog pages in the blog/ folder at least).

## Related Issues

Fixes #6641

### Check List


reeveng:
To fix this in the future, I could also add a "test" that fails the husky precommit, would basically check whether the directory corresponds to the category in the frontmatter.

AugustinMauroy (changes requested):
The problem is not the file name. And it **MUST** not be changed. This should come from the `withSidebarCrossLinks` component.

reeveng:
> The problem is not the file name. And it **MUST** not be changed. This should come from the `withSidebarCrossLinks` component.

Never changed the file name, just the files location. Besides, 2 people working on the same thing doesn't make sense. So since the other PR is more active, the other guy can take it over, not here to waste either of ours time.

PR #6662, opened 2024-04-17
fix: Optimizations for Turborepo setup

anthonyshew:
## Description

I dropped in to look at how the Node.js website team is using Turborepo and saw a few places I could help out with small optimizations.

## Validation

I've dropped a few comments in the file changes. To test things locally, you can try running the commands I've altered to see the new behaviors. Additionally, you can change files in the repo to see how the caching works across runs/tasks.

One thing I can't confidently test is the GitHub Action script that I've edited. You'll notice that I collapsed two steps into one to run more `turbo` tasks in parallel. I'll lean on a maintainer's help there to make sure I've not created an issues/missed any context. 

### Check List

### DIFF
--- a/.github/workflows/lint-and-tests.yml
+++ b/.github/workflows/lint-and-tests.yml
@@ -126,7 +126,7 @@ jobs:
             github.event.pull_request.head.ref != 'chore/crowdin')
         # We want to enforce that the actual `turbo@latest` package is used instead of a possible hijack from the user
         # the `${{ needs.base.outputs.turbo_args }}` is a string substitution happening from the base job
-        run: npx --package=turbo@latest -- turbo lint:js lint:md lint:css prettier ${{ needs.base.outputs.turbo_args }
+        run: npx --package=turbo@latest -- turbo lint:js lint:md lint:css prettier ${{ needs.base.outputs.turbo_args }}
 
       - name: Run `tsc build`
         # We want to ensure that the whole codebase is passing and successfully compiles with TypeScript
--- a/.github/workflows/lint-and-tests.yml
+++ b/.github/workflows/lint-and-tests.yml
@@ -60,7 +60,7 @@ jobs:
       (github.event_name == 'pull_request_target' &&
         github.event.label.name == 'github_actions:pull-request')
 
-    name: Lint
+    name: Quality checks
     runs-on: ubuntu-latest
     needs: [base]
 
@@ -116,8 +116,7 @@ jobs:
         # We also use `npm i` instead of `npm ci` so that the node_modules/.cache folder doesn't get deleted
         run: npm i --no-audit --no-fund --ignore-scripts --userconfig=/dev/null
 
-      - name: Run `turbo lint`
-        id: eslint-step
+      - name: Run quality checks with `turbo`
         # We run the ESLint and Prettier commands on all Workflow triggers of the `Lint` job, besides if
         # the Pull Request comes from a Crowdin Branch, as we don't want to run ESLint and Prettier on Crowdin PRs
         # Note: Linting and Prettifying of files on Crowdin PRs is handled by the `translations-pr.yml` Workflow
@@ -127,13 +126,7 @@ jobs:
             github.event.pull_request.head.ref != 'chore/crowdin')
         # We want to enforce that the actual `turbo@latest` package is used instead of a possible hijack from the user
         # the `${{ needs.base.outputs.turbo_args }}` is a string substitution happening from the base job
-        run: npx --package=turbo@latest -- turbo lint ${{ needs.base.outputs.turbo_args }}
-
-      - name: Run `turbo prettier`
-        if: steps.eslint-step.outcome == 'success'
-        # We want to enforce that the actual `turbo@latest` package is used instead of a possible hijack from the user
-        # the `${{ needs.base.outputs.turbo_args }}` is a string substitution happening from the base job
-        run: npx --package=turbo@latest -- turbo prettier ${{ needs.base.outputs.turbo_args }}
+        run: npx --package=turbo@latest -- turbo lint:js lint:md lint:css prettier ${{ needs.base.outputs.turbo_args }
 
       - name: Run `tsc build`
         # We want to ensure that the whole codebase is passing and successfully compiles with TypeScript
--- a/package.json
+++ b/package.json
@@ -23,8 +23,8 @@
     "lint:js": "eslint \"**/*.{js,mjs,ts,tsx}\" --cache --cache-strategy=content --cache-location=.eslintjscache",
     "lint:md": "eslint \"**/*.md?(x)\" --cache --cache-strategy=content --cache-location=.eslintmdcache",
     "lint:css": "stylelint \"**/*.css\" --allow-empty-input --cache --cache-strategy=content --cache-location=.stylelintcache",
-    "lint": "npm run lint:js && npm run lint:md && npm run lint:css",
-    "lint:fix": "npm run lint:js -- --fix && npm run lint:md -- --fix && npm run lint:css -- --fix",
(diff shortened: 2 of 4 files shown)


package.json
```diff
@@ -33,7 +33,6 @@
     "storybook:build": "cross-env NODE_NO_WARNINGS=1 storybook build --quiet --webpack-stats-json",
     "test:unit": "cross-env NODE_NO_WARNINGS=1 jest",
     "test:unit:watch": "npm run test:unit -- --watch",
-    "test": "npm run test:unit",
```
anthonyshew, line 36: I couldn't find anywhere that this command was used so I removed it as a small cleanup. If it does indeed get used, happy to bring it back.

.github/workflows/lint-and-tests.yml
```diff
@@ -127,13 +126,7 @@ jobs:
...
         # We want to enforce that the actual `turbo@latest` package is used instead of a possible hijack from the user
         # the `${{ needs.base.outputs.turbo_args }}` is a string substitution happening from the base job
-        run: npx --package=turbo@latest -- turbo lint ${{ needs.base.outputs.turbo_args }}
-
-      - name: Run `turbo prettier`
-        if: steps.eslint-step.outcome == 'success'
-        # We want to enforce that the actual `turbo@latest` package is used instead of a possible hijack from the user
-        # the `${{ needs.base.outputs.turbo_args }}` is a string substitution happening from the base job
-        run: npx --package=turbo@latest -- turbo prettier ${{ needs.base.outputs.turbo_args }}
+        run: npx --package=turbo@latest -- turbo lint:js lint:md lint:css prettier ${{ needs.base.outputs.turbo_args }
```
anthonyshew: These tasks can all be made to run in parallel, saving some time.

turbo.json
```diff
@@ -143,16 +147,6 @@
...
-    },
-    "test": {
-      "inputs": [
-        "{app,components,hooks,i18n,layouts,middlewares,pages,providers,types,util}/**/*.{ts,tsx,mjs}",
-        "{app,components,layouts,pages,styles}/**/*.css",
-        "{next-data,scripts,i18n}/**/*.{mjs,json}",
-        "{app,pages}/**/*.{mdx,md}",
-        "*.{md,mdx,json,ts,tsx,mjs,yml}"
-      ],
-      "outputs": ["coverage/**", "junit.xml"]
```
anthonyshew, line 155: This appears to be a duplicate of the `test:unit` task so we can remove it and run `test:unit` instead.

turbo.json
```diff
@@ -83,27 +83,31 @@
       ]
     },
     "lint:js": {
-      "cache": false,
+      "inputs": [
+        "!styles/**/*.css",
+        "!components/**/*.css",
+        "!layouts/**/*.css",
+        "!*.md",
+        "!pages/**/*.{md,mdx}"
```
anthonyshew: These glob patterns are the reverse of the `inputs` for the other two tasks below.

It would be great to do a glob more like `**/*.{js,mjs,ts,tsx}` but this significantly slows down the file discovery step. Even a full cache hit ends up taking multiple seconds, whereas a glob pattern like this results in full cache hits taking milliseconds.

turbo.json
```diff
@@ -83,27 +83,26 @@
...
     },
     "lint:css": {
-      "cache": false,
+      "inputs": ["{app,components,layouts,pages,styles}/**/*.css"],
       "outputs": [".stylelintcache"]
     },
-    "lint": {
-      "cache": false,
-      "outputs": [".eslintjscache", ".eslintmdcache", ".stylelintcache"]
-    },
```
anthonyshew, line 100: We don't end up needing this task because we can use the other linting tasks in parallel instead. An example of this can be found in the `package.json` changes.

package.json
anthonyshew: Found it in the pull request template!

turbo.json
```diff
@@ -83,41 +83,33 @@
...
     },
     "format": {
-      "cache": false,
-      "outputs": [
-        ".eslintjscache",
-        ".eslintmdcache",
-        ".stylelintcache",
-        ".prettiercache"
-      ]
+      "cache": false
```
anthonyshew, line 112: Any time `cache` is false, nothing will be sent to outputs. So, we can remove these lines for clarity.

anonrig (approved):

.github/workflows/lint-and-tests.yml
rluvaton: isn't it dangerous to use the latest version? what if breaking changes are made? we should use the version our website use

anthonyshew: I also wondered about this but wasn't going to take the liberty to question it, especially given that the comment right above describes the motivation. I figured a core team member consciously made this choice, carefully considering the tradeoffs - and that I'm not thinking about it deeply enough to understand the hijacking surface being described. I'd be interested in learning more!

rluvaton: I missed that comment 😅

ovflowd:
@anthonyshew apparently the new updated command causes a loop? 

```
x root task //#lint (turbo run lint:md lint:js lint:css) looks like it
  | invokes turbo and might cause a loop
```

anthonyshew:
@ovflowd Hm, that's weird. I changed the GitHub Actions run to not use `turbo lint` and, instead, call the individual tasks [here](https://github.com/nodejs/nodejs.org/pull/6662/files#r1570510682). Would there be any reason to believe that a stale version of the GitHub Action was run?

For what it's worth, I also see steps listed in the Action failure that have been removed as a part of this pull request.

ovflowd:
> @ovflowd Hm, that's weird. I changed the GitHub Actions run to not use `turbo lint` and, instead, call the individual tasks [here](https://github.com/nodejs/nodejs.org/pull/6662/files#r1570510682). Would there be any reason to believe that a stale version of the GitHub Action was run?
> 
> 
> 
> For what it's worth, I also see steps listed in the Action failure that have been removed as a part of this pull request.

Hmmm, yes? Because the actual workflows used on a PR are the ones from main branch for security reasons.

If you believe this PR is safe to merge we can give it a go and see if things pass on main

anthonyshew:
@ovflowd Makes sense! I was thinking that was set up but didn't want to assume.

As a brief explanation, the new Action will be using `turbo lint:md lint:js lint:css` directly so `turbo` won't be calling `turbo`, per the error message.

I'm confident that this is safe to merge in that regard and happy to play the part of on-call to quickly fix-forward anything that may come up.

.github/workflows/lint-and-tests.yml
rikhall1515: You forgot the second closing brace at the end in `${{ needs.base.outputs.turbo_args }`. 

Everything else looks good!

ovflowd: Right, I believe we could use commit pinning here; Bug the idea behind is, I might be wrong, that NPM checks what is the latest version on the registry, and then compares what the latest locally version is installed, so it complicates even if a tiny bit more, that the attacker must match both versions if they want to poison a GitHub Action cache.

ovflowd: @anthonyshew can you fix that one? 👀

ovflowd: Thanks @rikhall1515 for spotting this!!

anthonyshew: Oops! Thanks for that catch, @rikhall1515!

ovflowd:
Let's see if `main` branch passes, @anthonyshew :)

ovflowd:
FYI @nodejs/nodejs-website I forced a merge, since the checks would have failed regardless, since the Workflow was being updated.

PR #6666, opened 2024-04-20
Update twisted link in docs

matheusfelipeog:
The Twisted link used in the docs (https://twistedmatrix.com/trac) redirects to https://twisted.org. I just updated it to the final link.


mikeesto (approved):
LGTM, thanks

AugustinMauroy (approved):
LGTM ! Good catch

PR #6686, opened 2024-05-01
Fix: change the copy to clipboard behavior

### LINKED ISSUE #6685: Changes in Copy to ClipBoard button in main page
### URL:

https://nodejs.org/en

### Browser Name:

brave

### Browser Version:

1.65.123

### Operating System:

Windows 11

### The Problem: 
1. the current copy to clipboard has a display like this

image

making it not so UI friendly. and in smaller screen it is not even showing

### Approach:
- instead of the popup like the current version, we can so make it transition to a state of true and false when it is being clicked.
- so when it is click it transition to the clicked state.

TenzDelek:
Fixes #6685 
## Description
the current behavior of the copy button doesn't seems to work in the smaller screen. and also it is not appealing for a notification style when we click on it.
- changes
- made state based transition for the button allowing a smooth interaction for both smaller screen and large one
## Validation
before click:
image

after click:
image

### Check List

### DIFF
--- a/components/Common/CodeBox/index.tsx
+++ b/components/Common/CodeBox/index.tsx
@@ -6,10 +6,10 @@ import {
 } from '@heroicons/react/24/outline';
 import { useTranslations } from 'next-intl';
 import type { FC, PropsWithChildren, ReactNode } from 'react';
-import { Fragment, isValidElement, useRef } from 'react';
+import { Fragment, isValidElement, useRef, useState } from 'react';
 
 import Button from '@/components/Common/Button';
-import { useCopyToClipboard, useNotification } from '@/hooks';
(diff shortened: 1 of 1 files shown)


AugustinMauroy:
A word of advice: it's smarter to wait until you've reached a consensus on the issue before launching into the code. Because personally I'm very happy to have a toast. Also, your change applies to all codeBoxes, not just the one for the landing page.

TenzDelek:
> A word of advice: it's smarter to wait until you've reached a consensus on the issue before launching into the code. Because personally I'm very happy to have a toast. Also, your change applies to all codeBoxes, not just the one for the landing page.

Appreciate your concern, closing this repo until further discussion is reach.. :smiley:

PR #6680, opened 2024-04-30
fix: horizontal scrollbar appearing on small screen

theoludwig:
## Description

On small screen (e.g: width 412px), on some security blog post page (e.g: https://nodejs.org/en/blog/vulnerability/april-2024-security-releases), there is a horizontal scrollbar appearing because of the texts overflow in Table of Contents, and added padding in footer (we don't need this padding on small screens, as the "Trademark Policy", "Privacy Policy", etc. is centered). 

We can clearly see the horizontal (overflow-x) scrollbar on this screen:

image

## Validation

This PR fixes this issue by wrapping the texts:

image

### Check List

- [x] I have read the [Contributing Guidelines](https://github.com/nodejs/nodejs.org/blob/main/CONTRIBUTING.md) and made commit messages that follow the guideline.
- [x] I have run `npm run format` to ensure the code follows the style guide.
- [x] I have run `npm run test` to check if all tests are passing.
- [x] I have run `npx turbo build` to check if the website builds without errors.

PR #6712, opened 2024-05-05
fix: layout word break

### LINKED ISSUE #6711: Blog Word Break
### URL:

https://nodejs.org/en/blog/announcements/v22-release-announce

### Browser Name:

Arc

### Browser Version:

1.41.0 (49440)

### Operating System:

MacOS 14.4.1 (23E224)

The blog currently uses `word-break: break-all` for the body text. This seems like a bug to me. It makes it difficult to read the posts. I would recommend changing it to `word-break: normal`.

1. Go to https://nodejs.org/en/blog/announcements/v22-release-announce
2. Look at the end of each line of text. Many words will be chopped off, not even hyphenated.

It looks like it was introduced in #6680. I think that change was only supposed to affect small screens, like phones, but I am using a full desktop browser on a large screen and having the issue.

## Before
Notice the way that every line ends in a broken word.

image

## After (`word-break: normal`)
image

araujogui:
## Description

Change layout's word break to normal

## Validation

image

## Related Issues

Fixes #6711

### Check List

### DIFF
--- a/layouts/layouts.module.css
+++ b/layouts/layouts.module.css
@@ -215,7 +215,7 @@
     main {
       @apply max-w-[660px]
         gap-4
-        break-all;
+        break-normal;
     }
   }
 }
--- a/layouts/layouts.module.css
+++ b/layouts/layouts.module.css
@@ -215,7 +215,8 @@
     main {
       @apply max-w-[660px]
         gap-4
-        break-normal;
+        break-all
+        lg:break-normal;
     }
   }
 }


AugustinMauroy (approved):

mikeesto (approved):

aymen94 (approved):

canerakdas:
It seems to overflow content for low-resolution mobile browsers, for example;

https://nodejs-jow7mncue-openjs.vercel.app/en/blog/release/v20.12.0

IMO it might be better to use `xs:break-words` for mobile resolution

bmuenzenmeyer:
yeah - we have dueling solutions here now and we need to slow down and consider our use cases and options

canerakdas (approved):
LGTM

shanpriyan (approved):

NiedziolkaMichal:
You guys may consider `text-align: justify`:
image

krave1986:
I would suggest try

```
word-break: normal;
overflow-wrap: anywhere;
hyphens: auto;
```
This will both keep the result closest to the current design and bring hyphens to broken words at the end of the lines.

Another benefit would be that we don't need to concern media query for such cases further.

ovflowd:
Look we also already tried the hyphens approach. Each approach has its own cons and pros and different side effects.

I think is worth going over git history to see what was already attempted and what not and if whatever proposed works well on all scenarios

tniessen:
I came across this PR after running into this odd issue and trying to fix it myself. For me, personally, justified alignment and normal word break work fine in combination.

ovflowd:
@araujogui could you go over the considerations that both me, Caner and Brian mentioned? 👀

araujogui:
Found out that word-break was introduced in #6680, so I think we can assume that until then it was the browsers defaults (normal). The issue solved in #6680 didn't come back with this PR.

PR #6728, opened 2024-05-21
fix: added digit regex

TenzDelek:
- additional small changes for breadcrumb
the url of the ECMAScript in learn page contain digit after the dash ("https://nodejs.org/en/learn/getting-started/ecmascript-2015-es6-and-beyond") which is not handle resulting in below behavior

the current behavior:
image

after adding digit regex:
image

### DIFF
--- a/util/__tests__/stringUtils.test.mjs
+++ b/util/__tests__/stringUtils.test.mjs
@@ -1,6 +1,7 @@
 import {
   getAcronymFromString,
   parseRichTextIntoPlainText,
+  dashToCamelCase,
 } from '@/util/stringUtils';
 
 describe('String utils', () => {
@@ -58,5 +59,23 @@ describe('String utils', () => {
     const richText = '   Line 1   \n   Line 2   \n   Line 3   ';
     const result = parseRichTextIntoPlainText(richText);
     expect(result).toBe('Line 1\nLine 2\nLine 3');
+  });
+
+  it('should convert dash-separated strings to camelCase', () => {
+    expect(dashToCamelCase('hello-world')).toBe('helloWorld');
+    expect(dashToCamelCase('my-name-is')).toBe('myNameIs');
+    expect(dashToCamelCase('dash-to-camel-case')).toBe('dashToCamelCase');
+  });
+
+  it('should handle strings without dashes', () => {
+    expect(dashToCamelCase('hello')).toBe('hello');
+    expect(dashToCamelCase('world')).toBe('world');
+  });
+
+  it('should handle numbers in the string', () => {
+    expect(dashToCamelCase('my-name-is-007')).toBe('myNameIs007');
+    expect(dashToCamelCase('123-abc')).toBe('123Abc');
+    expect(dashToCamelCase('test-1-2-3')).toBe('test123');
+    expect(dashToCamelCase('test-123-abc')).toBe('test123Abc');
   });
 });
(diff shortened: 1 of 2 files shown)


AugustinMauroy:
I think it would be good to have a unit test for this would you do it? That would make it possible once and for all to see if all the use cases are correct.

TenzDelek:
> I think it would be good to have a unit test for this would you do it? That would make it possible once and for all to see if all the use cases are correct.

I believe I don't have prior experience with writing unit test for now, but I would love to do it . If it possible for any one of the members to guide me it can be helpful.

AugustinMauroy:
OK I can help you
We have a docs [General Guidelines for Unit Tests](https://github.com/nodejs/nodejs.org/blob/main/COLLABORATOR_GUIDE.md#general-guidelines-for-unit-tests) can help you

there are already a unit test file for `util/stringUtils.ts` it's `util/__tests__/stringUtils.test.ts`. 

to add the test:
1. import function
2. each use case of the function must be covered by an `it`.
For example : `it("dashToCamelCase should work with number", () => {`
3. need to use expect to "expect" what test should test. I recommend you to read this doc https://jestjs.io/docs/expect

`npm run test` to run test suit. if everything pass you can commit !
also you can use `npm run test:unit:watch` to start watching file so each time you will save it's will re-run the task

TenzDelek:
> OK I can help you We have a docs [General Guidelines for Unit Tests](https://github.com/nodejs/nodejs.org/blob/main/COLLABORATOR_GUIDE.md#general-guidelines-for-unit-tests) can help you
> 
> there are already a unit test file for `util/stringUtils.ts` it's `util/__tests__/stringUtils.test.ts`.
> 
> to add the test:
> 
> 1. import function
> 2. each use case of the function must be covered by an `it`.
>    For example : `it("dashToCamelCase should work with number", () => {`
> 3. need to use expect to "expect" what test should test. I recommend you to read this doc https://jestjs.io/docs/expect
> 
> `npm run test` to run test suit. if everything pass you can commit ! also you can use `npm run test:unit:watch` to start watching file so each time you will save it's will re-run the task

Thanks a lot !! will go through the docs and check

TenzDelek:
hello @AugustinMauroy  the testcases are passing for dashtocamel. (can you check once whether the testcase are fine. i used some random values)  thanks for the help. 

## also one thing to be note, during the test, I think one of the testcase (unrelated to this pr) is failing. 
image

AugustinMauroy:
@TenzDelek strange because here on CI it's pass

AugustinMauroy (approved):
LGTM !!!

aymen94 (approved):

TenzDelek:
@AugustinMauroy again thanks for the help, that's the one thing that I love about open source is that we can learn and explore new things. Happy learning 🙏🏼

AugustinMauroy:
> @AugustinMauroy again thanks for the help, that's the one thing that I love about open source is that we can learn and explore new things. Happy learning 🙏🏼

I try to do my best, but I'm not a teacher 😄.

bmuenzenmeyer:
https://nodejs-org-git-fork-tenzdelek-fix-added-digit-regex-openjs.vercel.app/en/learn/manipulating-files/working-with-different-filesystems still looks broken to me

maybe we need a combo of both solutions 😄

ovflowd:
As commented on the other PR, I'm closing this one in favour of @abizek's PR 🙇

PR #6734, opened 2024-05-24
feat(pride logo): introduce

AugustinMauroy:
## Description

Read issues #6733 to understand what is going on.

## Validation

Capture d’écran 2024-05-24 à 23 25 37

Capture d’écran 2024-05-24 à 23 25 25

## Related Issues

Related to #6733 

### Check List


components/Containers/NavBar/index.tsx
```diff
@@ -31,21 +33,32 @@ type NavbarProps = {
...
+            <>
+              <NodejsDarkPride className={style.nodejsLogoDark} />
+              <NodejsLightPride className={style.nodejsLogoLight} />
+            </>
+          ) : (
+            <>
+              <NodejsDark className={style.nodejsLogoDark} />
+              <NodejsLight className={style.nodejsLogoLight} />
+            </>
+          )}
```
canerakdas: What if we combine the `NodejsDark`, `NodejsLight`, `NodejsDarkPride`, and `NodejsLightPride` SVG components into a single component, perhaps called `NodejsLogo`?

Currently, we have four different versions, and if we plan to add a special logo in the future, it would be more efficient to manage these variations within a single component. This way, we can handle the logo's conditions directly within its component, rather than managing them with conditions in the NavBar container. I believe this approach would be more understandable and easier to edit 🤔

AugustinMauroy: I'll try that no dum 😄

AugustinMauroy: Updated

ovflowd: I'm not sure you addressed @canerakdas point? Afaik for what I understood it should just become one logo that encompasses all (and since they're a React component) you could simply have props that modify the parts that differ them.

Pretty much the main difference between dark/light are the stroke colours I'd assume

AugustinMauroy: I understood that the condition of variation had to take place somewhere other than in the navbar. Then I may have misunderstood the suggestion.

canerakdas: > I'm not sure you addressed @canerakdas point? Afaik for what I understood it should just become one logo that encompasses all (and since they're a React component) you could simply have props that modify the parts that differ them.
> 
> Pretty much the main difference between dark/light are the stroke colours I'd assume

I didn't think of it this way, but I think it's a better idea 😄

AugustinMauroy: So i understand what you propose.
So we'd have a `NodejsLogo` component that would change colour depending on the theme for the text and props  for the hexagon? Which means we'd have two other hex components, one green and one pride.

components/Common/NodejsLogo/index.tsx
```diff
@@ -0,0 +1,30 @@
...
+
+import NodejsDark from '@/components/Icons/Logos/NodejsDark';
+import NodejsDarkPride from '@/components/Icons/Logos/NodejsDarkPride';
+import NodejsLight from '@/components/Icons/Logos/NodejsLight';
+import NodejsLightPride from '@/components/Icons/Logos/NodejsLightPride';
+
+import style from './index.module.css';
+
+type NodejsLogoProps = {
+  isPrideEnabled?: boolean;
```
araujogui: Maybe something like
`variant: 'default' | 'pride'`
would fit better than booleans, considering that we may have more logos in the future.

araujogui (dismissed):

components/Common/NodejsLogo/index.tsx
ovflowd, line 1: What I'm trying to say is that you can merge all the Node.js logo variants into a single React Component, they are after all a React Component.

This (what you did below) is not what I was referring to, and does not reduce the duplication of code.

Try to reflect a bit on how to achieve what I'm saying. It's pretty simple actually ;)

AugustinMauroy: I thinks I go your idea just just want one `nodejsLogo` stroke of text will be apply with tailwind and two other component for Hexagon

bmuenzenmeyer:
I'm thinking we get this released sooner rather than later, even if we aren't in complete alignment on the approach.

bmuenzenmeyer (approved):
Content over purity of engineering

bmuenzenmeyer:
ugh - on mobile this all looked green - something's wrong post-merge - looking into it

AugustinMauroy:
IMG_0676
@bmuenzenmeyer ?

AugustinMauroy:
I thinks It’s small delay from Vercel deployment

bmuenzenmeyer:
@AugustinMauroy https://github.com/nodejs/nodejs.org/commits/main/

AugustinMauroy:
Oh my bad I thinks you talk about logo 😅

bmuenzenmeyer:
i'm glad its not outright broken, but apologies on any instability i rushed through - mobile is clearly not a good way to review CI status checks - thought i see here everything looks green.

i suggest we simply roll forward with a merge of a dependency update or something

PR #6739, opened 2024-05-26
Feat: Motion safe animations

canerakdas:
## Description

With this PR the animations we applied site-wide have been adjusted to be motion-safe for users sensitive to animations

- Additionally, the scroll behavior of articles has been changed to smooth
- The scroll behavior used for the body and articles has also been adjusted to be motion-safe

About reduced motion:

- [MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-reduced-motion)
- [Tailwind](https://tailwindcss.com/docs/hover-focus-and-other-states#prefers-reduced-motion)

## Validation

https://github.com/nodejs/nodejs.org/assets/14062599/f53c70a7-06de-40e5-b3d5-af74d468588f

### Check List

### DIFF
--- a/styles/base.css
+++ b/styles/base.css
@@ -1,6 +1,6 @@
 * {
   @apply subpixel-antialiased
-    transition-colors;
+    motion-safe:transition-colors;
 }
 
 html,
@@ -10,10 +10,10 @@ body {
 }
 
 body {
-  @apply scroll-smooth
-    bg-white
+  @apply bg-white
     font-open-sans
     text-neutral-950
+    motion-safe:scroll-smooth
     dark:bg-neutral-950
     dark:text-white;
 }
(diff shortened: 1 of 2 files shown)


styles/base.css
```diff
@@ -10,10 +10,10 @@ body {
 }
 
 body {
-  @apply scroll-smooth
-    bg-white
+  @apply bg-white
     font-open-sans
     text-neutral-950
+    motion-safe:scroll-smooth
```
AugustinMauroy, line 16: is it still need ? 'cause we make scrool smooth on layout.

canerakdas: I'm not sure if there is a place in the body where scroll behavior is used, it would be best to get other contributors comments and consider removing it 🤔 

Since I wasn't sure if it was used somewhere, I added motion-safe and left it as is 😄

bmuenzenmeyer (approved):
love this extra layer of ally work

PR #6750, opened 2024-05-30
Update tsx example

GeoffreyBooth:
## Description

This updates the `tsx` example to remove the mention of `--loader`, which is an [undocumented](https://nodejs.org/api/cli.html#--experimental-loadermodule) experimental flag that was created by accident and [will likely go away in the future, as the warning states](https://github.com/nodejs/node/issues/51196#issue-2045156881) when you use it. The stable, documented, recommended approach is to use `--import` instead. See https://github.com/nodejs/node/issues/51196#issuecomment-2089130867.

`ts-node` doesn’t (directly) support `--import`, so I’m leaving its example as just `npx ts-node`; there are [multiple](https://github.com/TypeStrong/ts-node/issues/2072) [open issues on their repo asking them to update](https://github.com/TypeStrong/ts-node/issues/1909#issuecomment-2089155346). I updated the `tsx` example to be more user-focused and to use `tsx` as the example of registering module hooks via `--import`.

## Validation

https://nodejs-org-git-fork-geoffreybooth-remove-loader-openjs.vercel.app/en/learn/getting-started/nodejs-with-typescript#running-typescript-code-with-tsx

image

## Related Issues

Resolves https://github.com/nodejs/node/issues/51196

### Check List

- [x] I’ve covered new added functionality with unit tests if necessary.

### DIFF
--- a/pages/en/learn/getting-started/nodejs-with-typescript.md
+++ b/pages/en/learn/getting-started/nodejs-with-typescript.md
@@ -148,20 +148,26 @@ Then you can run your TypeScript code like this:
 npx ts-node example.ts
 ```
 
-### Running TypeScript Code with a Node.js Loader
+### Running TypeScript Code with `tsx`
 
-Since Node.js v19.0.0, you can use a [custom loader](https://nodejs.org/api/cli.html#--importmodule). Download a loader such as `ts-node` or `tsx` or `nodejs-loaders`
+You can use [tsx](https://tsx.is/) to run TypeScript code directly in Node.js without the need to compile it first. But it's not typechecking your code. So we recommend to type check your code first with `tsc` and then run it with `tsx` before shipping it.
 
-First you need to install the loader:
+To use `tsx`, you need to install it first:
 
 ```bash
-npm i -D ts-node
+npm i -D tsx
 ```
 
 Then you can run your TypeScript code like this:
 
 ```bash
-node --loader=ts-node/esm example.ts
+npx tsx example.ts
+```
+
+If you want to use `tsx` via `node`, you can register `tsx` via `--import`:
+
+```bash
+node --import=tsx example.ts
 ```
 
 ## TypeScript in the Node.js world


pages/en/learn/getting-started/nodejs-with-typescript.md
```diff
@@ -148,20 +148,26 @@ Then you can run your TypeScript code like this:
 npx ts-node example.ts
 ```
 
-### Running TypeScript Code with a Node.js Loader
+### Running TypeScript Code with `tsx`
 
-Since Node.js v19.0.0, you can use a [custom loader](https://nodejs.org/api/cli.html#--importmodule). Download a loader such as `ts-node` or `tsx` or `nodejs-loaders`
```
AugustinMauroy, line 153: I'm not fan of removing "since of nodejs v19.0.0". 
Because this tells you which version the example is running on. But you could simply add the words “on the current LTS you can do ....”.

AugustinMauroy:
Another point I'm not a fan of is that only one loader (`tsx`) is featured, whereas before there were several that worked very well. I think `tsx` could be the example, but we need to come up with other alternatives.

GeoffreyBooth:
> Another point I’m not a fan of is that only one loader (`tsx`) is featured, whereas before there were several that worked very well. I think `tsx` could be the example, but we need to come up with other alternatives.

Previously it listed `ts-node`, `tsx` and “`nodejs-loaders`“. We still discuss the first two; the third didn’t get a description or link. Assuming it’s https://www.npmjs.com/package/nodejs-loaders, this is a package with 3 weekly downloads and it’s not specific to TypeScript.

Also this page isn’t about demonstrating how loaders work; the loaders docs do that. The purpose here is to explain to people how to use TypeScript.

pages/en/learn/getting-started/nodejs-with-typescript.md
GeoffreyBooth: Isn’t it assumed that the examples in the Learn pages are all for the current version of Node, or at least the current supported versions?

AugustinMauroy: Yes, it's true. Maybe we should mention it somewhere, because for the moment it's not written anywhere.

AugustinMauroy:
I agree with you that `nodejs-loaders` is not a very good example.
But it's also important to avoid the learn putting too much emphasis on a package, because we need to keep in mind that we have a certain status.

pages/en/learn/getting-started/nodejs-with-typescript.md
GeoffreyBooth: Sure, but I assume that should be a separate PR. Is there anything else about this or can I consider this resolved?

GeoffreyBooth:
> But it’s also important to avoid the learn putting too much emphasis on a package, because we need to keep in mind that we have a certain status.

I’m not sure what this means or what you’re asking me to change to resolve this note. The current page already emphasizes `ts-node` and `tsx` pretty strongly, and that remains in this PR. Maybe that should change, but the purpose of this PR is just to update the `tsx` example to remove the outdated reference to `--loader`. I agree that there could be other general improvements to this page, but perhaps those can come as follow-ups?

mikeesto (approved):
LGTM. I think it's important that we remove this reference to an undocumented & experimental flag that has already caused some confusion. We can discuss broader changes to the page elsewhere.

AugustinMauroy:
>  but perhaps those can come as follow-ups?

Yeah sure

AugustinMauroy (approved):
LGTM ! Thanks for updating learn section

PR #6814, opened 2024-06-04
CI: Orama cloud sync error handling/logs

canerakdas:
## Description

To check the developments in [this pull request](https://github.com/nodejs/nodejs.org/pull/6806) and ensure that it works successfully, error handling/logging has been added in the necessary areas to ensure that the build has the correct status in case of any issues during the build process

## Validation

We should see the `Orama Cloud sync completed successfully!` in the build logs

## Related Issues

Related to #6719

PR #6824, opened 2024-06-09
CI: Orama cloud sync based on environment

canerakdas:
## Description

In the previous PRs opened (#6814, #6806), we ensured that the Orama cloud sync script updates the preview environment on `push` and `pull_request_target` events.

With this PR, we aim to update the production indexes using this script on the `push` event instead of manually updating it and to update the preview environment on `pull_request_target` events.

Additionally, to prevent these scripts from running twice, we ensure they only run on `ubuntu-latest` runners. Since we are currently running it on `windows-latest` in the preview environment, the facets are currently not displayed correctly, likely due to differences in paths on Windows (You can check it in a sample preview build)

Before merging this development, the `ORAMA_PRODUCTION_INDEX_ID` and `ORAMA_PRODUCTION_SECRET_KEY` secrets need to be defined. I believe we should get support from @ovflowd or @bmuenzenmeyer for this 👀  I think we should first set these two values to point to the preview environment to ensure everything is working correctly, and then replace them with the production ones

Additionally, it would be helpful to ask the Orama team if indexing on every `pull_request_target` and `push` events would cause any issues on their side

## Related Issues

Related to #6719

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -125,9 +125,15 @@ jobs:
           NODE_OPTIONS: '--max_old_space_size=4096'
 
       - name: Sync Orama Cloud
+        # We only want to sync the Orama Cloud production indexes on `push` events.
+        # We also want to sync the Orama Cloud preview (deployment) indexes on `pull_request_target` events.
+        # We also want to ensure that the sync only happens on the `ubuntu-latest` runner to avoid duplicate syncs
+        # or Windows-based path issues.
         env:
-          ORAMA_INDEX_ID: ${{ secrets.ORAMA_INDEX_ID }}
-          ORAMA_SECRET_KEY: ${{ secrets.ORAMA_SECRET_KEY }}
-        if: github.ref == 'refs/heads/main'
+          ORAMA_INDEX_ID: ${{ github.event_name == 'push' && secrets.ORAMA_PRODUCTION_INDEX_ID || secrets.ORAMA_INDEX_ID }}
+          ORAMA_SECRET_KEY: ${{ github.event_name == 'push' && secrets.ORAMA_PRODUCTION_SECRET_KEY || secrets.ORAMA_SECRET_KEY }}
+        if: |
+          (matrix.os == 'ubuntu-latest') &&
+          ((github.event_name == 'push') || (github.event_name == 'pull_request_target'))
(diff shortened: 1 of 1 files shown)


ovflowd:
> Before merging this development, the ORAMA_PRODUCTION_INDEX_ID and ORAMA_PRODUCTION_SECRET_KEY secrets need to be defined. I believe we should get support from @ovflowd or @bmuenzenmeyer for this 👀 I think we should first set these two values to point to the preview environment to ensure everything is working correctly, and then replace them with the production ones

Secrets Added 🖖

ovflowd:
> Additionally, it would be helpful to ask the Orama team if indexing on every pull_request_target and push events would cause any issues on their side

cc @micheleriva

micheleriva:
@ovflowd that looks good to me!

bmuenzenmeyer (approved):

bmuenzenmeyer:
hmm - comparing experiences, the proposed search has odd headers

image

ovflowd:
Not sue why the discrepancies, this PR doesn't affect the website at all...

aymen94 (approved):

canerakdas:
> hmm - comparing experiences, the proposed search has odd headers

As far as I understand; the current script works for `windows-latest` and `ubuntu-latest`. However, since `windows-latest` runs later and the paths in the [file located at this path ](https://github.com/nodejs/nodejs.org/blob/main/scripts/orama-search/get-documents.mjs#L10) come as shown in the image below;

image

The [split in this line](https://github.com/nodejs/nodejs.org/blob/main/scripts/orama-search/get-documents.mjs#L64) does not work correctly. Therefore, the script that runs later overwrites the index created in `ubuntu-latest`, corrupting the existing indexes


Since we will [only run this script on `ubuntu-latest`](https://github.com/canerakdas/nodejs.org/blob/build/orama-environments/.github/workflows/build.yml#L136), I think there won't be such an issue after merging this PR. Since the current production secrets are replaced with the preview ones, it would be beneficial to run `pull_request_target` event in a different PR after this PR is merged and recheck the preview to ensure everything is working correctly 👀

bmuenzenmeyer:
Good analysis @canerakdas - I'm willing to get it a go today if we're all around to watch

PR #6828, opened 2024-06-10
Enhancement: New header and footer links design on hover and on active

mAmineChniti:
## Description
TLDR: The header now on hover and on active has a bottom border, the footer simply changes text color

## Validation

chrome_A34RBfK5qj
chrome_pzcti3Qmo9
chrome_XTf3UdGuCi
chrome_JK118W3ucO

### Check List

### DIFF
--- a/components/Containers/NavBar/NavItem/index.module.css
+++ b/components/Containers/NavBar/NavItem/index.module.css
@@ -27,9 +27,9 @@
 
     &:hover {
       @apply border-b-2
-        border-neutral-100
+        border-neutral-900
         dark:border-b-2
-        dark:border-neutral-900;
+        dark:border-neutral-100;
     }
 
     &.active {
--- a/components/Containers/NavBar/NavItem/index.module.css
+++ b/components/Containers/NavBar/NavItem/index.module.css
@@ -3,6 +3,7 @@
     items-center
     gap-2
     rounded
+    border-transparent
     px-3
     py-2;
 
@@ -38,7 +39,8 @@
         dark:border-green-700;
 
       .label {
-        @apply text-white;
+        @apply text-neutral-900
+          dark:text-white;
       }
 
       .icon {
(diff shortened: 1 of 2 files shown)


mAmineChniti:
@canerakdas @bmuenzenmeyer @ovflowd @mikeesto thoughts about this proposal?

bmuenzenmeyer:
I don't personally like it. I prefer the consistency of the other approach.
This one tries to do the right thing, I understand, but the underlines look odd to me with a rounded edge, and on the top right there is a non-link icon "button" to GH, that looks the same unfocused as the footer, but now behaves "like the buttons" next to it in the header but "like a link" in the footer (turning green using fill). GH and X both turn green, which is odd.

These are true statements to me:
1. I'm not a designer.
2. we shouldn't design by committee.
3. we had a designer design the original work
4. we are inconsistent in this approach

edit: Other opinions welcome.

mAmineChniti:
> the underlines look odd to me with a rounded edge, and on the top right there is a non-link icon "button" to GH, that looks the same unfocused as the footer, but now behaves "like the buttons" next to it in the header but "like a link" in the footer (turning green using fill). GH and X both turn green, which is odd.

- Fixed the rounded edges now its not rounded anymore (its just a simple straight underline now)
- Top right GH acts and feels like a link now
- Footer GH and X don't change color to green anymore

bmuenzenmeyer:
I'm 👎 on this
This causes layout shift on hover. There are ways to fix this, like using a transparent border, but I don't think we should pursue this.
https://github.com/nodejs/nodejs.org/pull/6775 is better

bmuenzenmeyer:
I'm closing this, in favor of https://github.com/nodejs/nodejs.org/pull/6775 which is merging now. If someone feels strongly about the UX - they can open an issue to discuss, but for now the navigation and footer are consistent and keyboard accessible

PR #6945, opened 2024-07-24
feat: create i18n package

araujogui:
## Description

Create i18n package

## Related Issues

Related to #5405

### Check List


packages/i18n/package.json
```diff
@@ -0,0 +1,6 @@
+{
+  "name": "@nodejs/i18n",
```
AugustinMauroy: I'm not a fan of this scope, we should have one specific to the website because this one is also used apr undici. The packages in `nodejs` are user packages, so it's not the best idea. But there's also `@node-core`, which doesn't fit either because it's intended for node core 😁. 
So for me it would be necessary to have a scope specific to the website project.

AugustinMauroy:
the config is in the `i18n` package but shouldn't it be dependent on the web app in question, i.e. there will be a config for nodejs.org but if there's another app it will have its own config. 

In this case, you could have a config for the app that just gives the language enabled or not.

packages/i18n/package.json
ovflowd: I think it should be "@node-core/website-i18n" and the current "@nodejs/website" on (apps/site) should be "@node-core/website"

ovflowd:
FYI, the i18n package is missing a build script

image

I'm not sure a build-step is needed, since right now the i18n package only has a bunch of JSONs; And I don't think the i18n package needs anything else 🤔 

cc @bmuenzenmeyer or @canerakdas any idea of what the i18n package should have else?

ovflowd:
> the config is in the `i18n` package but shouldn't it be dependent on the web app in question, i.e. there will be a config for nodejs.org but if there's another app it will have its own config.
> 
> In this case, you could have a config for the app that just gives the language enabled or not.

I think it is the other way around. site dependsOn i18n

AugustinMauroy:
> I think it is the other way around. site dependsOn i18n

you're not wrong. The question to make the right choice is: are the translation keys package/app dependent or shared, in simple terms do we have one JSON per language or one JSON per language per app/package. 
I think a JSON per language is more interesting. So yes, whether enabled or not should be in the I18n package.

apps/site/next.locales.mjs
```diff
@@ -1,6 +1,6 @@
 'use strict';
```
AugustinMauroy, line 1: IMO this file should be on package i18n. So it's can be share between multiple apps.

bmuenzenmeyer: agree

packages/i18n/package.json
bmuenzenmeyer: i am fine with that - one thing to keep in mind is that we cannot anticipate how people will use these packages once split out. `@node-core/website-i18n` could suddenly become very useful to someone downstream. That's not to say we should optimize for unknown, but I do like re-use.

PR #6986, opened 2024-08-13
NODE_ENV is not part of Node.js core and an antipattern

mcollina:
Setting NODE_ENV to anything but production is an antipattern should be avoided.

### Check List

### DIFF
--- a/apps/site/pages/en/learn/getting-started/nodejs-the-difference-between-development-and-production.md
+++ b/apps/site/pages/en/learn/getting-started/nodejs-the-difference-between-development-and-production.md
@@ -74,5 +74,5 @@ if (['production', 'staging'].includes(process.env.NODE_ENV)) {
 }
 ```
 
-While this might look harmless, it makes the production and staging environments different, thus making reliable testing impossible. For example a test and thus a functionality of your product could pass when `NODE_ENV` is set to development but fail when setting `NODE_ENV` to `production`.
+While this might look harmless, it makes the production and staging environments different, thus making reliable testing impossible. For example a test and thus a functionality of your product could pass when `NODE_ENV` is set to `development` but fail when setting `NODE_ENV` to `production`.
 Therefore, setting `NODE_ENV` to anything but `production` is considered an _antipattern_.
--- a/apps/site/pages/en/learn/getting-started/nodejs-the-difference-between-development-and-production.md
+++ b/apps/site/pages/en/learn/getting-started/nodejs-the-difference-between-development-and-production.md
@@ -1,15 +1,23 @@
 ---
 title: Node.js, the difference between development and production
 layout: learn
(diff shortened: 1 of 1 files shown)


apps/site/pages/en/learn/getting-started/nodejs-the-difference-between-development-and-production.md
```diff
@@ -50,14 +74,5 @@ if (['production', 'staging'].includes(process.env.NODE_ENV)) {
...
-```js
-if (process.env.NODE_ENV === 'development') {
-  app.use(express.errorHandler({ dumpExceptions: true, showStack: true }));
-}
-
-if (process.env.NODE_ENV === 'production') {
-  app.use(express.errorHandler());
-}
-```
+While this might look harmless, it makes the production and stanging environments different, making testing impossible.
```
Uzlopak:
```suggestion
While this might look harmless, it makes the production and staging environments different, thus making reliable testing impossible. For example a test and thus a functionality of your product could pass when `NODE_ENV` is set to development but fail when setting `NODE_ENV` to `production`.
```

AugustinMauroy:
@ovflowd Any hint for recurring Vercel failure

AugustinMauroy (commented):
Mostly LGTM, I think it should be mentioned that many tools use this environment variable. And which it can still use. For example webpack

jasnell (approved):

anonrig (approved):

apps/site/pages/en/learn/getting-started/nodejs-the-difference-between-development-and-production.md
```diff
@@ -50,14 +74,5 @@ if (['production', 'staging'].includes(process.env.NODE_ENV)) {
...
-```js
-if (process.env.NODE_ENV === 'development') {
-  app.use(express.errorHandler({ dumpExceptions: true, showStack: true }));
-}
-
-if (process.env.NODE_ENV === 'production') {
-  app.use(express.errorHandler());
-}
-```
+While this might look harmless, it makes the production and staging environments different, thus making reliable testing impossible. For example a test and thus a functionality of your product could pass when `NODE_ENV` is set to development but fail when setting `NODE_ENV` to `production`.
```
bmuenzenmeyer: suggest backticking the value `development` since `production` is likewise noted in the same sentence

```suggestion
While this might look harmless, it makes the production and staging environments different, thus making reliable testing impossible. For example a test and thus a functionality of your product could pass when `NODE_ENV` is set to `development` but fail when setting `NODE_ENV` to `production`.
```

bmuenzenmeyer (approved):
This is a really fantastic bit of clarification that Node.js should be using its platform to lead on - nice work!

PR #6994, opened 2024-08-20
fix: Improved understanding of setImmediate and setTimeout output

legendAhsan:
## Description

In the understanding-processnexttick.md file we have code example output for setImmediate and setTimeout. While that output holds true for ES modules case and have indeterministic behaviour in CommonJS case. I mentioned it to improve the understanding, incase a person tries to run the code and get confused.

### Check List

- [] I've covered new added functionality with unit tests if necessary. - Not required

### DIFF
--- a/apps/site/pages/en/learn/asynchronous-work/understanding-processnexttick.md
+++ b/apps/site/pages/en/learn/asynchronous-work/understanding-processnexttick.md
@@ -30,34 +30,34 @@ Use `nextTick()` when you want to make sure that in the next event loop iteratio
 console.log('Hello => number 1');
 
 setImmediate(() => {
-  console.log('Running before the timeout => number 3');
+  console.log('Running setImmediate in the check phase => number 3');
 });
 
 setTimeout(() => {
-  console.log('The timeout running last => number 4');
+  console.log('Running setTimeout in the timers phase => number 4');
 }, 0);
 
 process.nextTick(() => {
-  console.log('Running at next tick => number 2');
+  console.log('Running process.nextTick in the nextTick queue => number 2');
 });
 ```
 
 #### Example output:
 
-```bash
+```text
 Hello => number 1
-Running at next tick => number 2
-Running before the timeout => number 3
-The timeout running last => number 4
+Running process.nextTick in the nextTick queue => number 2
+Running setImmediate in the check phase => number 3
+Running setTimeout in the timers phase => number 4
 ```
 
 The output aforementioned holds true in ES Modules, e.g. mjs files, but keep in mind in CommonJS case, the output may be different:
 
-```bash
+```text
 Hello => number 1
-Running at next tick => number 2
-The timeout running last => number 4
-Running before the timeout => number 3
+Running process.nextTick in the nextTick queue => number 2
+Running setTimeout in the timers phase => number 4
+Running setImmediate in the check phase => number 3
 ```
 
-This is because the execution order of setImmediate and setTimeout is undeterministic in CommonJs.
+This is because the execution order of `setImmediate` and `setTimeout` is undeterministic in CommonJS.
--- a/apps/site/pages/en/learn/asynchronous-work/understanding-processnexttick.md
+++ b/apps/site/pages/en/learn/asynchronous-work/understanding-processnexttick.md
@@ -51,4 +51,13 @@ Running before the timeout => number 3
 The timeout running last => number 4
 ```
 
-The exact output may differ from run to run.
+The output aforementioned holds true in ES Modules, e.g. mjs files, but keep in mind in CommonJS case, the output may be different:
+
+```bash
+Hello => number 1
+Running at next tick => number 2
+The timeout running last => number 4
+Running before the timeout => number 3
+```
+
+This is because the execution order of setImmediate and setTimeout is undeterministic in CommonJs.


legendAhsan:
@mikeesto Can you look into this PR?

AugustinMauroy (approved):

apps/site/pages/en/learn/asynchronous-work/understanding-processnexttick.md
```diff
@@ -51,4 +51,13 @@ Running before the timeout => number 3
 The timeout running last => number 4
 ```
 
-The exact output may differ from run to run.
+The output aforementioned holds true in ES Modules, e.g. mjs files, but keep in mind in CommonJS case, the output may be different:
+
+```bash
```
aduh95: This is not valid bash syntax at all (would be worth fixing the above block as well)
````suggestion
```text
````

File: apps/site/pages/en/learn/asynchronous-work/understanding-processnexttick.md [hunk 2]
```diff
@@ -51,4 +51,13 @@ Running before the timeout => number 3
...
+Hello => number 1
+Running at next tick => number 2
+The timeout running last => number 4
+Running before the timeout => number 3
+```
+
+This is because the execution order of setImmediate and setTimeout is undeterministic in CommonJs.
```
aduh95: I don't think that's true, AFAIK CJS and ESM exposes the exact same `setImmediate` and `setTimeout` functions. In any case, here's a suggestion to fix the syntax.
```suggestion
This is because the execution order of `setImmediate` and `setTimeout` is undeterministic in CommonJS.
```

apps/site/pages/en/learn/asynchronous-work/understanding-processnexttick.md
AugustinMauroy: in addition the example output above can be updated

legendAhsan: @AugustinMauroy, Syntax revised.

apps/site/pages/en/learn/asynchronous-work/understanding-processnexttick.md
legendAhsan: Suggestion implemented. They may expose same function but it's deterministic in modules because of asynchronous nature.

RetrogradeDev (commented):
Looks good!

aymen94 (approved):

aduh95 (changes requested):
I tried to reproduce with the following script:

```js
import assert from 'node:assert'
import {spawn} from 'node:child_process'

let _1223344 = 0;
let _1224433 = 0;
let other = 0

for (let i = 0; i < 999; i++) {
  const cp = spawn(process.execPath, ['--input-type', process.argv[2], '-e', `console.log('Hello => number 1');

setImmediate(() => {
  console.log('Running before the timeout => number 3');
  console.log('Running setImmediate in the check phase => number 3');
});

setTimeout(() => {
  console.log('The timeout running last => number 4');
  console.log('Running setTimeout in the timers phase => number 4');
}, 0);

process.nextTick(() => {
  console.log('Running at next tick => number 2');
  console.log('Running process.nextTick in the nextTick queue => number 2');
});`]);
  cp.stdout.toArray().then(stdout => {
    stdout = Buffer.concat(stdout).toString();
    const numbers = stdout.trim().split('\n').map(line => Number(line.at(-1)));
    try {
      assert.deepStrictEqual(numbers, [1, 2, 2, 3, 3, 4, 4]);
      _1223344++;
      return;
    } catch {}
    try {
      assert.deepStrictEqual(numbers, [1, 2, 2, 4, 4, 3, 3]);
      _1224433++;
      return;
    } catch {}
    console.log({ stdout, numbers });
    other++;
  });
}

process.on('exit', () => {
  console.log({ _1223344, _1224433, other });
});
```

Here are the results:

```console
$ node repro.mjs commonjs
{ _1223344: 472, _1224433: 527, other: 0 }
$ node repro.mjs module  
{ _1223344: 311, _1224433: 688, other: 0 }
```

As you can see, the proportions are different in ESM vs CJS, but it seems very wrong to state that ESM output is guarenteed.

legendAhsan:
@aduh95 

```js
import assert from 'node:assert'

let order1223344 = 0;
let order1224433 = 0;
let other = 0
const totalRuns = 999

function runTest() {
  return new Promise((resolve) => {
    const output = [];
    
    console.log('Hello => number 1');
    output.push(1);
    
    setImmediate(() => {
      console.log('Running before the timeout => number 3');
      console.log('Running setImmediate in the check phase => number 3');
      output.push(3);
      output.push(3);
      if (output.length === 7) resolve(output);
    });
    
    setTimeout(() => {
      console.log('The timeout running last => number 4');
      console.log('Running setTimeout in the timers phase => number 4');
      output.push(4);
      output.push(4);
      if (output.length === 7) resolve(output);
    }, 1);
    
    process.nextTick(() => {
      console.log('Running at next tick => number 2');
      console.log('Running process.nextTick in the nextTick queue => number 2');
      output.push(2);
      output.push(2);
    });
  });
}

for (let i = 0; i < totalRuns; i++) {
  runTest().then(numbers => {
    try {
      assert.deepStrictEqual(numbers, [1, 2, 2, 3, 3, 4, 4]);
      order1223344++;
    } catch {
      try {
        assert.deepStrictEqual(numbers, [1, 2, 2, 4, 4, 3, 3]);
        order1224433++;
      } catch {
        console.log({ numbers });
        other++;
      }
    }
    
    if (i === totalRuns -1) {
      console.log({ order1223344, order1224433, other });
    }
  });
}
```

I feel that the above output you got is because Spawning childs may run like a CJS script while in module it is asynchronous.
Put the above code in module file like common.mjs and you will see that it will only increment 1223344 variable.

aduh95:
> I feel that the above output you got is because Spawning childs may run like a CJS script while in module it is asynchronous.
> Put the above code in module file like common.mjs and you will see that it will only increment 1223344 variable.

It seems to me it has nothing to do with ESM or CJS, rather how busy is the event loop. Anyway, it's good to know that sometimes it's deterministic I guess, but as I've shown in https://github.com/nodejs/nodejs.org/pull/6994#pullrequestreview-2335997716, it's not always the case.

bmuenzenmeyer:
Core Collaborators are ultimately responsible for approving content, not the website team, so I'd like @aduh95 to provide the final say on this.

aduh95:
As I said earlier, it would be wrong to use "is the source parsed as ESM/CJS" as a confounding variable for how those functions would behave. The exact same functions are exposed, therefore they behave exactly the same, and it would be wrong to document otherwise, and probably confusing for readers.

PR #7001, opened 2024-08-23
Fix: nodejs.org/learn leads to 404

### LINKED ISSUE #7000: nodejs.org/learn leads to 404
### URL:

https://nodejs.org/learn

### Browser Name:

Brave

### Browser Version:

Version 1.68.141 (arm64)

### Operating System:

macOS 14.6

Go to https://nodejs.org/learn
Observe 404

In addition, clicking the "learn" TL nav link does nothing. One must manually update the address to escape.

https://nodejs.org/learn should redirect to the "root" of learn.

Nirmalkumar6112:
## Description
Added the redirect path to the root of the Learn nav link

## Related Issues
Fixes #7000 nodejs.org/learn leads to 404

### Check List

### DIFF
--- a/apps/site/redirects.json
+++ b/apps/site/redirects.json
@@ -265,6 +265,10 @@
       "destination": "/:locale/about/get-involved/:path*"
     },
     {
+      "source": "/learn",
+      "destination": "/en/learn"
+    },
+    {
       "source": "/:locale/learn",
       "destination": "/:locale/learn/getting-started/introduction-to-nodejs"
     },


mAmineChniti:
@Nirmalkumar6112 maybe you should do `/:locale/learn` instead of `/en/learn` that way people get redirected to their correct locale pages and not the english one by default

bmuenzenmeyer:
> maybe you should do `/:locale/learn` instead of `/en/learn`

I looked at this, but it generates an error: `"destination" has segments not in "source"`
That seems straightforward enough, though googling for it suggests otherwise and our current code follows the as-is pattern. 

for reference, we have

```json
    {
      "source": "/guides",
      "destination": "/en/learn"
    },
    {
      "source": "/about",
      "destination": "/en/about"
    },
    {
      "source": "/community",
      "destination": "/en/get-involved"
    },
```
elsewhere in the [same file](https://github.com/nodejs/nodejs.org/blob/main/apps/site/redirects.json#L87-L98). I'd suggest we incrementally improve the experience via this PR and then pursue future action another time - this is already 2 weeks old

bmuenzenmeyer:
also - `learn` content is not internationalized anyway, per our decisions

bmuenzenmeyer (approved):

PR #7009, opened 2024-08-30
Update the-v8-javascript-engine.md

TheCaptain1810:
Opening links in the same tab is so annoying, hence tried to improve the markdown using {:target="_blank"} property.

## Description
I just made some small changes in the markdown by adding {:target="_blank"} property so the links will open in the new tab. Also, I suggest doing this in all other markdowns.


AugustinMauroy (commented):
-1 I'm not fan of this behaviour in content page

bmuenzenmeyer:
this doesn't render in GitHub or our current markdown/mdx tech

image

thanks for the thought - but I am going to close this for now

TheCaptain1810:
Alright 👍🏻
Pranav Pise

PR #7039, opened 2024-09-13
Fix dropdown cutoff issue for long text in mobile view (#7035)

### LINKED ISSUE #7035: Cut off dropdown for long form text
### URL:

### Browser Name:

brave

### Browser Version:

1.65.123

### Operating System:

Windows 11

in mobile view, the dropdown of long text is getting cutoff.

possible solution:
1. giving a scroll-y
2. expanding the width to full for the downdown.

image

aqibzaib:
## Description

**Fixes #7035**

This pull request resolves the issue where the dropdown menu was being cut off when displaying long text on mobile devices. The changes implemented include:

- Expanding the dropdown width to utilize the full available space on mobile devices.
- **Updated:** Removing horizontal scrolling and allowing the text to wrap to a new line when it exceeds the width of the dropdown.

### Before and After Videos

- **Before Fix**: [View Video](https://www.loom.com/share/5abae3c00c264d2cb13f2cfaec75b46a?sid=c38b5647-8a67-4639-8781-fc6011648022)
- **After Fix**: https://www.imghippo.com/i/s1xcH1726232116.png

### Validation

To verify that the issue has been resolved:

1. ✔️ Ensure that the dropdown now supports text wrapping for long text and does not get cut off.
2. ✔️ Confirm that the dropdown width is expanded appropriately on mobile devices.
3. ✔️ Review the provided before and after videos to validate the visual improvements.

### Checklist

- ✔️ I have read the [Contributing Guidelines](https://github.com/nodejs/nodejs.org/blob/main/CONTRIBUTING.md) and ensured that my commit messages follow the guidelines.
- ✔️ I have run `npm run format` to ensure the code follows the style guide.
- ✔️ I have run `npm run test` to confirm that all tests pass.
- ✔️ I have run `npx turbo build` to verify that the website builds without errors.


AugustinMauroy:
not fan of the new behavior. I thinks making a breakline when text is too long should be better than scrool

aqibzaib:
> Contributor

Here's a response you can use:

---

Thanks for your valuable feedback! I'll make the select options wrap to improve the user experience.

aqibzaib:
> not fan of the new behavior. I thinks making a breakline when text is too long should be better than scrool
Thanks for your valuable feedback! I'll make the select options wrap to improve the user experience.

aqibzaib:
> not fan of the new behavior. I thinks making a breakline when text is too long should be better than scrool
Thanks for your valuable feedback! 😊 I've updated the select options to wrap, so no more endless scrolling adventures. Easy reading from now on!

image

TenzDelek:
Why is the Vercel deployment failing (server error)on other page except homeimage

ovflowd:
> Why is the Vercel deployment failing (server error)on other page except homeimage

That is incompatible because you made changes on a server-side component by adding client-side specific APIs. (useRef, useEffect)

I'd suggest on finding alternatives without using client-side-specific APIs.

apps/site/components/Common/Select/index.tsx
```diff
@@ -66,13 +68,32 @@ const Select: FC<SelectProps> = ({
     return mappedValues;
   }, [values]);
 
+  useEffect(() => {
```
ovflowd, line 71: This sounds like a very inefficient way of solving an issue that could be solved only with CSS.

TenzDelek:
Haha , I thought it was some error from my device network 
@aqibzaib you can make changes as per suggested by @ovflowd . 
I suggest trying with ellipse property tailwind

apps/site/components/Common/Select/index.module.css
```diff
@@ -133,6 +133,10 @@
       dark:data-[disabled]:text-neutral-700;
   }
 
+  .textWrap {
+    text-wrap: wrap !important ;
+  }
```
canerakdas, line 138: Thank you for your contribution 🙇  

As Claudio stated, I also think we should solve these problems in CSS

We use Tailwind classes to style CSS files;
https://github.com/nodejs/nodejs.org/blob/main/COLLABORATOR_GUIDE.md#example-of-a-css-module

In addition, we are careful not to define `important` properties and classes unless necessary.

You can make an example arrangement like this;

```css
.inline {
  ... some code

  .text span > span {
    @apply max-w-fit
      text-wrap;
  }
}
```

mikeesto:
Thanks for your efforts here. I'm going to close this in favour of https://github.com/nodejs/nodejs.org/pull/7046 which has now landed.

PR #6492, opened 2024-03-20
feat: turtle surfin'

darcyclarke:
## Description

- adds the turtle rocket mascot to the homepage
- makes them surf

Fixes: #6480

https://github.com/nodejs/nodejs.org/assets/459713/6b810071-751c-425d-b55b-3818bbf055d3

### Check List

- [x] I have read the [Contributing Guidelines](https://github.com/nodejs/nodejs.org/blob/main/CONTRIBUTING.md) and made commit messages that follow the guideline.
- [x] I have run `npx turbo format` to ensure the code follows the style guide.
- [x] I have run `npx turbo test` to check if all tests are passing.
- [x] I have run `npx turbo build` to check if the website builds without errors.
- [x] I've covered new added functionality with unit tests if necessary.

PR #7056, opened 2024-09-23
6492 surfin 404

bmuenzenmeyer:
## Description

I spent the weekend chipping away at @darcyclarke's rocket turtle animation. This will land the longest open PR we have, and it's an important bit of style 😎 

- on the 404 page, as https://github.com/nodejs/nodejs.org/pull/6481 established
- converted to Tailwind
- added support for `prefers-reduced-motion`, per the [TW docs](https://tailwindcss.com/docs/animation#prefers-reduced-motion)

## Validation

Go to a non-existent page locally or on the preview.
Tested with 
Chrome, Firefox, Edge

| Existing | Chrome | Firefox | Edge | 
| - | - | - | - |
| image| image | image |image |

Mobile

| Existing | New |
| - | - |
|  |  |

Turn on `prefers reduced motion`, in your [operating system of choice.](https://stackoverflow.com/a/59709067)

After a bit, see the up and down animation cease to loop.

## Related Issues

closes #6492 
relates to #6480

### Check List


apps/site/app/[locale]/not-found.tsx
```diff
@@ -19,13 +19,15 @@ const NotFoundPage: FC = () => {
...
         404
         <h1 className="special -mt-4">{t('layouts.error.notFound.title')}</h1>
-        <div className="my-4 flex items-center justify-center">
-          <Image
-            src="/static/images/node-mascot.svg"
-            alt="The Node.js mascot"
-            height={114.69}
-            width={100}
-          />
+        <div className="my-4 flex h-[300] items-center justify-center">
```
Comment (bmuenzenmeyer) [inline apps/site/app/[locale]/not-found.tsx]: I'm not in love with this extra `h-300` class here, but it helped simplify the css greatly. There is no more translate x,y logic. 

One improvement would be to occupy less space on smaller viewports. I tried `sm:h-[150]` and it wouldn't work. This would improve the too much white space on mobile as observed via https://github.com/nodejs/nodejs.org/pull/7056#issuecomment-2368030378

apps/site/app/[locale]/not-found.tsx
```diff
@@ -19,13 +19,15 @@ const NotFoundPage: FC = () => {
...
+          <div className="turtle motion-safe:animate-surf motion-reduce:animate-none">
+            <Image
+              src="/static/images/node-mascot.svg"
+              alt="The Node.js mascot"
+              height={114.69}
+              width={100}
```
Comment (bmuenzenmeyer) [inline apps/site/app/[locale]/not-found.tsx:30]: Unsure how important these are in the newer approach, I suspect they could help with CLS or optimization, but the css changes the size anyway. 

The ported css is not mobile-first, that would be a potentially useful refactor.

apps/site/styles/effects.css
```diff
@@ -10,3 +10,47 @@ h1.special {
...
+    -z-10
+    block
+    h-[150]
+    w-[150]
+    rotate-[-90deg]
+    select-none
+    bg-[url('/static/images/smoke.gif')]
+    opacity-[0.15];
+  /* stylelint-disable-next-line declaration-empty-line-before */
+  content: '';
```
bmuenzenmeyer: Perhaps `content-['']` would work here. I thought I read that it wasn't supported.

apps/site/tailwind.config.ts
```diff
@@ -139,6 +139,18 @@ export default {
       spacing: { '4.5': '1.125rem', '18': '4.5rem' },
       aria: { current: 'current="page"' },
       maxWidth: { '8xl': '95rem' },
+      animation: {
```
bmuenzenmeyer, line 142: this needs to be defined in the config if you want to use it in the TW markup

apps/site/app/[locale]/not-found.tsx
```diff
@@ -19,13 +19,15 @@ const NotFoundPage: FC = () => {
...
         <h1 className="special -mt-4">{t('layouts.error.notFound.title')}</h1>
-        <div className="my-4 flex items-center justify-center">
-          <Image
-            src="/static/images/node-mascot.svg"
-            alt="The Node.js mascot"
-            height={114.69}
-            width={100}
-          />
+        <div className="my-4 flex h-[300] items-center justify-center">
+          <div className="turtle motion-safe:animate-surf motion-reduce:animate-none">
```
Comment (bmuenzenmeyer) [inline apps/site/app/[locale]/not-found.tsx:25]: this was the only way I could get the accessibility logic to work. Within CSS alone, it was not working for me.

bmuenzenmeyer (commented):
self-review, open to suggestions, especially on the tailwind conversion

bmuenzenmeyer:
demoting this down to a draft - the preview at https://nodejs-org-git-6492-surfin-404-openjs.vercel.app/en/does-not-exist is not rendering as I had it locally at all - that's frustrating.

image

ovflowd:
On mobile text is not centered and the image overlaps with the bottom text.

bmuenzenmeyer:
> On mobile text is not centered and the image overlaps with the bottom text.

yeah, I was getting there. the Chrome Devtool responsiveness testing is very rudimentary.

bmuenzenmeyer:
Samsung Galaxy S20+

bmuenzenmeyer:
> On mobile text is not centered and the image overlaps with the bottom text.

@ovflowd can you look again, this looks correct to me now on my phone. If you are concerned about the text, that is unrelated to the PR - this only alters the image

ovflowd:
Page could not be found still not centered for me 🙈

bmuenzenmeyer:
> Page could not be found still not centered for me 🙈

it never was, from what i can tell of current production. we can center it if you like.

apps/site/styles/effects.css
```diff
@@ -10,3 +10,47 @@ h1.special {
...
+  @apply animate-surf
+    absolute
+    z-10
+    translate-x-0
+    translate-y-0;
+}
+
+.turtle img {
+  @apply h-auto
+    w-[200px];
```
canerakdas: Instead of adding media query we can use it directly like this;
```suggestion
    w-24
    md:w-48;
```

Also, instead of using px, I tried to round to the nearest values ​​in the Tailwind's [spacings](https://github.com/tailwindlabs/tailwindcss/blob/main/stubs/config.full.js#L841). We can leave it as px if we don't want it to be scaled

apps/site/styles/effects.css
```diff
@@ -10,3 +10,47 @@ h1.special {
...
+    left-[-50%]
+    top-[50%]
+    -z-10
+    block
+    h-[150px]
+    w-[150px]
+    rotate-[-90deg]
+    select-none
+    bg-[url('/static/images/smoke.gif')]
+    opacity-[0.15];
```
canerakdas: same as above, more tailwindy way 😄 

```suggestion
    -left-full
    top-[20%]
    -z-10
    block
    h-36
    w-36
    -rotate-90
    select-none
    bg-[url('/static/images/smoke.gif')]
    opacity-[0.15]
    md:-left-1/2
    md:top-1/2;
```

apps/site/styles/effects.css
```diff
@@ -10,3 +10,47 @@ h1.special {
...
+
+  .turtle img {
+    @apply w-[100px];
+  }
+
+  .turtle::after {
+    @apply left-[-100%]
+      top-[20%];
+  }
+}
```
canerakdas: Since Tailwind responsive styling [utilities](https://tailwindcss.com/docs/responsive-design) implemented above, media query are no longer needed

```suggestion
```

canerakdas (commented):
Looks really cool 😍 I have added a few minor comments about the Tailwind conversion

apps/site/styles/effects.css
canerakdas: Since we are using it in `::after`, I think it should work, but if we are sure that it won't work somehow, it might be better to use it like Tailwind classes like this;


```css
  .foo::after {
    @apply [content:'']
      some
      tailwind
      classes;
  }
```

darcyclarke:
@bmuenzenmeyer thanks for helping get this over the finish line <3

ovflowd:
> > Page could not be found still not centered for me 🙈
> 
> it never was, from what i can tell of current production. we can center it if you like.

Let's center it :D

AugustinMauroy (approved):
LGTM! Cool new feature

canerakdas (approved):
LGTM!

PR #7058, opened 2024-09-24
Feat(banner) hide

### LINKED ISSUE #6292: Feat(banner): take place on screen, should we add the possibility of hide

AugustinMauroy:
## Description

A little while ago we discussed hiding the banner, and since then there have been no major objections, so I've opened this pr to implement it.

## Related Issues

close #6292

### Check List


aymen94 (approved):

apps/site/components/Common/Banner/index.tsx
```diff
@@ -8,17 +10,35 @@ import styles from './index.module.css';
...
+
+  return (
+    <div className={classNames(styles.banner, styles[type])}>
+      <span className={styles.content}>
+        {link ? <Link href={link}>{children}</Link> : children}
+        {link && <ArrowUpRightIcon />}
+      </span>
+      {onHiding && (
+        <button
+          arria-label={t('hide')}
```
bmuenzenmeyer:
```suggestion
          aria-label={t('hide')}
```

apps/site/components/Common/Banner/index.tsx
```diff
@@ -8,17 +10,35 @@ import styles from './index.module.css';
 type BannerProps = {
   link?: string;
   type?: 'default' | 'warning' | 'error';
+  onHiding?: () => void;
 };
 
 const Banner: FC<PropsWithChildren<BannerProps>> = ({
   type = 'default',
   link,
+  onHiding,
```
bmuenzenmeyer, line 19: I suggest a different name here. `onHiding` to me implies it's a handler, when it's really a conditional state or mode that the banner can opt-into. 

`isHideable` would not the boolean nature, and nod to the story name `Hideable`

packages/i18n/locales/en.json
```diff
@@ -159,6 +159,9 @@
       "previous": "Previous"
     },
     "common": {
+      "banner": {
+        "hide": "Hide this banner"
```
bmuenzenmeyer: This feels more appropriate to me. Dismiss is used by [bootstrap](https://getbootstrap.com/docs/5.3/components/alerts/) and [MUI](https://mui.com/material-ui/react-snackbar/)

```suggestion
        "hide": "Dismiss this banner"
```

apps/site/components/withBanner.tsx
```diff
@@ -1,21 +1,53 @@
...
+import { twoDateToUIID } from '@/util/stringUtils';
+
+type BannerState = {
+  uuid: string;
+  hideBanner: boolean;
+};
 
 const WithBanner: FC<{ section: string }> = ({ section }) => {
   const banner = siteConfig.websiteBanners[section];
+  const UUID = twoDateToUIID(banner.startDate, banner.endDate);
```
bmuenzenmeyer, line 19: nit: at first glance, I thought we were using the uuid package here due to the name. Not sure if it needs to change, but calling attention to the momentary confusion. I think the way you did this was clever overall.

apps/site/util/stringUtils.ts
```diff
@@ -28,3 +28,10 @@ export const dashToCamelCase = (str: string) =>
     // remove leftover - which don't match the above regex. Like 'es-2015'
     .replace(/-/g, '')
     .replace(/^[A-Z]/, chr => chr.toLowerCase());
+
+export const twoDateToUIID = (date1: string, date2: string): string => {
+  const date1String = date1.split('T')[0];
+  const date2String = date2.split('T')[0];
```
bmuenzenmeyer, line 34: nothing seems to enforce this as a a proper date string over within `site.json`, which could lead to bugs or errors here splitting on `T`

image

bmuenzenmeyer (changes requested):
functionally this works on the preview pretty simply
found one typo
have one small concern about relying on good JSON data in site.json

apps/site/components/Common/Banner/index.tsx
AugustinMauroy: gotcha

apps/site/util/stringUtils.ts
AugustinMauroy: I'll have to reread the split docs because split automatically gives you a table that can be unique.

apps/site/components/withBanner.tsx
AugustinMauroy: We can call it `generatekeyForBanner`. But UIID mean Universally unique identifier.

apps/site/util/stringUtils.ts
AugustinMauroy: But we can also use `new Date` and throw error if not right date

apps/site/hooks/react-client/useLocalStorage.ts
```diff
@@ -0,0 +1,22 @@
+'use client';
+import { useEffect, useState } from 'react';
+
+const useLocalStorage = <T>(key: string, initialValue?: T) => {
```
canerakdas, line 4: We can only access Local storage from the client components, which causes a layout shift;
image
image

AFAIK we can access [cookies](https://nextjs.org/docs/app/api-reference/functions/cookies) from the server components, which might be a better approach for this setup

Also, if there's someone we can consult before storing user preferences in local storage or cookies without their consent, it would be better to ask. To be honest, I don't know anything about this topic 🙇

AugustinMauroy: +1 for using cookies.
But as far as I know, the use of local storage is not subject to regulation, but using cookies is. 
NOTE: next-intl use cookie to store the local

canerakdas: Got it, then let's continue with local storage. Once we are sure, we can consider using cookies 👍

ovflowd:
I'm not sure about this PR. Sure, maybe being able to close the banner is nice... But how many people actually want this / or are frustrated with the banner?

Becazse as a technical change, I'm not a fan of using local storage just for this, and there are many other little technical details that are ehh, also not OK.

apps/site/components/withBanner.tsx
ovflowd, line 19: This feels so weird.

ovflowd (changes requested):
I'm not sure about this PR. Sure, maybe being able to close the banner is nice... But how many people actually want this / or are frustrated with the banner?

Becazse as a technical change, I'm not a fan of using local storage just for this, and there are many other little technical details that are ehh, also not OK, not to mention the whole hacky and weird UUID logic.

I just overall don't feel confident with these changes and how much garbage we are adding just to have a single X button. The benefits are just not there.

AugustinMauroy:
> also not OK, not to mention the whole hacky and weird UUID logic.

We can base the logic on dates and not a UIID. 

> I just overall don't feel confident with these changes and how much garbage we are adding just to have a single X button. The benefits are just not there.

I understand your point of view. I sent a message on slack to get a sample of opinions. And based on that: 
- Either we abandon the changes.
- Or we find a clean solution to implement it.

Note that Caner  had suggested using cookies, which has two advantages:
- the function for reading them is provided by next
- there's no more layout shift
But we don't know what this means from a legal point of view
ref: https://github.com/nodejs/nodejs.org/pull/7058#discussion_r1777421033

ovflowd:
> > also not OK, not to mention the whole hacky and weird UUID logic.
> 
> 
> 
> We can base the logic on dates and not a UIID. 
> 
> 
> 
> > I just overall don't feel confident with these changes and how much garbage we are adding just to have a single X button. The benefits are just not there.
> 
> 
> 
> I understand your point of view. I sent a message on slack to get a sample of opinions. And based on that: 
> 
> - Either we abandon the changes.
> 
> - Or we find a clean solution to implement it.
> 
> 
> 
> Note that Caner  had suggested using cookies, which has two advantages:
> 
> - the function for reading them is provided by next
> 
> - there's no more layout shift
> 
> But we don't know what this means from a legal point of view
> 
> ref: https://github.com/nodejs/nodejs.org/pull/7058#discussion_r1777421033

Im fine with the feature -- Im just pretty much saying your code is not good, and I won't allow it to get merged on the state it is, sorry.

I might have more bandwidth in the future; But still I'm not sure about how much value we get from hiding banners. And this is such a small change that genuinely speaking bothering core **collaborators to interfere here is overkill and unnecessary.**

Some banners should just stay there, and I don't think you should make an UUID based on the dates. You could simply create a hash based on the banner entry itself (the json blob)

I also don't have the time at the moment to propose code changes or to coach you here. For me it is more valuable for Caner to take over your PR if we decide to move forward with this. Which I doubt we will.

Sorry, Augustin 😅

ovflowd:
I forgot to comment regarding:

> But we don't know what this means from a legal point of view

Just a FYI although it is a cookie, we could consider these to be functional 1st party cookies as they store user preferences.

Nothing is really done with them, I doubt on a GDPR standpoint anything needs to be done. Just so you know, the "legal concerns" you have are not limited to cookies, they also apply to localStorage.

bmuenzenmeyer:
are we abandoning this for now then?

AugustinMauroy:
> are we abandoning this for now then?

I wanted to see how nextra and docausorus work. But I haven't taken the time for that yet.

ovflowd:
I'm closing this for now, we can always circle back.

PR #7086, opened 2024-10-03
Esp participation

joeeames:
## Description

This PR adds links and adjustments for participation in the OpenJS Ecosystem Sustainability Program (ESP). This includes a new top nav link, a warning on the download page when downloading EOL versions, and information on the "previous releases" page including a link to HeroDevs. 

Once merged, the graphic on the Previous Releases page will be updated, and the core node.js README on the github repo will be updated to summarize some of this information. This will be done with separate PR's.

## Validation

1. New top nav pointing to the Previous Releases page 
2. a warning box on the download page when selecting a version before 18
3. new text on the Previous Releases page

### Check List

- [ X] I have read the [Contributing Guidelines](https://github.com/nodejs/nodejs.org/blob/main/CONTRIBUTING.md) and made commit messages that follow the guideline.
- [ X] I have run `npm run format` to ensure the code follows the style guide.
- [ X] I have run `npm run test` to check if all tests are passing.
- [ X] I have run `npx turbo build` to check if the website builds without errors.
- [ X] I've covered new added functionality with unit tests if necessary.


apps/site/components/Downloads/Release/ReleaseCodeBox.tsx
```diff
@@ -28,14 +29,32 @@ const ReleaseCodeBox: FC = () => {
...
+    };
+  };
+
   const codeLanguage = os === 'WIN' ? 'PowerShell' : 'Bash';
 
   return (
     <div className="mb-2 mt-6 flex flex-col gap-2">
+      {parseVersion().major < 18 && (
+        <div
+          className="border-danger-400 bg-danger-100 relative rounded border px-4 py-3 text-red-700"
```
mikeesto: Missing styling here for dark mode - it's currently unreadable

apps/site/navigation.json
```diff
@@ -20,6 +20,10 @@
       "link": "https://nodejs.org/docs/latest/api/",
       "label": "components.containers.navBar.links.docs"
     },
+    "support": {
+      "link": "https://nodejs.org/en/about/previous-releases",
```
mikeesto: This needs to be a relative link. It's also unclear to me why we need to add previous releases to the top nav bar?

packages/i18n/locales/fr.json
mikeesto, line 1: We use CrowdIn for translations, the translations in this file & others can be removed

mikeesto (changes requested):
It's unclear to me how this fits in with https://github.com/nodejs/nodejs.org/issues/7037 ? If the underlying motivation is to market support for EOL versions through HeroDevs, that can be a short spiel on the Node.js Releases page. The changes to the Download page seem unproductive given it's about to be redesigned

apps/site/components/Downloads/Release/ReleaseCodeBox.tsx
```diff
@@ -28,14 +29,32 @@ const ReleaseCodeBox: FC = () => {
...
+      major: parseInt(versionParts[0]),
+      minor: parseInt(versionParts[1]),
+      patch: parseInt(versionParts[2]),
+    };
+  };
+
   const codeLanguage = os === 'WIN' ? 'PowerShell' : 'Bash';
 
   return (
     <div className="mb-2 mt-6 flex flex-col gap-2">
```
AugustinMauroy, line 36: Maybe we should put these styles in a css module to make them cleaner?

apps/site/components/Downloads/Release/ReleaseCodeBox.tsx
```diff
@@ -28,14 +29,32 @@ const ReleaseCodeBox: FC = () => {
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [release.versionWithPrefix, os, platform]);
 
+  const parseVersion = () => {
```
AugustinMauroy: on project we use `semver` that export a function call `semverParse`

apps/site/components/Downloads/Release/ReleaseCodeBox.tsx
```diff
@@ -28,14 +29,32 @@ const ReleaseCodeBox: FC = () => {
...
 
   return (
     <div className="mb-2 mt-6 flex flex-col gap-2">
+      {parseVersion().major < 18 && (
+        <div
+          className="border-danger-400 bg-danger-100 relative rounded border px-4 py-3 text-red-700"
+          role="alert"
+        >
+          {t('layouts.download.codeBox.unsupportedVersionWarning')}&nbsp;
+          <a href={`/${locale}/about/previous-releases/`}>Previous Releases</a>
```
AugustinMauroy: You use `import Link from '@/components/Link';` to automatically use right local (so no need to use the `useLocal` hook)

AugustinMauroy (commented):
+1 for this mention; +1 to display it as "alert box"

ovflowd:
> [ X] I have read the [Contributing Guidelines](https://github.com/nodejs/nodejs.org/blob/main/CONTRIBUTING.md?rgh-link-date=2024-10-03T22%3A51%3A24Z) and made commit messages that follow the guideline.

Have you actually read the guidelines? 🤔 -- because some of the changes you've done here, such as translations are mentioned there 🤔

ovflowd (commented):
Overall I'm fine with the change, although not ideal, but you definitely need to use the existing components from our design system please.

AugustinMauroy:
>  but you definitely need to use the existing components from our design system please.

I had the same feeling. But we don't have any red flash components at the moment, with the exception of `Banner`. If not there's `Stability` but that's an API component that hasn't been implemented yet.

ovflowd:
> > but you definitely need to use the existing components from our design system please.
> 
> I had the same feeling. But we don't have any red flash components at the moment, with the exception of `Banner`. If not there's `Stability` but that's an API component that hasn't been implemented yet.

Banner component is fine.

apps/site/navigation.json
joeeames: thanks for calling this out. I did want to use a relative link, but must have just looked at the closest neighbor. will address.

As to having the link in the top nav bar, this is part of the requirements for participation in the ESP program. The page showing the support and EOL policy for the project needs to be easy to find for full participation, generally meaning part of the main navigation.

joeeames:
thanks for all the review. I apologize for what I missed. Although I did read the contributing guidelines @ovflowd I did misunderstand the translation piece. I'll get changes made here and resubmitted.

apps/site/components/Downloads/Release/ReleaseCodeBox.tsx
joeeames: @AugustinMauroy Augustin, not sure what is meant by "alert box" can you clarify?

Also, not sure if changing to a CSS module is required or not. I'm not familiar with what that would entail.

AugustinMauroy: my question about css module is more for the website team. And alert box mean red box. But you should use Banner component like Claudio said in this [comment](https://github.com/nodejs/nodejs.org/pull/7086#issuecomment-2395559841)

AugustinMauroy:
I'm -1 to have a new entry on nav bar because it's not 100% usefull. It also introduces a strange effect on the navbar because it hasn't been designed to support two entries that have the same ‘section’.

Capture d’écran 2024-10-15 à 09 26 12

apps/site/components/Downloads/Release/ReleaseCodeBox.tsx
```diff
@@ -29,13 +32,17 @@ const ReleaseCodeBox: FC = () => {
   }, [release.versionWithPrefix, os, platform]);
 
   const codeLanguage = os === 'WIN' ? 'PowerShell' : 'Bash';
-
+  //{parseVersion().major < 18 && (
```
AugustinMauroy:
```suggestion
```

apps/site/components/Downloads/Release/ReleaseCodeBox.tsx
```diff
@@ -13,6 +15,7 @@ const memoizedShiki = shikiPromise.then(highlightToHtml);
 
 const ReleaseCodeBox: FC = () => {
   const { platform, os, release } = useContext(ReleaseContext);
+  // const locale = useLocale();
```
AugustinMauroy:
```suggestion
```

AugustinMauroy (commented):
Some thing can be clean

joeeames:
changes have been made to address all requested items (I believe)

apps/site/components/Downloads/Release/ReleaseCodeBox.tsx
joeeames: fixed

joeeames:
> I'm -1 to have a new entry on nav bar because it's not 100% usefull. It also introduces a strange effect on the navbar because it hasn't been designed to support two entries that have the same ‘section’.
> 
> Capture d’écran 2024-10-15 à 09 26 12

We need to find a resolution here. Participation in the ESP program is about making sure that users who need to understand what versions are supported and what are not, and consequently find out about solutions for versions that are not still supported, need a simple and easy way to find the page documenting this information. Currently it's buried under "about" and then "previous releases" which isn't at all straightforward for someone wanting to find out about support. I get the comment about it being "100% useful" but for someone trying to find out if the version they're on is supported, these links are critical. 

Some other options are to create an entirely new page as we've done with many other projects, but the previous releases page is by far more suited for this. Or to remove it from the about section and put it in its own section? We're open to anything that fits the criteria of "easy to find". So any ideas are welcome here.

apps/site/components/Downloads/Release/ReleaseCodeBox.tsx
```diff
@@ -29,13 +31,16 @@ const ReleaseCodeBox: FC = () => {
   }, [release.versionWithPrefix, os, platform]);
 
   const codeLanguage = os === 'WIN' ? 'PowerShell' : 'Bash';
-
   return (
     <div className="mb-2 mt-6 flex flex-col gap-2">
+      {semVer.lt(release.versionWithPrefix, '18.0.0') && (
```
AugustinMauroy: where does ‘18.0.0’ come from? and why ?

AugustinMauroy:
I don't agree with you that it's buried in the about section. It's fine like that. Because the section above is a navbar whose purpose is to navigate between sections and NOT to go to a specific article.
I've looked elsewhere and no-one puts ‘version support’ in the nav. I've checked for Deno, .net, java, php, ruby.

apps/site/components/Downloads/Release/ReleaseCodeBox.tsx
joeeames: yeah, sadly it's hardcoded. i didn't see that this was encoded anywhere in metadata, but 18 is the oldest "supported" version as of today. This is definitely not ideal as it would need to be updated whenever new versions fall out of support.

joeeames:
> I don't agree with you that it's buried in the about section. It's fine like that. Because the section above is a navbar whose purpose is to navigate between sections and NOT to go to a specific article. I've looked elsewhere and no-one puts ‘version support’ in the nav. I've checked for Deno, .net, java, php, ruby.

Sorry I wasn't meaning "projects in general" but "projects we've worked with". Meaning the purpose of the ESP is to improve the use case for users for whom support matters. Part of that is making the support information easy to find. So to clarify, on the other projects that HeroDevs has worked with (angular, jquery, bootstrap, vue 2, ESLint, etc) we've added links to this information usually in the main nav or on the home page. This is obviously not the only way to do this, but we work to make this page easy to find. 

It's very subjective of course, I just try to put myself in the shoes of someone who may not know a site intimately, can I find the information I'm looking for and if not what change would make it easy to find. Since I personally do this as I look to see what changes to recommend to make, I think I'm a good proxy for this user profile. So when I went looking for the information on the Node website I definitely had some trouble finding it. Again, this is very subjective. 

Thought I'd throw in some screenshots of ways some other project have addressed this. Two of these three did NOT go with a new top nav item but instead added something to the home page.

ESLINT:
image

jQuery:
image

Bootstrap:
image

marco-ippolito:
@mikeesto can you please take a look again? Please let us know how to move forward

mcollina (approved):
lgtm

bmuenzenmeyer:
I'm unsure why https://github.com/nodejs/nodejs.org/issues/7037 was first referenced, the OP doesn't seem to.

To be clear - the project and website maintainers fully support this effort. I'm happy to help land it to satisfaction of all parties.
I briefly looked into the rendering concerns that Augustin (rightly) mentioned within [this comment](https://github.com/nodejs/nodejs.org/pull/7086#issuecomment-2413107188) - I think that's the only real blocker from a UX perspective.

I'll try to move things along by end of weekend.

mikeesto (approved):
Thanks for taking our feedback on board @joeeames . Perhaps the nav bar item could be named "Releases" instead of "Version Support" ? I agree with others that it would be good to fix the UX here too before this lands

apps/site/components/Common/ActiveLink/index.tsx
```diff
@@ -26,7 +27,9 @@ const ActiveLink: FC<ActiveLocalizedLinkProps> = ({
       ? // When using allowSubPath we want only to check if
         // the current pathname starts with the utmost upper level
         // of an href (e.g. /docs/...)
-        pathname.startsWith(`/${href.toString().split('/')[1]}`)
+        pathname.startsWith(`/${href.toString().split('/')[1]}`) &&
+        // but not when this link is for the deep link shortcut to previous releases
+        href.toString() !== VERSION_SUPPORT_SHORTCUT
```
bmuenzenmeyer, line 32: if this doesn't merge by morning, i may add unit tests for the entire `allowSubpath` part of the ternary, which has no test coverage. it requires mocking `usePathname`

apps/site/next.constants.mjs
```diff
@@ -168,3 +168,18 @@ export const ORAMA_CLOUD_API_KEY = process.env.NEXT_PUBLIC_ORAMA_API_KEY || '';
...
+
+/**
+ * OpenJS Ecosystem Support Program (ESP) partners provide security updates and support for end-of-life and unsupported versions
+ *
+ * See https://openjsf.org/ecosystem-sustainability-program
+ *
+ * This is the minimum version Node.js support according to https://nodejs.org/en/about/previous-releases
+ */
+export const ESP_SUPPORT_THRESHOLD_VERSION =
+  process.env.ESP_SUPPORT_THRESHOLD_VERSION || '18.0.0';
```
bmuenzenmeyer, line 180: moving this to a ENV VAR or constant makes the magic string more obvious. with even versioning being the important milestone for prod usage, this is only really in need of touch twice a year, right?

we could automate it with this, as other portions of the code do

```
import getReleaseData from '@/next-data/releaseData';
```

bmuenzenmeyer (approved):
push a commit onto the branch, hope that's okay with you @joeeames

bmuenzenmeyer:
failing test is resolved on `main` via https://github.com/nodejs/nodejs.org/commit/851b085af21fe74612eccae9094803574c50e39c

apps/site/components/Common/ActiveLink/index.tsx
bmuenzenmeyer: added via https://github.com/nodejs/nodejs.org/pull/7086/commits/9847aef066a21499174b84b59e8876945601195a

PR #7092, opened 2024-10-06
Add mocking guide

JakobJingleheimer:
## Description

Add a guide for when, why, and how to mock in tests via node built-ins

## Validation

Read the article :)

## Related Issues

None

### Check List

### DIFF
--- a/apps/site/pages/en/learn/test-runner/mocking.md
+++ b/apps/site/pages/en/learn/test-runner/mocking.md
@@ -21,19 +21,19 @@ But first, there are several types of tests:
 | integration      | components fitting together               | -                                                                                              | external code, external system           |
 | end-to-end (e2e) | app + external data stores, delivery, etc | A fake user (ex a Playwright agent) literally using an app connected to real external systems. | none (do not mock)                       |
 
-There are varying schools of thought about when you should and should not mock, the broadstrokes of which are laid out below.
+There are different schools of thought about when to mock and when not to mock, the broad strokes of which are outlined below.
 
 ## When and not to mock
 
 There are 3 main mock candidates:
 
-- own code
-- external code
-- external system
+- Own code
+- External code
+- External system
 
 ### Own code
 
-This is what you/your project control.
+This is what your project controls.
 
 ```mjs displayName="your-project/main.mjs"
 import foo from './foo.mjs';
@@ -53,11 +53,11 @@ For a true unit test of `main`, `foo` should be mocked: you're testing that `mai
 
 Mocking `foo` can be more trouble than worth, especially when `foo` is simple, well-tested, and rarely updated.
 
-Others argue that not mocking `foo` is better because it's more authentic and increases coverage of `foo` (because `main`'s tests will also verify `foo`). This can, however, create noise: when `foo` breaks, a bunch of other tests will also break, so tracking down the problem is more tedious: if only the 1 test for the item ultimately responsible for the issue is failing, that's very easy to spot; whereas 100 tests failing creates a needle-in-a-haystack to find the real problem.
+Not mocking `foo` can be better because it's more authentic and increases coverage of `foo` (because `main`'s tests will also verify `foo`). This can, however, create noise: when `foo` breaks, a bunch of other tests will also break, so tracking down the problem is more tedious: if only the 1 test for the item ultimately responsible for the issue is failing, that's very easy to spot; whereas 100 tests failing creates a needle-in-a-haystack to find the real problem.
 
 ### External code
 
-This is what you/your project does not control.
+This is what your project does not control.
 
 ```mjs displayName="your-project/main.mjs"
 import bar from 'bar';
@@ -73,11 +73,11 @@ Uncontroversially, for unit tests, this should always be mocked. For component a
 
 #### Why
 
-Verifying that someone else's code works with your code is not the role of a unit test (and their code should already be tested).
+Verifying that code that your project does not maintain works is not the goal of a unit test (and that code should have its own tests).
 
 #### Why not
 
-Sometimes, it's just not realistic to mock. For example, you would almost never mock react (the medicine would be worse than the ailment).
+Sometimes, it's just not realistic to mock. For example, you would almost never mock a large framework such as react or angular (the medicine would be worse than the ailment).
 
 ### External system
 
@@ -118,7 +118,7 @@ describe('storage', { concurrency: true }, () => {
 
     const results = await read('a', true);
 
-    assert.equal(items.length, 1); // ensure read did not retrieve erroneous item
+    assert.equal(results.length, 1); // ensure read did not retrieve erroneous item
 
     assert.deepEqual(results[0], { key: 'good', val: 'item' });
   });
@@ -135,13 +135,13 @@ describe('storage', { concurrency: true }, () => {
 });
 ```
 
-In the above, the first and second cases (the `it`s) can sabotage each other because they are run concurrently and mutate the same store (a race condition): "save"'s insertion can cause the otherwise valid "read"'s test to fail its assertion on items found (and "read"'s can do the same thing to "save"'s).
+In the above, the first and second cases (the `it()` statements) can sabotage each other because they are run concurrently and mutate the same store (a race condition): `save()`'s insertion can cause the otherwise valid `read()`'s test to fail its assertion on items found (and `read()`'s can do the same thing to `save()`'s).
 
 ## What to mock
 
 ### Modules + units
 
-This leverages [`mock`](https://nodejs.org/api/test.html#class-mocktracker) from node's test runner.
+This leverages [`mock`](https://nodejs.org/api/test.html#class-mocktracker) from the Node.js test runner.
 
 ```mjs
 import assert from 'node:assert/strict';
@@ -167,8 +167,6 @@ describe('foo', { concurrency: true }, () => {
 
     // This MUST be a dynamic import because that is the only way to ensure the
     // import starts after the mock has been set up.
-    // There is a far more technical explanation,
-    // but just trust that this is logically necessary.
     ({ foo } = await import('./foo.mjs'));
   });
 
@@ -182,7 +180,7 @@ describe('foo', { concurrency: true }, () => {
 
 ### APIs
 
-A little-known fact: node has a builtin way to mock `fetch`. [`undici`](https://github.com/nodejs/undici) is the Node.js implementation of `fetch`. You do not have to install it—it's shipped with `node` by default.
+A little-known fact is that there is a builtin way to mock `fetch`. [`undici`](https://github.com/nodejs/undici) is the Node.js implementation of `fetch`. You do not have to install it—it's shipped with `node` by default.
 
 ```mjs displayName="endpoints.spec.mjs"
 import assert from 'node:assert/strict';
@@ -246,7 +244,7 @@ describe('endpoints', { concurrency: true }, () => {
 
 ### Time
 
-Like Doctor Strange, you too can control time. You would usually do this just for convience to avoid artificially protracted test runs (do you really want to wait 3 minutes for that setTimeout to trigger?). You may also want to travel through time. This leverages [`mock timers`](https://nodejs.org/api/test.html#class-mocktimers) from node's test runner.
+Like Doctor Strange, you too can control time. You would usually do this just for convenience to avoid artificially protracted test runs (do you really want to wait 3 minutes for that `setTimeout()` to trigger?). You may also want to travel through time. This leverages [`mock.timers`](https://nodejs.org/api/test.html#class-mocktimers) from the Node.js test runner.
 
 Note the use of time-zone here (`Z` in the time-stamps). Neglecting to include a consistent time-zone can (read: likely will) lead to unexpected restults.
 
--- a/apps/site/pages/en/learn/test-runner/mocking.md
+++ b/apps/site/pages/en/learn/test-runner/mocking.md
@@ -67,7 +67,7 @@ export function main() {
 }
 ```
 
-Here, `bar` is a node_module installed via npm (or yarn, etc).
+Here, `bar` is an external package, e.g. an npm dependency.
 
 Uncontroversially, for unit tests, this should always be mocked. For component and integration tests, whether to mock depends on what this is.
 
@@ -182,7 +182,7 @@ describe('foo', { concurrency: true }, () => {
 
 ### APIs
 
-A little-known fact is that there is a builtin way to mock `fetch`. [`undici`](https://github.com/nodejs/undici) is the Node.js implementation of `fetch`. You do not have to install it—it's shipped with `node` by default.
+A little-known fact is that there is a builtin way to mock `fetch`. [`undici`](https://github.com/nodejs/undici) is the Node.js implementation of `fetch`. It's shipped with `node`, but not currently exposed by `node` itself, so it must be installed (ex `npm install undici`).
 
 ```mjs displayName="endpoints.spec.mjs"
 import assert from 'node:assert/strict';
--- a/apps/site/pages/en/learn/test-runner/mocking.md
+++ b/apps/site/pages/en/learn/test-runner/mocking.md
@@ -248,7 +248,7 @@ describe('endpoints', { concurrency: true }, () => {
 
 Like Doctor Strange, you too can control time. You would usually do this just for convenience to avoid artificially protracted test runs (do you really want to wait 3 minutes for that `setTimeout()` to trigger?). You may also want to travel through time. This leverages [`mock.timers`](https://nodejs.org/api/test.html#class-mocktimers) from the Node.js test runner.
 
-Note the use of time-zone here (`Z` in the time-stamps). Neglecting to include a consistent time-zone can (read: likely will) lead to unexpected restults.
+Note the use of time-zone here (`Z` in the time-stamps). Neglecting to include a consistent time-zone will likely lead to unexpected restults.
 
 ```mjs displayName="master-time.spec.mjs"
 import assert from 'node:assert/strict';
--- a/apps/site/pages/en/learn/test-runner/mocking.md
+++ b/apps/site/pages/en/learn/test-runner/mocking.md
@@ -6,13 +6,13 @@ authors: JakobJingleheimer
 
 # Mocking in tests
 
-Tests should be deterministic: runnable in any order, any number of times, and always produce the same result. Proper setup and mocking make this possible.
-
 Mocking is a means of creating a facsimile, a puppet. This is generally done in a `when 'a', do 'b'` manner of puppeteering. The idea is to limit the number of moving pieces and control things that "don't matter". "mocks" and "stubs" are technically different kinds of "test doubles". For the curious mind, a stub is a replacement that does nothing (a no-op) but track its invocation. A mock is a stub that also has a fake implementation (the `when 'a', do 'b'`). Within this doc, the difference is unimportant, and stubs are referred to as mocks.
+
+Tests should be deterministic: runnable in any order, any number of times, and always produce the same result. Proper setup and mocking make this possible.
 
 Node.js provides many ways to mock various pieces of code.
 
-But first, there are several types of tests:
+This articles deals with the following types of tests:
 
 | type             | description                               | example                                                                                        | mock candidates                          |
 | :--------------- | :---------------------------------------- | :--------------------------------------------------------------------------------------------- | :--------------------------------------- |
--- a/apps/site/pages/en/learn/test-runner/mocking.md
+++ b/apps/site/pages/en/learn/test-runner/mocking.md
@@ -8,6 +8,8 @@ authors: JakobJingleheimer
 
 Tests should be deterministic, runnable in any order any number of times and always produce the same result. Proper setup and mocking make that possible.
 
+Mocking is a means of creating a facsimile, a puppet. This is generally done in a `when 'a', do 'b'` manner of puppeteering. The idea is to limit the number of moving pieces and control things that "don't matter". "mocks" and "stubs" are technically different kinds of "test doubles". For the curious mind, a stub is a replacement that does nothing (a no-op) but track its invocation. A mock is a stub that also has a fake implementation (the `when 'a', do 'b'`). Within this doc, the difference is unimportant, and stubs are referred to as mocks.
+
 Node.js provides many ways to mock various pieces of code.
 
 But first, there are several types of tests:
--- a/apps/site/pages/en/learn/test-runner/mocking.md
+++ b/apps/site/pages/en/learn/test-runner/mocking.md
@@ -6,7 +6,7 @@ authors: JakobJingleheimer
 
 # Mocking in tests
 
-Tests should be deterministic, runnable in any order any number of times and always produce the same result. Proper setup and mocking make that possible.
+Tests should be deterministic: runnable in any order, any number of times, and always produce the same result. Proper setup and mocking make this possible.
 
 Mocking is a means of creating a facsimile, a puppet. This is generally done in a `when 'a', do 'b'` manner of puppeteering. The idea is to limit the number of moving pieces and control things that "don't matter". "mocks" and "stubs" are technically different kinds of "test doubles". For the curious mind, a stub is a replacement that does nothing (a no-op) but track its invocation. A mock is a stub that also has a fake implementation (the `when 'a', do 'b'`). Within this doc, the difference is unimportant, and stubs are referred to as mocks.
 
@@ -69,7 +69,7 @@ export function main() {
 
 Here, `bar` is a node_module installed via npm (or yarn, etc).
 
-Uncontroversially, for unit tests, this should always be mocked. For component and integration tests, whether to mock depends on what it is.
+Uncontroversially, for unit tests, this should always be mocked. For component and integration tests, whether to mock depends on what this is.
 
 #### Why
 
@@ -83,7 +83,7 @@ Sometimes, it's just not realistic to mock. For example, you would almost never 
 
 These are things like databases, environments (Chromium or Firefox for a web app, an operating system for a node app, etc), file systems, memory store, etc.
 
-Ideally, mocking these would not be necessary. Aside from somehow creating isolated copies for each case (usually very impractical, due to cost, additional execution time, etc), the next best option is to mock. Without doing, tests sabotage each other:
+Ideally, mocking these would not be necessary. Aside from somehow creating isolated copies for each case (usually very impractical due to cost, additional execution time, etc), the next best option is to mock. Without mocking, tests sabotage each other:
 
 ```mjs displayName="storage.mjs"
 import { db } from 'db';
@@ -131,6 +131,8 @@ describe('storage', { concurrency: true }, () => {
     const items = await db.getAll();
 
     assert.equal(items.length, 1); // ensure save did not create duplicates
+
+    assert.deepEqual(items[0], { key: 'good', val: 'item' });
   });
 });
 ```
--- a/apps/site/pages/en/learn/test-runner/mocking.md
+++ b/apps/site/pages/en/learn/test-runner/mocking.md
@@ -1,0 +1,264 @@
+---
+title: Mocking in tests
+layout: learn
+authors: JakobJingleheimer
+---
+
+# Mocking in tests
+
+Tests should be deterministic, runnable in any order any number of times and always produce the same result. Proper setup and mocking make that possible.
+
+Node.js provides many ways to mock various pieces of code.
+
+But first, there are several types of tests:
+
+| type             | description                               | example                                                                                        | mock candidates                          |
+| :--------------- | :---------------------------------------- | :--------------------------------------------------------------------------------------------- | :--------------------------------------- |
+| unit             | the smallest bit of code you can isolate  | `const sum = (a, b) => a + b`                                                                  | own code, external code, external system |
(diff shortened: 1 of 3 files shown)


AugustinMauroy:
On first reading, it seems like a good direction to me. But I think adding a paragraph explaining the concept of ‘mock’ would be useful, as the learn section is supposed to be more explanatory than the API docs, which are very technical.

I also noticed that there is no link to the mock docs API, which could be very useful.

apps/site/pages/en/learn/test-runner/mocking.md
```diff
@@ -0,0 +1,264 @@
...
+
+    assert.deepEqual(await endpoints.save(endpoint), {
+      code,
+      data,
+    });
+  });
+});
+```
+
+### Time
```
AugustinMauroy, line 247: May be add mention of TZ. This can affect tests as part of a unit test of front-end code.

apps/site/pages/en/learn/test-runner/mocking.md
```diff
@@ -0,0 +1,264 @@
...
+
+```mjs displayName="your-project/main.mjs"
+import bar from 'bar';
+
+export function main() {
+  const f = bar();
+}
+```
+
+Here, `bar` is a node_module installed via npm (or yarn, etc).
```
aduh95:
```suggestion
Here, `bar` is an external package, e.g. an npm dependency.
```

apps/site/pages/en/learn/test-runner/mocking.md
```diff
@@ -0,0 +1,264 @@
...
+    barMock.mockImplementationOnce(function bar_mock() {/* … */});
+
+    assert.equal(foo(), 42);
+  });
+});
+```
+
+### Services
+
+A little-known fact: node has a builtin way to mock `fetch`. [`undici`](https://github.com/nodejs/undici) is the Node.js implementation of fetch. You do not have to install it—it's shipped with node by default.
```
aduh95:
```suggestion
A little-known fact: node has a builtin way to mock `fetch`. [`undici`](https://github.com/nodejs/undici) is the Node.js implementation of `fetch`. You do not have to install it — it's shipped with `node` by default.
```

JakobJingleheimer: Spaces around an em-dash are not correct

apps/site/pages/en/learn/test-runner/mocking.md
JakobJingleheimer: I think this is not correct? Does it require `npm install undici`?

mikeesto:
> But I think adding a paragraph explaining the concept of ‘mock’ would be useful, as the learn section is supposed to be more explanatory than the API docs, which are very technical.

+1 I think this would be helpful too - a short couple of sentences at the beginning

apps/site/pages/en/learn/test-runner/mocking.md
```diff
@@ -0,0 +1,270 @@
+---
+title: Mocking in tests
+layout: learn
+authors: JakobJingleheimer
+---
+
+# Mocking in tests
+
+Tests should be deterministic, runnable in any order any number of times and always produce the same result. Proper setup and mocking make that possible.
```
avivkeller:
```suggestion
Tests should be deterministic; They should be runnable in any order, any number of times, and always produce the same result. Proper setup and mocking make this possible.
```

apps/site/pages/en/learn/test-runner/mocking.md
```diff
@@ -0,0 +1,270 @@
...
+| :--------------- | :---------------------------------------- | :--------------------------------------------------------------------------------------------- | :--------------------------------------- |
+| unit             | the smallest bit of code you can isolate  | `const sum = (a, b) => a + b`                                                                  | own code, external code, external system |
+| component        | a unit + dependencies                     | `const arithmetic = (op = sum, a, b) => ops[op](a, b)`                                         | external code, external system           |
+| integration      | components fitting together               | -                                                                                              | external code, external system           |
+| end-to-end (e2e) | app + external data stores, delivery, etc | A fake user (ex a Playwright agent) literally using an app connected to real external systems. | none (do not mock)                       |
+
+There are varying schools of thought about when you should and should not mock, the broadstrokes of which are laid out below.
+
+## When and not to mock
```
avivkeller, line 26:
```suggestion
## When to mock
```

apps/site/pages/en/learn/test-runner/mocking.md
```diff
@@ -0,0 +1,270 @@
...
+
+There are 3 main mock candidates:
+
+- own code
+- external code
+- external system
+
+### Own code
+
+This is what you/your project control.
```
avivkeller:
```suggestion
This is what your project controls.
```

apps/site/pages/en/learn/test-runner/mocking.md
```diff
@@ -0,0 +1,270 @@
...
+
+```mjs displayName="your-project/main.mjs"
+import bar from 'bar';
+
+export function main() {
+  const f = bar();
+}
+```
+
+Here, `bar` is a node_module installed via npm (or yarn, etc).
```
avivkeller:
```suggestion
Here, `bar` is a package located in `node_modules` (e.g an npm library).
```

Any package in the node_modules folder will suffice, not only npm dependencies

apps/site/pages/en/learn/test-runner/mocking.md
```diff
@@ -0,0 +1,270 @@
...
+
+Verifying that someone else's code works with your code is not the role of a unit test (and their code should already be tested).
+
+#### Why not
+
+Sometimes, it's just not realistic to mock. For example, you would almost never mock react (the medicine would be worse than the ailment).
+
+### External system
+
+These are things like databases, environments (Chromium or Firefox for a web app, an operating system for a node app, etc), file systems, memory store, etc.
```
avivkeller, line 84:
```suggestion
These are things like databases, environments, file systems, memory store, etc.
```

apps/site/pages/en/learn/test-runner/mocking.md
```diff
@@ -0,0 +1,270 @@
...
+
+Ideally, mocking these would not be necessary. Aside from somehow creating isolated copies for each case (usually very impractical, due to cost, additional execution time, etc), the next best option is to mock. Without doing, tests sabotage each other:
```
avivkeller:
```suggestion
Ideally, mocking these would not be necessary. Aside from somehow creating isolated copies for each case (usually very impractical due to cost, additional execution time, etc), the next best option is to mock. Without mocking, tests sabotage each other:
```

apps/site/pages/en/learn/test-runner/mocking.md
```diff
@@ -0,0 +1,270 @@
...
+    assert.ok(id);
+
+    const items = await db.getAll();
+
+    assert.equal(items.length, 1); // ensure save did not create duplicates
+  });
+});
+```
+
+In the above, the first and second cases (the `it`s) can sabotage each other because they are run concurrently and mutate the same store (a race condition): "save"'s insertion can cause the otherwise valid "read"'s test to fail its assertion on items found (and "read"'s can do the same thing to "save"'s).
```
avivkeller:
```suggestion
In the above, the first and second cases (the `it()` statements) can sabotage each other because they are run concurrently and mutate the same store (a race condition): `save()`'s insertion can cause the otherwise valid `read()`'s test to fail its assertion on items found (and `read()`'s can do the same thing to `save()`'s).
```

apps/site/pages/en/learn/test-runner/mocking.md
```diff
@@ -0,0 +1,270 @@
...
+
+## What to mock
+
+### Modules + units
+
+This leverages [`mock`](https://nodejs.org/api/test.html#class-mocktracker) from node's test runner.
```
avivkeller:
```suggestion
This leverages [`mock`](https://nodejs.org/api/test.html#class-mocktracker) from the Node.js test runner.
```

apps/site/pages/en/learn/test-runner/mocking.md
```diff
@@ -0,0 +1,270 @@
...
+  let barMock = mock.fn();
+  let foo;
+
+  before(async () => {
+    const barNamedExports = await import('./bar.mjs')
+      // discard the original default export
+      .then(({ default, ...rest }) => rest);
+
+    // It's usually not necessary to manually call restore() after each
+    // nor reset() after all (node does this automatically).
```
avivkeller, line 163:
```suggestion
    // nor reset() after all (this is done automatically).
```

apps/site/pages/en/learn/test-runner/mocking.md
```diff
@@ -0,0 +1,270 @@
...
+    mock.module('./bar.mjs', {
+      defaultExport: barMock
+      // Keep the other exports that you don't want to mock.
+      namedExports: barNamedExports,
+    });
+
+    // This MUST be a dynamic import because that is the only way to ensure the
+    // import starts after the mock has been set up.
+    // There is a far more technical explanation,
+    // but just trust that this is logically necessary.
```
avivkeller: IMHO "just trust" isn't a very assuring statement for beginners, and the comment portrays the same info without it, so maybe just:

```suggestion
    // This MUST be a dynamic import because that is the only way to ensure the
    // import starts after the mock has been set up.
```

apps/site/pages/en/learn/test-runner/mocking.md
```diff
@@ -0,0 +1,270 @@
...
+    barMock.mockImplementationOnce(function bar_mock() {/* … */});
+
+    assert.equal(foo(), 42);
+  });
+});
+```
+
+### APIs
+
+A little-known fact: node has a builtin way to mock `fetch`. [`undici`](https://github.com/nodejs/undici) is the Node.js implementation of `fetch`. You do not have to install it—it's shipped with `node` by default.
```
avivkeller:
```suggestion
A little-known fact is that there is a builtin way to mock `fetch`. [`undici`](https://github.com/nodejs/undici) is the Node.js implementation of `fetch`. You do not have to install it—it's shipped with `node` by default.
```

> You do not have to install it—it's shipped with `node` by default.

Is this true? AFAICT it must be installed, despite being a dependency of Node.js.

apps/site/pages/en/learn/test-runner/mocking.md
```diff
@@ -0,0 +1,270 @@
...
+      code,
+      data,
+    });
+  });
+});
+```
+
+### Time
+
+Like Doctor Strange, you too can control time. You would usually do this just for convience to avoid artificially protracted test runs (do you really want to wait 3 minutes for that setTimeout to trigger?). You may also want to travel through time. This leverages [`mock timers`](https://nodejs.org/api/test.html#class-mocktimers) from node's test runner.
```
avivkeller:
```suggestion
Like Doctor Strange, you too can control time. You would usually do this just for convenience to avoid artificially protracted test runs (do you really want to wait 3 minutes for that `setTimeout()` to trigger?). You may also want to travel through time. This leverages [`mock.timers`](https://nodejs.org/api/test.html#class-mocktimers) from the Node.js test runner.
```

apps/site/pages/en/learn/test-runner/mocking.md
```diff
@@ -0,0 +1,270 @@
...
+
+Note the use of time-zone here (`Z` in the time-stamps). Neglecting to include a consistent time-zone can (read: likely will) lead to unexpected restults.
```
avivkeller:
```suggestion
Note the use of time-zone here (`Z` in the time-stamps). Neglecting to include a consistent time-zone will likely lead to unexpected restults.
```

apps/site/pages/en/learn/test-runner/mocking.md
```diff
@@ -0,0 +1,270 @@
...
+title: Mocking in tests
+layout: learn
+authors: JakobJingleheimer
+---
+
+# Mocking in tests
+
+Tests should be deterministic, runnable in any order any number of times and always produce the same result. Proper setup and mocking make that possible.
+
+Mocking is a means of creating a facsimile, a puppet. This is generally done in a `when 'a', do 'b'` manner of puppeteering. The idea is to limit the number of moving pieces and control things that "don't matter". "mocks" and "stubs" are technically different kinds of "test doubles". For the curious mind, a stub is a replacement that does nothing (a no-op) but track its invocation. A mock is a stub that also has a fake implementation (the `when 'a', do 'b'`). Within this doc, the difference is unimportant, and stubs are referred to as mocks.
```
avivkeller, line 9:
```suggestion
Mocking is a means to create a facsimile, a puppet. This is generally done in a `when 'a', do 'b'` manner of puppeteering. The idea is to limit the number of moving pieces and control things that "don't matter". "mocks" and "stubs" are technically different kinds of "test doubles". For the curious mind, a stub is a replacement that does nothing (a no-op) but track its invocation. A mock is a stub that also has a fake implementation (the `do 'b'`). Within this doc, the difference is unimportant, and stubs are referred to as mocks.
```

apps/site/pages/en/learn/test-runner/mocking.md
```diff
@@ -0,0 +1,270 @@
...
+
+There are varying schools of thought about when you should and should not mock, the broadstrokes of which are laid out below.
+
+## When and not to mock
+
+There are 3 main mock candidates:
+
+- own code
+- external code
+- external system
```
avivkeller:
```suggestion
- Own code
- External code
- External system
```

apps/site/pages/en/learn/test-runner/mocking.md
```diff
@@ -0,0 +1,270 @@
...
+
+#### Why
+
+For a true unit test of `main`, `foo` should be mocked: you're testing that `main` works, not that `main` + `foo` work (that's a different test).
+
+#### Why not
+
+Mocking `foo` can be more trouble than worth, especially when `foo` is simple, well-tested, and rarely updated.
+
+Others argue that not mocking `foo` is better because it's more authentic and increases coverage of `foo` (because `main`'s tests will also verify `foo`). This can, however, create noise: when `foo` breaks, a bunch of other tests will also break, so tracking down the problem is more tedious: if only the 1 test for the item ultimately responsible for the issue is failing, that's very easy to spot; whereas 100 tests failing creates a needle-in-a-haystack to find the real problem.
```
avivkeller:
```suggestion
Mocking `foo` can be more trouble than worth, especially when `foo` is simple, well-tested, and rarely updated.

Not mocking `foo` can be better because it's more authentic and increases coverage of `foo` (because `main`'s tests will also verify `foo`). This can, however, create noise: when `foo` breaks, a bunch of other tests will also break, so tracking down the problem is more tedious: if only the 1 test for the item ultimately responsible for the issue is failing, that's very easy to spot; whereas 100 tests failing creates a needle-in-a-haystack to find the real problem.
```

apps/site/pages/en/learn/test-runner/mocking.md
```diff
@@ -0,0 +1,270 @@
...
+
+### External code
+
+This is what you/your project does not control.
```
avivkeller:
```suggestion
This is what your project does not control.
```

apps/site/pages/en/learn/test-runner/mocking.md
```diff
@@ -0,0 +1,270 @@
...
+import bar from 'bar';
+
+export function main() {
+  const f = bar();
+}
+```
+
+Here, `bar` is a node_module installed via npm (or yarn, etc).
+
+Uncontroversially, for unit tests, this should always be mocked. For component and integration tests, whether to mock depends on what it is.
```
avivkeller: What _what_ is? The dep? The test? Can you be more specific?

apps/site/pages/en/learn/test-runner/mocking.md
```diff
@@ -0,0 +1,270 @@
...
+
+#### Why
+
+Verifying that someone else's code works with your code is not the role of a unit test (and their code should already be tested).
```
avivkeller:
```suggestion
Verifying that code that your project does not maintain works is not the goal of a unit test (and that code should have its own tests).
```

apps/site/pages/en/learn/test-runner/mocking.md
```diff
@@ -0,0 +1,270 @@
...
+
+#### Why not
+
+Sometimes, it's just not realistic to mock. For example, you would almost never mock react (the medicine would be worse than the ailment).
```
avivkeller:
```suggestion
Sometimes, it's just not realistic to mock. For example, you would almost never mock a large framework such as react or angular (the medicine would be worse than the ailment).
```

apps/site/pages/en/learn/test-runner/mocking.md
```diff
@@ -0,0 +1,270 @@
...
+describe('storage', { concurrency: true }, () => {
+  it('should retrieve the requested item', async () => {
+    await db.upsert('good', 'item'); // give it something to read
+    await db.upsert('erroneous', 'item'); // give it a chance to fail
+
+    const results = await read('a', true);
+
+    assert.equal(items.length, 1); // ensure read did not retrieve erroneous item
+
+    assert.deepEqual(results[0], { key: 'good', val: 'item' });
```
avivkeller: What is "items"? Did you mean:
```suggestion
    assert.equal(results.length, 1); // ensure read did not retrieve erroneous item

    assert.deepEqual(results[0], { key: 'good', val: 'item' });
```

Or even better:
```suggestion
    assert.deepEqual(results, [{ key: 'good', val: 'item' }]);
```

apps/site/pages/en/learn/test-runner/mocking.md
```diff
@@ -0,0 +1,270 @@
...
+  });
+
+  it('should save the new item', async () => {
+    const id = await save('good', 'item');
+
+    assert.ok(id);
+
+    const items = await db.getAll();
+
+    assert.equal(items.length, 1); // ensure save did not create duplicates
```
avivkeller, line 133: Can you assert the actual contents for consistency with the other test?

apps/site/pages/en/learn/test-runner/mocking.md
```diff
@@ -0,0 +1,270 @@
...
+| :--------------- | :---------------------------------------- | :--------------------------------------------------------------------------------------------- | :--------------------------------------- |
+| unit             | the smallest bit of code you can isolate  | `const sum = (a, b) => a + b`                                                                  | own code, external code, external system |
+| component        | a unit + dependencies                     | `const arithmetic = (op = sum, a, b) => ops[op](a, b)`                                         | external code, external system           |
+| integration      | components fitting together               | -                                                                                              | external code, external system           |
+| end-to-end (e2e) | app + external data stores, delivery, etc | A fake user (ex a Playwright agent) literally using an app connected to real external systems. | none (do not mock)                       |
+
+There are varying schools of thought about when you should and should not mock, the broadstrokes of which are laid out below.
```
AugustinMauroy:
```suggestion
There are different schools of thought about when to mock and when not to mock, the broad strokes of which are outlined below.
```

apps/site/pages/en/learn/test-runner/mocking.md
JakobJingleheimer: That changes the meaning of the heading to only 1 of 2 things, and the section it covers has both things

apps/site/pages/en/learn/test-runner/mocking.md
JakobJingleheimer: Why remove the examples?

apps/site/pages/en/learn/test-runner/mocking.md
JakobJingleheimer: huh?

apps/site/pages/en/learn/test-runner/mocking.md
JakobJingleheimer: The first part of the change is just stylistic (I prefer mine), and the second part is not correct (the mock includes both "when" and "do"; stubs have no "when" or "do").

apps/site/pages/en/learn/test-runner/mocking.md
avivkeller: For brevity, but these are all nitpicks, so feel free to disregard

apps/site/pages/en/learn/test-runner/mocking.md
avivkeller: 🤣 I made a typo in my suggestion, please have another look

apps/site/pages/en/learn/test-runner/mocking.md
```diff
@@ -0,0 +1,270 @@
...
+import { save } from './storage.mjs';
+
+describe('storage', { concurrency: true }, () => {
+  it('should retrieve the requested item', async () => {
+    await db.upsert('good', 'item'); // give it something to read
+    await db.upsert('erroneous', 'item'); // give it a chance to fail
+
+    const results = await read('a', true);
+
+    assert.equal(items.length, 1); // ensure read did not retrieve erroneous item
```
JakobJingleheimer, line 121:
```suggestion
    assert.equal(results.length, 1); // ensure read did not retrieve erroneous item
```

JakobJingleheimer:
```suggestion
    assert.equal(results.length, 1); // ensure read did not retrieve erroneous item
```

apps/site/pages/en/learn/test-runner/mocking.md
JakobJingleheimer: I agree the second one is better for an actual implementation; I split them up here to better demonstrate what and why.

apps/site/pages/en/learn/test-runner/mocking.md
JakobJingleheimer: I believe we're supposed to avoid the passive-voice.

apps/site/pages/en/learn/test-runner/mocking.md
JakobJingleheimer: Hmm, I just tried to access it, and indeed node claims undici is not there. Why… it's definitely present. @mcollina is it not possible to access the copy of undici that exists within node? 

If it's not possible to access node's copy:

* If it's installed via `npm`, won't that create 2 copies of it? (abusing disk space)
* Have we considered exposing node's copy? I thought that had been discussed and decided to indeed expose it.

apps/site/pages/en/learn/test-runner/mocking.md
JakobJingleheimer: I think it's extremely _likely_ (not extremely **un**likely)

apps/site/pages/en/learn/test-runner/mocking.md
JakobJingleheimer: The external code

apps/site/pages/en/learn/test-runner/mocking.md
avivkeller: In that case node should be replaced with Node.js.

apps/site/pages/en/learn/test-runner/mocking.md
avivkeller: I've modified the suggestion

apps/site/pages/en/learn/test-runner/mocking.md
avivkeller: IIRC it's been discussed before a few times, most recently https://github.com/nodejs/node/issues/43187#issuecomment-2321110677

apps/site/pages/en/learn/test-runner/mocking.md
JakobJingleheimer: I believe it should be `node`:

node = the software
Node.js = the project (and humans who make it)

Node the software is what automatically calls those cleanup methods

avivkeller: https://github.com/nodejs/node/blob/main/doc/README.md#nodejs-references

If you insist, it's okay, I just think Node.js is a better usage here, but at the end of the day, it really doesn't matter, it reads the same.

apps/site/pages/en/learn/test-runner/mocking.md
RedYetiDev: I think it is correct, it doesn't require the package to be installed via a package manager, the package could've just been placed in the node_modules folder.

avivkeller (approved):
Regardless of whether the rest of the nitpicks are added, it reads well, so LGTM.

The other members of @nodejs/test_runner might wanna take a look tho.

AugustinMauroy (approved):
LGTM ! Thanks to improve learn section

apps/site/pages/en/learn/test-runner/mocking.md
```diff
@@ -0,0 +1,270 @@
...
+    barMock.mockImplementationOnce(function bar_mock() {/* … */});
+
+    assert.equal(foo(), 42);
+  });
+});
+```
+
+### APIs
+
+A little-known fact is that there is a builtin way to mock `fetch`. [`undici`](https://github.com/nodejs/undici) is the Node.js implementation of `fetch`. You do not have to install it—it's shipped with `node` by default.
```
JakobJingleheimer:
```suggestion
A little-known fact is that there is a builtin way to mock `fetch`. [`undici`](https://github.com/nodejs/undici) is the Node.js implementation of `fetch`. It's shipped with `node`, but not currently exposed by `node` itself, so it must be installed (ex `npm install undici`).
```

JakobJingleheimer:
Thanks @aduh95 @AugustinMauroy @RedYetiDev !

I think it's ready to go, baring any requests from others on the test runner team. I'll wait a day to give them a chance to chime in.

mcollina (approved):
lgtm

aymen94 (approved):

apps/site/pages/en/learn/test-runner/mocking.md
```diff
@@ -0,0 +1,270 @@
...
+title: Mocking in tests
+layout: learn
+authors: JakobJingleheimer
+---
+
+# Mocking in tests
+
+Tests should be deterministic: runnable in any order, any number of times, and always produce the same result. Proper setup and mocking make this possible.
+
+Mocking is a means of creating a facsimile, a puppet. This is generally done in a `when 'a', do 'b'` manner of puppeteering. The idea is to limit the number of moving pieces and control things that "don't matter". "mocks" and "stubs" are technically different kinds of "test doubles". For the curious mind, a stub is a replacement that does nothing (a no-op) but track its invocation. A mock is a stub that also has a fake implementation (the `when 'a', do 'b'`). Within this doc, the difference is unimportant, and stubs are referred to as mocks.
```
aduh95, line 9:
```suggestion
Mocking is a mean of creating a facsimile, a puppet. This is generally done in a `when 'a', do 'b'` manner of puppeteering. The idea is to limit the number of moving pieces and control things that "don't matter". "mocks" and "stubs" are technically different kinds of "test doubles". For the curious mind, a stub is a replacement that does nothing (a no-op) but track its invocation. A mock is a stub that also has a fake implementation (the `when 'a', do 'b'`). Within this doc, the difference is unimportant, and stubs are referred to as mocks.
```

apps/site/pages/en/learn/test-runner/mocking.md
```diff
@@ -0,0 +1,270 @@
+---
+title: Mocking in tests
+layout: learn
+authors: JakobJingleheimer
+---
+
+# Mocking in tests
+
+Tests should be deterministic: runnable in any order, any number of times, and always produce the same result. Proper setup and mocking make this possible.
```
aduh95: nit: would it make more sense to start with the paragraph that defines mocking, then add this one which explains why mocking?

apps/site/pages/en/learn/test-runner/mocking.md
JakobJingleheimer: I think it is correct as written: https://dictionary.cambridge.org/dictionary/english/means

apps/site/pages/en/learn/test-runner/mocking.md
```diff
@@ -0,0 +1,270 @@
...
+
+Mocking is a means of creating a facsimile, a puppet. This is generally done in a `when 'a', do 'b'` manner of puppeteering. The idea is to limit the number of moving pieces and control things that "don't matter". "mocks" and "stubs" are technically different kinds of "test doubles". For the curious mind, a stub is a replacement that does nothing (a no-op) but track its invocation. A mock is a stub that also has a fake implementation (the `when 'a', do 'b'`). Within this doc, the difference is unimportant, and stubs are referred to as mocks.
+
+Node.js provides many ways to mock various pieces of code.
+
+But first, there are several types of tests:
```
aduh95:
```suggestion
This articles deals with the following types of tests:
```

apps/site/pages/en/learn/test-runner/mocking.md
JakobJingleheimer: Actually yeah that does make sense. Will update over lunch

apps/site/pages/en/learn/test-runner/mocking.md
aduh95: It seems the UK version takes an `s` in the singular form as you linked, but IIUC the US version doesn't (e.g. https://en.wiktionary.org/wiki/mean#Noun), and I think we tend to follow US conventions.

bmuenzenmeyer: its correct as `means`. to avoid the confusion, more active phasing could be used, like `mocking creates a fascimile, a puppet.`

not blocking this though

bmuenzenmeyer:
pushed a rebase

PR #7094, opened 2024-10-07
fix(accessibility): add right title/alt

### LINKED ISSUE #7093: Authorship displays only the user's GitHub profile image but no direct means to identify

Currently, the docs include authorship by means of the contributors' GitHub profile image. However, it does not cite the username or link to the profile. For a sighted user, this is very limited. This also creates an accessibility failure: authorship is effectively unavailable to a blind user.

A simple `title` attribute of the author's GitHub handle (or name) would be a good start. This could also get fancier beyond the browser's default display of `title` via a tooltip. On top of that, the image could link to the user's profile.

AugustinMauroy:
## Description

- Remove useless `getAcronymFromString` usage on `withMetaBar`
- Add a diffentitation bettwent `alt` and `fallback`
- Add title

## Validation

Title should be GitHub username. Alt should be GitHub username. And fallback should be the acronym.

## Related Issues

Partially fix #7093 (no new tooltips components added because the are no discussion on that) 

### Check List

### DIFF
--- a/apps/site/components/Common/AvatarGroup/Avatar/index.tsx
+++ b/apps/site/components/Common/AvatarGroup/Avatar/index.tsx
@@ -6,18 +6,20 @@ import styles from './index.module.css';
 export type AvatarProps = {
   src: string;
   alt: string;
+  fallback: string;
 };
 
-const Avatar: FC<AvatarProps> = ({ src, alt }) => (
+const Avatar: FC<AvatarProps> = ({ src, alt, fallback }) => (
   <RadixAvatar.Root className={styles.avatarRoot}>
     <RadixAvatar.Image
       loading="lazy"
       src={src}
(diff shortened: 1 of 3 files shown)


JakobJingleheimer (approved):
🙌

mikeesto (approved):
LGTM. I probably would have put `fallback: getAcronymFromString(username),` into `withMetaBar.tsx` but that's a small nit

AugustinMauroy:
> LGTM. I probably would have put `fallback: getAcronymFromString(username),` into `withMetaBar.tsx` but that's a small nit

It depends on how you look at `AvatarGroup`. I see it as any image with an alt which is the person and the acronym is part of the component itself. But the image ‘search’ logic isn't supposed to be in `AvatarGroup`.

PR #7123, opened 2024-10-16
feat: add streams guide

Ceres6:
## Description

This PR adds a stream guide for the learn section
## Validation

Lint passing and checked visually
## Related Issues

Closes nodejs/node#8646
### Check List


Ceres6:
I'm missing translation for the titles as I didn't want to use translator ones

package.json
```diff
@@ -30,7 +30,7 @@
     "lint": "turbo run lint",
     "lint:fix": "turbo run lint --force",
     "prettier": "prettier \"**/*.{js,mjs,ts,tsx,md,mdx,json,yml,css}\" --check --cache --cache-strategy=content --cache-location=.prettiercache",
-    "prettier:fix": "npm run prettier --write",
+    "prettier:fix": "npm run prettier -- --write",
```
AugustinMauroy: Why ?

Ceres6: I tried without it and it didn't write. I'm happy to remove it but I think it was broken

AugustinMauroy:
> I'm missing translation for the titles as I didn't want to use translator ones

You mustn't touch to translation fille it's handle with crowdin

Ceres6:
> > I'm missing translation for the titles as I didn't want to use translator ones
> 
> You mustn't touch to translation fille it's handle with crowdin

So should I remove the `en.json` change?

AugustinMauroy:
> > > I'm missing translation for the titles as I didn't want to use translator ones
> > 
> > 
> > You mustn't touch to translation fille it's handle with crowdin
> 
> So should I remove the `en.json` change?

I mean you just have to update `en.json` and don't touch to translated file

Ceres6:
cc @mcollina @Trott

mcollina:
Can you please add a sentence at the beginning or end saying that this guide is a derivative of https://blog.platformatic.dev/a-guide-to-reading-and-writing-nodejs-streams?

simoneb:
@mcollina do we want any corporate references to material that lands on the Node.js website? Asking because, in a similar vein, the article was reviewed by multiple people at Nearform, including myself, and we weren't going to ask @Ceres6 to attribute us. 

On the other hand, we were considering to publish the very same article on our company's blog, with a reference to the official documentation, as the authoring has in practice being supported by us.

mcollina:
> @mcollina do we want any corporate references to material that lands on the Node.js website? Asking because, in a similar vein, the article was reviewed by multiple people at Nearform, including myself, and we weren't going to ask @Ceres6 to attribute us.

@simoneb You can see there are _very few original contributions_ to what has been added here compared to what [I wrote in the article](https://blog.platformatic.dev/a-guide-to-reading-and-writing-nodejs-streams), as multiple paragraphs were taken verbatim, including the whole of the introduction.

By all intents and purposes, I'm the primary author of this PR, and I [gave permission](https://github.com/nodejs/node/issues/8646#issuecomment-2298245903) to use my original piece if I was included as a [co-author of it](https://github.com/nodejs/nodejs.org/pull/7123/commits/55ff4311a33db49cbc00aa30e679f6e27038bfc9) and kindly asked for a backlink if possible.

Adding a backlink is not unreasonable to ask. It's also ok if it's not added, but I would prefer it.

---

> do we want any corporate references to material that lands on the Node.js website? 

I think the question for the @nodejs/tsc is: 

> Should we add a backlink in case existing material is used in the Learn section of the website?

(We should not be using existing material without the author permission anyway).

mcollina:
Here is a proposal, how about you include in the commit description:

1. a link of the original piece
2. a mention of all people at NearForm that reviewed it

So we keep a record of the origin of this content.

Ceres6:
@mcollina I'm happy to add the backlink. I'm guessing now the TSC is tagged we should wait until that gets discussed, right?

EDIT: I saw that some resources in the learn section have an authors frontmatter prop, maybe that's another option to consider?

mcollina:
I think it's good if the TSC discuss this because I suspect it would come out more and more.

@Ceres6 you should definitely fill in the `authors` block in the frontmatter. It should appear something like:

Screenshot 2024-10-25 at 17 59 53

Ceres6:
Cool! I'll add authors then and wait for the TSC discussion on the backlink. Just one doubt, should I add Nearform reviewers as authors or are those not considered as such? @mcollina

mcollina:
I would list them all.

ovflowd:
Hey folks 👋 are we happy here?

Ceres6:
Hey @ovflowd, I think we're waiting on the next TSC meeting for the backlink issue.

mcollina:
@ovflowd do you know if we have a policy for content that was published elsewhere and brought in? That's what is missing in this case.

ovflowd:
> @ovflowd do you know if we have a policy for content that was published elsewhere and brought in? That's what is missing in this case.

The only content-related policy we have is this one https://github.com/nodejs/nodejs.org/blob/main/CONTENT_VS_CODE.md

Nothing else, such as content sourced from different places.

Ceres6:
@mcollina Should I add the backlink so we can merge this and we can revisit it if the tsc votes otherwise?

mhdawson:
My opinion is that we should include something like "This content was originally shared by Y in X"

mhdawson:
I think we should also document in the collaborator guide for the Node.js website that we should include references when pulling in content from other sources. That process should also be to make sure we document that the original author is ok with the content being pulled in, possibly through approving the PR that pulls it in.

apps/site/pages/en/learn/modules/how-to-use-streams.md
```diff
@@ -0,0 +1,595 @@
+---
+title: How to use Streams
+layout: learn
```
Ceres6, line 3:
```suggestion
layout: learn
authors: mcollina, ceres6, simoneb, codyzu
```

apps/site/pages/en/learn/modules/how-to-use-streams.md
```diff
@@ -0,0 +1,595 @@
...
+    for await (const chunk of source) {
+      yield chunk.toString().toUpperCase();
+    }
+  },
+  process.stdout
+);
+```
+
+Be aware that fetch body is a `ReadableStream<Uint8Array>` and therefore a [`TextDecoderStream`][] is needed to work with chunks as strings.
+
```
Ceres6, line 835:
```suggestion

This work is derived from the content published by [Matteo Collina][] in [Platformatic's Blog][]
```

apps/site/pages/en/learn/modules/how-to-use-streams.md
```diff
@@ -0,0 +1,595 @@
...
+[`WritableStream`]: https://nodejs.org/api/webstreams.html#class-writablestream
+[`TransformStream`]: https://nodejs.org/api/webstreams.html#class-transformstream
+[`WHATWG Streams Standard`]: https://streams.spec.whatwg.org/
+[`backpressure guide`]: /learn/modules/backpressuring-in-streams
+[`fs.readStream`]: https://nodejs.org/api/fs.html#class-fsreadstream
+[`http.IncomingMessage`]: https://nodejs.org/api/http.html#class-httpincomingmessage
+[`process.stdin`]: https://nodejs.org/api/process.html#processstdin
+[`fs.WriteStream`]: https://nodejs.org/api/fs.html#class-fswritestream
+[`process.stdout`]: https://nodejs.org/api/process.html#processstdout
+[`process.stderr`]: https://nodejs.org/api/process.html#processstderr
```
Ceres6, line 876:
```suggestion
[`process.stderr`]: https://nodejs.org/api/process.html#processstderr
[Matteo Collina]: https://github.com/mcollina
[Platformatic's Blog]: https://blog.platformatic.dev/a-guide-to-reading-and-writing-nodejs-streams
```

Ceres6:
Authors and original attribution added. Are we good to go, @mcollina?

apps/site/navigation.json
```diff
@@ -323,6 +323,10 @@
             "backpressuringInStreams": {
               "link": "/learn/modules/backpressuring-in-streams",
               "label": "components.navigation.learn.modules.links.backpressuringInStreams"
+            },
+            "howToUseStreams": {
```
ovflowd, line 328: This learn page should be before backpressure in streams IMO.

ovflowd (approved):
Content-wise (first glance) it looks great. @nodejs/tsc I see this is still in the TSC agenda. Any blocker here?

apps/site/pages/en/learn/modules/how-to-use-streams.md
```diff
@@ -0,0 +1,599 @@
...
+
+This section is just a reference of the history of streams in Node.js. Unless you’re working with a codebase written for a Node.js version prior to 0.11.5 (2013), you will rarely encounter older versions of the streams API, but the terms might still be in use.
+
+### Streams 0
+
+The first version of streams was released at the same time as Node.js. Although there wasn't a Stream class yet, different modules used the concept and implemented the `read`/`write` functions. `util.pump()` function was available to control the flow of data between streams.
+
+### Streams 1 (Classic)
+
+With the release of Node v0.4.0 in 2011 the Stream class was introduced as well as the `pipe()` method.
```
avivkeller:
```suggestion
With the release of Node v0.4.0 in 2011, the Stream class was introduced, as well as the `pipe()` method.
```

apps/site/pages/en/learn/modules/how-to-use-streams.md
```diff
@@ -0,0 +1,599 @@
...
+authors: mcollina, ceres6, simoneb, codyzu
+---
+
+# How To Use Streams
+
+Working with large amounts of data in Node.js applications can be a double-edged sword. The ability to handle massive amounts of data is extremely handy, but can also lead to performance bottlenecks and memory exhaustion. Traditionally, developers tackled this challenge by reading the entire dataset into memory at once. This approach, while intuitive for smaller datasets, becomes inefficient and resource-intensive for large data (e.g. files, network requests…).
+
+This is where Node.js streams come in. Streams offer a fundamentally different approach, allowing you to process data incrementally and optimize memory usage. By handling data in manageable chunks, streams empower you to build scalable applications that can efficiently tackle even the most daunting datasets. As popularly quoted, “streams are arrays over time”.
+
+In this guide we give an overview of the Stream concept, history and API as well as some recommendations on how to use and operate them.
```
avivkeller:
```suggestion
Working with large amounts of data in Node.js applications can be a double-edged sword. The ability to handle massive amounts of data is extremely handy but can also lead to performance bottlenecks and memory exhaustion. Traditionally, developers tackled this challenge by reading the entire dataset into memory at once. This approach, while intuitive for smaller datasets, becomes inefficient and resource-intensive for large data (e.g., files, network requests…).

This is where Node.js streams come in. Streams offer a fundamentally different approach, allowing you to process data incrementally and optimize memory usage. By handling data in manageable chunks, streams empower you to build scalable applications that can efficiently tackle even the most daunting datasets. As popularly quoted, “streams are arrays over time.”

In this guide, we give an overview of the Stream concept, history, and API as well as some recommendations on how to use and operate them.
```

apps/site/pages/en/learn/modules/how-to-use-streams.md
```diff
@@ -0,0 +1,599 @@
...
+
+## What are Node.js Streams?
+
+Node.js streams offer a powerful abstraction for managing data flow in your applications. They excel at processing large datasets like reading or writing from files and network requests without compromising performance.
+
+This approach differs from loading the entire dataset into memory at once. Streams process data in chunks, significantly reducing memory usage. All streams in Node.js inherit from the [`EventEmitter`][] class, allowing them to emit events at various stages of data processing. These streams can be readable, writable, or both, providing flexibility for different data handling scenarios.
```
avivkeller:
```suggestion
Node.js streams offer a powerful abstraction for managing data flow in your applications. They excel at processing large datasets, such as reading or writing from files and network requests, without compromising performance.

This approach differs from loading the entire dataset into memory at once. Streams process data in chunks, significantly reducing memory usage. All streams in Node.js inherit from the [`EventEmitter`][] class, allowing them to emit events at various stages of data processing. These streams can be readable, writable, or both, providing flexibility for different data-handling scenarios.
```

apps/site/pages/en/learn/modules/how-to-use-streams.md
```diff
@@ -0,0 +1,599 @@
...
+
+## Why use Streams?
+
+Streams provide three key advantages over other data-handling methods:
+
+- **Memory Efficiency**: Streams process data incrementally, consuming and processing data in chunks rather than loading the entire dataset into memory. This is a major advantage when dealing with large datasets, as it significantly reduces memory usage and prevents memory-related performance issues.
+- **Improved response time**: Streams allow for immediate data processing. When a chunk of data arrives, it can be processed without waiting for the entire payload or dataset to be received. This reduces latency and improves your application's overall responsiveness.
+- **Scalability for real-time processing**: By handling data in chunks, Node.js streams can efficiently handle large amounts of data with limited resources. This scalability makes streams ideal for applications that process high volumes of data in real time.
```
avivkeller:
```suggestion
- **Memory Efficiency**: Streams process data incrementally, consuming and processing data in chunks rather than loading the entire dataset into memory. This is a major advantage when dealing with large datasets, as it significantly reduces memory usage and prevents memory-related performance issues.
- **Improved Response Time**: Streams allow for immediate data processing. When a chunk of data arrives, it can be processed without waiting for the entire payload or dataset to be received. This reduces latency and improves your application's overall responsiveness.
- **Scalability for Real-Time Processing**: By handling data in chunks, Node.js streams can efficiently handle large amounts of data with limited resources. This scalability makes streams ideal for applications that process high volumes of data in real time.
```

apps/site/pages/en/learn/modules/how-to-use-streams.md
```diff
@@ -0,0 +1,599 @@
...
+
+These advantages make streams a powerful tool for building high-performance, scalable Node.js applications, particularly when working with large datasets or real-time data processing.
+
+### Note on performance
+
+If your application already has all the data readily available in memory, using streams might add unnecessary overhead, complexity and slow down your application.
```
avivkeller:
```suggestion
If your application already has all the data readily available in memory, using streams might add unnecessary overhead, complexity, and slow down your application.
```

apps/site/pages/en/learn/modules/how-to-use-streams.md
```diff
@@ -0,0 +1,599 @@
...
+
+## Stream history
+
+This section is just a reference of the history of streams in Node.js. Unless you’re working with a codebase written for a Node.js version prior to 0.11.5 (2013), you will rarely encounter older versions of the streams API, but the terms might still be in use.
```
avivkeller:
```suggestion
This section is a reference of the history of streams in Node.js. Unless you’re working with a codebase written for a Node.js version prior to 0.11.5 (2013), you will rarely encounter older versions of the streams API, but the terms might still be in use.
```

apps/site/pages/en/learn/modules/how-to-use-streams.md
```diff
@@ -0,0 +1,599 @@
...
+
+### Streams 0
+
+The first version of streams was released at the same time as Node.js. Although there wasn't a Stream class yet, different modules used the concept and implemented the `read`/`write` functions. `util.pump()` function was available to control the flow of data between streams.
```
avivkeller:
```suggestion
The first version of streams was released at the same time as Node.js. Although there wasn't a Stream class yet, different modules used the concept and implemented the `read`/`write` functions. The `util.pump()` function was available to control the flow of data between streams.
```

apps/site/pages/en/learn/modules/how-to-use-streams.md
```diff
@@ -0,0 +1,599 @@
...
+
+With the release of Node v0.4.0 in 2011 the Stream class was introduced as well as the `pipe()` method.
+
+### Streams 2
+
+In 2012, with the release of Node v0.10.0, Streams 2 were unveiled. This update brought new stream subclasses, including Readable, Writable, Duplex, and Transform. Additionally, the `readable` event was added. To maintain backwards compatibility, streams could be switched to the old mode by adding a `data` event listener or calling `pause()` or `resume()` methods.
+
+### Streams 3
+
+In 2013, Streams 3 were released with Node v0.11.5, to address the problem of a stream having both a `data` and `readable` event handlers. This removed the need to choose between 'current' and 'old' modes. Streams3 is the current version of streams in Node.js.
```
avivkeller:
```suggestion
In 2013, Streams 3 were released with Node v0.11.5, to address the problem of a stream having both a `data` and `readable` event handlers. This removed the need to choose between 'current' and 'old' modes. Streams 3 is the current version of streams in Node.js.
```

apps/site/pages/en/learn/modules/how-to-use-streams.md
```diff
@@ -0,0 +1,599 @@
...
+
+## Stream types
+
+### Readable
+
+[`Readable`][] is the class that we use to sequentially read a source of data. Typical examples of `Readable` streams in Node.js API are [`fs.ReadStream` ][] when reading files, [`http.IncomingMessage` ][] when reading HTTP requests, and [`process.stdin` ][] when reading from the standard input
```
avivkeller:
```suggestion
[`Readable`][] is the class that we use to sequentially read a source of data. Typical examples of `Readable` streams in Node.js API are [`fs.ReadStream` ][] when reading files, [`http.IncomingMessage` ][] when reading HTTP requests, and [`process.stdin` ][] when reading from the standard input.
```

apps/site/pages/en/learn/modules/how-to-use-streams.md
```diff
@@ -0,0 +1,599 @@
...
+
+A readable stream operates with several core methods and events that allow fine control over data handling:
+
+- **[`on('data')`][]**: This event is triggered whenever data is available from the stream. It is very fast, as the stream pushes data as quickly as it can handle, making it suitable for high-throughput scenarios.
+- **[`on('end')`][]**: Emitted when there is no more data to read from the stream. It signifies the completion of data delivery. This event is only fired when all the data from the stream has been consumed.
+- **[`on('readable')`][]**: This event is triggered when there is data available to read from the stream or when the end of the stream has been reached. It allows for more controlled data reading when needed.
+- **[`on('close')`][]**: This event is emitted when the stream and its underlying resources have been closed and indicates no more events will be emitted.
+- **[`on('error')`][]**: This event can be emitted at any point signalling that there was an error processing. A handler of this event can be used to avoid uncaught exceptions
+
+A demonstration of the use of this events can be seen in the following sections
```
avivkeller:
```suggestion
A readable stream operates with several core methods and events that allow fine control over data handling:

- **[`on('data')`][]**: This event is triggered whenever data is available from the stream. It is very fast, as the stream pushes data as quickly as it can handle, making it suitable for high-throughput scenarios.
- **[`on('end')`][]**: Emitted when there is no more data to read from the stream. It signifies the completion of data delivery. This event is only fired when all the data from the stream has been consumed.
- **[`on('readable')`][]**: This event is triggered when there is data available to read from the stream or when the end of the stream has been reached. It allows for more controlled data reading when needed.
- **[`on('close')`][]**: This event is emitted when the stream and its underlying resources have been closed and indicates that no more events will be emitted.
- **[`on('error')`][]**: This event can be emitted at any point, signaling that there was an error processing. A handler for this event can be used to avoid uncaught exceptions.

A demonstration of the use of these events can be seen in the following sections.
```

apps/site/pages/en/learn/modules/how-to-use-streams.md
```diff
@@ -0,0 +1,599 @@
...
+  while ((chunk = stream.read()) !== null) {
+    console.log(chunk.toString()); // Process the chunk
+  }
+});
+stream.on('end', () => console.log('>> end event'));
+```
+
+Here, the readable event is used to pull data from the stream as needed manually. The loop inside the readable event handler continues to read data from the stream buffer until it returns `null`, indicating that the buffer is temporarily empty or the stream has ended. Setting `highWaterMark` to 1 keeps the buffer size small, triggering the readable event more frequently and allowing more granular control over data flow.
+
+With the previous code you’ll get an output like
```
avivkeller:
```suggestion
Here, the readable event is used to pull data from the stream as needed manually. The loop inside the readable event handler continues to read data from the stream buffer until it returns `null`, indicating that the buffer is temporarily empty or the stream has ended. Setting `highWaterMark` to 1 keeps the buffer size small, triggering the readable event more frequently and allowing more granular control over the data flow.

With the previous code, you’ll get an output like
```

apps/site/pages/en/learn/modules/how-to-use-streams.md
```diff
@@ -0,0 +1,599 @@
...
+>> readable event: 3
+>> readable event: 4
+>> end event
+```
+
+Let’s try to digest that. When we attach the `on('readable')` event it does a first call to `read()` because that is what might triggers the emission of a `readable` event. After the emission of said event we call `read` on the first iteration of the `while` loop, that’s why we get the two first smileys in one row. After that we keep calling `read` until `null` is pushed. Each call to read programs the emission of a new `readable` event, but as we are in “flow” mode (i.e. using the `readable` event) the emission is programmed on `nextTick` that’s why we get them all at the end, when the synchronous code of the loop is finished.
+
+NOTE: You can try to run the code with `NODE_DEBUG=stream` to see the that `emitReadable` is triggered after each `push`.
+
+If we want to see readable events called before each smiley we can wrap `push` into a `setImmediate` or `process.nextTick` like this:
```
avivkeller:
```suggestion
Let’s try to digest that. When we attach the `on('readable')` event, it makes a first call to `read()` because that is what might trigger the emission of a `readable` event. After the emission of said event, we call `read` on the first iteration of the `while` loop. That’s why we get the first two smileys in one row. After that, we keep calling `read` until `null` is pushed. Each call to `read` programs the emission of a new `readable` event, but as we are in “flow” mode (i.e., using the `readable` event), the emission is scheduled for the `nextTick`. That’s why we get them all at the end, when the synchronous code of the loop is finished.

NOTE: You can try to run the code with `NODE_DEBUG=stream` to see that `emitReadable` is triggered after each `push`.

If we want to see readable events called before each smiley, we can wrap `push` into a `setImmediate` or `process.nextTick` like this:
```

apps/site/pages/en/learn/modules/how-to-use-streams.md
```diff
@@ -0,0 +1,599 @@
...
+:-)
+>> readable event: 5
+:-)
+>> readable event: 6
+>> end event
+```
+
+### Writable
+
+[`Writable`][] streams are useful for creating files, uploading data, or any task that involves sequentially outputting data. While readable streams provide the source of data, writable streams in Node.js act as the destination for your data. Typical examples of writable streams in Node.js API are [`fs.WriteStream` ][], [`process.stdout` ][] and [`process.stderr` ][]
```
avivkeller:
```suggestion
[`Writable`][] streams are useful for creating files, uploading data, or any task that involves sequentially outputting data. While readable streams provide the source of data, writable streams in Node.js act as the destination for your data. Typical examples of writable streams in the Node.js API are [`fs.WriteStream` ][], [`process.stdout` ][], and [`process.stderr` ][].
```

apps/site/pages/en/learn/modules/how-to-use-streams.md
```diff
@@ -0,0 +1,599 @@
...
+
+### Duplex
+
+[`Duplex`][] streams implement both the readable and writable interfaces.
+
+#### Key Methods and Events in Duplex Streams
+
+Duplex streams implement all the methods and events described in Readable and Writable Streams.
+
+A good example of duplex stream is the `Socket` class in the `net` module
```
avivkeller:
```suggestion
Duplex streams implement all the methods and events described in Readable and Writable Streams.

A good example of a duplex stream is the `Socket` class in the `net` module:
```

apps/site/pages/en/learn/modules/how-to-use-streams.md
```diff
@@ -0,0 +1,599 @@
...
+  });
+});
+
+// Start the server on port 8080
+server.listen(8080, () => {
+  console.log('Server listening on port 8080');
+});
+```
+
+The previous code will open a TCP socket in port 8080 and send `Hello from server!` to any connecting client and log any data received
```
avivkeller:
```suggestion
The previous code will open a TCP socket on port 8080, send `Hello from server!` to any connecting client, and log any data received.
```

apps/site/pages/en/learn/modules/how-to-use-streams.md
```diff
@@ -0,0 +1,599 @@
...
+  console.log(`Server says: ${data.toString()}`);
+});
+
+// Handle the server closing the connection
+client.on('end', () => {
+  console.log('Disconnected from server');
+});
+```
+
+The previous code will connect to the TCP socket and send a `Hello from client` and log any received data.
```
avivkeller:
```suggestion
The previous code will connect to the TCP socket, send a `Hello from client` message, and log any received data.
```

apps/site/pages/en/learn/modules/how-to-use-streams.md
```diff
@@ -0,0 +1,599 @@
...
+
+### Transform
+
+[`Transform`][] streams are duplex streams where the output is computed based on the input. As the name suggests, they are usually used between a readable and a writable stream to transform the data as it passes through.
```
avivkeller:
```suggestion
[`Transform`][] streams are duplex streams, where the output is computed based on the input. As the name suggests, they are usually used between a readable and a writable stream to transform the data as it passes through.
```

apps/site/pages/en/learn/modules/how-to-use-streams.md
```diff
@@ -0,0 +1,599 @@
...
+
+#### Key Methods and Events in Transform Streams
+
+Apart from all the methods and events in Duplex Streams there is
+
+- **[`_transform`][]**: This function is called internally to handle the flow of data between the readable and the writable parts. This MUST NOT be called by application code.
```
avivkeller:
```suggestion
Apart from all the methods and events in Duplex Streams, there is:

- **[`_transform`][]**: This function is called internally to handle the flow of data between the readable and writable parts. This MUST NOT be called by application code.
```

apps/site/pages/en/learn/modules/how-to-use-streams.md
```diff
@@ -0,0 +1,599 @@
...
+
+#### Creating a Transform Stream
+
+To create a new transform stream we can pass an `options` object to the `Transform` constructor including a `transform` function which handles how the output data is computed from the input data using the `push` method.
```
avivkeller:
```suggestion
To create a new transform stream, we can pass an `options` object to the `Transform` constructor, including a `transform` function that handles how the output data is computed from the input data using the `push` method.
```

apps/site/pages/en/learn/modules/how-to-use-streams.md
```diff
@@ -0,0 +1,599 @@
...
+    cb();
+  },
+});
+```
+
+This stream will take any input and output it in uppercase.
+
+## How to operate with streams
+
+When working with streams, we usually want to perform read from a source and write to a destination, possibly needing some transformation of the data in between. The following sections will cover different ways to do so.
```
avivkeller:
```suggestion
When working with streams, we usually want to read from a source and write to a destination, possibly needing some transformation of the data in between. The following sections will cover different ways to do so.
```

apps/site/pages/en/learn/modules/how-to-use-streams.md
```diff
@@ -0,0 +1,599 @@
...
+
+### .pipe()
```
avivkeller:
```suggestion
### `.pipe()`
```

apps/site/pages/en/learn/modules/how-to-use-streams.md
```diff
@@ -0,0 +1,599 @@
...
+
+The [`.pipe()`][] method concatenates one readable stream to a writable (or transform) stream. Although this seems a simple way to achieve our goal, it delegates all error handling to the programmer making it difficult to get it right.
+
+The following example shows a pipe trying to output the current file in upper case to the console.
```
avivkeller:
```suggestion
The [`.pipe()`][] method concatenates one readable stream to a writable (or transform) stream. Although this seems like a simple way to achieve our goal, it delegates all error handling to the programmer, making it difficult to get it right.

The following example shows a pipe trying to output the current file in uppercase to the console.
```

apps/site/pages/en/learn/modules/how-to-use-streams.md
```diff
@@ -0,0 +1,599 @@
...
+upper.on('error', err => {
+  console.error('\nError in transform stream:', err.message);
+});
+
+writeStream.on('close', () => {
+  console.log('Writable stream closed');
+});
+```
+
+After writing 10 characters `upper` will return an error in the callback, which will cause the stream to close. However, the other streams won’t be notified resulting in memory leaks. The output will be:
```
avivkeller:
```suggestion
After writing 10 characters, `upper` will return an error in the callback, which will cause the stream to close. However, the other streams won’t be notified, resulting in memory leaks. The output will be:
```

apps/site/pages/en/learn/modules/how-to-use-streams.md
```diff
@@ -0,0 +1,599 @@
...
+
+```bash
+CONST FS =
+Error in transform stream: BOOM!
+Transform stream closed
+```
+
+### pipeline()
```
avivkeller:
```suggestion
### `pipeline()`
```

apps/site/pages/en/learn/modules/how-to-use-streams.md
```diff
@@ -0,0 +1,599 @@
...
+
+To avoid the pitfalls low level complexity of the `.pipe()` method, in most cases it is recommended to use [`pipeline()`][] method. This method is a safer and more robust way to pipe streams together, handling errors and cleanup automatically.
+
+The following example demonstrates how using`pipeline()`prevents the pitfalls of the previous example:
```
avivkeller:
```suggestion
To avoid the pitfalls and low-level complexity of the `.pipe()` method, in most cases, it is recommended to use the [`pipeline()`][] method. This method is a safer and more robust way to pipe streams together, handling errors and cleanup automatically.

The following example demonstrates how using `pipeline()` prevents the pitfalls of the previous example:
```

apps/site/pages/en/learn/modules/how-to-use-streams.md
```diff
@@ -0,0 +1,599 @@
...
+
+pipeline(readStream, upper, writeStream, err => {
+  if (err) {
+    return console.error('Pipeline error:', err.message);
+  }
+  console.log('Pipeline succeeded');
+});
+```
+
+In this case all streams will be closed with the following output:
```
avivkeller:
```suggestion
In this case, all streams will be closed with the following output:
```

apps/site/pages/en/learn/modules/how-to-use-streams.md
```diff
@@ -0,0 +1,599 @@
...
+
+```bash
+CONST FS =
+Transform stream closed
+Writable stream closed
+Pipeline error: BOOM!
+Readable stream closed
+```
+
+The [`pipeline()`][] method also has an [`async pipeline()`][] version which doesn’t accept a callback but instead returns a promise which is rejected if the pipeline fails.
```
avivkeller:
```suggestion
The [`pipeline()`][] method also has an [`async pipeline()`][] version, which doesn’t accept a callback but instead returns a promise that is rejected if the pipeline fails.
```

apps/site/pages/en/learn/modules/how-to-use-streams.md
```diff
@@ -0,0 +1,599 @@
...
+
+### Async Iterators
+
+Async Iterators are recommended as the standard way of interfacing with the Streams API. Compared to all the streams primitives on the Web and Node.js, Async iterators are easier to understand and to use, contributing to fewer bugs and more maintainable code. In recent versions of Node.js, async iterators have emerged as a more elegant and readable way to interact with streams. Building upon the foundation of events, async iterators provide a higher-level abstraction that simplifies stream consumption.
+
+In Node.js, all readable streams are asynchronous iterables. This means you can use the for await...of syntax to loop through the stream's data as it becomes available, handling each piece of data with the efficiency and simplicity of asynchronous code.
```
avivkeller:
```suggestion
Async iterators are recommended as the standard way of interfacing with the Streams API. Compared to all the stream primitives in both the Web and Node.js, async iterators are easier to understand and use, contributing to fewer bugs and more maintainable code. In recent versions of Node.js, async iterators have emerged as a more elegant and readable way to interact with streams. Building upon the foundation of events, async iterators provide a higher-level abstraction that simplifies stream consumption.

In Node.js, all readable streams are asynchronous iterables. This means you can use the `for await...of` syntax to loop through the stream's data as it becomes available, handling each piece of data with the efficiency and simplicity of asynchronous code.
```

apps/site/pages/en/learn/modules/how-to-use-streams.md
```diff
@@ -0,0 +1,599 @@
...
+  async function* (source) {
+    for await (let chunk of source) {
+      yield chunk.toString().toUpperCase();
+    }
+  },
+  process.stdout
+);
+```
+
+This code achieves the same as the previous examples, without the need to define a new transform stream. The error from previous examples has been removed for the sake of brevity. The async version of pipeline has been used. It should be wrapped in a `try..catch` block to handle possible errors.
```
avivkeller:
```suggestion
This code achieves the same result as the previous examples, without the need to define a new transform stream. The error from the previous examples has been removed for the sake of brevity. The async version of the pipeline has been used, and it should be wrapped in a `try...catch` block to handle possible errors.
```

apps/site/pages/en/learn/modules/how-to-use-streams.md
```diff
@@ -0,0 +1,599 @@
...
+
+### Object mode
+
+By default streams can work with strings, [`Buffer`], [`TypedArray`], or [`DataView`], if an arbitrary value different than those (e.g. an object) is pushed into a stream a `TypeError` will be thrown. However, it is possible to work with objects by setting the `objectMode` option to `true`. This allows the stream to work with any JavaScript value, with the exception of `null`, which is used to signal the end of the stream. This means you can `push` and `read` any value in a readable stream and `write` any value in a writable stream.
```
avivkeller:
```suggestion
By default, streams can work with strings, [`Buffer`][], [`TypedArray`][], or [`DataView`][]. If an arbitrary value different from these (e.g., an object) is pushed into a stream, a `TypeError` will be thrown. However, it is possible to work with objects by setting the `objectMode` option to `true`. This allows the stream to work with any JavaScript value, except for `null`, which is used to signal the end of the stream. This means you can `push` and `read` any value in a readable stream, and `write` any value in a writable stream.
```

apps/site/pages/en/learn/modules/how-to-use-streams.md
```diff
@@ -0,0 +1,599 @@
...
+const readable = Readable({
+  objectMode: true,
+  read() {
+    this.push({ hello: 'world' });
+    this.push(null);
+  },
+});
+```
+
+When working in object mode it is important to remember that the `highWaterMark` option is the number of objects, not bytes.
```
avivkeller:
```suggestion
When working in object mode, it is important to remember that the `highWaterMark` option refers to the number of objects, not bytes.
```

apps/site/pages/en/learn/modules/how-to-use-streams.md
```diff
@@ -0,0 +1,599 @@
...
+
+### Backpressure
+
+When using streams it is important to make sure the producer doesn't overwhelm the consumer. For this the backpressure mechanism is used in all streams in the Node.js API and implementors are responsible to keep that behaviour.
+
+In any scenario where the data buffer has exceeded the [`highWaterMark`](https://nodejs.org/api/stream.html#stream_buffering) or the write queue is currently busy, [`.write()`](https://nodejs.org/api/stream.html#stream_writable_write_chunk_encoding_callback) will return `false`.
+
+When a `false` value is returned, the backpressure system kicks in. It will pause the incoming [`Readable`](https://nodejs.org/api/stream.html#stream_readable_streams) stream from sending any data and wait until the consumer is ready again. Once the data buffer is emptied, a [`'drain'`](https://nodejs.org/api/stream.html#stream_event_drain) event will be emitted and resume the incoming data flow.
+
+For a deeper understanding of backpressure, check the [`backpressure guide`][].
```
avivkeller:
```suggestion
When using streams, it is important to make sure the producer doesn't overwhelm the consumer. For this, the backpressure mechanism is used in all streams in the Node.js API, and implementors are responsible for maintaining that behavior.

In any scenario where the data buffer has exceeded the [`highWaterMark`](https://nodejs.org/api/stream.html#stream_buffering) or the write queue is currently busy, [`write()`](https://nodejs.org/api/stream.html#stream_writable_write_chunk_encoding_callback) will return `false`.

When a `false` value is returned, the backpressure system kicks in. It will pause the incoming [`Readable`](https://nodejs.org/api/stream.html#stream_readable_streams) stream from sending any data and wait until the consumer is ready again. Once the data buffer is emptied, a [`'drain'`](https://nodejs.org/api/stream.html#stream_event_drain) event will be emitted to resume the incoming data flow.

For a deeper understanding of backpressure, check the [`backpressure guide`][].
```

apps/site/pages/en/learn/modules/how-to-use-streams.md
```diff
@@ -0,0 +1,599 @@
...
+
+## Streams vs Web streams
+
+The stream concept is not exclusive to Node.js. In fact Node.js has a different implementation of the stream concept called [`Web Streams`][]. That implements the [`WHATWG Streams Standard`][]. Although the concepts behind them are similar, it is important to be aware that they have different APIs and are not directly compatible.
+
+[`Web Streams`][] implement [`ReadableStream`][], [`WritableStream`][] and [`TransformStream`][] classes, homologous to Node.js [`Readable`][], [`Writable`][] and [`Transform`][] streams.
```
avivkeller:
```suggestion
The stream concept is not exclusive to Node.js. In fact, Node.js has a different implementation of the stream concept called [`Web Streams`][], which implements the [`WHATWG Streams Standard`][]. Although the concepts behind them are similar, it is important to be aware that they have different APIs and are not directly compatible. 

[`Web Streams`][] implement the [`ReadableStream`][], [`WritableStream`][], and [`TransformStream`][] classes, which are homologous to Node.js's [`Readable`][], [`Writable`][], and [`Transform`][] streams.
```

apps/site/pages/en/learn/modules/how-to-use-streams.md
```diff
@@ -0,0 +1,599 @@
...
+
+### Interoperability of streams and Web Streams
+
+Node.js provides utility functions to convert to/from Web Streams and Node.js streams. This functions are implemented as `toWeb` and `fromWeb` methods in each stream class.
+
+The following example in the [`Duplex`][] class covers how to work with both readable and writable streams converted to Web Streams:
```
avivkeller:
```suggestion
Node.js provides utility functions to convert to/from Web Streams and Node.js streams. These functions are implemented as `toWeb` and `fromWeb` methods in each stream class.

The following example in the [`Duplex`][] class demonstrates how to work with both readable and writable streams converted to Web Streams:
```

apps/site/pages/en/learn/modules/how-to-use-streams.md
```diff
@@ -0,0 +1,599 @@
...
+  async function* (source) {
+    for await (const chunk of source) {
+      yield chunk.toString().toUpperCase();
+    }
+  },
+  process.stdout
+);
+```
+
+Be aware that fetch body is a `ReadableStream<Uint8Array>` and therefore a [`TextDecoderStream`][] is needed to work with chunks as strings.
```
avivkeller:
```suggestion
Be aware that the fetch body is a `ReadableStream<Uint8Array>`, and therefore a [`TextDecoderStream`][] is needed to work with chunks as strings.
```

apps/site/pages/en/learn/modules/how-to-use-streams.md
```diff
@@ -0,0 +1,599 @@
...
+
+This work is derived from the content published by [Matteo Collina][] in [Platformatic's Blog][]
```
avivkeller:
```suggestion
This work is derived from content published by [Matteo Collina][] in [Platformatic's Blog][].

```

apps/site/pages/en/learn/modules/how-to-use-streams.md
```diff
@@ -0,0 +1,599 @@
...
+[`WHATWG Streams Standard`]: https://streams.spec.whatwg.org/
+[`backpressure guide`]: /learn/modules/backpressuring-in-streams
+[`fs.readStream`]: https://nodejs.org/api/fs.html#class-fsreadstream
+[`http.IncomingMessage`]: https://nodejs.org/api/http.html#class-httpincomingmessage
+[`process.stdin`]: https://nodejs.org/api/process.html#processstdin
+[`fs.WriteStream`]: https://nodejs.org/api/fs.html#class-fswritestream
+[`process.stdout`]: https://nodejs.org/api/process.html#processstdout
+[`process.stderr`]: https://nodejs.org/api/process.html#processstderr
+[Matteo Collina]: https://github.com/mcollina
+[Platformatic's Blog]: https://blog.platformatic.dev/a-guide-to-reading-and-writing-nodejs-streams
```
avivkeller, line 878: Some links are inline, while others are like this, can you make them all uniform?

apps/site/pages/en/learn/modules/how-to-use-streams.md
```diff
@@ -0,0 +1,599 @@
...
+- **[`on('close')`][]**: This event is emitted when the stream and its underlying resources have been closed and indicates no more events will be emitted.
+- **[`on('error')`][]**: This event can be emitted at any point signalling that there was an error processing. A handler of this event can be used to avoid uncaught exceptions
+
+A demonstration of the use of this events can be seen in the following sections
+
+#### Basic Readable Stream
+
+Here's an example of a simple readable stream implementation that generates data dynamically:
+
+```cjs
```
avivkeller, line 83: Some examples are in CommonJS, while others are in ECMASCript Module Syntax, can you choose one or the other?

avivkeller (commented):
Reads well, just some minor grammar adjustments, among other things.

apps/site/pages/en/learn/modules/how-to-use-streams.md
AugustinMauroy: Note that you can use both. Two codeboxes one after the other can be put together to make a code tab.

apps/site/pages/en/learn/modules/how-to-use-streams.md
AugustinMauroy: The rule that should be applied is this. If the link is used once, we put it inline if it is used several times in the bottom

avivkeller: That's not what happening in this article. In this article, some are inline, some aren't. IMO one of the following should be done
1. All inline (preferably not)
2. All text-based references (I  like this the best)
3. Inline unless used multiple times

Ceres6: I think (almost) all of them are used only once. I like the idea of having them all at the bottom as it's like a bibliography. I haven't seen anything documented about this rule, but I'm happy to go either way as long as we agree

apps/site/pages/en/learn/modules/how-to-use-streams.md
Ceres6: Oh, yes, I intended to have both everywhere, but I was waiting until they were reviewed so I didn't have to change things twice. I'll update them tomorrow.

apps/site/pages/en/learn/modules/how-to-use-streams.md
avivkeller: IMO as you said, having them at the bottom is cleaner

avivkeller:
> Content-wise (first glance) it looks great. @nodejs/tsc I see this is still in the TSC agenda. Any blocker here?

According to the minutes from November 13th, they are still? deliberating on the policy of external content attribution.

Ceres6:
> > Content-wise (first glance) it looks great. @nodejs/tsc I see this is still in the TSC agenda. Any blocker here?
> 
> According to the minutes from November 13th, they are still? deliberating on the policy of external content attribution.

As far as I can see there were only opinions in favour of adding them, not sure if there should be a formal ruling or something

apps/site/navigation.json
Ceres6: I'll change it, I thought they needed to be in order of creation, but it makes sense to have this first as it is more basic/general

Ceres6: Done

apps/site/pages/en/learn/modules/how-to-use-streams.md
Ceres6: Done!

apps/site/pages/en/learn/modules/how-to-use-streams.md
Ceres6: I put them all at the end, hope that's okay

avivkeller:
It'll be discussed at every-ish meeting while it's on the agenda. Once they remove the label, they should come to a formal dicision.

(Also BTW this isn't the first time we've reposted external content with permission, see https://github.com/nodejs/nodejs.org/blob/83081efa4cde77626228c2709d2585cd554431af/apps/site/pages/en/blog/module/multi-server-continuous-deployment-with-fleet.md?plain=1#L9)

Ceres6:
@mcollina did you have time to review it?

apps/site/pages/en/learn/modules/how-to-use-streams.md
```diff
@@ -0,0 +1,878 @@
...
+});
+```
+
+```mjs
+import { Readable } from 'node:stream';
+
+class MyStream extends Readable {
+  #count = 0;
+  _read(size) {
+    this.push(':-)');
```
styfle, line 109: Why `:-)`? Could this be changed to something that helps the reader understand?

Also can we use `size` since this parameter is currently unused.

ovflowd: cc @Ceres6

ovflowd:
> @mcollina did you have time to review it?

Bump, @mcollina

simoneb:
Hey folks, this has been here for a couple of months, can we try to move it forward? It's just a guide to be added to the docs. 

I believe we agreed that we're going to credit @mcollina and Platformatic explicitly in the content, and myself and @codyzu in the markdown metadata.

I'm happy with that, so unless anybody disagrees, I would suggest to move this forward.

ovflowd:
> Hey folks, this has been here for a couple of months here, can we try to move it forward? It's just a guide to be added to the docs.
> 
> I believe we agreed that we're going to credit @mcollina and Platformatic explicitly in the content, and myself and @codyzu in the markdown metadata.
> 
> I'm happy with that, so unless anybody disagrees, I would suggest to move this forward.

Hey there, we can only move forward with explicit approval from the TSC. Hence we gotta wait for @mcollina

mcollina (approved):
lgtm

mcollina:
Sorry about the wait, I had a significant build-up of OSS work to crunch.

Ceres6:
So can we move this forward @ovflowd?

ovflowd:
Did you address @styfle's comment? 👀

ovflowd:
(I cannot merge until all discussions are resolved) -- hence you can either ignore his suggestion or address it 🙏

apps/site/pages/en/learn/modules/how-to-use-streams.md
Ceres6: I tries putting something different but I don't think it helps the reader, it's fairly simple to understand that we are printing the same symbol on every read operation. 

Regarding the size arg, it is not mandatory to use it, it's already documented in the docs and I think the guide is long enough to add that explanation.

Ceres6:
I answered the comment, we can always revisit, but I think it's good to go

PR #7134, opened 2024-10-22
fix: 404 on the 'package managers' link on home page

olehermanse:
## Description

[Nodejs.org](https://nodejs.org/) currently has a 404 link, when clicking on [_package managers_](https://nodejs.org/download/package-manager) (in all languages). [AFAICT](https://web.archive.org/web/20230314191504/https://nodejs.org/download/package-manager/), this link has been broken for about a year and a half.

This PR fixes the issue, by using the `Link` component, which already has the correct behavior - adding the locale at the beginning of the `href` path. The issue is fixed for all languages, linking to the correct download page, and I've added a redirect, so stored / cached URLs will also start working.

nodejs-1
nodejs-2

## Validation

Tested locally, with `npx turbo dev` and it fixes the issue locally. The new links with locale at the beginning are added and the redirect also works. I don't have the config for the nginx reverse proxy used in production, in Vercel, so I cannot verify this part, but I assume it will just work.

## Related Issues

Somewhat similar to #7001

### Check List

### DIFF
--- a/apps/site/next.mdx.use.client.mjs
+++ b/apps/site/next.mdx.use.client.mjs
@@ -20,6 +20,8 @@ export const clientMdxComponents = {
   Button: Button,
   // Links with External Arrow
   LinkWithArrow: LinkWithArrow,
+  // Regular links (without arrow)
+  Link: Link,
 };
 
 /**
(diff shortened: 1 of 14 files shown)


bmuenzenmeyer (approved):

bmuenzenmeyer:
wow - thanks so much for this @olehermanse 
what a great example of overlooking the obvious - something a maintainer on the site wouldn't regularly click on, why better testing is necessary, and the power first-time contributors have in their curiosity and agency. 🙇

bmuenzenmeyer:
due to the nature of this bug and fix, I'm proposing a fast-track and will merge when ✅ green

PR #7142, opened 2024-10-25
adds link to github profile

bmuenzenmeyer:
## Description

Authors are now GH links. currently this opens in the same tab. it could be enhanced to open elsewhere - but being a real, native link, the user has controls over that with mouse or keyboard controls

## Validation

See the deployed preview https://nodejs-org-git-7141-author-link-openjs.vercel.app/en/learn/getting-started/introduction-to-nodejs

## Related Issues

closes #7141

### Check List

- [x] I have read the [Contributing Guidelines](https://github.com/nodejs/nodejs.org/blob/main/CONTRIBUTING.md) and made commit messages that follow the guideline.
- [x] I have run `npm run format` to ensure the code follows the style guide.
- [x] I have run `npm run test` to check if all tests are passing.
- [x] I have run `npx turbo build` to check if the website builds without errors.
- [x] I've covered new added functionality with unit tests if necessary.

PR #7143, opened 2024-10-25
feat: Introducing avatar tooltip

canerakdas:
## Description

I saw it in Slack (OpenJS Foundation) [threads](https://openjs-foundation.slack.com/archives/CVAMEJ4UV/p1728220042564069) and thought it was a good way to support contributors

With this PR, the `Tooltip` component was created and minor refactors were made to `Avatar` and `AvatarGroup` components. In addition, the `WithAvatarGroup` component was created to make avatars easier to manage to simplify the process

I took care to use the elements/components in the existing design system as a design, I was inspired by (Search typing) in the Figma for the Tooltip component;

https://www.figma.com/design/pu1vZPqNIM7BePd6W8APA5/Node.js?node-id=464-6397&node-type=frame&t=dj6OGYBZLLo5cp4z-0

## Validation

- Avatars should continue to work as it is.
- A tooltip should appear for users in `authors.json` whose website field is exist

Example screenshot;
image

image

## Related Issues

Related to #7141 

### Check List


mcollina:
Does it work on mobile? I've tried from my iphone and it didn't work from the preview.

apps/site/components/Common/AvatarGroup/Avatar/index.module.css
```diff
@@ -15,9 +18,21 @@
     dark:text-neutral-300;
 }
 
-.avatarRoot {
+.container {
```
AugustinMauroy: -1 for using `container` because it's already a tailwind class

apps/site/components/Common/AvatarGroup/Overlay/index.tsx
```diff
@@ -0,0 +1,36 @@
+import { ArrowUpRightIcon } from '@heroicons/react/24/solid';
+import type { ComponentProps, FC } from 'react';
+
+import Avatar from '@/components/Common/AvatarGroup/Avatar';
+import Link from '@/components/Link';
+
+import styles from './index.module.css';
+
+export type AvatarOverlayProps = ComponentProps<typeof Avatar> & {
+  website?: string;
```
AugustinMauroy: Maybe `link` ?

AugustinMauroy (commented):
First look, it's great but maybe have separate type for avatar/ frontmatter/intermediate objects

AugustinMauroy:
After some test on desktop few point:
- For the learn section, all the authors have a GitHub username, so there's no need to go into the `authors.json` file.
- For the blog section, we'd have to try and standardise the data so that eventually it's just GitHub usernames. To reduce the complexity of things
- Like Matteo said need to found something for mobile

apps/site/components/Common/Tooltip/index.tsx
```diff
@@ -0,0 +1,43 @@
...
+  <TooltipPrimitive.Provider delayDuration={200}>
+    <TooltipPrimitive.Root>
+      <TooltipPrimitive.Trigger asChild={asChild}>
+        {children}
+      </TooltipPrimitive.Trigger>
+      <TooltipPrimitive.Portal>
+        <TooltipPrimitive.Content
+          side={side}
+          sideOffset={4}
+          className={`${styles[kind]} ${styles.content}`}
```
AugustinMauroy: may be use `classname` is used on other components

apps/site/components/Common/Tooltip/index.module.css
```diff
@@ -0,0 +1,41 @@
+.content {
+  @apply rounded-md
+    border
+    bg-white
+    text-sm
+    font-medium
+    text-neutral-900
+    shadow-lg
+    dark:bg-neutral-950
+    dark:text-white;
```
AugustinMauroy, line 10: add `m-1.5` to avoid the sticky edge of the mobile page.
Capture d’écran 2024-10-26 à 11 30 54

apps/site/components/Common/AvatarGroup/Overlay/index.tsx
AugustinMauroy, line 1: may be add story for that

apps/site/components/Common/AvatarGroup/Avatar/index.tsx
```diff
@@ -1,27 +1,40 @@
...
-      alt={alt}
-      title={alt}
+      src={image}
+      alt={name}
       className={styles.avatar}
     />
-    <RadixAvatar.Fallback delayMs={500} className={styles.avatar}>
+    <RadixAvatar.Fallback
+      delayMs={500}
+      className={`${styles.avatar} ${styles[size]}`}
```
AugustinMauroy: also here you can use `classname`

apps/site/components/Common/AvatarGroup/Avatar/index.tsx
```diff
@@ -1,27 +1,40 @@
...
 
 import styles from './index.module.css';
 
 export type AvatarProps = {
-  src: string;
-  alt: string;
-  fallback: string;
+  image?: string;
+  name?: string;
+  nickname: string;
```
AugustinMauroy, line 13: this type isn't used ?

canerakdas:
> Does it work on mobile? I've tried from my iphone and it didn't work from the preview.

This is the design choice of the radix tooltip component I use, and IMO it is the right choice (https://github.com/radix-ui/primitives/issues/955#issuecomment-960610209)

Here, as an alternative, it may be better to redirect directly when a profile image is clicked on mobile (like GitHub), but the current profile images are too small for mobile resolution, I will check the accessibility and try to implement a better solution👍

apps/site/components/Common/Tooltip/index.tsx
canerakdas: AFAIK, we are only using `classNames` if we are going to add a conditional class 🤔

apps/site/components/Common/AvatarGroup/Avatar/index.module.css
canerakdas: I totally forgot about that, thank you 🎉

AugustinMauroy:
> > Does it work on mobile? I've tried from my iphone and it didn't work from the preview.
> 
> This is the design choice of the radix tooltip component I use, and IMO it is the right choice ([radix-ui/primitives#955 (comment)](https://github.com/radix-ui/primitives/issues/955#issuecomment-960610209))
> 
> Here, as an alternative, it may be better to redirect directly when a profile image is clicked on mobile (like GitHub), but the current profile images are too small for mobile resolution, I will check the accessibility and try to implement a better solution👍

IMO, double can do the job and increase little bit the size of the avatar for mobile

apps/site/components/Common/Tooltip/index.tsx
AugustinMauroy: Oh, I didn't know. Personally I always use `classnames` when there are several `class` to avoid having a string with undifined in it.

ovflowd:
@canerakdas can you resolve the conflicts? 👀

apps/site/util/authorUtils.ts
```diff
@@ -0,0 +1,63 @@
+import authors from '@/authors.json';
```
AugustinMauroy: The problem that authors are missing from the `authors.json` comes from here, the functions don't differentiate between a frontmatter data that comes from the learn or the blog. As mentioned in the issue, the authors in the frontmatter of the learn section are valid, so we can just put their profile image and redirect to their GitHub user. But in the case of the blog it's more complex, because the frontmatter comes from the old website which just displayed the authors as such, which is why there's this intermediate json file which allows you to display an image and a correct name.

canerakdas: It made sense to me to read the people from the `authors.json` file while doing the development;

1. We can show the full name and GitHub known user name at the same time for the contributors
2. Everyone can contribute but I am not sure if every contributor would accept a redirect link to their GitHub profile 🤔 
3. It allows us to redirect from blog listing/detail pages as well

Maybe Im overthinking 😄

apps/site/components/Common/AvatarGroup/Avatar/index.tsx
canerakdas: We are now using it as an alternative to the name alt attribute of the image. I updated it because the name could be undefined

ovflowd:
FYI I think we should ditch the authors.json and use the GitHub user API to retrieve their profiles from username.

Then we just cache on the server side the resolution of the profile metadata :)

That can be within a follow-up PR btw, but would love that!

apps/site/util/authorUtils.ts
AugustinMauroy: for profile pictures it's easier GitHub. For links from a website other than GitHub, the problem is that if you have 0 hands on it, there could be content that doesn't follow our code of conduct.
What's more, if the author(s) don't want to be linked on the page, they won't put themselves in the frontmatter.

AugustinMauroy:
> FYI I think we should ditch the authors.json and use the GitHub user API to retrieve their profiles from username.
> 
> Then we just cache on the server side the resolution of the profile metadata :)
> 
> That can be within a follow-up PR btw, but would love that!

noted that there are authors who are not GitHub users ‘nodejs project ...’ those can remain in a constant

ovflowd:
> > FYI I think we should ditch the authors.json and use the GitHub user API to retrieve their profiles from username.
> 
> > 
> 
> > Then we just cache on the server side the resolution of the profile metadata :)
> 
> > 
> 
> > That can be within a follow-up PR btw, but would love that!
> 
> 
> 
> noted that there are authors who are not GitHub users ‘nodejs project ...’ those can remain in a constant 

True, authors.json can be a fallback file 🤷

canerakdas:
> > > FYI I think we should ditch the authors.json and use the GitHub user API to retrieve their profiles from username.
> > 
> > 
> > > 
> > 
> > 
> > > Then we just cache on the server side the resolution of the profile metadata :)
> > 
> > 
> > > 
> > 
> > 
> > > That can be within a follow-up PR btw, but would love that!
> > 
> > 
> > noted that there are authors who are not GitHub users ‘nodejs project ...’ those can remain in a constant
> 
> True, authors.json can be a fallback file 🤷

Sounds great! I'd love to add it, but I think this PR has already affected a lot of places. I'm in favor of creating a follow-up PR to this PR and working on it 🙇

apps/site/util/authorUtils.ts
canerakdas: Makes sense, thanks! I added GitHub redirect links by default for authors that are not in `authors.json` 👍 I think we can leave it as is, for now, I will look into improving it with GitHub APIs in the following PR

apps/site/components/Downloads/ChangelogModal/index.stories.tsx
```diff
@@ -180,10 +179,7 @@ export const Default: Story = {
     open: false,
     heading: 'Node v18.17.0',
     subheading: "2023-07-18, Version 18.17.0 'Hydrogen' (LTS), @danielleadams",
-    avatars: names.map(name => ({
-      src: getGitHubAvatarUrl(name),
-      alt: name,
-    })),
+    avatars: names,
```
AugustinMauroy: For strange rease there are username that display fallback instead of GitHub profile picture (it's valid username)

apps/site/components/Common/BlogPostCard/index.stories.tsx
```diff
@@ -11,12 +11,7 @@ export const Default: Story = {
     category: 'vulnerability',
     description:
       'Starting on March 15th and going through to March 17th (with much of the issue being mitigated on the 16th), users were receiving intermittent 404 responses when trying to download Node.js from nodejs.org, or even accessing parts of the website.',
-    authors: [
-      {
-        fullName: 'Hayden Bleasel',
-        src: 'https://avatars.githubusercontent.com/u/',
-      },
-    ],
+    authors: ['Hayden Bleasel'],
```
AugustinMauroy:
```suggestion
    authors: ['canerakdas'],
```

story with valid GitHub username

bmuenzenmeyer:
> Sounds great! I'd love to add it, but I think this PR has already affected a lot of places. I'm in favor of creating a follow-up PR to this PR and working on it 🙇

considering the complexity of this, should we consider this approach vs #7142 ??

apps/site/components/Common/BlogPostCard/index.stories.tsx
canerakdas: This component requires an author name, I updated it with one that is in the authors file

AugustinMauroy (approved):
LGMT ! Thanks it's super cool

apps/site/util/authorUtils.ts
```diff
@@ -0,0 +1,61 @@
...
+  names.map(name => {
+    if (Object.keys(authors).includes(name)) {
+      const author = authors[name as keyof typeof authors];
+
+      if (author) {
+        const {
+          id,
+          name,
+          website,
+        }: { id: string; name: string; website?: string } = author;
```
ovflowd: Can these be extracted to dedicated types?

apps/site/util/blogUtils.ts
```diff
@@ -15,18 +12,3 @@ export const mapBlogCategoryToPreviewType = (type: string): BlogPreviewType => {
       return 'announcements';
   }
 };
-
-// @todo: we should check about the future of GitHub avatars
-// and mapping them to the respective users
-// @see https://github.com/nodejs/nodejs.dev/blob/main/src/data/blog/authors.yaml
-export const mapAuthorToCardAuthors = (author: string) => {
-  const authors = author.split(/, | and |;| & | prepared by | by /i);
```
ovflowd, line 23: Is the removal of this safe? This splitting was intentional as some author strings within the markdown sources are well, complex, like in these examples.

AugustinMauroy: https://github.com/nodejs/nodejs.org/blob/1d1587d6697a015f30e26db055f616b13e3f266e/apps/site/util/authorUtils.ts#L6-L7

canerakdas: Thanks, @AugustinMauroy! I was a bit busy yesterday so I couldn't reply 🙇

ovflowd:
image

A bit broken in some scenarios. (https://nodejs-org-git-fork-canerakdas-feat-avatar-tooltip-openjs.vercel.app/en/blog/all/page/120)

ovflowd:
Also wondering if we can extract the usernames?

image

ovflowd:
@canerakdas on mobile clicking on the picture opens the profile, I think with the thumbnails clicking should not open the profile directly anymore, IMO.

canerakdas:
> A bit broken in some scenarios. (https://nodejs-org-git-fork-canerakdas-feat-avatar-tooltip-openjs.vercel.app/en/blog/all/page/120)

I didn't see that, thanks!

> Also wondering if we can extract the usernames?

Extracted the usernames here

> @canerakdas on mobile clicking on the picture opens the profile, I think with the thumbnails clicking should not open the profile directly anymore, IMO.

I removed the redirect/tooltip here, it makes more sense to have it on the detail pages 🙇 

----

Also, the tooltip is visible in the changelog modal, but it's not clickable, I am working on the fixing this issue 🤔

canerakdas:
> > Sounds great! I'd love to add it, but I think this PR has already affected a lot of places. I'm in favor of creating a follow-up PR to this PR and working on it 🙇
> 
> considering the complexity of this, should we consider this approach vs #7142 ??

I think it's more like approaches that are a continuation of each other rather than approaches that are against each other 🤔

mcollina:
I think we can show the full list of contributors instead only the first few:

Screenshot 2024-11-03 at 10 49 36


Also, it would be better to display all of them immediately instead of limiting to the first few. There is space, at least in the desktop variant.

Screenshot 2024-11-03 at 10 50 06

ovflowd:
> I think we can show the full list of contributors instead only the first few:
> 
> Screenshot 2024-11-03 at 10 49 36
> Also, it would be better to display all of them immediately instead of limiting to the first few. There is space, at least in the desktop variant.
> 
> Screenshot 2024-11-03 at 10 50 06

Maybe show full list if you click the +13?

canerakdas:
I agree with @mcollina that it would be better to show all contributors/authors, but it is not possible with the current design since we are listing them in a single line 🙇 

I tried to show the maximum avatar we can display for desktop and mobile resolutions.
Sample screenshots are as follows;

### Desktop

image

### Small Desktop / Tablet

image

### Mobile

image

ovflowd:
Hey @canerakdas can we hold this PR until I get the Next.js 15 PR done? 🙏 -- I also still want to provide code review here 🫡

avivkeller:
FYI Next.js v15 has landed, so feel free to continue development on this.

apps/site/components/Common/AvatarGroup/Avatar/index.tsx
```diff
@@ -1,27 +1,52 @@
...
+  const Wrapper = url ? Link : 'div';
+
+  return (
+    <RadixAvatar.Root
+      {...props}
+      className={classNames(styles.avatar, styles[size], props.className)}
+      ref={ref}
+    >
+      <Wrapper
+        {...(url ? { href: url, target: '_blank' } : {})}
```
ovflowd: I don't think I like this weird way of spreading

apps/site/components/Common/AvatarGroup/index.tsx
```diff
@@ -1,25 +1,31 @@
...
 import avatarstyles from '@/components/Common/AvatarGroup/Avatar/index.module.css';
-import { getAcronymFromString } from '@/util/stringUtils';
+import AvatarOverlay from '@/components/Common/AvatarGroup/Overlay';
+import Tooltip from '@/components/Common/Tooltip';
 
 import styles from './index.module.css';
 
 type AvatarGroupProps = {
-  avatars: Array<Omit<ComponentProps<typeof Avatar>, 'fallback'>>;
+  avatars: Array<AvatarProps & { url?: string }>;
```
ovflowd: Why is URL not part of AvatarProps?

apps/site/components/Common/AvatarGroup/index.tsx
```diff
@@ -30,21 +36,31 @@ const AvatarGroup: FC<AvatarGroupProps> = ({
...
-        <Avatar
-          src={avatar.src}
-          alt={avatar.alt}
-          fallback={getAcronymFromString(avatar.alt)}
-          key={index}
-        />
+      {renderAvatars.map(({ ...avatar }) => (
+        <Fragment key={avatar.nickname}>
+          {avatar.url ? (
+            <Tooltip
```
ovflowd: Let's always have the Tooltip component but with an optional argument of `disabled` so that the tree is the same, but tooltip just gets disabled.

apps/site/components/Common/Tooltip/index.tsx
```diff
@@ -0,0 +1,48 @@
+import * as TooltipPrimitive from '@radix-ui/react-tooltip';
```
ovflowd, line 1: Add this within next.config.mjs experimental optimizeImports

apps/site/components/withAvatarGroup.tsx
```diff
@@ -0,0 +1,36 @@
...
+const WithAvatarGroup: FC<WithAvatarGroupProps> = ({
+  usernames,
+  names,
+  clickable = true,
+  ...props
+}) => (
+  <AvatarGroup
+    avatars={
+      usernames
+        ? getAuthorWithId(usernames, clickable)
```
ovflowd: Do we need to nest?

apps/site/util/authorUtils.ts
```diff
@@ -0,0 +1,73 @@
+import authors from '@/authors.json';
```
ovflowd: JSON imports should come from our dedicated next.json file, and from there surround the type with a JSDoc type.

apps/site/util/authorUtils.ts
```diff
@@ -0,0 +1,73 @@
...
+      image: getGitHubAvatarUrl(username),
+      nickname: username,
+      fallback: getAcronymFromString(username),
+      url: hasUrl ? `https://github.com/${username}` : undefined,
+    };
+  });
+
+export const getAuthorWithName = (names: Array<string>, hasUrl: boolean) =>
+  names.map(name => {
+    if (Object.keys(authors).includes(name)) {
```
ovflowd: Extract this to an anonymous function so that the method becomes something like:

```ts
export const getAuthorWithName = (names: Array<string>, hasUrl: boolean) => {
  const mapWhateverBlaBlah = () => {};

  return names.map(mapWhateverBlaBlah);
} 
```

apps/site/util/authorUtils.ts
```diff
@@ -0,0 +1,73 @@
...
+      const author = authors[name as keyof typeof authors];
+
+      if (author) {
```
ovflowd:
```suggestion
      if (name in authors) {
```

apps/site/util/authorUtils.ts
```diff
@@ -0,0 +1,73 @@
...
+type Author = {
+  id: string;
+  name: string;
+  website?: string;
+};
+
+export const getAuthorWithId = (usernames: Array<string>, hasUrl: boolean) =>
+  usernames.map(username => {
+    const author = Object.values(authors).find(
+      ({ id }: { id: string }) => id.toLowerCase() === username.toLowerCase()
```
ovflowd: This should already be typed from the json import by moving the json import to the file where we import all json files

ovflowd (commented):
I've finally done the review :)

apps/site/components/Common/AvatarGroup/index.tsx
canerakdas: It was unnecessarily redefined, I removed it 👍

apps/site/components/Common/AvatarGroup/index.tsx
canerakdas: I solved it by adding the `pointer-events-none` class

apps/site/util/authorUtils.ts
canerakdas: I missed this, thank you! 🙇

avivkeller (approved):
LGTM!

apps/site/components/withAvatarGroup.tsx
canerakdas: By adding type checking, I removed the final `[]`. I could simplify the functions and move them inside a single function, but wouldn’t it be better to keep them separate functions? 🤔

avivkeller: ~~I think it's fine as is. @ovflowd, if this is your last concern, than this should be good to land :-)~~

edit: I've changed my mind. I agree that the nesting is not ideal.

ovflowd: I do usually prefer avoiding nested ?: chains as it makes the code harder to read

avivkeller: @canerakdas you could use an `??` statement, and then in the utilities, if the input is null, return null

AugustinMauroy: There's no need to debate this too long, because things are going to change soon, as there was a plan to standardise the blog's frontmatter to use github users only.

ovflowd: That doesn't mean we can avoid shipping tech debt from day 0 ;)

canerakdas: I was able to update it a little late 🫡 sorry for the delay ❤️

aymen94 (approved):

ovflowd (approved):

ovflowd:
Awesome work here, @canerakdas, thank you so much!! <3 <3

AugustinMauroy:
> Awesome work here, @canerakdas, thank you so much!! <3 <3

yeah it's super cool !

thanks @canerakdas !

PR #7157, opened 2024-11-01
fix: Return `locale` in `i18n.tsx`, also move default there

amannn:
The `locale` should now also be returned from `getRequestConfig`, in order to be able to change an incoming locale to a default in case it's not supported. I saw you already had this mechanism in `page.tsx`, therefore I moved it there.


AugustinMauroy (approved):
LGTM ! Thanks for taking time on that 🫶🏻

AugustinMauroy:
cc @ovflowd

aymen94 (approved):

ovflowd (approved):

PR #7229, opened 2024-11-14
feat(learn): add Package Configuration article

JakobJingleheimer:
This is an article I wrote a couple years ago. It's rather popular and addresses an issue @joyeecheung brought up at the collab summit.

I think canonical url needs to be set for SEO, so Google doesn't think it's plagiarised, but I don't know how to do that here.


apps/site/pages/en/learn/modules/package-configuration.md
```diff
@@ -0,0 +1,405 @@
+---
+title: Package Configuration
+layout: learn
```
AugustinMauroy: you can add you as authors (just GitHub) user name

AugustinMauroy:
> I think canonical url needs to be set for SEO, so Google doesn't think it's plagiarised, but I don't know how to do that here.

OH idk if we can handle that with ou current infra

JakobJingleheimer:
> > I think canonical url needs to be set for SEO, so Google doesn't think it's plagiarised, but I don't know how to do that here.
> 
> 
> 
> OH idk if we can handle that with ou current infra 

It's an html tag. But I dunno how to include that via markdown.

https://developers.google.com/search/docs/crawling-indexing/consolidate-duplicate-urls

AugustinMauroy:
we already have that for lang

JakobJingleheimer:
> we already have that for lang

Soooo… should I add something like

```
canonical: https://…
```

AugustinMauroy:
> Soooo… should I add something like

Sadly no because canonical for lang is handle by our i18n engine.

AugustinMauroy:
the only front matter that learn section handle is:
```md
--- 
title: A super cool title
layout: learn
authors: github_username, another_github_username
--- 
```

but why not open an issue if needed

apps/site/pages/en/learn/modules/package-configuration.md
```diff
@@ -0,0 +1,406 @@
...
+authors: JakobJingleheimer
+---
+
+# Package configuration
+
+Configuration is always a chore, but an unfortunately necessary evil. And configuring a package for CommonJS (CJS) and ES Modules (ESM) can be a waking nightmare—not least because it has changed a dozen times in half as many years.
+
+A frequent question is “how do I make this work!?” (often with angry tears); but yet more frequently we come across packages that are just misconfigured.
+
+All the provided `package.json` configurations (not specifically marked “does not work”) work in Node.js 12.22.x (v12 latest, the oldest supported line) and 17.2.0 (current latest at the time)<sup>[1](#footnotes)</sup>, and for grins, with webpack 5.53.0 and 5.63.0 respectively. These are available: [nodejs/package-examples](https://github.com/nodejs/package-examples/blob/main/config).
```
avivkeller: The markdown parser supports [footnotes](https://www.markdownguide.org/extended-syntax/#footnotes), IMO they should be used here, instead of the `<sup>...</sup>`

(E.G):
```
Some text[^1]

[^1]: Additional content
```

Some text[^1]

[^1]: Additional content

avivkeller: Can you ensure the examples work on 22 and 23, and update this sentence accordingly? (and throughout the file)

Those users are (IMO) the "target" audience.

joyeecheung:
Thanks for picking it up! I won't have time to review this until ~December, so feel free to go ahead without my approval if you want to get it out soonish. But some observations after a quick glance, which I think should be addressed given that we spent hours at the summit talking about the issues with outdated content...:

1. This doesn't reference a popular pattern that e.g. Babel uses to avoid the dual package hazard: https://github.com/nodejs/node/issues/52174 
2. Somewhere in the beginning of the article, it should make it clear when *no* configuration is necessary. Generally you don't need to use the exports field at all if you just have an CJS exports (use main) and a normal package.json.
3. It's missing instructions of shipping ESM on Node.js versions >= 23 (and soon >= 22 or >= 20) while maintaining compatibility for CJS users via require(esm).
4. When surveying the exports patterns in the ecosystem I have very rarely seen the package.json specified. On the other hand, types get specified quite a lot. Maybe that's something worth teaching here too.

apps/site/pages/en/learn/modules/package-configuration.md
```diff
@@ -0,0 +1,406 @@
...
+
+This the "Rum & Coke" of packages: pretty difficult to mess up. Essentially just declare the package’s exports via the [`"exports"`](https://nodejs.org/api/packages.html#conditional-exports) field/field-set.
+
+**Working example**: [cjs-with-cjs-distro](https://github.com/JakobJingleheimer/nodejs-module-config-examples/tree/main/packages/cjs/cjs-distro)
+
+```json
+{
+  "type": "commonjs", // current default, but may change
+  "engines": { "node": ">=12.22.7" }, // optional, but kind
+  "exports": {
```
joyeecheung: Is this necessary? At most it can use the explicit type field, but AFAICT you just need to point the main field and don't need to bother with exports.

JakobJingleheimer:
@joyeecheung there was talk a while ago about discouraging `"main"` in favour of `"exports"`. I'm not familiar with the rationale behind it ("main" _seems_ very okay to me without some of the drawbacks of "exports"); that's the only reason I use "exports" in the examples what aren't actually using any fanciness from exports (and "main" would be equivalent). Is that not the case? I have no preference whatsoever.

apps/site/pages/en/learn/modules/package-configuration.md
JakobJingleheimer: Maybe technically yes. But this guide is about _should_ not necessarily _must_.

apps/site/pages/en/learn/modules/package-configuration.md
```diff
@@ -0,0 +1,408 @@
...
+}
+```
+
+The [`.mjs`](https://nodejs.org/api/esm.html#enabling) file extension is a trump-card: it will override **any** other configuration and the file will be treated as ESM. Using this file extension is necessary because `packageJson.exports.import` does **NOT** signify that the file is ESM (contrary to common, if not universal, misperception), only that it is the file to be used when the package is imported (ESM _can_ import CJS. See [Gotchas](#gotchas) below).
+
+The [`"engines"`](https://nodejs.dev/learn/the-package-json-guide#engines) field provides both a human-friendly and a machine-friendly indication of with which version(s) of Node.js the package is compatible. Depending on the package manager used, an exception may be thrown causing the installation to fail when the consumer is using an incompatible version of Node.js (which can be very helpful to consumers). Including this field here will save a lot of headache for consumers with an older version of Node.js who cannot use the package.
+
+### CJS source and both CJS & ESM distribution
+
+You have a few options:
```
ljharb: this seems to presume you have multiple things to export - is there an example of when you only have one thing to export? (ie, a single-responsibility package)

apps/site/pages/en/learn/modules/package-configuration.md
```diff
@@ -0,0 +1,408 @@
+---
+title: Package Configuration
+layout: learn
+authors: JakobJingleheimer
+---
+
+# Package configuration
+
+Configuration is always a chore, but an unfortunately necessary evil. And configuring a package for CommonJS (CJS) and ES Modules (ESM) can be a waking nightmare—not least because it has changed a dozen times in half as many years.
```
AugustinMauroy:
```suggestion
Configuring a node.js project/package is always a chore, but an unfortunate necessary evil. And configuring a package for CommonJS (CJS) and ES Modules (ESM) can be a waking nightmare - not least because it has changed a dozen times in half as many years.
```

apps/site/pages/en/learn/modules/package-configuration.md
```diff
@@ -0,0 +1,408 @@
...
+
+An additional complication is bundlers, which historically managed much of this territory. However, much of what we previously needed bundle(r)s to manage is now native functionality; yet bundlers are still (and likely always will be) necessary for some things. Unfortunately, functionality bundlers no-longer need to provide is deeply ingrained in older bundlers’ implementations, so they can at times be too helpful, and in some cases, anti-pattern (bundling a library is often not recommended by bundler authors themselves). The hows and whys of that are an article unto itself.
+
+## Pick your poison
```
AugustinMauroy: I understand the idea behind the title, but the name Node.js doesn't really sell it. May be something like "Choosing what best suits our expectations "

apps/site/pages/en/learn/modules/package-configuration.md
```diff
@@ -0,0 +1,408 @@
...
+### CJS source and only ESM distribution
+
+The "Gin & Tonic" of packages: This takes a small bit of finesse but is also pretty straight-forward.
+
+**Working example**: [cjs-with-esm-distro](https://github.com/JakobJingleheimer/nodejs-module-config-examples/tree/main/packages/cjs/esm-distro)
+
+```json
+{
+  "type": "commonjs", // current default, but may change
+  "engines": { "node": ">=12.22.7" }, // optional, but kind
```
AugustinMauroy:
```suggestion
  "engines": { "node": ">=12.22.7" }, // optional, it lets you know which node version you're using and npm warns you when it's not the right version.
```

apps/site/pages/en/learn/modules/package-configuration.md
```diff
@@ -0,0 +1,408 @@
...
+}
+```
+
+Note that `packageJson.exports["."] = filepath` is shorthand for `packageJson.exports["."].default = filepath`
+
+### CJS source and only ESM distribution
+
+The "Gin & Tonic" of packages: This takes a small bit of finesse but is also pretty straight-forward.
+
+**Working example**: [cjs-with-esm-distro](https://github.com/JakobJingleheimer/nodejs-module-config-examples/tree/main/packages/cjs/esm-distro)
```
AugustinMauroy:
```suggestion
**Working example**: [cjs-with-esm-distro](https://github.com/JakobJingleheimer/nodejs-module-config-examples/tree/main/packages/cjs/esm-distro) _unofficial example node.js_
```

Add this mention to clarify that the project trust the source but it's not officially maintianed by node.js itself.
_yeah i know your are on node.js_ But I'm convicted that a good practice to explain when it's not a node.js official ressource

AugustinMauroy: I think there are other link in this article can be updated with this mention

apps/site/pages/en/learn/modules/package-configuration.md
JakobJingleheimer: Good idea but this comment is very long (and will cause horizontal scrolling). I think it would be better to add a general note somewhere pointing to the npm doc that explains what this field does.

apps/site/pages/en/learn/modules/package-configuration.md
JakobJingleheimer: Oh, good catch. I intend to move these into the `nodejs/package-examples` repo joyee started.

apps/site/pages/en/learn/modules/package-configuration.md
bmuenzenmeyer: i dont think this is better in this case

ovflowd (approved):
Soft LGTM, but I would appreciate more reviewers first.

ovflowd:
BTW I love your writing style, and would love if more existing content got updated with this nice writing style.

apps/site/pages/en/learn/modules/package-configuration.md
```diff
@@ -0,0 +1,408 @@
...
+
+For curious cats, [Preamble: How did we get here](#preamble-how-did-we-get-here) and [Down the rabbit-hole](#down-the-rabbithole) provide background and deeper explanations. If you're just looking for a solution, jump to [Pick your poison](#pick-your-poison) for the TLDR.
+
+## Preamble: How did we get here
+
+CommonJS (CJS) was created _long_ before ECMAScript Modules (ESM), back when JavaScript was still adolescent—CJS and jQuery were created just 3 years apart. CJS is not an official (TC39) standard and is supported by a limited few platforms (most notably, Node.js). ESM as a standard has been incoming for several years; it is currently supported by all major platforms (browsers, Deno, Node.js, etc), meaning it will run pretty much everywhere. As it became clear ESM would effectively succeed CJS (which is still very popular and widespread), many attempted to adopt early on, often before a particular aspect of the ESM specification was finalised. Because of this, those changed over time as better information became available (often informed by learnings/experiences of those eager beavers), going from best-guess to the aligning with the specification.
```
AugustinMauroy:
```suggestion
CommonJS (CJS) was created _long_ before ECMAScript Modules (ESM), back when JavaScript was still adolescent—CJS and jQuery were created just 3 years apart. CJS is not an official ([TC39](https://tc39.es)) standard and is supported by a limited few platforms (most notably, Node.js). ESM as a standard has been incoming for several years; it is currently supported by all major platforms (browsers, Deno, Node.js, etc), meaning it will run pretty much everywhere. As it became clear ESM would effectively succeed CJS (which is still very popular and widespread), many attempted to adopt early on, often before a particular aspect of the ESM specification was finalised. Because of this, those changed over time as better information became available (often informed by learnings/experiences of those eager beavers), going from best-guess to the aligning with the specification.
```

May be ad link to tc39

apps/site/pages/en/learn/modules/package-configuration.md
```diff
@@ -0,0 +1,408 @@
...
+  "exports": {
+    ".": "PATH/TO/DIST/CODE/ENTRYPOINT.mjs", // ex "./dist/index.mjs"
+    "./package.json": "./package.json" // ensure this file is importable
+  }
+}
+```
+
+The [`.mjs`](https://nodejs.org/api/esm.html#enabling) file extension is a trump-card: it will override **any** other configuration and the file will be treated as ESM. Using this file extension is necessary because `packageJson.exports.import` does **NOT** signify that the file is ESM (contrary to common, if not universal, misperception), only that it is the file to be used when the package is imported (ESM _can_ import CJS. See [Gotchas](#gotchas) below).
+
+The [`"engines"`](https://nodejs.dev/learn/the-package-json-guide#engines) field provides both a human-friendly and a machine-friendly indication of with which version(s) of Node.js the package is compatible. Depending on the package manager used, an exception may be thrown causing the installation to fail when the consumer is using an incompatible version of Node.js (which can be very helpful to consumers). Including this field here will save a lot of headache for consumers with an older version of Node.js who cannot use the package.
```
AugustinMauroy:
```suggestion
The [`"engines"`](https://nodejs.dev/learn/the-package-json-guide#engines) field provides both a human-friendly and a machine-friendly indication of with which version(s) of Node.js the package is compatible. Depending on the package manager used, an exception may be thrown causing the installation to fail when the consumer is using an incompatible version of Node.js (which can be very helpful to consumers). Including this field here will save a lot of headache for consumers with an older version of Node.js who cannot use the package.
```

"https://nodejs.dev/learn/the-package-json-guide#engines" = 404 
I'm looking if we still have this docs

AugustinMauroy: https://github.com/nodejs/nodejs.dev/blob/7d89c22668101d972bf463d713c09834e9586dc8/src/documentation/0019-package-json/index.md#engines

no sure if good idea to point to an old ressource like that

apps/site/pages/en/learn/modules/package-configuration.md
joyeecheung: That may be true before it's moved to the website, but now it's in the learn session I think we should target the audience differently (which I think was the core of the doc/ambassador program in the past two summits): we should teach more best practices that are approachable to beginners and up-to-date. Having the guide about package shipping very pro-configuration when it's not necessary and is not what the majority of the ecosystem does is deviating from that goal IMO.

Having to configure exports is already not great UX and it's only a compromise/advanced topic for those who needs more control over the package, but for many simpler cases, the package author should not need to pay for the mental burden of it at all. I have seen many complaints/jokes from end users and maintainers about the mental burden of the exports field, and I think we should make it clear in the website tutorials about when they don't have to pay for it to reduce barriers.

avivkeller:
Can you update the examples to use v22, which is the current LTS? I don't think it's helpful to users if this articles targets v12.

apps/site/pages/en/learn/modules/package-configuration.md
```diff
@@ -0,0 +1,408 @@
...
+      "default": "PATH/TO/DIST/CJS-CODE/ENTRYPOINT.js"
+    },
+    "./package.json": "./package.json"
+  }
+}
+```
+
+This does not work because `packageJson.exports["."].import` will get interpreted as CJS (but it’s actually ESM).
+
+## Footnotes
```
avivkeller:
```suggestion
```

avivkeller:
Hey,

https://github.com/nodejs/nodejs.org/pull/7123 is on the TSC agenda because it's from an external source, and this content is also from an external source. I don't think this can land until that conversation reaches consensus?

AugustinMauroy:
@RedYetiDev the orignal post was written by jakob who put it on the learn section. And Jakob propose that at the collab summit. So I don't think it's an problem here

JakobJingleheimer:
Yeah, I'm the original author, so posting external content is not an issue here: it's becoming internal content.

joyeecheung:
> there was talk a while ago about discouraging "main" in favour of "exports".

AFAIK the primary concern about "main" is that it does not forbid files to be loaded by dependencies, whereas "exports" can be used to guard off unwanted loads. But then that doesn't apply to all packages, especially the small ones that have no files that are internal to begin with. It would be pretty daunting for the package configuration guide to just start out with "let's do exports!" given that so many people complain about the mental burden of it. A tutorial should start small and simple and only reach to advanced practices as it goes IMO.

apps/site/pages/en/learn/modules/package-configuration.md
avivkeller: I think unfortunately -> unfortunate is better (seeing as unfortunate is an adjective, rather than an adverb).

apps/site/pages/en/learn/modules/package-configuration.md
JakobJingleheimer: How about 2 versions of each: "basic" and "advanced"?

apps/site/pages/en/learn/modules/package-configuration.md
JakobJingleheimer: > an unfortunately necessary evil

or

> an unfortunate, necessary evil

that latter feels cumbersome to me though

apps/site/pages/en/learn/modules/package-configuration.md
JakobJingleheimer: Ah. I'll add a note about it in general and remove the dead link.

JakobJingleheimer:
> Can you update the examples to use v22, which is the current LTS? I don't think it's helpful to users if this articles targets v12.

It doesn't target v12. The examples just all work in v12. I believe there is nothing to update in the examples for v22. I updated a note about `require(esm)`, which I think is the only new thing.

JakobJingleheimer:
I think everything has been addressed except the repo with working examples hasn't yet been transferred into nodejs's github. They'll be ported wholesale (no changes), so in the interest of helping users sooner, I propose handling that subsequently and then updating the links in this article with a tiny PR.

apps/site/pages/en/learn/modules/package-configuration.md
```diff
@@ -0,0 +1,563 @@
...
+| ESM source code using `import`    | CJS: consumers `require()` your package                          | [ESM source with only CJS distribution](#esm-source-with-only-cjs-distribution)<br />[ESM source and distribution](#esm-source-and-distribution)             |
+| ESM source code using `import`    | ESM: consumers `import` your package                             | [ESM source and distribution](#esm-source-and-distribution)                                                                                                  |
+| ESM: source code uses `import`    | CJS & ESM: consumers either `require()` or `import` your package | [ESM source and both CJS & ESM distribution](#esm-source-and-both-cjs-amp-esm-distribution)<br />[ESM source and distribution](#esm-source-and-distribution) |
+
+### CJS source and distribution
+
+This the "Rum & Coke" of packages: pretty difficult to mess up. You _technially_ may not need _any_ package configuration beyond [`"name"`](https://nodejs.org/api/packages.html#name). But the less arcane, the better: Essentially just declare the package’s exports via the `"exports"` field/field-set.
```
joyeecheung: For official tutorials, references to alcoholic drinks seem inappropriate. It may depend on the readers, but repeated references throughout the post like these can increase the chance of being a trigger for people struggling with sobriety. Also as a non-drinker I feel these references just add an extra layer of confusion for me when I read it...(even if I do drink, people from my background generally drink very different things and I think these references just sets up cultural barriers for us who may already have a harder time reading English..)

apps/site/pages/en/learn/modules/package-configuration.md
```diff
@@ -0,0 +1,563 @@
...
+  "name": "cjs-source-and-distribution"
+  // "main": "./index.js"
+}
+```
+
+```json displayName="Advanced (verbose) package.json"
+{
+  "name": "cjs-source-and-distribution",
+  "type": "commonjs", // current default, but may change
+  "engines": { "node": ">=12.22.7" },
```
joyeecheung:
```suggestion
```

This doesn't seem necessary?

apps/site/pages/en/learn/modules/package-configuration.md
```diff
@@ -0,0 +1,563 @@
...
+  "exports": {
+    ".": "./dist/index.js",
+    "./package.json": "./package.json"
```
joyeecheung: A "./package.json" export seems rather uncommon, shouldn't they be left to some extra sections?

apps/site/pages/en/learn/modules/package-configuration.md
```diff
@@ -0,0 +1,563 @@
...
+}
+```
+
+Note that `packageJson.exports["."] = filepath` is shorthand for `packageJson.exports["."].default = filepath`
+
+### CJS source and only ESM distribution
+
+The "Gin & Tonic" of packages: This takes a small bit of finesse but is also pretty straight-forward.
+
+**Working example**: [cjs-with-esm-distro](https://github.com/JakobJingleheimer/nodejs-module-config-examples/tree/main/packages/cjs/esm-distro)
```
joyeecheung: This seems rather uncommon, I am not sure if I have seen any real world packages authored in CJS and then ship as transpiled ESM...

apps/site/pages/en/learn/modules/package-configuration.md
```diff
@@ -0,0 +1,563 @@
...
+```
+
+```json displayName="Advanced (verbose) package.json"
+{
+  "name": "cjs-with-alt-full-dual-distro",
+  "type": "commonjs",
+  "engines": { "node": ">=12.22.7" },
+  "exports": {
+    ".": {
+      "default": "./dist/esm/index.mjs",
```
joyeecheung: I am pretty sure it's incorrect to put default before any other exports?

apps/site/pages/en/learn/modules/package-configuration.md
```diff
@@ -0,0 +1,563 @@
...
+layout: learn
+authors: JakobJingleheimer
+---
+
+# Package configuration
+
+Configuration is always a chore, but an unfortunately necessary evil. And configuring a package for CommonJS (CJS) and ES Modules (ESM) can be a waking nightmare—not least because it has changed a dozen times in half as many years.
+
+A frequent question is “how do I make this work!?” (often with angry tears); but yet more frequently we come across packages that are just misconfigured.
+
```
joyeecheung:
```suggestion
```

These feel more like personal takes that don't really belong to the official website..

apps/site/pages/en/learn/modules/package-configuration.md
```diff
@@ -0,0 +1,563 @@
...
+All the provided `package.json` configurations (not specifically marked “does not work”) work in Node.js 12.22.x (v12 latest, the oldest supported line) and 17.2.0 (current latest at the time)[^1], and for grins, with webpack 5.53.0 and 5.63.0 respectively. These are available: [nodejs/package-examples](https://github.com/nodejs/package-examples/blob/main/config).
+
+For curious cats, [Preamble: How did we get here](#preamble-how-did-we-get-here) and [Down the rabbit-hole](#down-the-rabbithole) provide background and deeper explanations. If you're just looking for a solution, jump to [Pick your fix](#pick-your-fix) for the TLDR.
+
+## General notes
+
+[Syntax detection](https://nodejs.org/api/packages.html#syntax-detection) is _**not**_ a replacement for proper package configuration; syntax detection is not fool-proof and it has [significant performance cost](https://github.com/nodejs/node/pull/55238).
```
joyeecheung: The notes feel like something that should be left to the end, instead of in the beginning, as beginners might not even know what these fields are..

apps/site/pages/en/learn/modules/package-configuration.md
```diff
@@ -0,0 +1,563 @@
...
+
+When using [`"exports"`](https://nodejs.org/api/packages.html#conditional-exports) in package.json, it is generally a good idea to include `"./package.json": "./package.json"` so that it can be imported ([`module.findPackageJSON`](https://nodejs.org/api/module.html#modulefindpackagejsonspecifier-base) is not affected by this limitation, but `import` may be more convenient).
```
joyeecheung: Is it for `import 'pkg/package.json'`? I think that's rather uncommon?

apps/site/pages/en/learn/modules/package-configuration.md
```diff
@@ -0,0 +1,563 @@
...
+
+## Preamble: How did we get here
+
+CommonJS (CJS) was created _long_ before ECMAScript Modules (ESM), back when JavaScript was still adolescent—CJS and jQuery were created just 3 years apart. CJS is not an official (TC39) standard and is supported by a limited few platforms (most notably, Node.js). ESM as a standard has been incoming for several years; it is currently supported by all major platforms (browsers, Deno, Node.js, etc), meaning it will run pretty much everywhere. As it became clear ESM would effectively succeed CJS (which is still very popular and widespread), many attempted to adopt early on, often before a particular aspect of the ESM specification was finalised. Because of this, those changed over time as better information became available (often informed by learnings/experiences of those eager beavers), going from best-guess to the aligning with the specification.
```
joyeecheung: I feel that the tutorial doesn't need to give a history lesson, but if it does, at least it should give reference the CommonJS standard: https://wiki.commonjs.org/wiki/Modules

apps/site/pages/en/learn/modules/package-configuration.md
```diff
@@ -0,0 +1,563 @@
...
+
+An additional complication is bundlers, which historically managed much of this territory. However, much of what we previously needed bundle(r)s to manage is now native functionality; yet bundlers are still (and likely always will be) necessary for some things. Unfortunately, functionality bundlers no-longer need to provide is deeply ingrained in older bundlers’ implementations, so they can at times be too helpful, and in some cases, anti-pattern (bundling a library is often not recommended by bundler authors themselves). The hows and whys of that are an article unto itself.
+
+## Pick your fix
+
+This article covers configuration of all possible combinations in modern Node.js (v12+). If you are trying to decide which options are ideal, it is better to avoid dual packages, so either:
```
joyeecheung:
```suggestion
This article covers configuration of all possible combinations in currently supported Node.js versions. If you are trying to decide which options are ideal, it is better to avoid dual packages, so either:
```

`modern Node.js (v12+)` is somewhat weird, a more commonly used span would be all the currently supported Node.js versions. Also it seems a bit too ambitious to claim that this cover "all possible combinations" (there is `module-sync` left uncovered, for example).

apps/site/pages/en/learn/modules/package-configuration.md
AugustinMauroy: I'm not a fan of redirecting there because it's not a great resource (not well maintained).

apps/site/pages/en/learn/modules/package-configuration.md
JakobJingleheimer: Does anyone disagree (with what the content is itself saying)?

apps/site/pages/en/learn/modules/package-configuration.md
JakobJingleheimer: I put them at the beginning because I think they address distractions readers may be fixated on. For instance "Why do I even need this when I have syntax detection?"

Also, I believe you had previously asked for this section at the beginning :)

apps/site/pages/en/learn/modules/package-configuration.md
JakobJingleheimer: Yes, and also yes. But not never, and you never know what a user may want from it.

apps/site/pages/en/learn/modules/package-configuration.md
JakobJingleheimer: The wiki article seems okay to me 🤔

I think the history is interesting, but it's not super important. I included it mostly as support for commonjs.

apps/site/pages/en/learn/modules/package-configuration.md
JakobJingleheimer: Ah, yes, that was a product of its time (the article was originally written 2–3 years ago. I think we were on v17 and v12 had just hit EoL.

`module-sync` is new (it did not exist when this was written). Should it be included?

I think there are no other possibilities, but happy to adjust the wording to say "most" instead of "all". Would that work for you?

apps/site/pages/en/learn/modules/package-configuration.md
JakobJingleheimer: They could be virgin drinks—nowhere does it mention alcohol :)

> Also as a non-drinker I feel these references just add an extra layer of confusion for me

I don't understand why. These are not at all keystones in explanations—merely passing humour that do perhaps also provide a tiny bit of insight.

The inclusion of these is to add some humour to an otherwise dry and boring topic. A lot of readers specifically cited enjoying these. Just because some people don't enjoy them doesn't mean others shouldn't get the chance.

Perhaps sobriety is a reason, but this is hardly someone's only exposure to it. And again, there is no mention of actual alcohol.

We can remove them, but I think it would be much to the detriment.

apps/site/pages/en/learn/modules/package-configuration.md
JakobJingleheimer: Nope. That's explained at the top; these are also hidden by default per your suggestion to keep things simple by default:

Default: minimal package.json without 'engines'

Non-default: advanced package.json with 'engines'

apps/site/pages/en/learn/modules/package-configuration.md
JakobJingleheimer: That's exactly where this is :) see screenshots above.

apps/site/pages/en/learn/modules/package-configuration.md
JakobJingleheimer: I have definitely seen this. DM me if you need specific examples (they are not public).

apps/site/pages/en/learn/modules/package-configuration.md
JakobJingleheimer: Do they? What makes it seem there are multiple or single exports?

apps/site/pages/en/learn/modules/package-configuration.md
JakobJingleheimer: OH! You're right! But also, how and why does this possibly matter 🤔

apps/site/pages/en/learn/modules/package-configuration.md
joyeecheung: I don't recall asking about syntax detection per-se, but generally I think it should be an implementation detail that package authors rarely know about as for anyone who has a `package.json`, the most no-brainer thing to do is to just use explicit type fields. The article assumes that readers know what even is syntax detection, but I have seen several experienced maintainers not even knowing it's a thing, let alone beginners (and to answer what even is syntax detection, one needs to know about `.js` rules, which are pretty daunting for beginners, even though they could've just used the type field and not waste any brain energy on it).

apps/site/pages/en/learn/modules/package-configuration.md
AugustinMauroy: +1 to leave the text as it is

AugustinMauroy (approved):
LTGM! I can't review more than this. It's out of my competence.

apps/site/pages/en/learn/modules/package-configuration.md
joyeecheung: For this specific case, the dual module doesn't work on versions of Node.js that doesn't support `require(esm)`, because default gets picked first (and on versions of Node.js that support `require(esm)`, you don't need any of these).

apps/site/pages/en/learn/modules/package-configuration.md
joyeecheung: Are there any note-worthy pacakges on npm doing this? Should we even document patterns that can't be found on npm?

apps/site/pages/en/learn/modules/package-configuration.md
joyeecheung: oh right, missed the explanation about tabs, sorry.

apps/site/pages/en/learn/modules/package-configuration.md
joyeecheung: I think jokes are great, but in-jokes aren't appropriate for tutorials that are supposed to be inclusive and un-assuming. It suggests to me "you are a lame non-drinker, so you won't be part of us who understand this joke". I can understand the one about wine, but the other repeated references that I don't understand makes me feel dumb when reading the article.

apps/site/pages/en/learn/modules/package-configuration.md
```diff
@@ -0,0 +1,563 @@
...
+
+#### Attach named exports directly onto `exports`
+
+The "French 75" of packages: Classic but takes some sophistication and finesse.
+
+Pros:
+
+- Smaller package weight
+- Easy and simple (probably least effort if you don't mind keeping to a minor syntax stipulation)
+- Precludes the Dual-Package Hazard
```
joyeecheung: This assumes users know what `Dual-Package Hazard` even is..

apps/site/pages/en/learn/modules/package-configuration.md
```diff
@@ -0,0 +1,563 @@
...
+
+module.exports = someObject;
+```
+
+Instead, do this:
+
+```cjs
+module.exports.foo = function foo() {};
+module.exports.foo = function bar() {};
+module.exports.foo = function qux() {};
```
joyeecheung:
```suggestion
module.exports.bar = function bar() {};
module.exports.qux = function qux() {};
```

apps/site/pages/en/learn/modules/package-configuration.md
```diff
@@ -0,0 +1,563 @@
...
+      "default": "./dist/cjs/index.js"
+    },
+    "./package.json": "./package.json"
+  }
+}
+```
+
+In order to support named exports from the CJS bundle for an ESM consumer, this will need a bit of gymnastics from a bundler but is conceptually very simple.
+
+In certain conditions, CJS exports an object (which gets aliased to ESM's `default`); that object, like any object, is destructure-able. You can leverage that to pluck all the members of the object out, and then re-export them so the ESM consumer is none the wiser.
```
joyeecheung:
```suggestion
When CJS exports an object (which gets aliased to ESM's `default`), you can save references to all the members of the object locally in the wrapper, and then re-export them so the ESM consumer can access all of them by name.
```

(The paragraph is very difficult to read for me).

I feel that this needs to at least explain that this pattern, while simple enough, also breaks the live binding if the module re-assigns any of its exports, so should be avoided by modules that need to do that...

apps/site/pages/en/learn/modules/package-configuration.md
```diff
@@ -0,0 +1,563 @@
...
+  "type": "commonjs", // current default, but may change
+  "engines": { "node": ">=12.22.7" },
+  "exports": {
+    ".": "./dist/cjs/index.js",
+    "./package.json": "./package.json"
+  }
+}
+```
+
+Typically, you would see `module.exports` assigned to something (be it an object or a function) like this:
```
joyeecheung:
```suggestion
Sometimes, a CJS module may re-assign `module.exports` to something else (be it an object or a function) like this:
```

(I am not sure how typical the re-assigning actually is..)

apps/site/pages/en/learn/modules/package-configuration.md
```diff
@@ -0,0 +1,563 @@
...
+const someObject = {
+  foo() {},
+  bar() {},
+  qux() {},
+};
+
+module.exports = someObject;
+```
+
+Instead, do this:
```
joyeecheung:
```suggestion
Node.js detects the named exports in CJS via [static analysis that look for certain patterns](https://github.com/nodejs/cjs-module-lexer/tree/main?tab=readme-ov-file#parsing-examples), which the example above evades. To make the named exports detectable, do this:
```

There are several caveats of the named exports detection, and I think it should at least link to https://github.com/nodejs/cjs-module-lexer/tree/main?tab=readme-ov-file#parsing-examples because it can be rather easy to miss the detection. Also, I think this deserves a bit of explanation otherwise beginners would have no idea how this makes any difference and how to extrapolate it properly to their own exports.

apps/site/pages/en/learn/modules/package-configuration.md
```diff
@@ -0,0 +1,563 @@
...
+      "import": "./dist/esm/wrapper.mjs",
+      "require": "./dist/cjs/index.js",
+      "default": "./dist/cjs/index.js"
+    },
+    "./package.json": "./package.json"
+  }
+}
+```
+
+In order to support named exports from the CJS bundle for an ESM consumer, this will need a bit of gymnastics from a bundler but is conceptually very simple.
```
joyeecheung:
```suggestion
When the CJS output from the bundler evades the named exports detection in Node.js, a ESM wrapper can be used to explicitly re-export the known named exports for ESM consumers.
```

I feel the original paragraph generalizes it a bit too much about the use case and can be confusing for beginners who don't understand whether this is something they should go with - only when the named exports detection simply doesn't work would you need to write a ESM wrapper to give Node.js a bit of help.

apps/site/pages/en/learn/modules/package-configuration.md
```diff
@@ -0,0 +1,563 @@
...
+      "import": "./dist/esm/index.mjs",
+      "require": "./dist/cjs/index.js",
+      "default": "./dist/cjs/index.js"
+    },
+    "./package.json": "./package.json"
+  }
+}
+```
+
+Alternatively, you can use `"default"` and `"node"` keys, which are less counter-intuitive: Node.js will always choose the `"node"` option (which always works), and non-Node.js tooling will choose `"default"` when configured to target something other than node. **This precludes the dual-package hazard.**
```
joyeecheung:
```suggestion
Alternatively, you can use `"node"` and `"default"` keys:  Node.js could pick up the file pointed by the `"node"` condition, which should point to CJS and is shared by both CJS and ESM consumers, and non-Node.js tooling could pick up `"default"` which can point to ESM. **This precludes the dual-package hazard** because only one copy is ever picked up in the application.
```

I feel that the explanation is missing the point, the reason why this works is that only one copy gets picked in one applicaiton no matter how it's consumed, so you wouldn't end up with two copies being loaded into the same graph. Also, it should be a bit more precise - the pick up is done by insertion order, Node.js doesn't always choose `"node"` if `"node"` is not the first one.

(Why is it in this section? This section explicitly says "Vulnerable to the Dual-Package Hazard" in the beginning, but this pattern avoids that, it seems contradictory).

apps/site/pages/en/learn/modules/package-configuration.md
```diff
@@ -0,0 +1,563 @@
...
+    ".": {
+      "node": "./dist/cjs/index.js",
+      "default": "./dist/esm/index.mjs"
+    },
+    "./package.json": "./package.json"
+  }
+}
+```
+
+### ESM source and distribution
```
joyeecheung: I think this should be placed right after the CJS-CJS configuration, because soon for all packages that only support LTS, they can stop reading after the CJS-CJS and ESM-ESM examples as all the others are workarounds prior to `require(esm)` support and will be legacy notes that they shouldn't spend much brain energy on.

apps/site/pages/en/learn/modules/package-configuration.md
```diff
@@ -0,0 +1,563 @@
...
+### CJS source and distribution
+
+This the "Rum & Coke" of packages: pretty difficult to mess up. You _technially_ may not need _any_ package configuration beyond [`"name"`](https://nodejs.org/api/packages.html#name). But the less arcane, the better: Essentially just declare the package’s exports via the `"exports"` field/field-set.
+
+**Working example**: [cjs-with-cjs-distro](https://github.com/JakobJingleheimer/nodejs-module-config-examples/tree/main/packages/cjs/cjs-distro)
+
+```json displayName="Minimal package.json"
+{
+  "name": "cjs-source-and-distribution"
+  // "main": "./index.js"
```
joyeecheung: Why is the main field commented? It's probably more common to have the main field than not to have it, as most package manager initializers initialize it. If commented then this might warrant an explanation about how it still works without a `main` field and which file would get picked then, which feels a bit too much for beginners..

apps/site/pages/en/learn/modules/package-configuration.md
```diff
@@ -0,0 +1,563 @@
...
+  "engines": { "node": ">=12.22.7" },
+  "exports": {
+    ".": "./dist/index.js",
+    "./package.json": "./package.json"
+  }
+}
+```
+
+Note that ESM now _is_ “backwards” compatible with CJS: a CJS module now _can_ [`require()` an ES Module](https://nodejs.org/api/modules.html#loading-ecmascript-modules-using-require) without a flag as of 23.0.0 and 22.12.0.
+
```
joyeecheung: Since the import CJS sections mention rules about named exports detection, I think the require ESM sections should also mention `export { something as "module.exports" }` as that's another quirk for the interop to work properly.

apps/site/pages/en/learn/modules/package-configuration.md
```diff
@@ -0,0 +1,563 @@
...
+The "Piña Colada" of packages: Complicated setup and difficult to get the balance right.
+
+Pros:
+
+- Smaller package weight
+
+Cons:
+
+- Likely requires complicated bundler gymnastics (we could not find any existing option to automate this in Webpack).
```
joyeecheung: What is this "complicated bundler gymnastics" being required? Bundler settings? Isn't this pattern specifically used to deal with the case where bundler settings aren't good enough to support named exports?

Also an obvious cons is that one needs to manually keep the exports in-sync.

apps/site/pages/en/learn/modules/package-configuration.md
```diff
@@ -0,0 +1,563 @@
...
+
+When source code is written in non-JavaScript (ex TypeScript), options can be limited due to needing to use file extension(s) specific to that language (ex `.ts`) and there may be no `.mjs` equivalent.
+
+Similar to [CJS source and both CJS & ESM distribution](#cjs-source-and-both-cjs-amp-esm-distribution), you have the same options.
+
+#### Publish only a CJS distribution with property exports
+
+The "Mojito" of packages: Tricky to make and needs good ingredients.
+
+This option is almost identical to the [CJS source with CJS & ESM distribution's property exports](#attach-named-exports-directly-onto-raw-exports-endraw-) above. The only difference is in package.json: `"type": "module"`.
```
joyeecheung: I feel that this is lacking some explanation somewhere about what `"type": "module"` is actually for (the authored code).

apps/site/pages/en/learn/modules/package-configuration.md
```diff
@@ -0,0 +1,563 @@
...
+    "./package.json": "./package.json"
+  }
+}
+```
+
+💡 Using `"type": "module"`[^2] paired with the `.cjs` file extension (for commonjs files) yields best results. For more information on why, see [Down the rabbit-hole](#down-the-rabbithole) and [Gotchas](#gotchas) below.
+
+#### Publish a CJS distribution with an ESM wrapper
+
+The "Pornstar Martini" of packages: There's a lot going on here. Unlike the cocktail, this is usually not the best.
```
joyeecheung: For an non-native English speaker from Eastern Asian background I feel the word "Pornstar" showing up in a technical documentation is very inappropriate, even if it seems to be just the name of a drink, but choosing a drink of this name to make a joke almost feels sexist to me (and I know many other women from my background would just call it sexist, people boycott brands for less) and makes me want to stop reading...

apps/site/pages/en/learn/modules/package-configuration.md
JakobJingleheimer: Let me think about how to fix it without losing the humour.

apps/site/pages/en/learn/modules/package-configuration.md
JakobJingleheimer: I don't know if there are any on npm's public registry. I suppose we could tell companies with huge collections of private libraries to figure it out themselves. But why?

apps/site/pages/en/learn/modules/package-configuration.md
JakobJingleheimer: > I don't recall asking about syntax detection per-se

True, you had mentioned it in regard to other things. This seemed in the same general FYI vein.

If they don't know what syntax detection is though, is that a problem? They can safely ignore it (I could add a note like "if you don't know what this is, don't worry about it"); whereas someone who is aware of it could benefit from the forewarning.

apps/site/pages/en/learn/modules/package-configuration.md
JakobJingleheimer: The article previously linked to it in the module docs, but that section has been removed. I was going to ask you if you'd consider returning that specific piece back to the API docs or a Learn article (perhaps even this article?). I think it being in that repo makes it too far removed for anyone to find it.

apps/site/pages/en/learn/modules/package-configuration.md
JakobJingleheimer: Ooow! That is a good point. Destructuring does break the reference. Buuut I think this will not?

```js
import cjs from '../cjs/index.js';

export const a = cjs.a;
export const b = cjs.b;
export const c = cjs.c;
/* … */
```

apps/site/pages/en/learn/modules/package-configuration.md
JakobJingleheimer: My first thought is yes, it should. Buuuut this is covered rather well (and succinctly) in the linked API doc, whereas the stuff I mentioned in the CJS section is not covered in the CJS/Modules API docs (at least not where I expect to find it: [`module.exports`](https://nodejs.org/api/modules.html#moduleexports)).

My only hesitation is that this article is already super long, so repeating what is already well covered in the docs seems undesirable.

If you think this has a decent chance of getting missed in the API doc and needs a shout-out here too, I'm happy to oblige :)

apps/site/pages/en/learn/modules/package-configuration.md
JakobJingleheimer: > Also an obvious cons is that one needs to manually keep the exports in-sync.

That's precisely the complicated bundler gymnastics 😅

JakobJingleheimer:
```suggestion
- Likely requires complicated bundler gymnastics to keep exports in-sync (we could not find any existing option to automate this in Webpack).
```

apps/site/pages/en/learn/modules/package-configuration.md
JakobJingleheimer: I think somewhere earlier in the article this is linked to the API doc explaining pjson type. The approach I've been taking with that is to link on the first reference only (otherwise, there could end up being a LOT of links to the same thing, which is a bit of a maintenance burden).

Do you think it's worth the trouble to add links to all occurrences of things like this?

P.S. If there isn't already a link to pjson type, I'm happy to add one. I'll look tomorrow.

apps/site/pages/en/learn/modules/package-configuration.md
JakobJingleheimer: Oof, it is an controversial name (it is indeed the name of a cocktail). I can find a different drink if these drink names are kept.

apps/site/pages/en/learn/modules/package-configuration.md
```diff
@@ -0,0 +1,563 @@
+---
+title: Package Configuration
+layout: learn
+authors: JakobJingleheimer
```
JakobJingleheimer:
```suggestion
authors: JakobJingleheimer, joyeecheung
```

@joyeecheung you've contributed quite significantly to this. Shall I add you as one of the authors?

apps/site/pages/en/learn/modules/package-configuration.md
ljharb: i believe it will still break it, because the "live binding" would be on the a, b, and c, which are `const`, and as such can't update anyways - that's not a live binding to `cjs.a`, for example.

apps/site/pages/en/learn/modules/package-configuration.md
joyeecheung: We talked about information flow during the collab summit for beginners, and I think the issue with throwing details like this in the beginning of a tutorial is that it makes the whole thing a lot more intimidating for beginners. The ideal flow I think for such a guide is:

1. If you author in CJS, ship CJS and this is how you do it.
2. If you author in ESM, ship ESM and this is how you do it if you don't care about EOL Node.js versions

At this point, >80% of the reader should get what they are looking for in the tutorial and can already leave. And the remaining <20% can continue reading about the legacy practices for supporting EOL Node.js versions and hacky ways to transpile CJS into ESM and all their pitfalls and gotchas. People who need to know how to do other methods have to know how 1 and 2 is done to get some basics going anyway, but for people who only need to know about 1 and 2 the others are just some legacy noise or hacks they may never need to know about.

joyeecheung: And, in the same vein, if we don't mention at all "what happens when you have neither main nor exports field", then it doesn't seem to worth mentioning "what happens when you have no type field". Though personally I find both rather irrelevant for a beginner tutorial that just teach people to always use type field, and one of main/exports field. The information about what happens when you don't declare them is mostly for resolver implementers, but not for most people who are just trying to configure a package correctly (which requires using these fields).

JakobJingleheimer: Hmm, that sounds like a _major_ restructuring of this huge article. Do you have a specific proposal?

apps/site/pages/en/learn/modules/package-configuration.md
ljharb: the “Long Island ice tea”?

apps/site/pages/en/learn/modules/package-configuration.md
JakobJingleheimer:
```js
const x = { a: 'hello' };

module.exports = x;

setTimeout(() => { x.a = 'bonjour' });
```
```js
export * from './index.cjs';
```
```js
import * as reexports from './reexport.mjs';

setTimeout(() => console.log(reexports));
/*
[Module: null prototype] {
  'module.exports': { a: 'bonjour' }
}
*/
```

That `'module.exports'` is quite unideal, but IIR, as designed—I can't remember the reason—it's very unergonomic, so I think it must have been a technical limitation?

apps/site/pages/en/learn/modules/package-configuration.md
JakobJingleheimer: It's commented to show it's not needed because node will pick it up automatically (assuming the most basic setup). I did it because you wanted a minimal package.json with the least config possible (this is how you get that).

apps/site/pages/en/learn/modules/package-configuration.md
JakobJingleheimer: Long Island Ice Tea is already used 😅 I chose "Pornstar Martini" because it's surprisingly packed (whereas a Long Island is famously A,MF).

apps/site/pages/en/learn/modules/package-configuration.md
joyeecheung: I don't feel comfortable adding my name to a document full of inappropriate cocktail jokes TBH.

apps/site/pages/en/learn/modules/package-configuration.md
ovflowd: For the sake of letting everyone here be comfortable, and as a mediator, let's keep references to anything that might be only for a targeted audience (be it alcohol or anything else) out of here.

I understand the intent, but for better or for worse, let's keep these out please 🙇

apps/site/pages/en/learn/modules/package-configuration.md
joyeecheung: If a pattern is a popular enough then surely they should have some representation in the public registration to be noticed. Otherwise this may just be some specific hacks that certain companies do internally while no one outside their bubbles do it, and should be left to their internal documentation. (I know certain things that Chinese companies use commonly, e.g. use of "node" in the dependencies to lock Node.js versions, but that's not something I would suggest to document in official tutorials for beginners either).

apps/site/pages/en/learn/modules/package-configuration.md
joyeecheung: I think the dual package hazard can be explained in this post directly, since most examples about dual shipping are related to it. That information certainly feels more useful compared to some of the opening paragraphs...

apps/site/pages/en/learn/modules/package-configuration.md
joyeecheung:
```suggestion
- Requires manually maintained code to keep exports in-sync.
```

I think this should be enough, at least to me I still don't know how this can be interpreted as `complicated bundler gymnastics`. This pattern is only relevant when the bundlers can't generate detectable code, so the "Likely" should be removed (it's a precondition). Also in this case the user no longer needs to worry about the patterns emitted by a bundler anymore because they are doing the chores themselves, which means more like "no bundler gymnastics required, because we are not relying on bundler to output detectable code at all anymore" to me.

apps/site/pages/en/learn/modules/package-configuration.md
joyeecheung: `"module.exports"` on an imported namespace is new and is not something that users should rely on, it's a marker meant for bundlers, because there were some claims (https://github.com/nodejs/node/pull/53848) that adding this to the imported CJS facade helps bundlers that transpile CJS -> ESM, though I have yet to see any living bundlers actually proving that claim. As far as a non-bundler user is concerned, this is as good as an empty object.

joyeecheung: Also, if the original CJS is assigning `exports.xxx`, then using an ESM wrapper that does `export *` won't help you with live binding either, because the CJS module facade uses a SyntheticModule that only evaluates and assigns the named exports once during evaluation, so it's as good as "export const a = cjs.a" - still broken.

apps/site/pages/en/learn/modules/package-configuration.md
joyeecheung: I meant that this only shows `"type": "module"` and `.cjs` files in the package.json. It should explain that somewhere in this package's source form, there are `.js` files containing ESM. Otherwise one wonders why is  `"type": "module"`  needed at all (you could also have all your source files in `.mjs` and then ship `"type": "commonjs"` + `".js"` containing CJS, which was what the "almost identical" example before was doing)

apps/site/pages/en/learn/modules/package-configuration.md
joyeecheung: I agree that no cocktail jokes belong here.

(Also FYI "Long Island Ice Tea" is associated with some myths about rapist traps for women in my background - not everyone knows that it would taste alcoholic enough that you probably won't be tricked, but I also know people who would think this is promoting rapist tricks that trivialize alcohol content in cocktails. In general, using references that are biased towards an audience in a way that has nothing to do with the primary topic of a technical article is just a mistake IMO).

JakobJingleheimer: @ovflowd as can be seen from my previous https://github.com/nodejs/nodejs.org/pull/7229#discussion_r1864679437 a mediator is not necessary.

apps/site/pages/en/learn/modules/package-configuration.md
joyeecheung: I think in practice, removing the main field already means extra configuration for people. From what I can tell, at least >50% of the high impact packages use the main field instead of the exports field, which is likely due to the fact that `npm init` just adds the main field by default, so removing that requires extra actions (and knowledge about what that entails). I would say the "minimal configuration" should be using the main field instead.

apps/site/pages/en/learn/modules/package-configuration.md
joyeecheung: > My only hesitation is that this article is already super long

I think the article is strangely prioritizing information that aren't super helpful (complaints about what a nightmare it is, cocktail jokes, history lessons, implementation details about features that most examples here would not hit) over actually useful information (how do you make ESM consumed properly by CJS). If length is a concern, I would suggest something that I suggested earlier - put the most useful patterns at front (CJS -> CJS, ESM -> ESM), and the obsolete practice in the back, or even in a separate article.

apps/site/pages/en/learn/modules/package-configuration.md
JakobJingleheimer: Ah, okay. So I think it is indeed impossible so best to document the limitation. There are better options anyway.

apps/site/pages/en/learn/modules/package-configuration.md
marco-ippolito: Id remove personal takes, agree with Joyee

JakobJingleheimer:
@ovflowd it looks like the table is not rendering in the article; I switched it from markdown to html so it could leverage `rowspan`. That's allowed in markdown, but the site seems to not like it. Is there something I need to do to get it to play nice?

apps/site/pages/en/learn/modules/package-configuration.md
JakobJingleheimer: Done

ovflowd:
> @ovflowd it looks like the table is not rendering in the article; I switched it from markdown to html so it could leverage `rowspan`. That's allowed in markdown, but the site seems to not like it. Is there something I need to do to get it to play nice?

Rename file to .mdx -- use props as react (ie rowSpan)

JakobJingleheimer:
References to drinks have been removed and the article re-arranged to put the most common options at the start. Perhaps subsequently, move the other options into a separate article.

apps/site/pages/en/learn/modules/package-configuration.mdx
```diff
@@ -0,0 +1,597 @@
...
+
+### ESM source and both CJS & ESM distribution
+
+When source code is written in non-JavaScript (ex TypeScript), options can be limited due to needing to use file extension(s) specific to that language (ex `.ts`) and there may be no `.mjs` equivalent.
+
+Similar to [CJS source and both CJS & ESM distribution](#cjs-source-and-both-cjs-amp-esm-distribution), you have the same options.
+
+#### Publish only a CJS distribution with property exports
+
+The "Mojito" of packages: Tricky to make and needs good ingredients.
```
marco-ippolito: Is this a cocktail reference?

apps/site/pages/en/learn/modules/package-configuration.mdx
```diff
@@ -0,0 +1,597 @@
...
+    "./package.json": "./package.json"
+  }
+}
+```
+
+💡 Using `"type": "module"`[^2] paired with the `.cjs` file extension (for commonjs files) yields best results. For more information on why, see [Down the rabbit-hole](#down-the-rabbithole) and [Gotchas](#gotchas) below.
+
+#### Publish a CJS distribution with an ESM wrapper
+
+The "Pornstar Martini" of packages: There's a lot going on here. Unlike the cocktail, this is usually not the best.
```
marco-ippolito: Same here

apps/site/pages/en/learn/modules/package-configuration.mdx
```diff
@@ -0,0 +1,597 @@
...
+    "./package.json": "./package.json"
+  }
+}
+```
+
+💡 Using `"type": "module"`[^2] paired with the `.cjs` file extension (for commonjs files) yields best results. For more information on why, see [Down the rabbit-hole](#down-the-rabbithole) and [Gotchas](#gotchas) below.
+
+#### Publish both full CJS & ESM distributions
+
+The "Tokyo Tea" of packages: Chuck in a bunch of stuff (with a surprise) and hope for the best. This is probably the most common and easiest of the ESM to CJS & ESM options, but you pay for it. This is rarely a good idea.
```
marco-ippolito: Ditto

apps/site/pages/en/learn/modules/package-configuration.mdx
JakobJingleheimer: What the…I removed these…

Thanks for catching!

apps/site/pages/en/learn/modules/package-configuration.mdx
```diff
@@ -0,0 +1,597 @@
+---
+title: Package Configuration
+layout: learn
+authors: JakobJingleheimer
+---
+
+# Package configuration
+
+All the provided `package.json` configurations (not specifically marked “does not work”) work in Node.js 12.22.x (v12 latest, the oldest supported line) and 17.2.0 (current latest at the time)[^1], and for grins, with webpack 5.53.0 and 5.63.0 respectively. These are available: [nodejs/package-examples](https://github.com/nodejs/package-examples/blob/main/config).
```
marco-ippolito: this link https://github.com/nodejs/package-examples/blob/main/config is a 404 for me

apps/site/pages/en/learn/modules/package-configuration.mdx
```diff
@@ -0,0 +1,597 @@
+---
+title: Package Configuration
+layout: learn
+authors: JakobJingleheimer
+---
+
+# Package configuration
```
marco-ippolito: I'd start the guide with the topic this article is talking about:
How to publish a package, which might be obvious for some but is not clearly stated

marco-ippolito (commented):
I have the feeling this article is not very beginner friendly and there are so many caveats and notes along the way that make it very hard to follow for someone who is not 100% into it

apps/site/pages/en/learn/modules/package-configuration.mdx
```diff
@@ -0,0 +1,597 @@
...
+      <td><a href="#esm-source-with-only-cjs-distribution">ESM source with only CJS distribution</a></td>
+    </tr>
+    <tr>
+      <td>CJS & ESM: consumers either <code>require()</code> or <code>import</code> your package</td>
+      <td><a href="#esm-source-and-both-cjs-amp-esm-distribution">ESM source and both CJS & ESM distribution</a></td>
+    </tr>
+  </tbody>
+</table>
+
+### General notes
```
marco-ippolito: Id move the general notes section at the end since it breaks the reading flow. you just explained there are 2 ways of publishing, lets dive into them and add considerations at the end. Also it requires to know some concepts that are explained after.

apps/site/pages/en/learn/modules/package-configuration.mdx
```diff
@@ -0,0 +1,597 @@
...
+
+`"exports"` can be advisable over [`"main"`](https://nodejs.org/api/packages.html#main) because it prevents external access to internal code (so you can be relatively sure users are not depending on things they shouldn't). If you don't need that, `"main"` is simpler and may be a better option for you.
+
+The `"engines"` field provides both a human-friendly and a machine-friendly indication of which version(s) of Node.js the package is compatible. Depending on the package manager used, an exception may be thrown causing the installation to fail when the consumer is using an incompatible version of Node.js (which can be very helpful to consumers). Including this field will save a lot of headache for consumers with an older version of Node.js who cannot use the package.
+
+### CJS source and distribution
+
+You _technially_ may not need _any_ package configuration beyond [`"name"`](https://nodejs.org/api/packages.html#name). But the less arcane, the better: Essentially just declare the package’s exports via the `"exports"` field/field-set.
```
marco-ippolito:
```suggestion
You _technially_ may not need _any_ package configuration beyond [`"name"`](https://nodejs.org/api/packages.html#name). But the less arcane, the better: Essentially just declare the package’s exports via the `"exports"` field/field-set.
```

I'd rather say:
The most minimal configuration is: (just names) but it's better to explictly declare exports because....

apps/site/pages/en/learn/modules/package-configuration.mdx
```diff
@@ -0,0 +1,597 @@
...
+}
+```
+
+Note that `packageJson.exports["."] = filepath` is shorthand for `packageJson.exports["."].default = filepath`
+
+### ESM source and distribution
+
+Simple, tried, and true.
+
+Note that since Node.js v23.0.0, it is possible to `require` static ESM (code that does not use top-level `await`). See [Loading ECMAScript modules using `require()`](https://nodejs.org/api/modules.html#loading-ecmascript-modules-using-require) for details.
```
marco-ippolito: I'd move the note to the end of the paragraph with other notes

apps/site/pages/en/learn/modules/package-configuration.mdx
```diff
@@ -0,0 +1,597 @@
...
+    "./package.json": "./package.json"
+  }
+}
+```
+
+Note that ESM now _is_ “backwards” compatible with CJS: a CJS module now _can_ [`require()` an ES Module](https://nodejs.org/api/modules.html#loading-ecmascript-modules-using-require) without a flag as of 23.0.0 and 22.12.0.
+
+### CJS source and only ESM distribution
+
+This takes a small bit of finesse but is also pretty straight-forward.
```
marco-ippolito: This section contains little to no explanation other than examples.
I'd write some explanation at the beginning

apps/site/pages/en/learn/modules/package-configuration.mdx
```diff
@@ -0,0 +1,597 @@
...
+    "./package.json": "./package.json"
+  }
+}
+```
+
+The [`.mjs`](https://nodejs.org/api/esm.html#enabling) file extension is a trump-card: it will override **any** other configuration and the file will be treated as ESM. Using this file extension is necessary because `packageJson.exports.import` does **NOT** signify that the file is ESM (contrary to common, if not universal, misperception), only that it is the file to be used when the package is imported (ESM _can_ import CJS. See [Gotchas](#gotchas) below).
+
+### CJS source and both CJS & ESM distribution
+
+You have a few options:
```
marco-ippolito: A paragraph imho should start with some explaination about the content (example)
To be able to distrubute you CJS source code in ESM &  CJS you have a few options:

apps/site/pages/en/learn/modules/package-configuration.mdx
```diff
@@ -0,0 +1,597 @@
...
+
+### ESM source and both CJS & ESM distribution
+
+When source code is written in non-JavaScript (ex TypeScript), options can be limited due to needing to use file extension(s) specific to that language (ex `.ts`) and there may be no `.mjs` equivalent.
+
+Similar to [CJS source and both CJS & ESM distribution](#cjs-source-and-both-cjs-amp-esm-distribution), you have the same options.
+
+#### Publish only a CJS distribution with property exports
+
+Tricky to make and needs good ingredients.
```
marco-ippolito:
```suggestion
```
I'd also remove all these catch phrases since imho they don't add too much value

apps/site/pages/en/learn/modules/package-configuration.mdx
```diff
@@ -0,0 +1,597 @@
...
+- Easy and simple (probably least effort if you don't mind keeping to a minor syntax stipulation)
+- Precludes [the Dual-Package Hazard](#the-dual-package-hazard)
+
+Cons:
+
+- Requires very specific syntax (either in source code and/or bundler gymnastics).
+
+**Working example**: [cjs-with-dual-distro (properties)](https://github.com/JakobJingleheimer/nodejs-module-config-examples/tree/main/packages/cjs/dual/property-distro)
+
+```json displayName="Minimal package.json"
```
marco-ippolito: I'd move example first so the reader understand what we are talking about and then pros and cons, also I'd put some explanation at the beginning of the paragraph on what `Attach named exports directly onto `exports`` means

apps/site/pages/en/learn/modules/package-configuration.mdx
JakobJingleheimer: Do you mean move it to a footnote?

marco-ippolito: I mean at the end of the paragraph

apps/site/pages/en/learn/modules/package-configuration.mdx
JakobJingleheimer: People have told me they're funny

JakobJingleheimer:
Thanks for the polish suggestions! I have a subsequent Learn article I need to write though, which depends on this. Could we set v1.0.0 goals for this article, and then subsequently improve it?

JakobJingleheimer:
I haven't heard anything in a few days, so I assume there are no more requests. I'm gonna merge. If there are others, please let me know and we can iterate on the article.

PR #7241, opened 2024-11-19
chore: use `toUpperCase` in `detectOS`

CodingRule:
Instead of using the switch statement, we use a faster way with an object. Hope it's not useless tho.

### DIFF
--- a/apps/site/util/detectOS.ts
+++ b/apps/site/util/detectOS.ts
@@ -6,7 +6,6 @@ export const detectOsInUserAgent = (userAgent: string | undefined): UserOS => {
   return osMatch ? (osMatch[1].toUpperCase() as UserOS) : 'OTHER';
 };
 
-
 // Since `navigator.appVersion` is deprecated, we use the `userAgent``
 // https://developer.mozilla.org/en-US/docs/Web/API/Navigator/appVersion
 export const detectOS = (): UserOS => detectOsInUserAgent(navigator?.userAgent);
--- a/apps/site/util/detectOS.ts
+++ b/apps/site/util/detectOS.ts
@@ -1,21 +1,16 @@
 import type { UserOS } from '@/types/userOS';
 
 export const detectOsInUserAgent = (userAgent: string | undefined): UserOS => {
-  const osMatch = userAgent?.match(/(Win|Mac|Linux)/);
-  switch (osMatch && osMatch[1]) {
-    case 'Win':
-      return 'WIN';
-    case 'Mac':
-      return 'MAC';
-    case 'Linux':
-      return 'LINUX';
-    case 'AIX':
-      return 'AIX';
-    default:
-      return 'OTHER';
-  }
+  const osMatch = userAgent?.match(/(Win|Mac|Linux|AIX)/);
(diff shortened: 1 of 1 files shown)


CodingRule:
Also I can change it at utils/getNodeDownloadUrl.ts too

apps/site/util/detectOS.ts
```diff
@@ -1,21 +1,16 @@
...
-      return 'OTHER';
-  }
+  const osMatch = userAgent?.match(/(Win|Mac|Linux|AIX)/);
+  const osMap: Record<string, UserOS> = {
+    Win: 'WIN',
+    Mac: 'MAC',
+    Linux: 'LINUX',
+    AIX: 'AIX',
+  };
+  return osMap[osMatch?.[1] ?? ''] || 'OTHER';
```
avivkeller: if we are already matching, why not just use toUpperCase if we find a match, instead of a map?

apps/site/util/detectOS.ts
```diff
@@ -1,21 +1,16 @@
...
 };
 
-// Since `navigator.appVersion` is deprecated, we use the `userAgent``
+// Since navigator.appVersion is deprecated, we use the userAgent
```
avivkeller: Please revert this change

CodingRule: ok!

CodingRule: I did revert it I hope It updated.

apps/site/util/detectOS.ts
CodingRule: 1 sec Ill be doing that rn

CodingRule:
@RedYetiDev Done Thanks for pointing out to me that I can directly do it with uppercase and I have updated it in your sayings. Thanks! Hope this contribution will help.

apps/site/util/detectOS.ts
```diff
@@ -1,21 +1,12 @@
...
-      return 'AIX';
-    default:
-      return 'OTHER';
-  }
+  // Match OS names and convert to uppercase directly if there's a match
+  const osMatch = userAgent?.match(/(Win|Mac|Linux|AIX)/);
+  return osMatch ? (osMatch[1].toUpperCase() as UserOS) : 'OTHER';
 };
 
+
```
avivkeller:
```suggestion
```

avivkeller (approved):
See my one comment, then LGTM

CodingRule:
Haha thanks! I saw that one coming tbh (=

CodingRule:
> See my one comment, then LGTM

thanks good sir!

CodingRule:
also quick question @RedYetiDev would I be able to apply the same logic here too?

utils/getNodeDownloadUrl.ts too

avivkeller:
In a seperate PR, preferably

CodingRule:
yes ofc

aymen94 (approved):

avivkeller:
Thanks for your contribution, @CodingRule! It's be merged shortly.

CodingRule:
Ty!

PR #7242, opened 2024-11-19
Update getNodeDownloadUrl.ts

CodingRule:
Technically the same thing I did to PR #7241

### DIFF
--- a/apps/site/util/getNodeDownloadUrl.ts
+++ b/apps/site/util/getNodeDownloadUrl.ts
@@ -10,64 +10,45 @@ export const getNodeDownloadUrl = (
   const baseURL = `${DIST_URL}${versionWithPrefix}`;
 
   if (kind === 'source') {
+    // Prepares a downloadable Node.js source code link
     return `${baseURL}/node-${versionWithPrefix}.tar.gz`;
   }
 
-  switch (os) {
-    case 'MAC':
-      // Prepares a downloadable Node.js installer link for the x64, ARM64 platforms
-      if (kind === 'installer') {
-        return `${baseURL}/node-${versionWithPrefix}.pkg`;
-      }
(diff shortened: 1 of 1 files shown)


CodingRule:
If the comments are wrote wrongly please tell me!

apps/site/util/getNodeDownloadUrl.ts
avivkeller, line 1: Looking at the comparison, I think the switch statement is much cleaner.

CodingRule: Ah okay

CodingRule: Any good paths I could take into making this better than the switch cases ? (;

avivkeller: What's wrong with the switch statement?

CodingRule: Yeah true ig.
Thx anyway for review

avivkeller: No problem! Feel free to look at the issues tab for things that need to be fixed. "Help wanted" and "good first issue" are places to start.

AugustinMauroy: > Looking at the comparison, I think the switch statement is much cleaner.

Same feeling

avivkeller (changes requested):
Based on my comment above, I dont think this should land in its current form.

ovflowd (commented):
Not gonna lie, I actually was a fan of this PR, it looked way cleaner than the switch statement :)

Feel free to reopen it.

avivkeller:
I'm still -1. I think that the switch statement is cleaner, but if you disagree, I won't block.

PR #7266, opened 2024-11-20
meta: skip lighthouse CI when no vercel preview detected

avivkeller:
## Description

This changes the Lighthouse CI to skip when a Vercel preview is not detected

## Related Issues

#7253

### DIFF
--- a/.github/workflows/lighthouse.yml
+++ b/.github/workflows/lighthouse.yml
@@ -27,12 +27,39 @@ permissions:
   pull-requests: write
 
 jobs:
-  lighthouse-ci:
+  get-vercel-preview:
     # We want to skip our lighthouse analysis on Dependabot PRs
     if: |
       startsWith(github.event.pull_request.head.ref, 'dependabot/') == false &&
       github.event.label.name == 'github_actions:pull-request'
+    name: Get Vercel Preview
+    runs-on: ubuntu-latest
+    outputs:
+      deployment_found: ${{ steps.check_deployment.outputs.deployment_found }}
+      url: ${{ steps.check_deployment.outputs.url }}
+    steps:
+      - name: Capture Vercel Preview
+        id: check_deployment
+        uses: patrickedqvist/wait-for-vercel-preview@06c79330064b0e6ef7a2574603b62d3c98789125 # v1.3.2
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
+          max_timeout: 300 # timeout after 5 minutes
+          check_interval: 10 # check every 10 seconds
+        continue-on-error: true
+      - name: Set Outputs
+        if: always()
+        id: set_outputs
+        run: |
+          if [[ -z "${{ steps.check_deployment.outputs.url }}" ]]; then
+            echo "deployment_found=false" >> $GITHUB_OUTPUT
+          else
+            echo "deployment_found=true" >> $GITHUB_OUTPUT
+            echo "url=${{ steps.check_deployment.outputs.url }}" >> $GITHUB_OUTPUT
+          fi
 
+  lighthouse-ci:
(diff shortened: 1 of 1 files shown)


avivkeller:
The commit message didn't save :thinking:, regardless, the PR title will be used, right? If not, feel free to force push and fix it.

avivkeller:
image
 Without a Vercel preview, the Lighthouse CI is skipped, as shown above.

.github/workflows/lighthouse.yml
```diff
@@ -76,11 +93,11 @@ jobs:
...
-            ${{ steps.vercel_preview_url.outputs.url }}/en
-            ${{ steps.vercel_preview_url.outputs.url }}/en/about
-            ${{ steps.vercel_preview_url.outputs.url }}/en/about/previous-releases
-            ${{ steps.vercel_preview_url.outputs.url }}/en/download
-            ${{ steps.vercel_preview_url.outputs.url }}/en/blog
+            ${{ needs.get-vercel-preview.outputs.url }}/en
+            ${{ needs.get-vercel-preview.outputs.url }}/en/about
+            ${{ needs.get-vercel-preview.outputs.url }}/en/about/previous-releases
+            ${{ needs.get-vercel-preview.outputs.url }}/en/download
+            ${{ needs.get-vercel-preview.outputs.url }}/en/blog
```
AugustinMauroy, line 100: need to add learn here

avivkeller: Shouldn't that be a followup, given that this is specifically for skipping the workflow?

.github/workflows/lighthouse.yml
```diff
@@ -27,12 +27,39 @@ permissions:
   pull-requests: write
 
 jobs:
-  lighthouse-ci:
+  get-vercel-preview:
     # We want to skip our lighthouse analysis on Dependabot PRs
```
bmuenzenmeyer, line 31: this comment needs to move with the ordering

.github/workflows/lighthouse.yml
bmuenzenmeyer: this list was semi-intelligently designed not to be exhaustive of all pages, but to try to hit the major features

- homepage (global)
- markdown rendering (about)
- previous-releases (the release data / table)
- download (release data / os detection / lots of complexity)
- blog (blog data / rendering)

AugustinMauroy: Imo we should test the learn section to ensure everything is right. Yeah it's look like about but it's not same eg there are avatars in the learn page

.github/workflows/lighthouse.yml
avivkeller: What do you mean? It's commenting on the line below (skipping the Dependabot).

bmuenzenmeyer: to me the comment nesting seems to suggest it is documenting the job directly below it, which is no longer lighthouse but get vercel preview.

it is probably just a small nit - sorry for noise

.github/workflows/lighthouse.yml
ovflowd: Lighthouse is not about everything being right, but sampling some of it and measuring performance. Realistically speaking we don't want to sample everything nor people should take it too seriously.

bmuenzenmeyer: thanks @ovflowd - i've had this open all day struggling to find the words. 
if we ever get to that point, we should use playwright to make e2e tests. i suspect chromatic could do more too

bmuenzenmeyer (approved):

avivkeller:
👋 follow-up https://github.com/nodejs/nodejs.org/pull/7361

PR #7279, opened 2024-11-23
feat(learn): add article for publishing a typescript package

JakobJingleheimer:
## Description

Document the recommended way to publish a typescript package

## Related Issues

https://github.com/nodejs/typescript/pull/19

Depends on https://github.com/nodejs/nodejs.org/pull/7229

### Check List


AugustinMauroy:
Jacob you should:
-  remove lint thing
- explain `tsc` cli (may be redirect to the previous page)
- link to `--run` https://nodejs.org/en/learn/command-line/run-nodejs-scripts-from-the-command-line#run-a-task-with-nodejs
- remove comment on GA example `# glob is not backported below 22.x` that from nodejs-loaders 😁
- Also on GA example you put `node --run test`, but on `package.json` example it's missing
- If you keep `test` thing maybe add link to the test runner part.

apps/site/pages/en/learn/typescript/publishing.md
```diff
@@ -0,0 +1,183 @@
...
+    strategy:
+      matrix:
+        node:
+          - version: 23.x
+          - version: 22.x
+          # glob is not backported below 22.x
+      fail-fast: false # prevent a failure in one version cancelling other runs
+
+    steps:
+      - uses: actions/checkout@v4
```
ljharb: these version refs will likely get outdated pretty fast; what’s the plan for keeping them updated?

apps/site/pages/en/learn/typescript/publishing.md
```diff
@@ -0,0 +1,183 @@
...
+
+```json displayName="package.json"
+{
+  "version": "0.0.0",
+  "name": "example-ts-pkg",
+  "scripts": {
+    "lint": "…",
+    "types:check": "tsc --noEmit"
+  },
+  "optionalDependencies": {
```
ljharb: this is incorrect; optional deps are runtime deps, and will always be installed - it just won’t fail the overarching install if it fails to compile. This will definitely install typescript locally, and for all consumers. There is no way to have optional dev deps, and “optional” wouldn’t be the same as it means for peerDependenciesMeta.

apps/site/pages/en/learn/typescript/publishing.md
```diff
@@ -0,0 +1,183 @@
...
+}
+```
+
+```json displayName="tsconfig.json"
+{
+  "compilerOptions": {
+    "declarationMap": true,
+    "declaration": true,
+    "emitDeclarationOnly": true,
+    "esModuleInterop": true, // Flux Capacitor: The universe breaks without it, but nobody knows what it does.
```
ljharb: it makes ts’s module system not be horribly broken :-)

apps/site/pages/en/learn/typescript/publishing.md
```diff
@@ -0,0 +1,183 @@
...
+
+      # You can probably ignore the boilerplate config above
+
+      - name: Generate types
+        run: node --run types:generate
+
+      - name: Publish with provenance
+        env:
+          NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
+        run: npm publish --access public --provenance
```
ljharb: this publishes with one factor, which is much less secure than publishing with 2fa; i strongly suggest node not in any way encourage doing this.

karlhorky: Is there any way to set up automated version publishing with 2FA? That would be ideal - automated and secure.

ljharb: No, not without an external server - until GitHub adds a native way for a workflow to pause and wait for user input to continue.

AugustinMauroy: Oula, I don't agree with you 100%. First of all you can ask for the 2FA for the provenance depending on the NPM token used. Also I think Node.js should recommend packages with provenance for security reasons.

JakobJingleheimer:
Thank you all, but if I may direct your attention to the huge note at the top of the PR 😜

This is not ready for feedback

Feedback is very welcome soon 🙏 but not now. Ex the GHA samples are copy+pasted from another project and have not been updated yet.

apps/site/pages/en/learn/typescript/publishing.md
JakobJingleheimer: It's possible via [`--omit`](https://docs.npmjs.com/cli/v11/commands/npm-install#omit), I just haven't got to that yet :)

apps/site/pages/en/learn/typescript/publishing-a-ts-project.md
```diff
@@ -0,0 +1,225 @@
...
+  "name": "example-ts-pkg",
+  "scripts": {
+    "test": "node --test",
+    "types:check": "tsc --noEmit"
+  },
+  "optionalDependencies": {
+    // This is used only in CI.
+    // Avoid bloating your local node_modules
+    // (where you probably won't use it) by
+    // running `npm install --omit="optional"`
```
AugustinMauroy: this must be written in the markdown not in the code snippet because if the user want to copy past it's will not work

apps/site/navigation.json
```diff
@@ -209,6 +209,10 @@
             "runNatively": {
               "link": "/learn/typescript/run-natively",
               "label": "components.navigation.learn.typescript.links.runNatively"
+            },
+            "publishingTSProject": {
+              "link": "/learn/typescript/publishing-a-ts-project",
+              "label": "components.navigation.learn.typescript.links.publishing-a-ts-project"
```
AugustinMauroy:
```suggestion
              "label": "components.navigation.learn.typescript.links.publishingTSProject"
```

apps/site/pages/en/learn/typescript/publishing.md
ljharb: I'm not sure what you mean - the only way to do two-factor is with two factors. Using an npm token is only ever a single factor.

As for provenance, I'm happy to discuss that separately, but whether it has value on its own or not is irrelevant - having 2FA is universally recognized as table stakes for security, which is why github and npm are making it required for everyone. They're not even planning to ever require provenance information, which should tell you about the relative security value of each.

apps/site/pages/en/learn/typescript/publishing.md
ljharb: For an application, sure, but then doesn't that force consumers to use it?

JakobJingleheimer: No, why? As you say, it won't do anything magical on its own.

apps/site/pages/en/learn/typescript/publishing.md
JakobJingleheimer: @ljharb could you point me to some docs on the alternative? From googling, I just get instructions on how to set up 2FA on my npm account (which I already have), or [use it interactively via CLI](https://docs.npmjs.com/configuring-two-factor-authentication). And on the GHA side, I get only info about adding 2FA to my GH account (again, already have it).

But if you say GHA can't wait for something, I don't see how it's possible.

ljharb: The only secure alternative is to publish locally, unless you have the resources to maintain an external server (like https://github.com/GoogleCloudPlatform/wombat-dressing-room) or unless you want to trust a third party server (like https://github.com/step-security/wait-for-secrets).

apps/site/pages/en/learn/typescript/publishing.md
ljharb: Anything in `optionalDependencies` will automatically be installed for consumers precisely the same as if it's in `dependencies`.

JakobJingleheimer: Yes…? Exactly?

apps/site/pages/en/learn/typescript/publishing.md
JakobJingleheimer: After asking around, I'm not getting a clear answer here. One security expert said use 2FA to secure the account and provenance to authenticate the release (slightly paraphrased, because it wasn't in English).

This _seems_ logical and sufficient to me because you can't do the things triggered with the token without first passing the 2FA on the account doing them.

@nodejs/security could you please advise? Both things seem important yet seemingly cannot be combined during a publish step.

apps/site/pages/en/learn/typescript/publishing.md
ljharb: and why would typescript be a runtime dependency? it should only ever be a peerDependency of a package, at most, and a package that isn't directly about typescript shouldn't be requiring consumers to install typescript at all.

apps/site/pages/en/learn/typescript/publishing.md
ljharb: Unfortunately, to do both is impossible at the moment due to a limitation of all CI systems I'm aware of, and because npm provenance only works from a "trusted" CI system. It's possible GitLab offers this capability, but since Github doesn't, I'm not sure how it makes sense to recommend users do it.

ljharb: also, the way npm publish/automation tokens work, 2FA is _bypassed entirely_ - that's my point.

apps/site/pages/en/learn/typescript/publishing-a-ts-project.md
```diff
@@ -0,0 +1,225 @@
...
+
+# Publishing a TypeScript project
+
+This article augments TypeScript's own [Publishing guide](https://www.typescriptlang.org/docs/handbook/declaration-files/publishing.html) with specifics for native node support.
+
+Some important things to note:
+
+- Everything from [Publishing a package](../modules/publishing-a-package) applies here.
+
+- Node runs TypeScript code via a process called "[type stripping](https://nodejs.org/api/typescript.html#type-stripping)", wherein node (via [Amaro](https://github.com/nodejs/amaro)) removes TypeScript-specific syntax, leaving behind vanilla JavaScript (which node already understands). This behaviour is enabled by default of node version 23.6.0.
```
mcollina: This will become outdated once we backport to v22 (it will eventually happen).

apps/site/pages/en/learn/typescript/publishing-a-ts-project.md
```diff
@@ -0,0 +1,225 @@
...
+
+  - Node does **not** strip types in `node_modules` because it can cause significant performance issues for the official TypeScript compiler (`tsc`), so the TypeScript maintainers would like to discourage people publishing raw TypeScript, at least for now.
```
mcollina: I would add that it slow down the VSCode language server as  well.

apps/site/pages/en/learn/typescript/publishing-a-ts-project.md
```diff
@@ -0,0 +1,225 @@
...
+```json displayName="package.json"
+{
+  "name": "example-ts-pkg",
+  "scripts": {
+    "test": "node --test './src/**/*.test.ts'",
+    "types:check": "tsc --noEmit"
+  },
+  "optionalDependencies": {
+    "typescript": "^5.7.2"
+  }
```
mcollina: Why is this an optionalDependency?

apps/site/pages/en/learn/typescript/publishing-a-ts-project.md
```diff
@@ -0,0 +1,225 @@
...
+    "resolvePackageJsonImports": true,
+    "target": "ESNext"
+  },
+  // These may be different for your repo:
+  "include": ["./src"],
+  "exclude": ["**/*/*.test.*", "**/fixtures/**"]
+}
+```
+
+Pro-tip: The TypeScript executable (`tsc`) is likely used only in CI. Avoid bloating your local node_modules (where you probably won't use it) by adding [`--omit="optional"`](https://docs.npmjs.com/cli/v11/commands/npm-install#omit) when you run `npm install` locally: `npm install --omit="optional"`
```
mcollina: This is actually not something I would recommend. Move typescript to a devDependency

apps/site/pages/en/learn/typescript/publishing-a-ts-project.md
JakobJingleheimer: For the same reason, right?

apps/site/pages/en/learn/typescript/publishing-a-ts-project.md
JakobJingleheimer: For the reasons discussed in this PR 😅

apps/site/pages/en/learn/typescript/publishing-a-ts-project.md
JakobJingleheimer: Why would you not recommend?

apps/site/pages/en/learn/typescript/publishing-a-ts-project.md
mcollina: yes exactly

apps/site/pages/en/learn/typescript/publishing-a-ts-project.md
AugustinMauroy: ref: https://github.com/nodejs/nodejs.org/pull/7279#discussion_r1899161376

apps/site/pages/en/learn/typescript/publishing-a-ts-project.md
AugustinMauroy: IMO, optional is something not needed for the project. Dev is used for dev part and classic for runtime. So to transpile we need to put it in dev

mcollina: Exactly that. Otherwise ts will be installed every time.

marco-ippolito: I agree we should advice it as devDependency

apps/site/pages/en/learn/typescript/publishing.md
styfle: @JakobJingleheimer Can we change it to devDependencies instead?

apps/site/pages/en/learn/typescript/publishing-a-ts-project.md
JakobJingleheimer: The people have spoken 😅👍

apps/site/pages/en/learn/typescript/publishing.md
JakobJingleheimer: I think "bypassed _entirely_", whilst technically possible, is not realistic. Yes, GitHub has the token and _could_ do it. That possibility seems…remote. As you say, it is a trusted CI system. Aside from that, both sides are guarded by 2FA, and left or right are the only realistic access-points—when both actors are trusted, it seems reasonable to trust whatever they're doing, and provenance assures there has been no slight of hand / man-in-the-middle.

If we could have both, yeah, let's do it!

I don't know enough to authoritatively say what we should do since we apparently must choose one or the other, so I think we should defer to our own team of security experts (@marco-ippolito 👀).

ljharb: GitHub could be exploited, separate from whether one of the tens of thousands of Microsoft employees was a bad actor.

I’m suggesting node simply not recommend a publishing method rather than wade into a conversation that industry security experts don’t even agree on.

JakobJingleheimer: Let's give the security team a chance to weigh in. If we don't get a definitive answer, then let's come up with an alternative (I would be inclined to list both options and, possibly in an addendum, explain some of the key advantages and disadvantages—especially since this info was difficult for even us to gather).

marco-ippolito: Generally speaking I'd omit giving an advice on this matter completely and leave the user to do their own research on how to publish on npm since its out of the scope of this guide. Ssince there is no out of the box alternative I'm ok with suggesting 1F (which we use on many of our @nodejs packages)

apps/site/pages/en/learn/typescript/publishing-a-ts-project.md
himself65, line 1: we should alert that the published npm pacakge is still run with javascript +dts files?

apps/site/pages/en/learn/typescript/publishing-a-ts-project.md
```diff
@@ -0,0 +1,224 @@
...
+  }
+}
+```
+
+```json displayName="tsconfig.json"
+{
+  "compilerOptions": {
+    "allowArbitraryExtensions": true,
+    "allowImportingTsExtensions": true,
+    "baseUrl": "./",
```
andrewbranch: This allows TS to resolve `"foo.js"` in place of `"./foo.js"` and so is incorrect for Node.js. (It was made for AMD loaders in the browser and isn’t recommended for anyone anymore.)

apps/site/pages/en/learn/typescript/publishing-a-ts-project.md
JakobJingleheimer: Isn't that covered at the beginning and also in "Generate type declarations"?

apps/site/pages/en/learn/typescript/publishing-a-ts-project.md
```diff
@@ -0,0 +1,224 @@
...
+    "declaration": true,
+    "declarationMap": true,
+    "esModuleInterop": true, // Flux Capacitor: The universe breaks without it, but nobody knows exactly what it does.
```
andrewbranch: `esModuleInterop`, `moduleResolution`, `resolvePackageJsonExports`, `resolvePackageJsonImports`, and `target` are all redundant here with `"module": "NodeNext"`. I personally usually drop every option I can because tsconfigs are big and complicated enough already 😄

apps/site/pages/en/learn/typescript/publishing-a-ts-project.md
```diff
@@ -0,0 +1,224 @@
...
+    "typescript": "^5.7.2"
+  }
+}
+```
+
+```json displayName="tsconfig.json"
+{
+  "compilerOptions": {
+    "allowArbitraryExtensions": true,
+    "allowImportingTsExtensions": true,
```
andrewbranch: This needs to be `rewriteRelativeImportExtensions` instead of `allowImportingTsExtensions`. The latter only prevents TS from erroring on `import {} from "./foo.ts"`, but the emitted JavaScript will still contain `"import {} from "./foo.ts"`, which will crash. With the former option, the import will be rewritten to `import {} from "./foo.js"`. This is the only way (with `tsc`) the package author can run their TS files in Node.js while still getting functional JS files for publishing. The option is only available in TypeScript 5.7+, which probably deserves a mention somewhere in here.

apps/site/pages/en/learn/typescript/publishing-a-ts-project.md
```diff
@@ -0,0 +1,224 @@
...
+    "baseUrl": "./",
+    "declaration": true,
+    "declarationMap": true,
+    "esModuleInterop": true, // Flux Capacitor: The universe breaks without it, but nobody knows exactly what it does.
+    "lib": ["ESNext"],
+    "module": "NodeNext",
+    "moduleResolution": "NodeNext",
+    "outDir": "./",
```
andrewbranch: This is the default, but we generally recommend setting `outDir` to something else. (`"rootDir": "src", "outDir": "dist"` is common.)

andrewbranch (dismissed):
Thanks for writing this up! I have some tsconfig feedback. Also, is it worth mentioning anything about package.json fields (i.e. that they should reference `.js` files, not `.ts` files)?

apps/site/pages/en/learn/typescript/publishing-a-ts-project.md
JakobJingleheimer: Ah, so `"baseUrl"` should just be removed entire? I vaguely recall other fields requiring it 🤔 perhaps that was a long time ago.

apps/site/pages/en/learn/typescript/publishing-a-ts-project.md
JakobJingleheimer: w00t! thanks

apps/site/pages/en/learn/typescript/publishing-a-ts-project.md
jakebailey: Yeah, I was also about to comment that when I noticed the flat layout above.

apps/site/pages/en/learn/typescript/publishing-a-ts-project.md
```diff
@@ -0,0 +1,224 @@
...
++   "prepack": "tsc",
+    "types:check": "tsc --noEmit"
+  }
+}
+```
+
+```text displayName=".npmignore"
+*.ts
+!*.d.ts
+*.fixture.*
```
jakebailey: If `rootDir` is `src`, `src` can be ignored and then everything else will work.

These patterns will not correctly ignore any `cts`/`mts` files.

apps/site/pages/en/learn/typescript/publishing-a-ts-project.md
JakobJingleheimer: Ah, yes, thank you for catching.

apps/site/pages/en/learn/typescript/publishing-a-ts-project.md
JakobJingleheimer: What is the reason for that recommendation?

`"outDir": "dist"` will cause that folder to get created and published though, whereas I want it to be in just the root to avoid unnecessary drilling.

So I think I want neither `"rootDir": "src"` nor `"outDir": "dist"`, because that will double what I want to avoid.

apps/site/pages/en/learn/typescript/publishing-a-ts-project.md
JakobJingleheimer: > These patterns will not correctly ignore any `cts`/`mts` files.

Sorry, I don't understand why those are special cases. Could you please explain?

jakebailey: The intent of this npmignore appears to be to prevent inclusion of source files; but one can handwrite `foo.mts`, which emits as `foo.mjs` and `foo.d.mts`, and so these globs will not handle them.

It's moot if you just ignore `src`, though.

apps/site/pages/en/learn/typescript/publishing-a-ts-project.md
jakebailey: Well, it at least implies that any large project is just going to dump a load of files at the root, which are going to be very annoying to `gitignore`/`npmignore`/`eslitignore`/`prettierignore`/etc, and visually ignore in the repo. If you output to `dist`, they're all in one place, easy to ignore, and notably, less likely to be accidentally loaded by TS or something.

I haven't personally seen a project which used a `src` and then dumped files at the root, only `src` -> `dist`, or, no `outDir` and allow the emitted JS to live next to the TS.

apps/site/pages/en/learn/typescript/publishing-a-ts-project.md
```diff
@@ -0,0 +1,218 @@
...
+  ├ some-util.ts
+  └ some-util.test.ts
+├ LICENSE
+├ package.json
+└ README.md
+```
+
+## What to do with your types
+
+### Treat types like a test
```
mcollina: We should add a note: types do not sobstitute unit testing

mcollina: We should add a note: types do not sobstitute unit testing

apps/site/pages/en/learn/typescript/publishing-a-ts-project.md
andrewbranch: `paths` used to require it, but hasn’t in a long time, and `paths` won’t play well with Node either

apps/site/pages/en/learn/typescript/publishing-a-ts-project.md
```diff
@@ -0,0 +1,218 @@
...
+├ src/
+  ├ foo.fixture.js
+  ├ main.ts
+  ├ main.test.ts
+  ├ some-util.ts
+  └ some-util.test.ts
+├ LICENSE
+├ package.json
+└ README.md
+```
```
AugustinMauroy: these info need to be reverse/inverted because it's more logical to have "what you write" the "what you get (published)"

apps/site/pages/en/learn/typescript/publishing-a-ts-project.md
```diff
@@ -0,0 +1,218 @@
...
+
+This article augments TypeScript's own [Publishing guide](https://www.typescriptlang.org/docs/handbook/declaration-files/publishing.html) with specifics for native node support.
+
+Some important things to note:
+
+- Everything from [Publishing a package](../modules/publishing-a-package) applies here.
+
+- Node runs TypeScript code via a process called "[type stripping](https://nodejs.org/api/typescript.html#type-stripping)", wherein node (via [Amaro](https://github.com/nodejs/amaro)) removes TypeScript-specific syntax, leaving behind vanilla JavaScript (which node already understands). This behaviour is enabled by default as of node version 23.6.0.
+
+  - Node does **not** strip types in `node_modules` because it can cause significant performance issues for the official TypeScript compiler (`tsc`) and parts of VS Code, so the TypeScript maintainers would like to discourage people publishing raw TypeScript, at least for now.
```
DanielRosenwasser:
```suggestion
- Node does **not** strip types in `node_modules` because it can cause significant performance issues for the official TypeScript compiler (`tsc`) and parts of VS Code, so the TypeScript maintainers would like to discourage people publishing raw TypeScript, at least for now.
```

apps/site/pages/en/learn/typescript/publishing-a-ts-project.md
```diff
@@ -0,0 +1,218 @@
...
+└ some-util.js
+```
+
+That would be derived from a repository looking something like:
+
+```text displayName="Source of the example TypeScript package (directory overview)"
+example-ts-pkg/
+├ .github/
+  ├ workflows/
+    ├ ci.yml
```
DanielRosenwasser: It it intentional that inner connecting lines are omitted?

apps/site/pages/en/learn/typescript/publishing-a-ts-project.md
```diff
@@ -0,0 +1,218 @@
...
+
+```ts
+const foo = 'a';
+const bar: number = 1 + foo;
+//    ^^^ Type 'string' is not assignable to type 'number'.
+```
+
+TypeScript has warned that the above code will not behave as intended, just like a unit test warns that code does not behave as intended.
+
+Your IDE (ex VS Code) likely has built-in support for TypeScript, displaying errors as you work. If not, and/or you missed those, CI will have your back.
```
DanielRosenwasser:
```suggestion
Your editor (e.g. VS Code) likely has built-in support for TypeScript, displaying errors as you work. If not, and/or you missed those, CI will have your back.
```

apps/site/pages/en/learn/typescript/publishing-a-ts-project.md
DanielRosenwasser: It also makes it easier to throw away all the files for a clean build. With package.json `exports`, it feels like there really isn't much reason to publish at the root anymore.

apps/site/pages/en/learn/typescript/publishing-a-ts-project.md
```diff
@@ -0,0 +1,218 @@
...
+
+#### Breaking this down
+
+Generating type declarations is deterministic: you'll get the same output from the same input, every time. So there is no need to commit these to git.
+
+[`npm publish`](https://docs.npmjs.com/cli/v11/commands/npm-publish) grabs everything applicable and available at the moment the command is run; so generating type declarations immediately before means those are available and will get picked up.
+
+By default, `npm publish` grabs (almost) everything (see [Files included in package](https://docs.npmjs.com/cli/v11/commands/npm-publish#files-included-in-package)). In order to keep your published package minimal (see the "Heaviest Objects in the Universe" meme about `node_modules`), you want to exclude certain files (like tests and test fixtures) from from packaging. Add these to the opt-out list specified in [`.npmignore`](https://docs.npmjs.com/cli/v11/using-npm/developers#keeping-files-out-of-your-package); ensure the `!*.d.ts` exception is listed, or the generated type declartions will not be published! Alternatively, you can use [package.json "files"](https://docs.npmjs.com/cli/v11/configuring-npm/package-json#files) to create an opt-in list.
```
DanielRosenwasser: > By default, `npm publish` grabs (almost) everything (see [Files included in package](https://docs.npmjs.com/cli/v11/commands/npm-publish#files-included-in-package)).

Similar to what @andrewbranch and @jakebailey said above, if you specify an `--outDir`, then you can use the `package.json` `"files"` array to avoid other hazards.

apps/site/pages/en/learn/typescript/publishing-a-ts-project.md
```diff
@@ -0,0 +1,218 @@
...
+    "allowArbitraryExtensions": true,
+    "declaration": true,
+    "declarationMap": true,
+    "lib": ["ESNext"],
+    "module": "NodeNext",
+    "outDir": "./",
+    "resolveJsonModule": true,
+    "rewriteRelativeImportExtensions": true
+  },
+  // These may be different for your repo:
```
DanielRosenwasser: Make sure that if you're using a comment, it's getting colorized correctly. I think Shiki might want `jsonc` for these sorts of blocks.

apps/site/pages/en/learn/typescript/publishing-a-ts-project.md
JakobJingleheimer: > It also makes it easier to throw away all the files for a clean build.

This is definitely true: I only generate type dec files during publication, and when I ran it locally, it was annoying to remove because they were gitignored.

JakobJingleheimer: > any large project is just going to dump a load of files at the root

Doesn't it otherwise keep the structure though, and basically un-nest "src"? (If not and it collapses everything, then I definitely withdraw my suggestion because that would be awful indeed.)

For source code, "src" makes sense because you usually have a lot of other files (various tooling configs, build config, maybe some test scripts, workflow config, etc). Those aren't distributed though, so at the root, you'd always have almost no files—aside from the dumped ones, just README.md and package.json and maybe LICENSE, no?

```text
bigger-repo
├ dist/
  ├ foo/
    ├ …
    └ …
  ├ bar/
    ├ …
    └ …
  ├ quz/
    ├ …
    └ …
  ├ main.js
  ├ main.d.ts
  └ main.d.ts.map
├ LICENSE
├ README.md
└ package.json
```

vs

```text
bigger-repo
├ foo/
  ├ …
  └ …
├ bar/
  ├ …
  └ …
└ quz/
  ├ …
  └ …
├ main.js
├ main.d.ts
├ main.d.ts.map
├ LICENSE
├ README.md
└ package.json
```

In a small repo, this looks kinda silly to me:

```text
@nodejs-loaders/alias/
├ dist/
  ├ alias.mjs
  ├ alias.d.ts
  └ alias.d.ts.map
├ LICENSE
├ README.md
└ package.json
```

---

> I haven't personally seen a project which used a `src` and then dumped files at the root, only `src` → `dist`, or, no `outDir` and allow the emitted JS to live next to the TS.

That doesn't happen in any way a human would see though (usually) because the js files only exist briefly in CI (a person likely would not run this locally—and if they do, they wouldn't want to persist the generated files, so they would comment out the gitignore and `git reset --hard HEAD`).

apps/site/pages/en/learn/typescript/publishing-a-ts-project.md
JakobJingleheimer: OH! You're talking about line 211. The code samples aren't using those extensions, so I didn't account for them to keep things simple and explicit.

Buuut that is a good idea. If we go this route btw, I think I should explain why this is a good idea.

jakebailey: > OH! You're talking about line 211

Yeah, it's a multi-line code review comment 😅

apps/site/pages/en/learn/typescript/publishing-a-ts-project.md
JakobJingleheimer: Yes: I had them, but on nodejs.org they got rendered weird and didn't align, so I removed them 😞

apps/site/pages/en/learn/typescript/publishing-a-ts-project.md
```diff
@@ -0,0 +1,218 @@
...
+  "include": ["./src"],
+  "exclude": ["**/*/*.test.*", "**/*.fixture.*"]
```
DanielRosenwasser: What exactly is `.fixture.` supposed to imply? I haven't seen this convention.

apps/site/pages/en/learn/typescript/publishing-a-ts-project.md
JakobJingleheimer: nodejs.org uses `jsonc` for `json` (and I think doesn't like `jsonc` explicitly used—nodejs.org has only a subset of codeblock languages enabled), so it gets rendered angrily only here in GitHub's UI (take a look at the preview env and you can see it looks fine).

apps/site/pages/en/learn/typescript/publishing-a-ts-project.md
ljharb: Publishing from "not the root" creates tons of hard-to-debug issues; `exports` doesn't change the calculus there imo.

apps/site/pages/en/learn/typescript/publishing-a-ts-project.md
JakobJingleheimer: You can see it in the previous samples. It's like `syntax-error.fixture.mjs`. Sometimes fixtures are all within a `fixtures` directory, but if you have only 1 or 2 fixtures and won't have more, a directory might be bloating.

JakobJingleheimer:
> I have some tsconfig feedback.

Thank you! TIL several things 😁

> is it worth mentioning anything about package.json fields (i.e. that they should reference `.js` files, not `.ts` files)?

Oh! Yes!

apps/site/pages/en/learn/typescript/publishing-a-ts-project.md
jakebailey: > Doesn't it otherwise keep the structure though, and basically un-nest "src"? (If not and it collapses everything, then I definitely withdraw my suggestion because that would be awful indeed.)

It does keep the structure, because it has to, but now your root will be full of stuff depending on how many things are directly within `src`. You also make `tsc` have to verify that none of the outputs are overwriting or turning into inputs again.

> For source code, "src" makes sense because you usually have a lot of other files (various tooling configs, build config, maybe some test scripts, workflow config, etc). Those aren't distributed though, so at the root, you'd always have almost no files—aside from the dumped ones, just README.md and package.json and maybe LICENSE, no?

Yes, but if you ever run `tsc` directly, now you have a whole bunch of other stuff at the root you need to figure out how to ignore.

> In a small repo, this looks kinda silly to me:

That's honestly exactly what I want it to look like.

> That doesn't happen in any way a human would see though (usually) because the js files only exist briefly in CI (a person likely would not run this locally—and if they do, they wouldn't want to persist the generated files, so they would comment out the gitignore and `git reset --hard HEAD`).


I know the npm script says `tsc --noEmit`, but IIRC this is a bad idea because there are actually diagnostics which are only emitted during emit. It's very atypical for anyone to run `tsc --noEmit` and then _also_ run `tsc` and publish its outputs. There's also diagnostics which are only given during emit which would not be visible until later.

That and if you want to use any other extra features like incremental mode / tsbuildinfo, project references, you can't use `--noEmit` as-is.

apps/site/pages/en/learn/typescript/publishing-a-ts-project.md
```diff
@@ -0,0 +1,218 @@
...
+        run: npm ci
+      - run: node --run test
+```
+
+```json displayName="package.json"
+{
+  "name": "example-ts-pkg",
+  "scripts": {
+    "test": "node --test './src/**/*.test.ts'",
+    "types:check": "tsc --noEmit"
```
jakebailey: I noted this in the other thread, but I would be cautious about this as a default; I really only see people setting noEmit when they're doing a quick check, or are using a bundler or something.

apps/site/pages/en/learn/typescript/publishing-a-ts-project.md
jakebailey: Can you say what those issues are? The only thing I can think of is compat with old node releases that do not support exports, or trying to support people who want to deep import when the package author does not want that.

apps/site/pages/en/learn/typescript/publishing-a-ts-project.md
JakobJingleheimer: The indentation wasn't an accident/typo, if that's what you're thinking. This could be root-level I suppose, but I was thinking it was an addendum to the item above it. I'm splitting hairs though don't feel strongly. If people thing it's better to keep root-level, sure.

apps/site/pages/en/learn/typescript/publishing-a-ts-project.md
DanielRosenwasser: Yes, it preserves structure, but that feels beside the point. But by and large, the recommendation we've had (and really the de facto norm at this point) is to use a dedicated output directory.

apps/site/pages/en/learn/typescript/publishing-a-ts-project.md
JakobJingleheimer: This is the type "test" command. Why would you want to emit a compilation?

Maybe the names could be better? When I have unit and end-to-end tests with different setups, I split those into different commands like:
* `test:unit`
* `test:e2e`

So in that scenario, it could make sense to name `types:check` → `test:types`.

But in the sample, there's no differentiation between units and e2e, so then what do I call what is currently `test`?

jakebailey: I mean, I guess it's fine, I am just wary of cases where `tsc` and `tsc --noEmit` output different errors because the former is doing more. Maybe you'd hit it on `prepack` and that's okay, but it's a little unfortunate to only hit an error when you go to release...

apps/site/pages/en/learn/typescript/publishing-a-ts-project.md
JakobJingleheimer: > It's very atypical for anyone to run `tsc --noEmit` and then _also_ run `tsc` and publish its outputs.

Maaybe there's a misunderstanding: these would not be run one right after the other:

* On a PR, `tsc --noEmit` would run.
* On a release, `tsc` would run.

apps/site/pages/en/learn/typescript/publishing-a-ts-project.md
DanielRosenwasser: Maybe just avoid mentioning "treat types like tests" in the first place? I don't really see how it weaves into the blog post (even in this section)

apps/site/pages/en/learn/typescript/publishing-a-ts-project.md
JakobJingleheimer: Is that likely? I've been doing this for years and never encountered that—am I just very lucky?

apps/site/pages/en/learn/typescript/publishing-a-ts-project.md
DanielRosenwasser: I mean more like using a connecting lines like so

```
a/
├── b/
│   └── c
└── d
```

whereas you have something closer to

```
a/
├── b/
    └── c
└── d
```

apps/site/pages/en/learn/typescript/publishing-a-ts-project.md
jakebailey: I understand that, I'm just saying that the two may behave differently as there are errors that can only happen when `tsc` is emitting, and you would never see them until `tsc` is run without `--noEmit`.

apps/site/pages/en/learn/typescript/publishing-a-ts-project.md
JakobJingleheimer: Yes, I understood 😉 That is what I originally had, and it renders wonky on nodejs.org.

I think you meant to respond to this thread though? https://github.com/nodejs/nodejs.org/pull/7279#discussion_r1907899993

apps/site/pages/en/learn/typescript/publishing-a-ts-project.md
```diff
@@ -0,0 +1,219 @@
...
+    "declarationMap": true,
+    "lib": ["ESNext"],
+    "module": "NodeNext",
+    "outDir": "./",
+    "resolveJsonModule": true,
+    "rewriteRelativeImportExtensions": true
+  },
+  // These may be different for your repo:
+  "include": ["./src"],
+  "exclude": ["**/*/*.test.*", "**/*.fixture.*"]
```
jakebailey: Why would you exclude your tests from the tsconfig? Shouldn't they be typechecked?

apps/site/pages/en/learn/typescript/publishing-a-ts-project.md
DanielRosenwasser: Sorry, I misread and responded to the wrong comment. 😅

apps/site/pages/en/learn/typescript/publishing-a-ts-project.md
DanielRosenwasser: > On a PR, `tsc --noEmit` would run.

Let me just say that a release-blocker because of `.d.ts` emit problems is better caught as early as possible, and the delta in CI time is not worth it.

apps/site/pages/en/learn/typescript/publishing-a-ts-project.md
jakebailey: I'm not sure how to gauge "likely", probably unlikely, but they're cases like "tsc failed to write the files", along with potentially some declaration transform errors. (The latter _shouldn't_ actually end up mattering by my reading of the code, though.)

DanielRosenwasser: The specific issues that come up that I can think of are when a declaration file can't reliably be generated because doing so might require referencing entities that are private or non-exported. Trying to figure out why this error is happening can be pretty frustrating, especially if you've been relying a specific pattern over time. Having a divergence between publish/CI probably just makes this even more confusing since most people outside of the person who set up the build won't be aware of any differences.

apps/site/pages/en/learn/typescript/publishing-a-ts-project.md
DanielRosenwasser: They actually might need a separate `tsconfig.json` if you don't want them in `outDir`, right?

jakebailey: Yes, though I personally opt to put my tests in a specific folder like `__tests__` just so I can exclude them and their tests easily, but I know some people make a second tsconfig and then build mode and so on (it's just too many steps for me to feel good about it).

apps/site/pages/en/learn/typescript/publishing-a-ts-project.md
ljharb: or `.npmignore`, which avoids hazards endemic to `files` :-)

apps/site/pages/en/learn/typescript/publishing-a-ts-project.md
AugustinMauroy: IMO, we should treat types as a part of code quality checking. Like lintting and formatting

apps/site/pages/en/learn/typescript/publishing-a-ts-project.md
```diff
@@ -0,0 +1,219 @@
...
+Since these are generated based on source code, they can be built as part of your publication process and do not need to be checked into your repository.
+
+Take the following example, where the type declarations are generated just before publishing to the NPM registry.
+
+```yaml displayName=".github/workflows/publish.yml"
+name: Publish to NPM
+on:
+  push:
+    tags:
+      - '**@*'
```
AugustinMauroy:
```suggestion
name: Publish to NPM
on:
  release:
    types: [published]
```

We can propose to just use release not tag. In some case tag is useful for example when you have monorepo

ljharb: ? Package releases should always have a git tag, no exceptions - single-package repos too.

AugustinMauroy: npm sugest to use release
https://docs.npmjs.com/generating-provenance-statements#example-github-actions-workflow

apps/site/pages/en/learn/typescript/publishing-a-ts-project.md
JakobJingleheimer: > tsc failed to write the files

That does sound like an issue that you legitimately wouldn't catch with a dry-run, but also seems a very unlikely issue (and one that could still occur at publishing even if you were using non-dry-run for the test step). It seems an edge-case worth noting but not worth taking a hit every time to avoid something that likely will never happen (if it does, there are a very limited number of causes—two? permissions and storage availability).

> The specific issues that come up that I can think of are when a declaration file can't reliably be generated because doing so might require referencing entities that are private or non-exported.

That sounds very detectable for `--noEmit`; if it doesn't do that, that sounds like a defect in `tsc`? Why would you need to writing to disk in order to discover it?

-----

But maybe let's take a step back for a second: The reason I wrote the setup this way is because of performance—but perhaps my information is outdated. Last I heard, `tsc --noEmit` was significantly faster than with emit.

apps/site/pages/en/learn/typescript/publishing-a-ts-project.md
JakobJingleheimer: I would strongly advocate against a test folder. Node has that, and it is an atrocious experience. If the reason to do it is to simplify tooling/configuration, there's an X-Y problem (tooling exists to make life easier, not to make life harder). I think that's a different discussion though.

apps/site/pages/en/learn/typescript/publishing-a-ts-project.md
JakobJingleheimer: I definitely think a release should have a tag, but I think different points are being argued here: It's not saying a release shouldn't have a tag, but is instead saying make this relative to the release.

ljharb: ahhh i see, i was confused. yeah i think it's fine to trigger on release.

apps/site/pages/en/learn/typescript/publishing-a-ts-project.md
```diff
@@ -0,0 +1,223 @@
+---
+title: Publishing a TypeScript project
+layout: learn
+authors: JakobJingleheimer
+---
+
+# Publishing a TypeScript project
+
+This article augments TypeScript's own [Publishing guide](https://www.typescriptlang.org/docs/handbook/declaration-files/publishing.html) with specifics for native node support.
```
jakebailey: I am sort of questioning linking to this page; reading it, it's pretty out of date and is part of the declaration file section, so sort of misses out on other important stuff.

apps/site/pages/en/learn/typescript/publishing-a-ts-project.md
jakebailey: > That sounds very detectable for `--noEmit`; if it doesn't do that, that sounds like a defect in `tsc`? Why would you need to writing to disk in order to discover it?

I think as of recent releases (since `--isolatedDeclarations`), we've actually always checked declaration errors without emit, so I think the only errors one can see differently are just the errors that happen while writing, which may not really be important except in the case where you've somehow accidentally marked output paths as readonly in the FS or something.

> But maybe let's take a step back for a second: The reason I wrote the setup this way is because of performance—but perhaps my information is outdated. Last I heard, `tsc --noEmit` was significantly faster than with emit.

It can be, though I think at the scale of this demo, it's definitely not a big difference.

apps/site/pages/en/learn/typescript/publishing-a-ts-project.md
jakebailey: FWIW you cannot create a GitHub release without pointing it at a tag; the UI will force you to either select one or make one, so a tag will definitely have existed.

apps/site/pages/en/learn/typescript/publishing-a-ts-project.md
jakebailey: I think the problem is still what I said originally; the setup currently in the PR will not typecheck your tests.

apps/site/pages/en/learn/typescript/publishing-a-ts-project.md
JakobJingleheimer: Happy to change it to a different one—is there one you had in mind?

apps/site/pages/en/learn/typescript/publishing-a-ts-project.md
JakobJingleheimer: Is the take-away then that what I have _is_ fine? There is a potentially (and probably likely) significant perf savings, and there're basically no type-related errors this won't catch?

apps/site/pages/en/learn/typescript/publishing-a-ts-project.md
JakobJingleheimer: Good point. I'll cover that specifically :)

apps/site/pages/en/learn/typescript/publishing-a-ts-project.md
```diff
@@ -0,0 +1,288 @@
...
+    # they are platform and node-version independent
+    # and need be run only once.
+
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-node@v4
+        with:
+          node-version-file: '.nvmrc'
```
AugustinMauroy:
```suggestion
          node-version-file: '.nvmrc' # Runs on a selected development version
```

apps/site/pages/en/learn/typescript/publishing-a-ts-project.md
```diff
@@ -0,0 +1,288 @@
...
+        run: npm ci
+      # You may want to run a lint check here too
+      - run: node --run types:check
+
+  test:
+    runs-on: ubuntu-latest
+
+    strategy:
+      matrix:
+        node:
```
AugustinMauroy:
```suggestion
        node: # Runs on supported version to ensure that the packages work on every supported version
```

apps/site/pages/en/learn/typescript/publishing-a-ts-project.md
```diff
@@ -0,0 +1,288 @@
...
+!*.d.ts
+*.fixture.*
+```
+
+```text displayName="(alt: dist/) .npmignore"
+src
+test
+```
+
+`npm publish` will automatically run [`prepack` beforehand](https://docs.npmjs.com/cli/v11/using-npm/scripts#npm-publish). `npm` will also run `prepack` automatically before `npm pack --dry-run` (so you can easily see what your published package will be without actually publishing it). **Beware**, [`node --run` does _not_ do that](../command-line/run-nodejs-scripts-from-the-command-line.md#using-the---run-flag). You can't use `node --run` for this step, so that caveat does not apply here, but it can for other steps.
```
AugustinMauroy:
```suggestion
`npm publish` will automatically run [`prepack` beforehand](https://docs.npmjs.com/cli/using-npm/scripts). `npm` will also run `prepack` automatically before `npm pack --dry-run` (so you can easily see what your published package will be without actually publishing it). **Beware**, [`node --run` does _not_ do that](../command-line/run-nodejs-scripts-from-the-command-line.md#using-the---run-flag). You can't use `node --run` for this step, so that caveat does not apply here, but it can for other steps.
```

Let's point to lasted version of npm to unsure that link will not become depressed

JakobJingleheimer: oow, snap. I did not know that was possible. there are other references here, so i'll push a commit to get them all.

apps/site/pages/en/learn/typescript/publishing-a-ts-project.md
JakobJingleheimer: I'm gonna switch this to Jordan's action (which handles the keeping the various LTS +OS versioning in sync), so this will be abstracted away.

apps/site/pages/en/learn/typescript/publishing-a-ts-project.md
ljharb: is this not self-evident?

apps/site/pages/en/learn/typescript/publishing-a-ts-project.md
ljharb: is this also not self-evident?

JakobJingleheimer: I think this is covered by the docs already linked here?

apps/site/pages/en/learn/typescript/publishing-a-ts-project.md
```diff
@@ -0,0 +1,288 @@
...
+        with:
+          node-version-file: '.nvmrc'
+          registry-url: 'https://registry.npmjs.org'
+      - run: npm ci
+
+      # You can probably ignore the boilerplate config above
+
+      - name: Publish with provenance
+        env:
+          NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
```
mcollina: Is `NPM_TOKEN` automatically provided somehow? I think we should mention how to get this or avoid automated publishing accordingly. having a separate guide on automated publishing is worthwhile.

AugustinMauroy: t's not provide automatically. But this GA example from https://docs.npmjs.com/generating-provenance-statements#example-github-actions-workflow

JakobJingleheimer: Would we have enough content for a dedicated guide on automated publishing?

For now, I think let's add a note explaining what the token is and where it comes from: 0f7f993

AugustinMauroy: +1 for note

ljharb: There are a ton of problems with automated publishing (discussed above) and I think a separate guide that can fully explain the pros and cons is definitely warranted.

JakobJingleheimer: In that case, sure, happy to do subsequently (subsequently because otherwise we end up with a web of ½ finished, inter-dependent guides, and this doesn't seem like a show-stopper—unless someone strongly thinks that it needs to happen first).

ljharb: I've made the case above that it would be better to give no publishing guidance prior to such a properly nuanced guide, then to provide instructions absent nuance - iow, I suggest removing this section entirely and hand-wave over publishing for the time being.

I don't think we can avoid interdependent guides; the reality is too complex to allow for a single "does everything" document imo.

AugustinMauroy: why not just redirect to npm docs ?

JakobJingleheimer: Oh for sure they'll be inter-dependent. I just want to avoid a bunch of _unfinished_ inter-dependent guides that are stuck in a spiderman meme.

AugustinMauroy: So let's remove provenance and add note or let the security team write another article about it

JakobJingleheimer: works for me

done: 6056802e

apps/site/pages/en/learn/typescript/publishing-a-ts-project.md
JakobJingleheimer: @jakebailey is there a better link for typescript's own publishing guide?

jakebailey: I do not think we have one.

JakobJingleheimer: Is there anything tangentially related that could be useful?

jakebailey:
All of the threads about it appear to have been marked "resolved", but I really can't express enough how much I do not like the idea of having a `src` dir but then also tell `tsc` to dump its output into the package root _above_ `src`. There are so many downsides with that approach, and I have never seen that in the wild in any TypeScript project.

I do not think a publishing guide we'd write for our own website would suggest something like it. And in a future release, we are expecting to simplfiy `tsc --init` and have it pre-populate `tsconfig.json` with `rootDir=./src`, `outDir=./dist` (https://github.com/microsoft/TypeScript/issues/58420). That is the least problematic way to do it (both for TS's handling but also when dealing with other tools).

I would even prefer not having a `src` dir and just letting the JS files be dumped next to the TS, becuase that would be easier to explain / avoid problems with than the mismatch.

apps/site/pages/en/learn/typescript/publishing-a-ts-project.md
jakebailey: The most up to date ish related thing are the extensive modules docs (https://www.typescriptlang.org/docs/handbook/modules/introduction.html) which talk about export mapping and such, but it's not really a publishing guide at all.

Potentially https://www.typescriptlang.org/docs/handbook/migrating-from-javascript.html, but a lot of this content is very out of date and we just don't have the time to rewrite it.

JakobJingleheimer: Hmm, I think those could be useful in other node typescript articles, but I don't see much relevant for publishing. I think I'll just remove this paragraph then.

JakobJingleheimer:
@jakebailey I don't understand: there are examples with `src/` + `test/` + `dist/`. Here are screenshots from the deployed article:

image
image

---

@AugustinMauroy do you have any idea why the vertical line characters are rendering weirdly?

vertical line characters off-centre

jakebailey:
> I don't understand: there are examples with `src/` + `test/` + `dist/`. Here are screenshots from the deployed article:

There's one example as a tree, yes, but that is not the tsconfig provided in the post, which the one people are going to copy without really understanding the impliciations (since `dist` is not mentioned anywhere else except the ASCII diagram). The `npmignore` doesn't reflect `dist`, and still don't handle `.mts`/`.cts`.

JakobJingleheimer:
> not the tsconfig provided in the post, […]. The `npmignore` doesn't reflect `dist`, and still don't handle `.mts`/`.cts`.

Ahh, I'll add those (hopefully either tomorrow or Sunday).

AugustinMauroy:
> @AugustinMauroy do you have any idea why the vertical line characters are rendering weirdly?

Strange I'll dedicate time on that

apps/site/pages/en/learn/typescript/publishing-a-ts-project.md
JakobJingleheimer: @jakebailey I believe I have added the necessary samples to cover this now. Please LMK if I missed something.

JakobJingleheimer:
@AugustinMauroy the deployment is returning 500 again (no indication of the problem). I _think_ it may be due to setting a codeblock lang to `ini` (which may not be enabled).

AugustinMauroy:
Like said on opened issue when there are a 500 it's an error on server side so take a look to terminal

JakobJingleheimer:
> Like said on opened issue when there are a 500

Do you mean like https://github.com/nodejs/nodejs.org/issues/7464 🙂

AugustinMauroy:
Bump for codetab, I'll open an issue because it's an error on our side not on shiky or your docs

apps/site/pages/en/learn/typescript/publishing-a-ts-project.md
```diff
@@ -0,0 +1,328 @@
+---
+title: Publishing a TypeScript project
+layout: learn
+authors: JakobJingleheimer
+---
+
+# Publishing a TypeScript project
+
+This article covers items regarding typescript publishing specifically.
```
AugustinMauroy: Here we can write a little more about the fact that we're publishing a package that will go on the npm registery.
It's not like an express app that you run. package ≠ app

AugustinMauroy (commented):
Also need to mention that ts target allow the user to support "old" js version.

JakobJingleheimer:
> Also need to mention that ts target allow the user to support "old" js version.

Good point! What ES version does node 18 support? I think probably es2022? https://compat-table.github.io/compat-table/es2016plus/#node16_11

ljharb:
It's not a clean mapping; unfortunately TS doesn't have a `babel-preset-env`-like setting where it can infer the parser settings from your supported envs.

JakobJingleheimer:
> It's not a clean mapping; unfortunately TS doesn't have a `babel-preset-env`-like setting where it can infer the parser settings from your supported envs.

Looks like there is _something_: https://www.typescriptlang.org/tsconfig/#target → https://www.npmjs.com/package/@tsconfig/node18

ljharb:
That's great, thanks!

apps/site/pages/en/learn/typescript/publishing-a-ts-project.md
```diff
@@ -0,0 +1,332 @@
+---
+title: Publishing a TypeScript project
```
AugustinMauroy:
```suggestion
title: Publishing a TypeScript package
```

Why not using word package ?

AugustinMauroy (approved):
LGTM !!!! WOW after some interation

apps/site/pages/en/learn/typescript/publishing-a-ts-project.md
JakobJingleheimer: Ah, yes that's better. Should also change the file name and json refs.

bmuenzenmeyer:
merging per a conversation with @JakobJingleheimer

bmuenzenmeyer (approved):

marco-ippolito (approved):

JakobJingleheimer:
Thanks all! Happy to iterate if needed 🙂

apps/site/pages/en/learn/typescript/publishing-a-ts-package.md
```diff
@@ -0,0 +1,332 @@
...
+
+  - Node does **not** strip types in `node_modules` because it can cause significant performance issues for the official TypeScript compiler (`tsc`) and parts of VS Code, so the TypeScript maintainers would like to discourage people publishing raw TypeScript, at least for now.
+
+- Consuming TypeScript-specific features like `enum` in node still requires a flag ([`--experimental-transform-types`](https://nodejs.org/api/typescript.html#typescript-features)). There are often better alternatives for these anyway.
+
+  - To ensure TypeScript-specific features are _not_ present (so your code can just run in node), set the [`erasableSyntaxOnly`](https://devblogs.microsoft.com/typescript/announcing-typescript-5-8-beta/#the---erasablesyntaxonly-option) config option in TypeScript version 5.8+.
+
+- Use [dependabot](https://docs.github.com/en/code-security/dependabot) to keep your dependencies current, including those in github actions. It's a very easy set-and-forget configuration.
+
+- `.nvmrc` comes from [NVM](https://github.com/nvm-sh/nvm), a multi-version manager for node. It allows you to specify the version of node the project should generally use.
```
ljharb, line 29:
```suggestion
- `.nvmrc` comes from [`nvm`](https://github.com/nvm-sh/nvm), a multi-version manager for node. It allows you to specify the version of node the project should generally use.
```

apps/site/pages/en/learn/typescript/publishing-a-ts-package.md
```diff
@@ -0,0 +1,332 @@
...
+
+```ts
+const foo = 'a';
+const bar: number = 1 + foo;
+//    ^^^ Type 'string' is not assignable to type 'number'.
+```
+
+TypeScript has warned that the above code will not behave as intended, just like a unit test warns that code does not behave as intended. They are complementary and verify different things—you should have both.
+
+Your editor (ex VS Code) likely has built-in support for TypeScript, displaying errors as you work. If not, and/or you missed those, CI will have your back.
```
ljharb, line 139:
```suggestion
Your editor (eg VS Code) likely has built-in support for TypeScript, displaying errors as you work. If not, and/or you missed those, CI will have your back.
```

apps/site/pages/en/learn/typescript/publishing-a-ts-package.md
```diff
@@ -0,0 +1,332 @@
...
+Since these are generated based on source code, they can be built as part of your publication process and do not need to be checked into your repository.
+
+Take the following example, where the type declarations are generated just before publishing to the NPM registry.
+
+```yaml displayName=".github/workflows/publish.yml"
+# yaml-language-server: $schema=https://json.schemastore.org/github-workflow.json
+
+# This is mostly boilerplate.
+
+name: Publish to NPM
```
ljharb, line 275:
```suggestion
name: Publish to npm
```
(npm should always be lowercased)

apps/site/pages/en/learn/typescript/publishing-a-ts-package.md
```diff
@@ -0,0 +1,332 @@
...
+
+You'll want to publish a package compiled to support all Node.js LTS versions since you don't know which version the consumer will be running; the `tsconfig`s in this article support node 18.x and later.
+
+`npm publish` will automatically run [`prepack` beforehand](https://docs.npmjs.com/cli/using-npm/scripts#npm-publish). `npm` will also run `prepack` automatically before `npm pack --dry-run` (so you can easily see what your published package will be without actually publishing it). **Beware**, [`node --run` does _not_ do that](../command-line/run-nodejs-scripts-from-the-command-line.md#using-the---run-flag). You can't use `node --run` for this step, so that caveat does not apply here, but it can for other steps.
+
+The steps to actually publish to npm will be included in a separate article (there are several pros and cons beyond the scope of this article).
+
+#### Breaking this down
+
+Generating type declarations is deterministic: you'll get the same output from the same input, every time. So there is no need to commit these to git.
```
ljharb, line 328: i'm pretty sure this isn't actually true, but either way generated output should never be committed to source control.

apps/site/pages/en/learn/typescript/publishing-a-ts-package.md
```diff
@@ -0,0 +1,332 @@
...
+
+[`npm publish`](https://docs.npmjs.com/cli/commands/npm-publish) grabs everything applicable and available at the moment the command is run; so generating type declarations immediately before means those are available and will get picked up.
+
+By default, `npm publish` grabs (almost) everything (see [Files included in package](https://docs.npmjs.com/cli/commands/npm-publish#files-included-in-package)). In order to keep your published package minimal (see the "Heaviest Objects in the Universe" meme about `node_modules`), you want to exclude certain files (like tests and test fixtures) from from packaging. Add these to the opt-out list specified in [`.npmignore`](https://docs.npmjs.com/cli/using-npm/developers#keeping-files-out-of-your-package); ensure the `!*.d.ts` exception is listed, or the generated type declartions will not be published! Alternatively, you can use [package.json "files"](https://docs.npmjs.com/cli/configuring-npm/package-json#files) to create an opt-in list.
```
ljharb, line 332:
```suggestion
By default, `npm publish` grabs (almost) everything (see [Files included in package](https://docs.npmjs.com/cli/commands/npm-publish#files-included-in-package)). In order to keep your published package minimal (see the "Heaviest Objects in the Universe" meme about `node_modules`), you want to exclude certain files (like tests and test fixtures) from from packaging. Add these to the opt-out list specified in [`.npmignore`](https://docs.npmjs.com/cli/using-npm/developers#keeping-files-out-of-your-package); ensure the `!*.d.ts` exception is listed, or the generated type declartions will not be published! Alternatively, you can use [package.json "files"](https://docs.npmjs.com/cli/configuring-npm/package-json#files) to create an opt-in list - but if a mistake is made and a file accidentally omitted, your package may be broken for downstream users, so this is a less safe option.
```

ljharb (commented):
Overall looks great! Just a few comments (sorry they came in post-merge)

JakobJingleheimer:
Thanks! I'll open a follow-up soon (on quasi-holiday til this weekend, so not sure if/when I'll have time before then).

apps/site/pages/en/learn/typescript/publishing-a-ts-package.md
JakobJingleheimer: 🤔 Oh? I can't think of how it isn't—could you enlighten me?

apps/site/pages/en/learn/typescript/publishing-a-ts-package.md
JakobJingleheimer: These are the same 😜 One is english, the other is latin:

`ex` → `example`
`eg` → [`exempli gratia`](https://en.wikipedia.org/wiki/Exempli_gratia)

apps/site/pages/en/learn/typescript/publishing-a-ts-package.md
ljharb: I'm sure it's semantically deterministic, but I would be very surprised if the output bytes were actually deterministic (mangled variable names, etc) - and either way, nondeterminism isn't why they shouldn't be committed, "they're generated" is.

apps/site/pages/en/learn/typescript/publishing-a-ts-package.md
ljharb: right, but "ex" reads to me as "formerly", "eg" doesn't have the same confusion

apps/site/pages/en/learn/typescript/publishing-a-ts-package.md
JakobJingleheimer: This is type def generation, no transpiling the typescript (which is definitely not deterministic, as indeed variable name mangling etc). Type names are not changed when the def files are generated (that would break the hell out of the type defs).

> nondeterminism isn't why they shouldn't be committed, "they're generated" is.

True, but if it's nondeterministic, I shouldn't say it is deterministic.

ljharb: ahhh that's fair. you might be right that the type defs are deterministic! but i'd still say that's not important to mention because it's not why they shouldn't be committed.

apps/site/pages/en/learn/typescript/publishing-a-ts-package.md
JakobJingleheimer: Ah, okay that's better then 🙂

Issue #7275, opened 2024-11-22
Visually differentiate H2 from H3 headings in Metabar Table of Contents


bmuenzenmeyer:
Visually indicate the difference between an H2 and H3 heading.
This could be indentation or other textual formatting. Whatever it is, make sure it is responsive and does not create a lot of flow problems. For instance, do not indent many many pixels, as that just makes the headings wrap and become longer.

---

> You should use smaller headings or we should limit depth of what goes on Table of Contents

I like the heading hierarchy of the content - it'd be nice if we visually indicated the indentation level. I think H2 and H3 only.

Look at content like https://nodejs.org/en/learn/test-runner/mocking for similar needs.

_Originally posted by @bmuenzenmeyer in https://github.com/nodejs/nodejs.org/issues/7215#issuecomment-2492842051_

AugustinMauroy:
+1 for that

nafisreza:
Hello, I am interested to work on this. Do you think adding ``text-indent: 10px`` on the h3 will solve it?

![image](https://github.com/user-attachments/assets/80d3cd3a-1dc6-41b8-91b6-3edc18f940d0)

bmuenzenmeyer:
visually perhaps, but the headings are not rendered as h2, h3 etc in the metabar, it's a definition list entry

see the code at https://github.com/nodejs/nodejs.org/blob/fbd443d4b354ea1530cb0f2e262d2282086fc1d2/apps/site/components/Containers/MetaBar/index.tsx#L41-L53 - which does have the heading context - but i am yet unsure if it knows of the heading size - some plumbing or additional context values may be necessary when tracing it all the way back

see https://nodejs.org/en/learn/test-runner/mocking

![image](https://github.com/user-attachments/assets/c84d2460-5cb2-4d13-a4c2-a56b0c07b100)


```html
<div class="MetaBar_wrapper__Th_dl">
  <dl>
    <dt>Table of Contents</dt>
    <dd>
      <ol>
        <li>
          <a href="#when-and-not-to-mock">When and not to mock</a>
        </li>
        <li>
          <a href="#own-code">Own code</a>
        </li>
        <li>
          <a href="#why">Why</a>
        </li>
        <li>
          <a href="#why-not">Why not</a>
        </li>
        <li>
          <a href="#external-code">External code</a>
        </li>
        <li>
          <a href="#why-1">Why</a>
        </li>
        <li>
          <a href="#why-not-1">Why not</a>
        </li>
        <li>
[...]

PR #7286, opened 2024-11-28
add changes to MetaBar index.modul.css

faridomarAf:
# MetaBar modified

MetaBar index.module.css modifie
added different font-weight to the 'dd' and 'a' of the css file, to make difference between H2 & H3 at Metabar Table
## Validation
Screenshot 2024-11-27 at 5 22 31 PM

## Related Issues

https://github.com/nodejs/nodejs.org/issues/7275

### DIFF
--- a/apps/site/components/Containers/MetaBar/index.module.css
+++ b/apps/site/components/Containers/MetaBar/index.module.css
@@ -23,7 +23,7 @@
   dt {
     @apply mb-2
       text-sm
-      font-medium
+      font-[700]
       text-neutral-800
       dark:text-neutral-200;
   }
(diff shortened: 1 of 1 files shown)


AugustinMauroy:
That doesn't fix the issue ?
@bmuenzenmeyer You can confirm

bmuenzenmeyer:
I'm mobile now,  but no i don't think it does

bmuenzenmeyer (changes requested):
confirmed this does not solve the problem outlined within https://github.com/nodejs/nodejs.org/issues/7275

we need to approach the headings as a data structure we can differentiate h2 and h3s from

AugustinMauroy:
Maybe a left margin or something like that

PR #7296, opened 2024-11-30
always show the latest version of nvm

ljharb:
## Description

This PR attempts to ensure that the version of nvm users are instructed to install is always the latest one.

I had to make the "get snippets" function async for this, but I eagerly create the promise that will hold the latest version, so that the user has to wait as little time as possible. Another alternative would be to convert this function into a react component, and immediately render a hardcoded nvm version, but then later rerender when the fetch promise returns. This would be a larger change, though, so I didn't start out going that route.

## Validation

I'm not sure how to test this locally, but I'm happy to do so if someone can give me pointers :-)

## Related Issues

N/A

### Check List

I'll write tests once the implementation direction is confirmed :-)

### DIFF
--- a/apps/site/app/[locale]/next-data/nvm-data/route.ts
+++ b/apps/site/app/[locale]/next-data/nvm-data/route.ts
@@ -1,0 +1,31 @@
+import provideNvmData from '@/next-data/providers/nvmData';
+import { defaultLocale } from '@/next.locales.mjs';
+
+// This is the Route Handler for the `GET` method which handles the request
+// for generating static data for the latest nvm version
+// @see https://nextjs.org/docs/app/building-your-application/routing/router-handlers
+export const GET = async () => {
+  const nvmData = provideNvmData();
+
+  return Response.json(nvmData);
+};
+
+// This function generates the static paths that come from the dynamic segments
+// `[locale]/next-data/nvm-data/` and returns an array of all available static paths
+// This is used for ISR static validation and generation
+export const generateStaticParams = async () => [
+  { locale: defaultLocale.code },
+];
+
+// Enforces that only the paths from `generateStaticParams` are allowed, giving 404 on the contrary
+// @see https://nextjs.org/docs/app/api-reference/file-conventions/route-segment-config#dynamicparams
+export const dynamicParams = false;
+
+// Enforces that this route is used as static rendering
+// @see https://nextjs.org/docs/app/api-reference/file-conventions/route-segment-config#dynamic
+export const dynamic = 'error';
+
+// Ensures that this endpoint is invalidated and re-executed every X minutes
+// so that when new deployments happen, the data is refreshed
+// @see https://nextjs.org/docs/app/api-reference/file-conventions/route-segment-config#revalidate
+export const revalidate = 300;
--- a/apps/site/components/Downloads/Release/ReleaseCodeBox.tsx
+++ b/apps/site/components/Downloads/Release/ReleaseCodeBox.tsx
@@ -21,13 +21,20 @@ const ReleaseCodeBox: FC = () => {
   const t = useTranslations();
 
   useEffect(() => {
-    const updatedCode = getNodeDownloadSnippet(release, os, t)[platform];
-    // Docker and NVM support downloading tags/versions by their full release number
-    // but usually we should recommend users to download "major" versions
-    // since our Download Buttons get the latest minor of a major, it does make sense
-    // to request installation of a major via a package manager
-    memoizedShiki.then(shiki => shiki(updatedCode, 'bash')).then(setCode);
-    // Only react when the specific release number changed
+    async function getSnippet() {
+      const [shiki, { [platform]: updatedCode }] = await Promise.all([
+        memoizedShiki,
+        getNodeDownloadSnippet(release, os, t),
+      ]);
+
+      // Docker and nvm support downloading tags/versions by their full release number
+      // but usually we should recommend users to download "major" versions since
+      // our Download Buttons get the latest minor of a major, it does make sense
+      // to request installation of a major via a package manager
+      setCode(await shiki(updatedCode, 'bash'));
+      // Only react when the specific release number changed
+    }
+    getSnippet();
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [release.versionWithPrefix, os, platform]);
 
--- a/apps/site/next-data/generators/nvmData.mjs
+++ b/apps/site/next-data/generators/nvmData.mjs
@@ -1,0 +1,24 @@
+'use strict';
+
+const latestKnownVersion = 'v0.40.1';
+
+/**
+ * Fetches the latest NVM version
+ * @returns {Promise<`v${string}`>} Latest NVM version
+ */
+export default async function generateNvmData() {
+  return fetch('https://latest.nvm.sh', { redirect: 'manual' })
+    .then(({ headers }) => {
+      const url = headers.get('location');
+      if (!url) {
+        throw new Error('No redirect location found');
+      }
+      return fetch(url, { redirect: 'manual' });
+    })
+    .then(x => {
+      const url = x.headers.get('location');
+      const version = url?.slice(url.lastIndexOf('/') + 1);
+      return version || latestKnownVersion;
+    })
+    .catch(() => latestKnownVersion);
+}
--- a/apps/site/util/getNodeDownloadSnippet.ts
+++ b/apps/site/util/getNodeDownloadSnippet.ts
@@ -1,15 +1,16 @@
 import dedent from 'dedent';
 import type { TranslationValues } from 'next-intl';
 
+import getNvmData from '@/next-data/nvmData';
 import type { NodeRelease } from '@/types';
 import type { PackageManager } from '@/types/release';
 import type { UserOS } from '@/types/userOS';
 
-export const getNodeDownloadSnippet = (
+export async function getNodeDownloadSnippet(
   release: NodeRelease,
   os: UserOS,
   t: (key: string, values?: TranslationValues) => string
-) => {
+) {
   const snippets: Record<PackageManager, string> = {
     NVM: '',
     FNM: '',
@@ -39,7 +40,7 @@ export const getNodeDownloadSnippet = (
   if (os === 'MAC' || os === 'LINUX') {
     snippets.NVM = dedent`
       # ${t('layouts.download.codeBox.installsNvm')}
-      curl -o- https://raw.githubusercontent.com/nvm-sh/nvm/v0.40.0/install.sh | bash
+      curl -o- https://raw.githubusercontent.com/nvm-sh/nvm/${await getNvmData()}/install.sh | bash
 
       # ${t('layouts.download.codeBox.downloadAndInstallNodejsRestartTerminal')}
       nvm install ${release.major}
@@ -118,4 +119,4 @@ export const getNodeDownloadSnippet = (
   }
 
   return snippets;
-};
+}
--- a/apps/site/next-data/nvmData.ts
+++ b/apps/site/next-data/nvmData.ts
@@ -1,0 +1,17 @@
+import {
+  ENABLE_STATIC_EXPORT,
+  IS_DEVELOPMENT,
+  NEXT_DATA_URL,
+  VERCEL_ENV,
+} from '@/next.constants.mjs';
+
+export default async function getNvmData(): Promise<`v${string}`> {
+  if (ENABLE_STATIC_EXPORT || (!IS_DEVELOPMENT && !VERCEL_ENV)) {
+    const { default: provideNvmData } = await import(
+      '@/next-data/providers/nvmData'
+    );
+    provideNvmData();
+  }
+
+  return fetch(`${NEXT_DATA_URL}nvm-data`).then(r => r.json());
+}
--- a/apps/site/next-data/providers/nvmData.ts
+++ b/apps/site/next-data/providers/nvmData.ts
@@ -1,0 +1,7 @@
+import { cache } from 'react';
+
+import generateNvmData from '@/next-data/generators/nvmData.mjs';
+
+const nvmData = await generateNvmData();
+
+export default cache(() => nvmData);


apps/site/util/getNodeDownloadSnippet.ts
```diff
@@ -5,11 +5,23 @@ import type { NodeRelease } from '@/types';
...
+  .then(({ headers }) => {
+    const url = headers.get('location');
+    if (!url) { throw null; }
+    return fetch(url, { redirect: 'manual' });
+  })
+  .then(x => {
+    const url = x.headers.get('location');
+    return url?.slice(url?.lastIndexOf('/'));
+  })
+  .catch(() => 'v0.40.1');
```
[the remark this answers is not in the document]
ljharb: this entire PR _is_ that more dynamic way; the hardcoded version is necessary as a fallback. that comment seems useless to me.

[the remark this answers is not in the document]
ljharb: that's because it's not an error message, it's a fallback _value_

apps/site/util/getNodeDownloadSnippet.ts
```diff
@@ -5,11 +5,23 @@ import type { NodeRelease } from '@/types';
 import type { PackageManager } from '@/types/release';
 import type { UserOS } from '@/types/userOS';
 
-export const getNodeDownloadSnippet = (
+const latestVersion = fetch('https://latest.nvm.sh', { redirect: 'manual' })
```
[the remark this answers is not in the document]
ljharb: That's intentional, so that every invocation of the function shares the same version value. The `await` would reduce concurrency and make things slower.

mikeesto:
This isn't working in [staging](https://nodejs-org-git-fork-ljharb-nvm-latest-openjs.vercel.app/en/download/package-manager) because of CORS - is that expected?

`Access to fetch at 'https://latest.nvm.sh/' from origin 'https://nodejs-org-git-fork-ljharb-nvm-latest-openjs.vercel.app' has been blocked by CORS policy: No 'Access-Control-Allow-Origin' header is present on the requested resource.`

ljharb:
ahhhh ofc, i didn't think about that.

ideally this version would be fetched on the server, then, so it doesn't need to be done on the client. how is that done with the current setup?

mikeesto:
It's mostly here in https://github.com/nodejs/nodejs.org/tree/main/apps/site/next-data  . Here's a very rough patch

```diff
diff --git a/apps/site/app/[locale]/next-data/nvm-data/route.ts b/apps/site/app/[locale]/next-data/nvm-data/route.ts
new file mode 100644
index 00000000..aaf79a1b
--- /dev/null
+++ b/apps/site/app/[locale]/next-data/nvm-data/route.ts
@@ -0,0 +1,31 @@
+import provideNvmData from '@/next-data/providers/nvmData';
+import { defaultLocale } from '@/next.locales.mjs';
+
+// This is the Route Handler for the `GET` method which handles the request
+// for generating static data related to the Node.js Release Data
+// @see https://nextjs.org/docs/app/building-your-application/routing/router-handlers
+export const GET = async () => {
+  const nvmData = provideNvmData();
+
+  return Response.json(nvmData);
+};
+
+// This function generates the static paths that come from the dynamic segments
+// `[locale]/next-data/release-data/` and returns an array of all available static paths
+// This is used for ISR static validation and generation
+export const generateStaticParams = async () => [
+  { locale: defaultLocale.code },
+];
+
+// Enforces that only the paths from `generateStaticParams` are allowed, giving 404 on the contrary
+// @see https://nextjs.org/docs/app/api-reference/file-conventions/route-segment-config#dynamicparams
+export const dynamicParams = false;
+
+// Enforces that this route is used as static rendering
+// @see https://nextjs.org/docs/app/api-reference/file-conventions/route-segment-config#dynamic
+export const dynamic = 'error';
+
+// Ensures that this endpoint is invalidated and re-executed every X minutes
+// so that when new deployments happen, the data is refreshed
+// @see https://nextjs.org/docs/app/api-reference/file-conventions/route-segment-config#revalidate
+export const revalidate = 300;
diff --git a/apps/site/components/Downloads/Release/ReleaseCodeBox.tsx b/apps/site/components/Downloads/Release/ReleaseCodeBox.tsx
index ca465eb9..4a47199f 100644
--- a/apps/site/components/Downloads/Release/ReleaseCodeBox.tsx
+++ b/apps/site/components/Downloads/Release/ReleaseCodeBox.tsx
@@ -21,13 +21,18 @@ const ReleaseCodeBox: FC = () => {
   const t = useTranslations();
 
   useEffect(() => {
-    const updatedCode = getNodeDownloadSnippet(release, os, t)[platform];
-    // Docker and NVM support downloading tags/versions by their full release number
-    // but usually we should recommend users to download "major" versions
-    // since our Download Buttons get the latest minor of a major, it does make sense
-    // to request installation of a major via a package manager
-    memoizedShiki.then(shiki => shiki(updatedCode, 'bash')).then(setCode);
-    // Only react when the specific release number changed
+    async function getSnippet() {
+      const snippets = await getNodeDownloadSnippet(release, os, t);
+      const updatedCode = snippets[platform];
+
+      // Docker and NVM support downloading tags/versions by their full release number
+      // but usually we should recommend users to download "major" versions
+      // since our Download Buttons get the latest minor of a major, it does make sense
+      // to request installation of a major via a package manager
+      memoizedShiki.then(shiki => shiki(updatedCode, 'bash')).then(setCode);
+      // Only react when the specific release number changed
+    }
+    getSnippet();
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [release.versionWithPrefix, os, platform]);
 
diff --git a/apps/site/next-data/generators/nvmData.mjs b/apps/site/next-data/generators/nvmData.mjs
new file mode 100644
index 00000000..ad0f4b3e
--- /dev/null
+++ b/apps/site/next-data/generators/nvmData.mjs
@@ -0,0 +1,26 @@
+// File: next-data/generators/nvmData.mjs
+
+'use strict';
+
+/**
+ * Fetches the latest NVM version
+ * @returns {Promise<string>} Latest NVM version
+ */
+const generateNvmData = () => {
+  return fetch('https://latest.nvm.sh', { redirect: 'manual' })
+    .then(({ headers }) => {
+      const url = headers.get('location');
+      if (!url) {
+        throw new Error('No redirect location found');
+      }
+      return fetch(url, { redirect: 'manual' });
+    })
+    .then(x => {
+      const url = x.headers.get('location');
+      console.log('URL!!!', url?.slice(url?.lastIndexOf('/')));
+      return url?.slice(url?.lastIndexOf('/')) || 'v0.40.1';
+    })
+    .catch(() => 'v0.40.1');
+};
+
+export default generateNvmData;
diff --git a/apps/site/next-data/nvmData.ts b/apps/site/next-data/nvmData.ts
new file mode 100644
index 00000000..bdb1e52b
--- /dev/null
+++ b/apps/site/next-data/nvmData.ts
@@ -0,0 +1,18 @@
+import {
+  ENABLE_STATIC_EXPORT,
+  IS_DEVELOPMENT,
+  NEXT_DATA_URL,
+  VERCEL_ENV,
+} from '@/next.constants.mjs';
+
+const getNvmData = (): Promise<string> => {
+  if (ENABLE_STATIC_EXPORT || (!IS_DEVELOPMENT && !VERCEL_ENV)) {
+    return import('@/next-data/providers/nvmData').then(
+      ({ default: provideNvmData }) => provideNvmData()
+    );
+  }
+
+  return fetch(`${NEXT_DATA_URL}nvm-data`).then(r => r.json());
+};
+
+export default getNvmData;
diff --git a/apps/site/next-data/providers/nvmData.ts b/apps/site/next-data/providers/nvmData.ts
new file mode 100644
index 00000000..ed8eb6f2
--- /dev/null
+++ b/apps/site/next-data/providers/nvmData.ts
@@ -0,0 +1,9 @@
+import { cache } from 'react';
+
+import generateNvmData from '@/next-data/generators/nvmData.mjs';
+
+const nvmData = await generateNvmData();
+
+const provideNvmData = cache(() => nvmData);
+
+export default provideNvmData;
diff --git a/apps/site/util/getNodeDownloadSnippet.ts b/apps/site/util/getNodeDownloadSnippet.ts
index b1874f60..f9c0b727 100644
--- a/apps/site/util/getNodeDownloadSnippet.ts
+++ b/apps/site/util/getNodeDownloadSnippet.ts
@@ -1,11 +1,12 @@
 import dedent from 'dedent';
 import type { TranslationValues } from 'next-intl';
 
+import getNvmData from '@/next-data/nvmData';
 import type { NodeRelease } from '@/types';
 import type { PackageManager } from '@/types/release';
 import type { UserOS } from '@/types/userOS';
 
-export const getNodeDownloadSnippet = (
+export const getNodeDownloadSnippet = async (
   release: NodeRelease,
   os: UserOS,
   t: (key: string, values?: TranslationValues) => string
@@ -18,6 +19,9 @@ export const getNodeDownloadSnippet = (
     CHOCO: '',
   };
 
+  const nvmData = await getNvmData();
+  console.log(nvmData);
+
   if (os === 'WIN' || os === 'MAC' || os === 'LINUX') {
     snippets.DOCKER = dedent`
       # ${t('layouts.download.codeBox.noteWithColon')}
@@ -39,7 +43,7 @@ export const getNodeDownloadSnippet = (
   if (os === 'MAC' || os === 'LINUX') {
     snippets.NVM = dedent`
       # ${t('layouts.download.codeBox.installsNvm')}
-      curl -o- https://raw.githubusercontent.com/nvm-sh/nvm/v0.40.0/install.sh | bash
+      curl -o- https://raw.githubusercontent.com/nvm-sh/nvm${nvmData}/install.sh | bash
 
       # ${t('layouts.download.codeBox.downloadAndInstallNodejsRestartTerminal')}
       nvm install ${release.major}

```

ljharb:
@mikeesto amazing, thanks! i've updated the PR (but still need to write tests)

apps/site/app/[locale]/next-data/nvm-data/route.ts
```diff
@@ -0,0 +1,31 @@
+import provideNvmData from '@/next-data/providers/nvmData';
+import { defaultLocale } from '@/next.locales.mjs';
+
+// This is the Route Handler for the `GET` method which handles the request
+// for generating static data related to the Node.js Release Data
```
Comment (mikeesto) [inline apps/site/app/[locale]/next-data/nvm-data/route.ts]:
```suggestion
// for generating static data for the latest nvm version
```

apps/site/app/[locale]/next-data/nvm-data/route.ts
```diff
@@ -0,0 +1,31 @@
...
+// @see https://nextjs.org/docs/app/building-your-application/routing/router-handlers
+export const GET = async () => {
+  const nvmData = provideNvmData();
+
+  return Response.json(nvmData);
+};
+
+// This function generates the static paths that come from the dynamic segments
+// `[locale]/next-data/release-data/` and returns an array of all available static paths
```
Comment (mikeesto) [inline apps/site/app/[locale]/next-data/nvm-data/route.ts]:
```suggestion
// `[locale]/next-data/nvm-data/` and returns an array of all available static paths
```

apps/site/components/Downloads/Release/ReleaseCodeBox.tsx
```diff
@@ -21,12 +21,20 @@ const ReleaseCodeBox: FC = () => {
...
+        memoizedShiki,
+        getNodeDownloadSnippet(release, os, t),
+      ]);
+
+      // Docker and nvm support downloading tags/versions by their full release number
+      // but usually we should recommend users to download "major" versions since
+      // our Download Buttons get the latest minor of a major, it does make sense
+      // to request installation of a major via a package manager
+      setCode(await shiki(updatedCode, 'bash'));
+      // Only react when the specific release number changed
```
mikeesto, line 35:
```suggestion
```

ovflowd:
@ljharb apologies, could you add some screenshots of what the changes are? 🙈

ljharb:
@ovflowd the only visible change will be that the version number in the nvm install URL will be up to date (it's currently out of date). I'm happy to provide screenshots tho, can you help me get the site running locally? (or i suppose one could click the vercel preview and look there?)

apps/site/next-data/generators/nvmData.mjs
```diff
@@ -0,0 +1,24 @@
...
+const latestKnownVersion = 'v0.40.1';
+
+/**
+ * Fetches the latest NVM version
+ * @returns {Promise<`v${string}`>} Latest NVM version
+ */
+export default async function generateNvmData() {
+  return fetch('https://latest.nvm.sh', { redirect: 'manual' })
+    .then(({ headers }) => {
+      const url = headers.get('location');
```
ovflowd, line 12: I'm afraid that on static exports Next.js will complain about this

ljharb: ok, why would it complain?

ovflowd (changes requested):
I appreciate the patch, but all of this feels something that should be done by something like `https://raw.githubusercontent.com/nvm-sh/nvm/latest/install.sh`

I'm unfortunately against adding vendor-specific code for a single version manager here. Imagine if every single version manager had some curious requirement such as this one 😅 

Could there be a latest.nvm.sh/install.sh?

apps/site/next-data/generators/nvmData.mjs
ovflowd: Ah, I misread; I thought you were trying to get the current environment (route) headers. Whose Next.js would complain as trying to access runtime-specific info.

ovflowd:
> @ovflowd the only visible change will be that the version number in the nvm install URL will be up to date (it's currently out of date). I'm happy to provide screenshots tho, can you help me get the site running locally? (or i suppose one could click the vercel preview and look there?)

Ah, thanks for the explanation No screenshots needed 🫡

ljharb:
I don't think "a version number" is all too vendor-specific?

nvm intentionally doesn't provide a "latest" URL because that would allow people to *update* to an nvm version that may have breaking changes, which would be dangerous - akin to putting `*` or `latest` in package.json. In this case, though, the people viewing these instructions are new users, so they should be getting the most recent version of the software (just like how you have logic here to do `nvm install 22` so that users always get the latest v22). Unfortunately github doesn't have as much flexibility as nvm or npm. Either way, concerns about vendor-specific things don't seem to really apply, since node's website isn't trying to be vendor-agnostic - it's actively making choices and recommending them - so vendor-specific code seems perfectly reasonable to me.

The alternative to a PR like this is automation that opens a PR to change the hardcoded version number every time nvm does a release, which will mean that nvm is out of date until a human reviews, merges, and deploys that PR.

ovflowd:
> In this case, though, the people viewing these instructions are new users, so they should be getting the most recent version of the software (just like how you have logic here to do nvm install 22 so that users always get the latest v22). 

That's precisely why the latest/install.sh shortcut should exist; people here are installing NVM for the first time. Also, the install script should probably tell people (if they already have nvm) that running the latest/install.sh could override their existing installation; regardless, this is an nvm problem, not ours.

I don't want to pollute this repository with one piece of version manager-specific code and add yet another piece that needs to be maintained and is dependent on NVM infrastructure or something else.

> The alternative to a PR like this is automation that opens a PR to change the hardcoded version number every time nvm does a release, which will mean that nvm is out of date until a human reviews, merges, and deploys that PR.

I would also be against that; I don't want to add specific code to one version manager. This is not an us problem, and the solution shouldn't be at the responsibility of the Node.js website... Sorry Jordan :/

ljharb:
It's a problem for nvm users, which makes it a node problem.

I can certainly make the automation myself, but it'll still be a PR that needs manual review.

In other words, it feels like you're prioritizing ideological code purity over user's needs here.

(as for "nvm infrastructure", it's openjs infrastructure, which is precisely as reliable as node's own infra since it's owned by the same organization)

ovflowd:
> I would also be against that; I don't want to add specific code to one version manager. This is not an us problem, and the solution shouldn't be at the responsibility of the Node.js website... Sorry Jordan :/

Although I'd be willing to accept a patch for that, given the version is stored within a said JSON file or something that we could read from during build time.

ovflowd:
> It's a problem for nvm users, which makes it a node problem.

I don't think it is fair to say it is a Node problem by being a constraint you're intentionally creating for your own users. No other version manager has such constraint on their install script (that we have listed here on the CodeBoxes)

> In other words, it feels like you're prioritizing ideological code purity over user's needs here.

I'm prioritising DX and removing nvm-specific knowledge as a burden to my contributors. Which also removes the burden on you...

> (as for "nvm infrastructure", it's openjs infrastructure, which is precisely as reliable as node's own infra since it's owned by the same organization)

Touche.

ovflowd:
Anyhow, I'm sorry, but I won't remove my blocker as this stands, at least now. Let me mutter internally with myself and reflect on this PR. Feel free to open an alternative PR in the meantime.

ljharb:
To restate, the alternative approach is that I make a workflow on nvm's own repo, and have it submit an automated PR to this repo every time nvm does a release. If you'd prefer an automated PR, and a delay until a human gets around to landing and deploying the PR before users get the latest version of nvm, then I can certainly set that up in the meantime. It's somewhat pressing since this is stemming from user-reported issues.

ovflowd:
> To restate, the alternative approach is that I make a workflow on nvm's own repo, and have it submit an automated PR to this repo every time nvm does a release. If you'd prefer an automated PR, and a delay until a human gets around to landing and deploying the PR before users get the latest version of nvm, then I can certainly set that up in the meantime. It's somewhat pressing since this is stemming from user-reported issues.

I think it is fine opening a Workflow in this repository that reacts on new releases to nvm/nvm.

Issue #7294, opened 2024-11-29
 Map and add frontmatter on learn pages tying to core apis


bmuenzenmeyer:
> Broken out of #7197 

Precursor work to for #7292

Map and add frontmatter on learn pages tying to core apis

@AugustinMauroy suggested

> I'd be more in favour of having an MDX component that puts the displayName and stability index linked to the version (idea above of having it in the frontmatter) and that also allows links on the page.

I think the ideas can feed off of each other. With any solution, a static declaration of the content / api module is necessary.

AugustinMauroy:
to start with, I propose to have a frontmatter with the version and a text in the metabar that links to the docs.
The second part is to have an MDX component that takes the version into account for the docs.

PR #7310, opened 2024-12-04
meta: add security-wg as owners of security practices

RafaelGSS:


bmuenzenmeyer (approved):
I gave `@nodejs/security-wg` write access to the repo.
CODEOWNERS only works for teams that _OWN_ code, meaning, they can write.

This is a pattern we will likely repeat as part of #7294

AugustinMauroy (approved):
LGTM

RafaelGSS:
> I gave `@nodejs/security-wg` write access to the repo. CODEOWNERS only works for teams that _OWN_ code, meaning, they can write.
> 
> This is a pattern we will likely repeat as part of #7294

Initially, my idea was to ping @nodejs/security-wg when changing the repo. It's fine to not have write access, but having a ping is ideal.

marco-ippolito (approved):

PR #7327, opened 2024-12-12
Add CSS structure for country-specific style overrides (#7326)

### LINKED ISSUE #7326: Proposal for a CSS Structure to Support Country-Specific Style Overrides
## Description
To improve the accessibility and readability of the Node.js website for non-English languages, especially Korean, I propose a CSS structure that supports country-specific style overrides. This approach addresses issues such as narrow line spacing and unnatural line breaks, enhancing readability for languages like Korean. By providing a mechanism for locale-specific style customization, we can improve the user experience for all audiences.

## Problem
The current CSS structure does not account for the typographic and layout differences required for various languages and locales. For example:

* Korean (ko): The default line height and text wrapping behavior are not optimized for Korean typography, resulting in cramped text and poor readability.
* Other languages: Similar issues may arise for other languages or regions, such as Arabic, Japanese, or Chinese.

Introducing a country-specific CSS override mechanism would allow localized adjustments without compromising global design consistency.

## Proposed Solution

1. Add a `locals.css` file
    * Create a dedicated locals.css file to include locale-specific overrides.
2. Use the html[lang] attribute
    * Define country-specific styles using the lang attribute:
```css
html[lang='ko'] {  
  overflow-wrap: anywhere;  
  word-break: keep-all;  
  line-height: 1.8;  
}

/* Define other locale-specific styles */  
```
3. Update the CSS import structure
    * Include locals.css at the end of the main CSS file:
```css
@import 'tailwindcss/base';
@import 'tailwindcss/components';
@import 'tailwindcss/utilities';
@import './base.css';
@import './markdown.css';
@import './effects.css';
@import './locals.css';
```

### **Before and After Comparison**
| As-Is (Current) | To-Be (Proposed) |
|------------------|------------------|
| as-is | to-be |

## Expected Benefits

* Improved readability for non-English locales, particularly Korean.
* Enhanced user experience through better typographic adjustments.
* Flexibility to add and modify styles for additional locales as needed.

## Potential Challenges

* Ensuring that new styles do not unintentionally override global styles.
* Maintaining consistency across locales while allowing for necessary differences.

----
This structure provides a clear and scalable solution to address the typographic and layout needs of different languages. If you have any feedback or suggestions, I’d be happy to collaborate further.

mpruser:
## Description

This PR introduces a CSS structure to support country-specific style overrides. The new structure allows defining localized styles efficiently while maintaining global consistency.

## Validation
- Verified that global styles are unaffected by the changes in `locals.css`.  
- Tested that the styles (`line-height: 1.8`, `word-break: keep-all`, `word-wrap: break-word`) are correctly applied when the HTML `lang='ko'` attribute is present.  
- Visually compared the text layout before and after applying the Korean-specific styles to ensure proper implementation.  
- Ensured that styles for other languages (e.g., `lang='en'`) remain unchanged.

## Related Issues

Fixes #7326

### Check List

### DIFF
--- a/apps/site/styles/locals.css
+++ b/apps/site/styles/locals.css
@@ -1,5 +1,3 @@
 html[lang='ko'] {
-  line-height: 1.8;
-  word-break: keep-all;
-  word-wrap: break-word;
+  @apply break-words break-keep leading-[1.8];
 }
--- a/apps/site/styles/locals.css
+++ b/apps/site/styles/locals.css
@@ -1,3 +1,10 @@
+/**
+ * To enhance readability for Korean users, line spacing is increased,
+ * line breaks in the middle of words are prevented, and long words are
+ * managed to avoid disrupting the layout.
+ */
 html[lang='ko'] {
-  @apply break-words break-keep leading-[1.8];
+  @apply break-words
+    break-keep
+    leading-7;
 }
--- a/apps/site/styles/locals.css
+++ b/apps/site/styles/locals.css
@@ -1,0 +1,5 @@
+html[lang='ko'] {
+  line-height: 1.8;
+  word-break: keep-all;
+  word-wrap: break-word;
+}
--- a/apps/site/styles/index.css
+++ b/apps/site/styles/index.css
@@ -12,3 +12,4 @@
 @import './base.css';
 @import './markdown.css';
 @import './effects.css';
+@import './locals.css';


apps/site/styles/locals.css
```diff
@@ -0,0 +1,5 @@
+html[lang='ko'] {
+  line-height: 1.8;
+  word-break: keep-all;
+  word-wrap: break-word;
```
AugustinMauroy: You have to use tailwind classes with `@apply` directive

AugustinMauroy (commented):
is there are other languages need this change ?

lucas-rxc:
> is there are other languages need this change ?

I'm not certain if there are additional languages that would require these specific changes. However, implementing such functionality provides a pathway to address usability issues caused by linguistic differences. I believe this approach is neither excessive nor unnecessary, as it allows for improved accessibility and adaptability in a multilingual context. 🙏

mpruser:
> is there are other languages need this change ?

I'm not certain if there are additional languages that would require these specific changes. However, implementing such functionality provides a pathway to address usability issues caused by linguistic differences. I believe this approach is neither excessive nor unnecessary, as it allows for improved accessibility and adaptability in a multilingual context. 🙏

apps/site/styles/locals.css
mpruser: Thanks for the feedback! I'll make the changes and update the PR shortly😀

apps/site/styles/locals.css
```diff
@@ -0,0 +1,3 @@
+html[lang='ko'] {
+  @apply break-words break-keep leading-[1.8];
```
mpruser: @AugustinMauroy 
I've made the changes. Since I'm not very familiar with Tailwind, I'd really appreciate it if you could take a look and let me know if there's anything that needs improvement.😀

AugustinMauroy: You need to add break line and indentation between each tailwind classes. You can take a look to other css file on the codebase.
Also why specify leading with 1.8 ? why not use `leading-7` ?

apps/site/styles/locals.css
```diff
@@ -0,0 +1,3 @@
+html[lang='ko'] {
```
AugustinMauroy, line 6:
```suggestion
/* Comment to explain why, this tile for ko */
html[lang='ko'] {
```

Can you document why, you put theses style for Korean

mpruser: Comments have been added!

apps/site/styles/locals.css
mpruser: It didn't have to be 1.8. I just chose a value that seemed appropriate from the recommended ones, but I've changed it to `leading-7` as you suggested. I'm still new to using Tailwind, so I wasn't very experienced. Thanks for letting me know!

AugustinMauroy (approved):
LGTM !

aymen94 (approved):

mikeesto (approved):
LGTM, thanks

PR #7336, opened 2024-12-18
Update asynchronous-flow-control.md

mrwasii:
### Check List


mikeesto (commented):
Hi. I think this code block is just meant to be an illustrative example of what "callback hell" looks like. This change adds a lot of lines for not a lot of benefit in the context of the page. Does that make sense?

bmuenzenmeyer:
I'm 👎 on this change - the intent of the content, IMO, is brief explanation of what we mean. Expanding the code to be more complete actually hurts that comprehension.

mikeesto:
Thanks for the PR but I don't think this will be landing

PR #7340, opened 2024-12-19
Fix FR about page

ghost:
Remove repetitions and change a couple of words to use the technical idiom in French.

## Description
I am French (so a native french speaker)
Reading the About page in French, I found a lot of repetitions, maybe due to a commit/merge issue. I also took the freedom to change a couple of words for their more appropriate french technical idiom (child process > processus fils).

## Validation

This only a text update so I assume it should not break anything. Not knowing how the whole NPM stuff works, and not having it installed, I could not run all the tests of the checklist.

### Check List

### DIFF
--- a/apps/site/pages/fr/about/index.mdx
+++ b/apps/site/pages/fr/about/index.mdx
@@ -6,8 +6,8 @@ layout: about
 À propos de Node.js®
 
 En tant que moteur d'exécution JavaScript asynchrone piloté par les événements, Node.js est conçu pour construire des applications réseau évolutives.
-des applications réseau évolutives. Dans l'exemple suivant, "hello world", de nombreuses
-peuvent être gérées simultanément. À chaque connexion, le rappel est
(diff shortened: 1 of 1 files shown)


TenzDelek:
i dont think we are suppose to make changes in localization here. it is done through crowdin

ghost:
> i dont think we are suppose to make changes in localization here. it is done through crowdin

I clicked the link "Edit this page" from the website... and ended on the GitHub editor from where I made my edits.
You may want to remove the link if it's not the correct method to update the page.

ghost:
The french about page still needs to be fixed.

AugustinMauroy:
> The french about page still needs to be fixed.

First of all, it's not the term “fix” that should be used, but “correct”, because fix is the term for a bug and correct for typography/grammar.

As mentioned above and on the `translation.md` document, all questions concerning translation are answered on crowdin. So if you think there's something to revise, say so there.

PR #7374, opened 2024-12-31
Increase font size of headings in metabar table of contents

### LINKED ISSUE #7275: Visually differentiate H2 from H3 headings in Metabar Table of Contents

yaten2302:
## Description
This PR increases the font size of the headings in table of contents.

## Validation
Before:
image

After:
image

## Related Issues
Fixes #7275 

### Check List

### DIFF
--- a/apps/site/components/Containers/MetaBar/index.module.css
+++ b/apps/site/components/Containers/MetaBar/index.module.css
@@ -22,7 +22,7 @@
 
   dt {
     @apply mb-2
-      text-sm
+      text-[17px]
       font-medium
       text-neutral-800
       dark:text-neutral-200;


bmuenzenmeyer:
sorry, this is not the approach we need - the original issue should be amended to be more prescriptive I guess.

We need to go way deeper into the mdx compilation / data gathering step to determine the heading level of each heading, `h2` vs `h3`, etc, so that they can rendered differently. this and https://github.com/nodejs/nodejs.org/pull/7286 both do not achieve that

PR #7385, opened 2025-01-03
Visually differentiate table of content heading levels

### LINKED ISSUE #7275: Visually differentiate H2 from H3 headings in Metabar Table of Contents

bmuenzenmeyer:
## Description

Previous attempts to fix this did not use `head.depth`. My quick looks at the code made it seem that this data was not available, but luckily it is! I should have known, because we constrain the headings we show already (omitting h1, h5, and h6). 

Finding this out was as quick as relaying it to contributors to fix - so I added a story and tests for visibility.

## Validation

Here's a before (left) and after (right) from a recently published Learn article

image

Note the component wraps fine on smaller viewports

image

See the preview.

## Related Issues

fixes #7275 

### Check List

### DIFF
--- a/apps/site/components/Containers/MetaBar/__tests__/index.test.mjs
+++ b/apps/site/components/Containers/MetaBar/__tests__/index.test.mjs
@@ -1,0 +1,60 @@
+import { render } from '@testing-library/react';
+
+import MetaBar from '..';
+
+describe('MetaBar', () => {
+  it('does not render h5s in the table of contents', () => {
+    const { queryByText, getByText } = render(
+      <MetaBar
+        items={{}}
+        headings={{
+          items: [
+            {
+              value: 'Heading Level 1',
+              depth: 1,
+              data: { id: 'heading-1' },
+            },
+            {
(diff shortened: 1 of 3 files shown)


ovflowd (approved):
SGTM

AugustinMauroy (approved):
LGMT !

canerakdas (approved):
LGTM!

PR #7387, opened 2025-01-05
feat(learn): add section for dynamically generating test cases

JakobJingleheimer:
### Check List

### DIFF
--- a/apps/site/pages/en/learn/test-runner/using-test-runner.md
+++ b/apps/site/pages/en/learn/test-runner/using-test-runner.md
@@ -66,30 +66,63 @@ Each example below was taken from real-world projects; they may not be appropria
 
 ## Dynamically generating test cases
 
-Some times, you may want to dynamically generate test-cases. For instance, you want to test the same thing across a bunch of files. This is possible, albeit slightly arcane. You must use `test` (you cannot use `describe`) + `await testContext.test`:
+Some times, you may want to dynamically generate test-cases. For instance, you want to test the same thing across a bunch of files. This is possible, albeit slightly arcane. You must use `test` (you cannot use `describe`) + `testContext.test`:
 
-```js
+```js displayName="simple example (prior to 23.8.0)"
 import assert from 'node:assert/strict';
-import { globSync } from 'node:fs';
 import { test } from 'node:test';
-import { fileURLToPath } from 'node:url';
+
+import { detectOsInUserAgent } from '…';
+
+const userAgents = [
+  { ua: '…', os: 'WIN' },
+  // …
+];
+
+test('Detect OS via user-agent', { concurrency: true }, async t => {
+  const cases = userAgents.map(({ os, ua }) => {
+    t.test(ua, () => assert.equal(detectOsInUserAgent(ua), os));
+  });
+
+  await Promise.allSettled(cases);
+});
+```
+
+```js displayName="simple example (23.8.0 and later)"
+import assert from 'node:assert/strict';
+import { test } from 'node:test';
+
+import { detectOsInUserAgent } from '…';
+
+const userAgents = [
+  { ua: /* … */, os: 'WIN' },
+  // …
+];
+
+test('Detect OS via user-agent', { concurrency: true }, t => {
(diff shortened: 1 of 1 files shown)


AugustinMauroy:
I'm not fan of the provided example because it's based on experimental api.

maybe do same things as this test (do with jest). But I mean based on hand written variable
https://github.com/nodejs/nodejs.org/blob/main/apps/site/util/__tests__/detectOS.test.mjs

JakobJingleheimer:
~Hmm, I think that example isn't quite as good. The one I have is unknowable.~

~I'm happy to just switch the assertion (and far less work 😅)~

EDIT: ohhh! Wait, is the experimental API you're concerned about `globSync`? (Rather than `assert.partialDeepStrictEqual`)

apps/site/pages/en/learn/test-runner/using-test-runner.md
```diff
@@ -64,6 +64,40 @@ Then for each setup, create a dedicated `setup` file (ensuring the base `setup.m
...
+    ).map(path => import(path, { with: { type: 'json' } }))
+  );
+
+  const cases = pjsons.map(pjson =>
+    // ⚠️ `t.test`, NOT `test`
+    t.test(`Ensure fields are properly set: ${pjson.name}`, () => {
+      assert.partialDeepStrictEqual(pjson.keywords, [
+        'node.js',
+        'sliced bread',
+      ]);
```
JakobJingleheimer:
```suggestion
      assert.match(pjson.name, /@nodejs\/[a-z \-]+/);
```

apps/site/pages/en/learn/test-runner/using-test-runner.md
```diff
@@ -64,6 +64,40 @@ Then for each setup, create a dedicated `setup` file (ensuring the base `setup.m
...
+  const pjsons = await Promise.all(
+    globSync(
+      // Get all the package.json files 1-level deep within ./workspaces
+      // ⚠️ Passing a file URL string, like from import.meta.resolve, causes glob* to fail silently
+      fileURLToPath(import.meta.resolve('./workspaces/*/package.json'))
+    ).map(path => import(path, { with: { type: 'json' } }))
+  );
+
+  const cases = pjsons.map(pjson =>
+    // ⚠️ `t.test`, NOT `test`
```
AugustinMauroy, line 145: and if you use the describe it syntax what happened ?

AugustinMauroy:
> EDIT: ohhh! Wait, is the experimental API you're concerned about globSync? (Rather than assert.partialDeepStrictEqual)

Yup and and using a variable is easier because the reader can copy past and run.

apps/site/pages/en/learn/test-runner/using-test-runner.md
JakobJingleheimer: It does nothing (it does not notice the dynamically generated test-cases).

JakobJingleheimer:
> using a variable is easier because the reader can copy past and run

A variable for what?

AugustinMauroy:
> A variable for what?

```js
import { isBiscuit } from "factory";

const valid_tests_case = [
  "cookies",
  "speculoos"
];

test("isBiscuit", (t) => {
   for() {
   // loop for dynamic test 
   t.test(`${e} - should be true`)
   }
});
```

in this case it's use Biscuit because it's fun

But [here](https://github.com/nodejs/nodejs.org/blob/main/apps/site/util/__tests__/detectOS.test.mjs) it's test user agent

ovflowd (approved):
Feel free to merge whenever the PR becomes ready.

JakobJingleheimer:
Hmm, I've given this some more thought, and I think the experimental API used is okay because it does not affect the main issue of the article. Thoughts on that?

How about both options? The user-agent one is simple, and the glob is advanced; I know I frequently see docs cover a simple example but not an advanced one, which leaves an advanced user on their own to figure out something complicated (which is very frustrating).

AugustinMauroy (approved):
LGTM ! maybe open an issue after this pr merged to explore the possibility of making one article per use case, as this one is starting to get long.

PR #7419, opened 2025-01-24
docs: #7416 - Removes winget from documentation because of unreliability

### LINKED ISSUE #7416: NodeJS in winget is not up to date
### URL:

https://nodejs.org/en/download/package-manager/all#windows-1

### Browser Name:

Chrome

### Browser Version:

128.0.6613.85

### Operating System:

Windows

The documentations is referencing WinGet as possiblity to install nodejs on windows.
Unfortunately, winget is only having version up to 19
https://winget.run/pkg/OpenJS/NodeJS 

Can you please publish more recent version to winget

stefanseeger:
Remove winget from documentation because it is unreliable

## Description

NodeJS versions are unreliably published to winget. Therefore, this reference falsy gives the impression that it will contain the latest versions.

## Validation

Current LTS is 22.13.1
Image

which is non existing

```
winget install --id=OpenJS.NodeJS -v "22.13.1" -e
No version found matching: 22.13.1
```

# Result of change
image

### Check List

### DIFF
--- a/apps/site/pages/en/download/package-manager/all.md
+++ b/apps/site/pages/en/download/package-manager/all.md
@@ -364,17 +364,6 @@ Download the [Windows Installer](/#home-downloadhead) directly from the [nodejs.
 
 ### Alternatives
 
-Using **[Winget](https://aka.ms/winget-cli)**:
-
-```bash
-winget install OpenJS.NodeJS
-# or for LTS
-winget install OpenJS.NodeJS.LTS
-```
-
-After running one of the two commands above, it may be necessary to restart the
-terminal emulator before the `node` CLI command becomes available.
-
 Using **[Chocolatey](https://chocolatey.org/)**:
 
 ```bash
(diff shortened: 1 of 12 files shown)


AugustinMauroy:
Do not change translation because it's handle by crowdin.

AugustinMauroy:
also why not put winged on "main" download page ?

cc @ovflowd

aduh95:
I’m not sure we have enough information to claim winger is unreliable, it looks like the latest LTS should be available as it was added in https://github.com/microsoft/winget-pkgs/pull/214141.

stefanseeger:
Okay, I think now I understand what is happening.
OpenJS.NodeJS.LTS contains only the LTS versions.
OpenJS.NodeJS other versions?

But even for OpenJS.NodeJS.LTS I'm missing the patched LTS 20.18.2 only 20.18.0 is available.

Obviously I will close this PR, but I still think that this versioning is confusing and the published versions are incomplete.

PR #7426, opened 2025-01-25
Remove reference-style markdown links from translated content

### LINKED ISSUE #7425: Links in  i18n
Currently, there are elements that are links but are not rendering correctly because we translate the link text, which causes markdown to not be able to redirect. For example, in Spanish, it renders as follows:

Image

Solution: not translating the link text, but I wouldn't like to mix Spanish with English when it's not necessary.

The same happens for other languages, for example, in French.

Image

bmuenzenmeyer:
## Description

[Reference style links](https://daringfireball.net/projects/markdown/syntax), those that look like this:

```md
This is [an example][id] reference-style link.
...
[id]: http://example.com/  "Optional Title Here"
```

have proven to be problematic for i18n efforts, as the `id`, when omitted, defaults to the text being used as the reference, such as `an example` above. When this happens, translators are not aware, nor would they really have the tools in crowdin, to maintain this association. We are overloading the label with the reference. This is odd and has broken `id` and `es` locales for quite some time.

The fix is to inline our markdown links. We don't need reference-style markdown links here, or nearly anywhere. This should solve the [cited manual fixes](https://github.com/nodejs/nodejs.org/pull/7418#discussion_r1929601532) once and for all, because to my knowledge crowdin is smart enough to allow users to translate the text, but not the link. Reference-style links had no such association.

Omitted in this PR is fixing some that I found in the `learn/` section. That content is not translated currently.

## Validation

- https://nodejs-org-git-normalize-markdown-links-openjs.vercel.app/en/about
- https://nodejs-org-git-normalize-markdown-links-openjs.vercel.app/en/about/governance

## Related Issues

closes #7425
relates to https://github.com/nodejs/nodejs.org/pull/7418

### Check List

### DIFF
--- a/apps/site/pages/en/about/governance.md
+++ b/apps/site/pages/en/about/governance.md
@@ -7,31 +7,24 @@ layout: about
 
 ## Consensus Seeking Process
 
-The Node.js project follows a [Consensus Seeking][] decision making model.
+The Node.js project follows a [Consensus Seeking](https://en.wikipedia.org/wiki/Consensus-seeking_decision-making) decision making model.
 
 ## Collaborators
 
-The [nodejs/node][] core GitHub repository is maintained by the Collaborators
+The [nodejs/node](https://github.com/nodejs/node) core GitHub repository is maintained by the Collaborators
 who are nominated by other existing Collaborators on an ongoing basis.
 
 Individuals making significant and valuable contributions are made Collaborators
 and given commit-access to the project. These individuals are identified by other
 Collaborators and their nomination is discussed with the existing Collaborators.
 
-For the current list of Collaborators, see the project's [README.md][].
+For the current list of Collaborators, see the project's [README.md](https://github.com/nodejs/node/blob/main/README.md#current-project-team-members).
 
-A guide for Collaborators is maintained at [collaborator-guide.md][].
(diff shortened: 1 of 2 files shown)


bjohansebas (commented):
LGTM :)

AugustinMauroy (approved):
LGMT !

AugustinMauroy:
This markdown feature comme from https://github.com/remarkjs/remark-gfm?tab=readme-ov-file#remark-gfm should we remove this plugins.

I mean there are still thing that use theses features ?

bmuenzenmeyer:
> This markdown feature comme from https://github.com/remarkjs/remark-gfm?tab=readme-ov-file#remark-gfm should we remove this plugins.
> 
> I mean there are still thing that use theses features ?

i would imagine we use github flavored markdown all over the place - this is just one particular pattern that is usually nice on its own (DRY dont repeat yourself) but leads to translation problems because we overload the usage of the text

AugustinMauroy:
> i would imagine we use github flavored markdown all over the place - this is just one particular pattern that is usually nice on its own (DRY dont repeat yourself) but leads to translation problems because we overload the usage of the text

For example `<details>` isn't parse on our site because this plugin remove them, also `<cite>` tag is removed.

So any strong opinion on this plugin. but we need to compare the replication/usage of the plugin

mikeesto (approved):

ovflowd (approved):

aymen94 (approved):

PR #7439, opened 2025-02-01
[DRAFT] - attempt to fix windows CI

mikeesto:
### Check List

### DIFF
--- a/apps/site/app/[locale]/next-data/og/[category]/[title]/route.tsx
+++ b/apps/site/app/[locale]/next-data/og/[category]/[title]/route.tsx
@@ -55,7 +55,7 @@ export const generateStaticParams = async () => [
   {
     locale: defaultLocale.code,
     category: 'announcement',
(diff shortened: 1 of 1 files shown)


mikeesto:
I'm going to close this as I don't have a Windows machine to reliably test on. From what I can tell, the issue seems to be related to these lines in the vercel/og package:

```js
var fontData = fs.readFileSync(fileURLToPath(join(import.meta.url, "../noto-sans-v27-latin-regular.ttf")));
var yoga_wasm = fs.readFileSync(fileURLToPath(join(import.meta.url, "../yoga.wasm")));
var resvg_wasm = fs.readFileSync(fileURLToPath(join(import.meta.url, "../resvg.wasm")));
```

Perhaps import.meta.url is not resolving to a valid/expected path? This package doesn't seem to be OSS either so that complicates things too.

AugustinMauroy:
What if we use `next/og`?

PR #7463, opened 2025-02-01
fix: remove confusing close label for navigation menu

aduh95:
## Description

As I was trying around navigating the website with VoiceOver, I noticed the current design is kinda confusing:

- when the nav menu is closed, and the user focuses the checkbox, VoiceOver says "Show navigation menu, tickbox, unticked" (so far so good).
- when the nav menu is open, and the user focuses the checkbox, VoiceOver says "Close navigation menu, tickbox, ticked", which may be kind of confusing. If the user is not able to get visual clues to know if the menu is actually currently opened, they might assume the menu is actually closed (since the checkbox is ticked).

TBH I'm not at all confident which version is the correct one, if someone who are a screen reader user could chime in, that'd be much appreciated.
An alternative would be to use a more verbose label (such as "Navigation menu is currently closed, tick the box to open it", and "Navigation menu is currently open, untick the box to close it"), not sure if this is preferable.

## Related Issues

https://github.com/nodejs/nodejs.org/pull/7427

### Check List

### DIFF
--- a/apps/site/components/Containers/NavBar/index.tsx
+++ b/apps/site/components/Containers/NavBar/index.tsx
@@ -70,9 +70,7 @@ const NavBar: FC<NavbarProps> = ({
         id="sidebarItemToggler"
         type="checkbox"
         onChange={e => setIsMenuOpen(() => e.target.checked)}
-        aria-label={t(
-          `components.containers.navBar.controls.${isMenuOpen ? 'close' : 'open'}`
-        )}
+        aria-label={t(`components.containers.navBar.controls.open`)}
       />
 
       <div className={`${style.main} peer-checked:flex`}>
--- a/apps/site/components/Containers/NavBar/index.tsx
+++ b/apps/site/components/Containers/NavBar/index.tsx
@@ -70,7 +70,7 @@ const NavBar: FC<NavbarProps> = ({
         id="sidebarItemToggler"
         type="checkbox"
         onChange={e => setIsMenuOpen(() => e.target.checked)}
-        aria-label={t(`components.containers.navBar.controls.open`)}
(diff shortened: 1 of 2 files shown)


AugustinMauroy (commented):
maybe something like toggle ?

aduh95:
> maybe something like toggle ?

I don't understand your comment, can you clarify?

packages/i18n/locales/en.json
```diff
@@ -12,8 +12,7 @@
       },
       "navBar": {
         "controls": {
-          "open": "Show navigation menu",
-          "close": "Close navigation menu"
+          "open": "Show navigation menu"
```
AugustinMauroy: > I don't understand your comment, can you clarify?

Here we can use "toggle" because the open button it's also the one that close

aduh95:
```suggestion
          "toggle": "Toggle navigation menu"
```

apps/site/components/Containers/NavBar/index.tsx
```diff
@@ -70,9 +70,7 @@ const NavBar: FC<NavbarProps> = ({
         id="sidebarItemToggler"
         type="checkbox"
         onChange={e => setIsMenuOpen(() => e.target.checked)}
-        aria-label={t(
-          `components.containers.navBar.controls.${isMenuOpen ? 'close' : 'open'}`
-        )}
+        aria-label={t(`components.containers.navBar.controls.open`)}
```
aduh95:
```suggestion
        aria-label={t(`components.containers.navBar.controls.toggle`)}
```

AugustinMauroy (approved):
LGMT !

PR #7495, opened 2025-02-13
feat: display the reading time in a formatted for the locale

okuryu:
## Description

The display of the reading time for each page is currently based on the assumption that the page is in English. This change makes the Intl API is used to display the reading time in a format appropriate for the locale of the translated language.

### Check List

### DIFF
--- a/apps/site/components/withMetaBar.tsx
+++ b/apps/site/components/withMetaBar.tsx
@@ -20,6 +20,10 @@ const WithMetaBar: FC = () => {
   const lastUpdated = frontmatter.date
     ? formatter.dateTime(new Date(frontmatter.date), DEFAULT_DATE_FORMAT)
     : undefined;
+  const readingTimeText = formatter.number(Math.round(readingTime.minutes), {
+    style: 'unit',
+    unit: 'minute',
+  });
 
   const usernames =
     frontmatter.authors?.split(',').map(author => author.trim()) ?? [];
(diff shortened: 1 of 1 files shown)


apps/site/components/withMetaBar.tsx
```diff
@@ -20,6 +20,10 @@ const WithMetaBar: FC = () => {
   const lastUpdated = frontmatter.date
     ? formatter.dateTime(new Date(frontmatter.date), DEFAULT_DATE_FORMAT)
     : undefined;
+  const readingTimeText = formatter.number(Math.round(readingTime.minutes), {
```
araujogui: Why `Math.round`? As far as I know, `Intl.NumberFormat` rounds automatically.

okuryu: As far as I have tried, I needed `Math.round()` as follows.

***
image

AugustinMauroy: why not use [next-intl api](https://next-intl.dev/docs/usage/numbers) ?

okuryu: I’m using next-intl for this, is something wrong?

AugustinMauroy: Oh yeah I didn't see it. Oups my bad. So yeah we should try to avoid `Math`

okuryu: 6cfa18a14b58211fb7aef2c59ee6876b4e4ed1c3

I'm not sure why we should avoid the standard Math functions, but I used the `maximumFractionDigits` option instead. How does this look?

araujogui: > [6cfa18a](https://github.com/nodejs/nodejs.org/commit/6cfa18a14b58211fb7aef2c59ee6876b4e4ed1c3)
> 
> I'm not sure why we should avoid the standard Math functions, but I used the `maximumFractionDigits` option instead. How does this look?

lgtm

bjohansebas (commented):
LGTM :)

bmuenzenmeyer (approved):

AugustinMauroy (approved):

bmuenzenmeyer:
Windows still failing

PR #7566, opened 2025-03-24
feat(actions): publish `@node-core/ui-components`

### LINKED ISSUE #236: How are we going to import `@node-core/ui-components`?
npm does not support importing packages from subdirectories of git repositories.

According to https://github.com/npm/npm/issues/2974#issuecomment-169451651:  
> This isn't something that the npm CLI team is interested in adding to the tool.  

There are third-party solutions like [GitPkg](https://gitpkg.vercel.app/about) that allow this functionality in npm, but I'm not sure we should rely on a third-party tool to handle imports–lest something bad happen. 

**However**, both [Yarn](https://github.com/yarnpkg/yarn/issues/4725#issuecomment-753416391) (although, yarn requires the website to also use yarn) and [pnpm](https://github.com/pnpm/pnpm/pull/7487) do support this, so we *could* convert this repository to use one of them instead.

This would also require any project that runs `api-docs-tooling` use the same package manager (i.e. `pnpm dlx` instead of `npx` for execution).  

Reference: https://github.com/nodejs/nodejs.org/pull/7401#issuecomment-2748180596

avivkeller:
It's hard to properly test this, as it makes use of GitHub tokens, but based on my testing of each step independently, this'll publish the `@node-core/ui-components` package.

This PR publishes `@node-core/ui-components to the GitHub registry. **Before merging, we need to verify that we have access to this scope**.

For security reasons, several checks are in place to prevent malicious execution:
- The commit must've passed all checks on `main`'s `Linting and Tests` workflow, and that check must've been triggered by a `push` event to `main`, or the workflow was manually dispatched
- The commit **must** be verified.
- Unless explicitly specified in the `workflow_dispatch` event, the commit **must** come from the merge queue.
	- i.e. a push directly to `main` will not be published automatically 

Additionally, the commit must've modified something in `packages/ui-components`.

---

Ref: #7401
Closes https://github.com/nodejs/api-docs-tooling/issues/236

### DIFF
--- a/.github/workflows/publish-ui-components.yml
+++ b/.github/workflows/publish-ui-components.yml
@@ -39,7 +39,7 @@ jobs:
             exit 1
           fi
 
-          if [[ "${{ github.event.inputs.skip_github_check }}" != "true" && "$COMMITTER" != *"noreply@github.com" ]]; then
+          if [[ "${{ github.event.inputs.skip_github_check }}" != "true" && "$COMMITTER" != "noreply@github.com" ]]; then
             echo "❌ Unauthorized committer! Aborting."
             exit 1
           fi
--- a/.github/workflows/publish-ui-components.yml
+++ b/.github/workflows/publish-ui-components.yml
@@ -1,0 +1,76 @@
+name: Publish UI Components Package
+
+on:
+  workflow_run:
+    workflows: ['Linting and Tests']
+    types: [completed]
+    branches: [main]
+  workflow_dispatch:
+    inputs:
+      skip_github_check:
+        description: 'Skip GitHub commit author check'
+        required: false
+        default: false
+        type: boolean
+
+permissions:
+  contents: read
+  packages: write
+
+env:
+  COMMIT_SHA: ${{ github.event.workflow_run.head_sha || github.sha }}
+
+jobs:
+  publish:
+    if: github.event_name == 'workflow_dispatch' || (github.event.workflow_run.conclusion == 'success' && github.event.workflow_run.event == 'push')
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+
+      - name: Verify commit authenticity
+        run: |
+          COMMIT_DATA=$(gh api repos/${{ github.repository }}/commits/$COMMIT_SHA)
+          VERIFIED=$(echo "$COMMIT_DATA" | jq -r '.commit.verification.verified')
+          COMMITTER=$(echo "$COMMIT_DATA" | jq -r '.commit.committer.email')
+
+          if [[ "$VERIFIED" != "true" ]]; then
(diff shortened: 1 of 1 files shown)


.github/workflows/publish-ui-components.yml
```diff
@@ -0,0 +1,76 @@
...
+          COMMIT_DATA=$(gh api repos/${{ github.repository }}/commits/$COMMIT_SHA)
+          VERIFIED=$(echo "$COMMIT_DATA" | jq -r '.commit.verification.verified')
+          COMMITTER=$(echo "$COMMIT_DATA" | jq -r '.commit.committer.email')
+
+          if [[ "$VERIFIED" != "true" ]]; then
+            echo "❌ Unverified commit! Aborting."
+            exit 1
+          fi
+
+          if [[ "${{ github.event.inputs.skip_github_check }}" != "true" && "$COMMITTER" != *"noreply@github.com" ]]; then
```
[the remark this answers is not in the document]
avivkeller: This is the intended behavior. I'd like to **as specific as possible**, to prevent anyone except merge-queue from triggering this.

.github/workflows/publish-ui-components.yml
```diff
@@ -0,0 +1,76 @@
...
+
+      - name: Generate version
+        if: github.event_name == 'workflow_dispatch' || steps.check_changes.outputs.changed == 'true'
+        run: npm version --no-git-tag-version 0.0.0-$COMMIT_SHA --workspace=packages/ui-components
+
+      - name: Publish package
+        if: github.event_name == 'workflow_dispatch' || steps.check_changes.outputs.changed == 'true'
+        run: npm publish --workspace=packages/ui-components
+        env:
+          NODE_AUTH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
AugustinMauroy, line 76:
```suggestion
          NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
```

REF: https://docs.npmjs.com/generating-provenance-statements#example-github-actions-workflow

AugustinMauroy (commented):
I didn't get if you are trying to publish it npm registry or GitHub packages

.github/workflows/publish-ui-components.yml
avivkeller: We aren't publishing to the npm registry. We are publishing to GitHub Packages, which authenticates with the GITHUB_TOKEN

avivkeller:
> I didn't get if you are trying to publish it npm registry or GitHub packages

We are publishing to GitHub Packages.

.github/workflows/publish-ui-components.yml
AugustinMauroy: got it ! that wasn't written in pr description

avivkeller:
We don't control the `@node-core` scope on Github Packages, see https://github.com/nodejs/api-docs-tooling/issues/236#issuecomment-2752011612

avivkeller:
Per discussion in other thread

PR #7567, opened 2025-03-24
chore: update chromatic and storybook

avivkeller:
This PR updates Chromatic and Storybook, as it looks like our version of Chromatic is significantly out of date.

Before merging, I'd like to verify the baselines.


bjohansebas (approved):

ovflowd:
Is the PR ready? Im fine fast tracking it :)

avivkeller:
Yes, it's ready, but I don't think it solves the problem

avivkeller:
Actually, I'd like to run a few more tests

avivkeller:
@ovflowd @bjohansebas please re-review. **This solves all the Chromatic issues :-)**

packages/ui-components/.storybook/preview.tsx
```diff
@@ -4,7 +4,7 @@ import type { Preview, ReactRenderer } from '@storybook/react';
 
 import { STORYBOOK_MODES, STORYBOOK_SIZES } from './constants';
 
-import '../styles/index.css';
+import '../../../apps/site/styles/index.css';
```
avivkeller: IMO it's fine that this is from a different part of the monorepo, because this is only for testing. We can always change it in the future, but for now, I think it's important to get the tests fixed.

ovflowd: But does it need all the styles?

packages/ui-components/tailwind.config.ts
```diff
@@ -2,7 +2,7 @@ import type { Config } from 'tailwindcss';
 import plugin from 'tailwindcss/plugin';
 
 export default {
-  content: ['./**/*.tsx'],
+  content: ['./Common/**/*.tsx', './Containers/**/*.tsx', './Icons/**/*.tsx'],
```
ovflowd: Interesting. But why?

packages/ui-components/.storybook/preview.tsx
ovflowd: Feels weird for Package A to depend on Package B whose depends on Package A

ovflowd (approved):

packages/ui-components/.storybook/preview.tsx
bjohansebas: I don’t really like this change. Why would it be necessary to import that file from outside the package instead of within it? Is there an import missing?

packages/ui-components/tailwind.config.ts
avivkeller: Tailwind wasn't happy that the glob included node_modules, so I figured I'd might as well fix it

packages/ui-components/.storybook/preview.tsx
avivkeller: Yes I know it's weird, and it's definitely something I want to change in the near future, but right now I'd like to get the tests back up and running

bjohansebas: I see that the file **markdown.css** is not being imported [here](https://github.com/nodejs/nodejs.org/blob/main/packages/ui-components/styles/index.css). Maybe that could help.

avivkeller: See @ovflowd's review

bjohansebas:
@avivkeller 
The header is broken
image

avivkeller:
Okay. I'm going to slow my roll here, get all the styles imported correctly, and we should be able to merge this tomorrow.

Sorry sorry for the rush, the broken test stressed me out.

.github/workflows/lint-and-tests.yml
```diff
@@ -202,7 +202,7 @@ jobs:
             startsWith(github.event.pull_request.head.ref, 'dependabot/') == false &&
             github.event.pull_request.head.ref != 'chore/crowdin')
         # sha reference has no stable git tag reference or URL. see https://github.com/chromaui/chromatic-cli/issues/797
-        uses: chromaui/action@30b6228aa809059d46219e0f556752e8672a7e26
+        uses: chromaui/action@c93e0bc3a63aa176e14a75b61a31847cbfdd341c
```
bmuenzenmeyer, line 205: BTW, I double checked https://github.com/marketplace/actions/publish-to-chromatic is a verified publisher, meaning we can upgrade safely without having to allow-list individual shas - updated the actions settings accordingly

avivkeller:
Closing. I'll open a seperate PR to fix the regression, Storybook will be updated during a Dependabot update, and Chromatic isn't broken, but I'll upgrade it once we confirm the baselines

PR #7574, opened 2025-03-26
adds llm.txt file to docs

iyush05:
## Description

Adds a llm.txt file to the docs at /docs/llm route

## Validation

Screenshot 2025-03-26 at 18 50 01

## Related Issues

Fixes #57587

### Check List


avivkeller (changes requested):
This should be a generator in `api-docs-tooling`, or a generator in node core, not just a concatenation of the current documentation.

Also, it's worth noting there is an open issue in both `api-docs-tooling` and `nodejs/node` discussing this, and so the potential implementation can be subject to change during discussion.

araujogui:
I'm closing this since it should be a generator in `api-docs-tooling`. https://github.com/nodejs/api-docs-tooling/issues/226

PR #7626, opened 2025-04-06
fix: convert incorrect breadcrumb links to opaque spans

dario-piotrowicz:
## Description

It's pretty minor but I figured I could just open a PR to fix this small UI bug I noticed :slightly_smiling_face: 

The fact is that in pages with breadcrumbs at the bottom of the screen most of the time the first breadcrumbs link is not really a valid link and clicking it results in a hard reload of the app:

See:
issue

My changes here make breadcrumb items with falsy hrefs simple text elements instead of a links, for good measure they are also dimmed to make it more evident that they are not clickable

## Validation

I manually validated this check with `npm run dev` and also `npm run build && npm run start`

### Check List

### DIFF
--- a/packages/ui-components/Common/Breadcrumbs/index.stories.tsx
+++ b/packages/ui-components/Common/Breadcrumbs/index.stories.tsx
@@ -24,6 +24,25 @@ export const Default: Story = {
   },
 };
 
+export const Linkless: Story = {
+  args: {
+    links: [
+      {
+        label: 'Learn',
+        href: '',
+      },
+      {
+        label: 'Getting Started',
+        href: '',
+      },
+      {
+        label: 'Introduction to Node.js',
+        href: '',
+      },
+    ],
+  },
+};
+
 export const Truncate: Story = {
   args: {
     links: [
--- a/packages/ui-components/Common/Breadcrumbs/index.tsx
+++ b/packages/ui-components/Common/Breadcrumbs/index.tsx
@@ -51,9 +51,13 @@ const Breadcrumbs: FC<BreadcrumbsProps> = ({
             hideSeparator={isLastItem}
             position={position + +!hideHome}
           >
-            <BreadcrumbLink as={as} href={link.href} active={isLastItem}>
-              {link.label}
-            </BreadcrumbLink>
+            {link.href ? (
+              <BreadcrumbLink as={as} href={link.href} active={isLastItem}>
+                {link.label}
+              </BreadcrumbLink>
+            ) : (
+              <span className="opacity-70">{link.label}</span>
+            )}
           </BreadcrumbItem>
         );
       }),
--- a/packages/ui-components/Common/Breadcrumbs/index.tsx
+++ b/packages/ui-components/Common/Breadcrumbs/index.tsx
@@ -51,8 +51,12 @@ const Breadcrumbs: FC<BreadcrumbsProps> = ({
             hideSeparator={isLastItem}
             position={position + +!hideHome}
           >
-            {link.href ? (
-              <BreadcrumbLink as={as} href={link.href} active={isLastItem}>
+            {link.href || isLastItem ? (
+              <BreadcrumbLink
+                as={as}
+                href={link.href || undefined}
+                active={isLastItem}
+              >
                 {link.label}
               </BreadcrumbLink>
             ) : (


avivkeller:
Can you add a linkless story?

dario-piotrowicz:
> Can you add a linkless story?

Thanks for the suggestion :slightly_smiling_face: 

Sure, done :slightly_smiling_face::+1:

dario-piotrowicz:
mh... actually I might need to amend something, the last item should be styled as active even if it doesn't have an `href` right? :thinking: 

Screenshot at 2025-04-07 01-23-44

The `About Node.js` breadcrumb item should be styled as active, right? :thinking:

dario-piotrowicz:
> mh... actually I might need to amend something, the last item should be styled as active even if it doesn't have an `href` right? 🤔

addressed :+1:  https://github.com/nodejs/nodejs.org/pull/7626/commits/614899a0a2d98f07ab69d9a9e1d82b4a22decd87 (the solution is a bit ugly but I am not sure if much better can be done here :thinking:)

ovflowd (changes requested):
This is weird. All breadcrumb entries should be links. The fact that About Node.js is causing a hard refresh is a bug.

We shouldn't do this; we should investigate why the link wasn't a link to begin with.

dario-piotrowicz:
> why the link wasn't a link to begin with.

It's because of the site navigation structure, for example see:

Screenshot at 2025-04-07 10-50-00

Screenshot at 2025-04-07 10-51-12



there is actually no `getting started` page (see: https://nodejs.org/en/learn/getting-started/)

so if the breadcrumb for `getting started` were a link what page would it point to?

ovflowd:
Right, but I do have the feeling that something else is at play here :thinking: like all these breadcrumbs should link to actual pages. They should correlate to something within the tree of the path structure. So something could be borken on the breadcrumb generation.

dario-piotrowicz:
> Right, but I do have the feeling that something else is at play here 🤔 like all these breadcrumbs should link to actual pages. They should correlate to something within the tree of the path structure. So something could be borken on the breadcrumb generation.

mh... I feel like this is a UI/design issue more than a code related one, so if we again take https://nodejs.org/en/learn/getting-started/differences-between-nodejs-and-the-browser as an example and you look at the breadcrumbs at the bottom, what do you think they should look like? should `Getting Started` be there? (since this page is part of the getting starter section) or shouldn't it be there? (since there is no page for that specific section).

If `Getting Started` should be there, maybe a new `/getting-started` route/page should added to make the breadcrumbs' link can point to a real page?

I feel like we first need to answer these design questions before we can agree/understand if there's something actually wrong with the code

ovflowd:
> If Getting Started should be there, maybe a new /getting-started route/page should added to make the breadcrumbs' link can point to a real page?

We should make the first article of getting-started also be available under /getting-started, that could fix this issue (we can make internal redirects on redirects.json) wdyt?

ovflowd (approved):
LGTM!

ovflowd:
> > If Getting Started should be there, maybe a new /getting-started route/page should added to make the breadcrumbs' link can point to a real page?
> 
> We should make the first article of getting-started also be available under /getting-started, that could fix this issue (we can make internal redirects on redirects.json) wdyt?

We can open a follow-up issue for this if you feel like?

dario-piotrowicz:
> We should make the first article of getting-started also be available under /getting-started, that could fix this issue (we can make internal redirects on redirects.json) wdyt?

I did try something like that and it worked pretty ok however this was a bit awkward in the first article since the link there would point to the exact same page,  so let's take https://nodejs.org/en/learn/getting-started/introduction-to-nodejs there you'd have:
Screenshot at 2025-05-04 14-06-16

where the last two breadcrumb links would point to the exact same page, which is not the end of the world but if does feel pretty awkward to me :thinking:

dario-piotrowicz:
> We can open a follow-up issue for this if you feel like?

Sorry but I feel like the solution here and what you're suggesting are not really compatible (we either do make the breadcrumb a valid working link or make it not a link at all, I don't think that we can apply both).

If you prefer your solution I am totally happy to just close this PR and open a new one with that solution instead :slightly_smiling_face: (however please do keep in mind my [comment](https://github.com/nodejs/nodejs.org/pull/7626#issuecomment-2849210456) above)

ovflowd:
It's fine to keep as is (your solution)

dario-piotrowicz:
> It's fine to keep as is (your solution)

alright, thanks! :slightly_smiling_face:

avivkeller:
@dario-piotrowicz Once you rebase, this'll be ready to merge :-)

PR #7634, opened 2025-04-09
chore(deps): re-sync lockfile

avivkeller:
## Description

This is a chore re-sync of the lockfile. It's been a while since this file was re-generated, and it's important (IMO) to do it every once in a while.

All tests pass, storybook looks fine, and nothing crashes AFAICT on Chrome + Safari.

## Validation

Nothing should change, and no tests should fail

### Check List

[^1]: Except for `eslint-plugin-mdx`, as it should match the version of `eslint-mdx`


AugustinMauroy (approved):

draclyr:
good work❤️

bmuenzenmeyer (commented):
I'm -1 on the practice, as it's imprecise and very hard to review.

AugustinMauroy:
I didn't get what is the issue why "updating" the lock file ?

avivkeller:
Regenerating the lockfile ensures that the package versions we install are the best ones available. Over time, the file can slightly differ from its optimal state, mainly because Dependabot updates only a few packages at a time in isolation, not taking into account if a specific package has a better version available that suits our specified requirements.

bmuenzenmeyer (changes requested):
putting an explicit block on this til we have consensus

MattIPv4:
I am -1 to this practice -- I have never encountered it before as a recommended thing to do, and it seems incredibly risky / complex to review. If we're going to do this, why have a lockfile at all? I'd much rather see PRs to bump individual transitive dependencies that we know need updating for a specific reason, and beyond that just let them naturally update as direct dependencies require.

avivkeller:
I think enough of us have said -1 to this PR that it shouldn't land, and we should continue discussion on https://openjs-foundation.slack.com/archives/CVAMEJ4UV/p1744401326173889.

Sorry for the noise, everyone!

PR #7643, opened 2025-04-11
meta: redirect Corepack documentation page

aduh95:
## Description

To address https://github.com/nodejs/node/pull/57663#issuecomment-2788722443.

This page is being removed, however it's quite cheap for us not to break the existing links.

### Check List

### DIFF
--- a/apps/site/redirects.json
+++ b/apps/site/redirects.json
@@ -9,6 +9,14 @@
       "destination": "/api"
     },
     {
+      "source": "/api/corepack.html",
+      "destination": "https://github.com/nodejs/corepack"
+    },
+    {
+      "source": "/docs/latest/api/corepack.html",
+      "destination": "https://github.com/nodejs/corepack"
+    },
+    {
       "source": "/changelog.html",
       "destination": "https://github.com/nodejs/node/blob/HEAD/CHANGELOG.md"
     },
@@ -343,3 +351,4 @@
   ],
   "internal": []
 }
+


apps/site/redirects.json
```diff
@@ -8,6 +8,14 @@
...
       "destination": "/api"
     },
+    {
+      "source": "/api/corepack.html",
+      "destination": "https://github.com/nodejs/corepack"
+    },
+    {
+      "source": "/docs/latest/api/corepack.html",
+      "destination": "https://github.com/nodejs/corepack"
+    },
```
avivkeller, line 18: Since we're referring to the API docs, shouldn't this redirect to a removal notice instead of Corepack? We don't redirect to docs for other Node.js packages on `/api`, and I don’t think we should break that precedent, but I'm not partial either way

Maybe redirect to https://github.com/nodejs/TSC/pull/1697#issuecomment-2737093616, or to a new blog post?

apps/site/redirects.json
```diff
@@ -8,6 +8,14 @@
       "source": "/api.html",
       "destination": "/api"
     },
+    {
+      "source": "/api/corepack.html",
```
trivikr, line 12: The redirect, if implemented for latest, should be done after corepack is not available in that Node.js version
i.e. when v25.0.0 is released.

While corepack is included in the latest, it should highlight the decision in bold and provide a call to action, like it's suggested in https://github.com/nodejs/node/pull/57825

trivikr, line 12: The redirect, if implemented for latest, should be done after corepack is not available in that Node.js version
i.e. when v25.0.0 is released.

While corepack is included in the latest, it should highlight the decision in bold and provide a call to action, like it's suggested in https://github.com/nodejs/node/pull/57825

trivikr (dismissed):

flakey5 (changes requested):
The website doesn't serve `/api` or `/docs`, so these redirects wouldn't be functional. These would either need to be set up in Cloudflare or in the release worker

AugustinMauroy:
@flakey5 if you open an issue/pr on right repo we can close this one

trivikr:
> if you open an issue/pr on right repo we can close this one

I think we shouldn't introduce the redirect till Corepack is removed from Node.js latest, which will happen in 25.0.0 in October 2025.

I've called it out in https://github.com/nodejs/nodejs.org/pull/7643#discussion_r2040008198

apps/site/redirects.json
bjohansebas: I’d really prefer to be sent to the Corepack documentation rather than the issue where the vote happened or a potential blog post.

As a user, I want to get the information from the documentation

avivkeller: I'm not polarized either way, so whatever you all think is best

apps/site/redirects.json
aduh95: Whatever call to action you'd like to see, why does it need to be maintained in nodejs/node rather than nodejs/corepack?

trivikr: The call to action is for existing Node.js versions which ship corepack.
It's similar to [punycode](https://nodejs.org/api/punycode.html) which was deprecated since v7.0.0 but call to action still exists.

I don't suggest keeping corepack call to action forever. It needs to be minimally in Node.js versions which ship experimental corepack, i.e. v24. It'll be helpful if the call to action is in latest doc till <=v28. But dropping it in v27 is also okay.

Other options are mentioned in https://github.com/nodejs/node/pull/57663#issuecomment-2766906437

trivikr: Dismissed my review, and provided summary https://github.com/nodejs/nodejs.org/pull/7643#event-17240639556

trivikr:
> These would either need to be set up in Cloudflare or in the release worker

The next action is to setup redirect in Cloudflare or in the release worker, and close this PR.

avivkeller:
I'm closing this PR since, no matter what happens, this cannot merge in this repository, however, feel free to continue discussing the future in this PR, if needed.

avivkeller:
Also, for posterity,

@trivikr was referencing https://openjs-foundation.slack.com/archives/C06LTLUK9NZ/p1744489893809369

PR #7644, opened 2025-04-11
feat: introduce stylelint plugin

canerakdas:
## Description

While trying to identify indentation inconsistencies in the styling files, I ended up writing a basic stylelint plugin for it 😄 
I believe it's a valuable addition to maintaining code standards in the project and helping new contributors adapt more easily

In editor;
Screenshot 2025-04-11 at 20 11 23

When running `lint:css`

```bash
layouts/layouts.module.css
  269:3  ✖  Each line should contain only one Tailwind token; "some tailwind   classes"  ui-components/one-tailwind-token-per-line

✖ 1 problem (1 error, 0 warnings)
  1 error potentially fixable with the "--fix" option.
```

Also, it's auto fixable rule

----

## Validation

The `lint:css` and `lint:fix` scripts should work seamlessly across Windows, macOS, and Linux distributions, without affecting any correctly styled code

### Check List

### DIFF
--- a/apps/site/.stylelintrc.mjs
+++ b/apps/site/.stylelintrc.mjs
@@ -23,7 +23,11 @@ const ONLY_ALLOW_CAMEL_CASE_SELECTORS = [
 
 export default {
   extends: ['stylelint-config-standard'],
-  plugins: ['stylelint-order', 'stylelint-selector-bem-pattern'],
+  plugins: [
+    'stylelint-order',
+    'stylelint-selector-bem-pattern',
+    '@node-core/ui-components/stylelint.plugin.mjs',
+  ],
   rules: {
     // Enforces Element Class Names to be camelCase
     'selector-class-pattern': ONLY_ALLOW_CAMEL_CASE_SELECTORS,
@@ -45,5 +49,6 @@ export default {
     'import-notation': 'string',
     // Allow the `@apply` at rule as its part of Tailwind
     'at-rule-no-deprecated': [true, { ignoreAtRules: CUSTOM_AT_RULES }],
+    'ui-components/one-tailwind-token-per-line': true,
   },
 };
--- a/apps/site/.stylelintrc.mjs
+++ b/apps/site/.stylelintrc.mjs
@@ -26,7 +26,7 @@ export default {
   plugins: [
     'stylelint-order',
     'stylelint-selector-bem-pattern',
-    '@node-core/ui-components/stylelint/one-mixin-per-line.mjs',
+    '@node-core/ui-components/stylelint/one-utility-class-per-line.mjs',
   ],
   rules: {
     // Enforces Element Class Names to be camelCase
@@ -49,6 +49,6 @@ export default {
     'import-notation': 'string',
     // Allow the `@apply` at rule as its part of Tailwind
     'at-rule-no-deprecated': [true, { ignoreAtRules: CUSTOM_AT_RULES }],
-    'nodejs/one-mixin-per-line': true,
+    'nodejs/one-utility-class-per-line': true,
   },
 };
--- a/apps/site/.stylelintrc.mjs
+++ b/apps/site/.stylelintrc.mjs
@@ -26,7 +26,7 @@ export default {
   plugins: [
     'stylelint-order',
     'stylelint-selector-bem-pattern',
-    '@node-core/ui-components/stylelint.plugin.mjs',
+    '@node-core/ui-components/stylelint/one-mixin-per-line.mjs',
   ],
   rules: {
     // Enforces Element Class Names to be camelCase
@@ -49,6 +49,6 @@ export default {
     'import-notation': 'string',
     // Allow the `@apply` at rule as its part of Tailwind
     'at-rule-no-deprecated': [true, { ignoreAtRules: CUSTOM_AT_RULES }],
-    'ui-components/one-tailwind-token-per-line': true,
+    'nodejs/one-mixin-per-line': true,
   },
 };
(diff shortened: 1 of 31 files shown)


packages/ui-components/stylelint.plugin.mjs
```diff
@@ -0,0 +1,116 @@
+import stylelint from 'stylelint';
+
+const {
+  createPlugin,
+  utils: { report, ruleMessages },
+} = stylelint;
+
+const name = 'ui-components/one-tailwind-token-per-line';
```
avivkeller:
```suggestion
const name = 'nodejs/one-mixin-per-line';
``` 

I think this name better represents it, right?

apps/site/.stylelintrc.mjs
```diff
@@ -23,7 +23,11 @@ const ONLY_ALLOW_CAMEL_CASE_SELECTORS = [
 
 export default {
   extends: ['stylelint-config-standard'],
-  plugins: ['stylelint-order', 'stylelint-selector-bem-pattern'],
+  plugins: [
+    'stylelint-order',
+    'stylelint-selector-bem-pattern',
+    '@node-core/ui-components/stylelint.plugin.mjs',
```
avivkeller: nit: should we name the file something else to better represent the rule?

packages/ui-components/stylelint.plugin.mjs
```diff
@@ -0,0 +1,116 @@
...
+
+/**
+ * Cleans the input string by removing unnecessary whitespace and newlines.
+ *
+ * @param {string} params - The class names string to be cleaned.
+ * @returns {string} The cleaned class names string.
+ */
+const cleanClassNames = params => {
+  return params
+    .replace(/\n+/g, '') // Remove new lines
```
avivkeller:
```suggestion
    .replace(/[\n\r]+/g, '') // Remove new lines
```

avivkeller:
Can you add a test?

Maybe make a stylelint directory with the plugin and the test file?

Also, should this maybe go at repo root? It's a plugin for the entire repository, not just UI components? (But I'd need another opinion on that)

packages/ui-components/stylelint.plugin.mjs
ovflowd, line 1: This so cool!!!!

packages/ui-components/stylelint.plugin.mjs
ovflowd: cc @canerakdas

ovflowd (approved):
LGTM but feel free to apply code suggestions, first.

packages/ui-components/stylelint.plugin.mjs
AugustinMauroy: yeah I love that great idea

packages/ui-components/stylelint.plugin.mjs
AugustinMauroy: +1 for this name

bjohansebas (commented):
It's not a blocker, and it’s probably for another PR, but we should have a configuration package, both to share this plugin and to share configurations that are being repeated, like the Stylint configuration

bjohansebas (approved):

packages/ui-components/stylelint/__tests__/index.test.mjs
avivkeller, line 1: Can you please add a test for Nested CSS?

By the way, this is amazing!

avivkeller (approved):

packages/ui-components/stylelint/one-mixin-per-line.mjs
```diff
@@ -0,0 +1,64 @@
...
+const {
+  createPlugin,
+  utils: { report, ruleMessages },
+} = stylelint;
+
+const name = 'nodejs/one-mixin-per-line';
+
+const messages = ruleMessages(name, {
+  className: tokens =>
+    `Each line should contain only one Tailwind token; "${tokens}"`,
```
avivkeller:
```suggestion
    `Each line should contain only one mixin; Saw "${tokens}"`,
```

packages/ui-components/stylelint/one-mixin-per-line.mjs
```diff
@@ -0,0 +1,64 @@
...
+  spacing: tokens =>
+    `Each class used with @apply should be indented by 2 spaces on a new line; ${tokens}`,
```
avivkeller:
```suggestion
    `Each mixin should be indented two spaces from the \`@apply\` call; Saw "${tokens}"`,
```

packages/ui-components/stylelint/__tests__/index.test.mjs
canerakdas: These functions do not use nested utility classes in any way. Stylelint returns all `@apply` used places individually for us. See; https://github.com/nodejs/nodejs.org/pull/7644/files#diff-74200f8ed70511b104692fcdc4fe43f9e5a9e388bc6663d5d5ac7ad6d84d71beR26

packages/ui-components/stylelint/one-mixin-per-line.mjs
canerakdas: I looked into it a bit, but I'm not sure if our usage — or any usage in general — is actually considered a `mixin`.
In Tailwind's own documentation, the `@apply` directive is described as being used with utility classes.
See: https://tailwindcss.com/docs/functions-and-directives#apply
I'll update the other parts accordingly based on this document.

AugustinMauroy (approved):
LGMT !

PR #7639, opened 2025-04-10
feat(learn): add article on tuning memory

avivkeller:
## Description

This PR adds an article on understanding and tuning Memory in Node.js. It discusses the V8 memory types (Stack and Heap), along with Old vs New space.

CC @nodejs/diagnostics 

## Validation

After building the package, the article should exist at `/learn/diagnostics/memory/understanding-and-tuning-memory`

## Related Issues

Fixes #7637 

### Check List

- [x] I have read the [Contributing Guidelines](https://github.com/nodejs/nodejs.org/blob/main/CONTRIBUTING.md) and made commit messages that follow the guideline.
- [x] I have run `npm run format` to ensure the code follows the style guide.
- [x] I have run `npm run test` to check if all tests are passing.
- [x] I have run `npx turbo build` to check if the website builds without errors.
- [x] I've covered new added functionality with unit tests if necessary.

PR #7645, opened 2025-04-12
feat(learn): add article on promises

### LINKED ISSUE #7200: Add a Learn section about Promises/queueMicrotask and promises
This issue is about creating a new learn page (async work) regarding Promises on Node.js and queueMicroTask

avivkeller:
## Description

This PR adds an explanation of JavaScript / Node.js Promises to the website.

I took some feedback from #7639, and tried to incorporate specific examples and real-world scenarios.

## Related Issues

Fixes: #7200

### Check List

### DIFF
--- a/apps/site/pages/en/learn/asynchronous-work/discover-promises-in-nodejs.md
+++ b/apps/site/pages/en/learn/asynchronous-work/discover-promises-in-nodejs.md
@@ -1,10 +1,10 @@
 ---
-title: Promises in Node.js
+title: Discover Promises in Node.js
 layout: learn
 authors: avivkeller
 ---
 
-# Promises in Node.js
+# Discover Promises in Node.js
 
 A **Promise** is a special object in JavaScript that represents the eventual completion (or failure) of an asynchronous operation and its resulting value. Think of a Promise as a placeholder for a value that is not yet available but will be in the future.
 
@@ -96,7 +96,7 @@ promise1
 One of the best ways to work with Promises in modern JavaScript is using **async/await**. This allows you to write asynchronous code that looks synchronous, making it much easier to read and maintain.
 
 - `async` is used to define a function that returns a Promise.
-- `await` is used inside an `async` function to pause execution until the Promise is resolved or rejected.
+- `await` is used inside an `async` function to pause execution until a Promise settles. However, in [ECMAScript Modules](https://nodejs.org/api/esm.html), you can use [`await` at the top level](https://nodejs.org/api/esm.html#top-level-await) without needing an `async` function.
 
 ```js
 async function performTasks() {
--- a/apps/site/pages/en/learn/asynchronous-work/discover-promises-in-nodejs.md
+++ b/apps/site/pages/en/learn/asynchronous-work/discover-promises-in-nodejs.md
@@ -181,7 +181,7 @@ In this example, `fs.promises.readFile()` returns a Promise, which we handle usi
 
 JavaScript's `Promise` global provides several powerful methods that help manage multiple asynchronous tasks more effectively:
 
-### **`Promise.all()`**:
+### `Promise.all()`
 
 This method accepts an array of Promises and returns a new Promise that resolves once all the Promises are fulfilled. If any Promise is rejected, `Promise.all()` will immediately reject. However, even if rejection occurs, the Promises continue to execute. When handling a large number of Promises, especially in batch processing, using this function can strain the system's memory.
 
@@ -200,7 +200,7 @@ Promise.all([fetchData1, fetchData2])
   });
 ```
 
-### **`Promise.allSettled()`**:
+### `Promise.allSettled()`
 
 This method waits for all promises to either resolve or reject and returns an array of objects that describe the outcome of each Promise.
 
@@ -216,7 +216,7 @@ Promise.allSettled([promise1, promise2]).then(results => {
 
 Unlike `Promise.all()`, `Promise.allSettled()` does not short-circuit on failure. It waits for all promises to settle, even if some reject. This provides better error handling for batch operations, where you may want to know the status of all tasks, regardless of failure.
 
-### **`Promise.race()`**:
+### `Promise.race()`
 
 This method resolves or rejects as soon as the first Promise settles, whether it resolves or rejects. Regardless of which promise settles first, all promises are fully executed.
 
@@ -231,7 +231,7 @@ Promise.race([task1, task2]).then(result => {
 });
 ```
 
-### **`Promise.any()`**:
+### `Promise.any()`
 
 This method resolves as soon as one of the Promises resolves. If all promises are rejected, it will reject with an `AggregateError`.
 
@@ -251,7 +251,7 @@ Promise.any([api1, api2, api3])
   });
 ```
 
-### **`Promise.reject()` and `Promise.resolve()`**:
+### `Promise.reject()` and `Promise.resolve()`
 
 These methods create a rejected or resolved Promise directly.
 
@@ -261,7 +261,7 @@ Promise.resolve('Resolved immediately').then(result => {
 });
 ```
 
-### **`Promise.try()`**:
+### `Promise.try()`
 
 `Promise.try()` is a method that executes a given function, whether it's synchronous or asynchronous, and wraps the result in a promise. If the function throws an error or returns a rejected promise, `Promise.try()` will return a rejected promise. If the function completes successfully, the returned promise will be fulfilled with its value.
 
@@ -286,7 +286,7 @@ Promise.try(mightThrow)
 
 In this example, `Promise.try()` ensures that if `mightThrow()` throws an error, it will be caught in the `.catch()` block, making it easier to handle both sync and async errors in one place.
 
-### **`Promise.withResolvers()`**:
+### `Promise.withResolvers()`
 
 This method creates a new promise along with its associated resolve and reject functions, and returns them in a convenient object. This is used, for example, when you need to create a promise but resolve or reject it later from outside the executor function.
 
@@ -339,7 +339,7 @@ performTask();
 
 In addition to Promises, Node.js provides several other mechanisms for scheduling tasks in the event loop.
 
-### **`queueMicrotask()`**
+### `queueMicrotask()`
 
 `queueMicrotask()` is used to schedule a microtask, which is a lightweight task that runs after the currently executing script but before any other I/O events or timers. Microtasks include tasks like Promise resolutions and other asynchronous operations that are prioritized over regular tasks.
 
@@ -353,7 +353,7 @@ console.log('Synchronous task is executed');
 
 In the above example, "Microtask is executed" will be logged after "Synchronous task is executed," but before any I/O operations like timers.
 
-### **`process.nextTick()`**
+### `process.nextTick()`
 
 `process.nextTick()` is used to schedule a callback to be executed immediately after the current operation completes. This is useful for situations where you want to ensure that a callback is executed as soon as possible, but still after the current execution context.
 
@@ -365,7 +365,7 @@ process.nextTick(() => {
 console.log('Synchronous task executed');
 ```
 
-### **`setImmediate()`**:
+### `setImmediate()`
 
 `setImmediate()` is used to execute a callback after the current event loop cycle finishes and all I/O events have been processed. This means that `setImmediate()` callbacks run after any I/O callbacks, but before timers.
 
@@ -377,7 +377,7 @@ setImmediate(() => {
 console.log('Synchronous task executed');
 ```
 
-### **When to Use Each**:
+### When to Use Each
 
 - Use `queueMicrotask()` for tasks that need to run immediately after the current script and before any I/O or timer callbacks, typically for Promise resolutions.
 - Use `process.nextTick()` for tasks that should execute before any I/O events, often useful for deferring operations or handling errors synchronously.
--- a/apps/site/pages/en/learn/asynchronous-work/discover-promises-in-nodejs.md
+++ b/apps/site/pages/en/learn/asynchronous-work/discover-promises-in-nodejs.md
@@ -261,7 +261,7 @@ Promise.resolve('Resolved immediately').then(result => {
 });
 ```
 
-### **`Promise.try()`**
+### **`Promise.try()`**:
 
 `Promise.try()` is a method that executes a given function, whether it's synchronous or asynchronous, and wraps the result in a promise. If the function throws an error or returns a rejected promise, `Promise.try()` will return a rejected promise. If the function completes successfully, the returned promise will be fulfilled with its value.
 
@@ -285,6 +285,24 @@ Promise.try(mightThrow)
 ```
 
 In this example, `Promise.try()` ensures that if `mightThrow()` throws an error, it will be caught in the `.catch()` block, making it easier to handle both sync and async errors in one place.
+
+### **`Promise.withResolvers()`**:
+
+This method creates a new promise along with its associated resolve and reject functions, and returns them in a convenient object. This is used, for example, when you need to create a promise but resolve or reject it later from outside the executor function.
+
+```js
+const { promise, resolve, reject } = Promise.withResolvers();
+
+setTimeout(() => {
+  resolve('Resolved successfully!');
+}, 1000);
+
+promise.then(value => {
+  console.log('Success:', value);
+});
+```
+
+In this example, `Promise.withResolvers()` gives you full control over when and how the promise is resolved or rejected, without needing to define the executor function inline. This pattern is commonly used in event-driven programming, timeouts, or when integrating with non-promise-based APIs.
 
 ## Error Handling with Promises
 
@@ -375,7 +393,7 @@ In short, the execution order is as follows:
    If a microtask queues another microtask, the new one is also executed before moving on.
 4. **Timers** (e.g., `setTimeout()`, `setInterval()`)
 5. **I/O callbacks** (e.g., network and file system operations)
-   Some operations, like `close` events, may use `process.nextTick()` internally and are executed in the next tick.
+   Some operations, like `close` and `error` events, may use `process.nextTick()` internally and are executed in the next tick.
 6. **`setImmediate()`** callbacks
 
 `process.nextTick()` can execute at any stage in the execution order, as it is not bound by the event loop.
--- a/apps/site/pages/en/learn/asynchronous-work/discover-promises-in-nodejs.md
+++ b/apps/site/pages/en/learn/asynchronous-work/discover-promises-in-nodejs.md
@@ -6,7 +6,7 @@ authors: avivkeller
 
 # Discover Promises in Node.js
 
-A **Promise** is a special object in JavaScript that represents the eventual completion (or failure) of an asynchronous operation and its resulting value. Think of a Promise as a placeholder for a value that is not yet available but will be in the future.
+A **Promise** is a special object in JavaScript that represents the eventual completion (or failure) of an asynchronous operation and its resulting value. Essentially, a Promise is a placeholder for a value that is not yet available but will be in the future.
 
 Think of a Promise like ordering a pizza: you don't get it right away, but the delivery person promises to bring it to you later. You don't know _exactly_ when, but you know the outcome will either be "pizza delivered" or "something went wrong."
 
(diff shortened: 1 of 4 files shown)


avivkeller:
See https://nodejs-org-git-feat-learnpromises-openjs.vercel.app/en/learn/asynchronous-work/discover-promises-in-nodejs

apps/site/pages/en/learn/asynchronous-work/promises-in-nodejs.md
```diff
@@ -0,0 +1,264 @@
...
+    console.error(error); // If any Promise is rejected, catch the error
+  });
+```
+
+In this example:
+
+- The first Promise resolves after 1 second, and its result is logged.
+- The second Promise resolves after another second, and its result is logged after the first one finishes.
+
+## Using Async/Await with Promises
```
mcollina: This whole paragraph is a bit too simplistic. async/await is a bit more complex than that.

You can refer to the talk/workshop @jasnell did when at NearForm: https://www.youtube.com/watch?v=XV-u_Ow47s0. I would recommend adding it as a reference, too.

apps/site/pages/en/learn/asynchronous-work/promises-in-nodejs.md
```diff
@@ -0,0 +1,264 @@
...
+
+This method takes an array of Promises and resolves them all. It only resolves when all Promises are fulfilled. If any of the Promises is rejected, `Promise.all()` will reject immediately.
+
+```js
+const fetchData1 = new Promise(resolve =>
+  setTimeout(() => resolve('Data from API 1'), 1000)
+);
+const fetchData2 = new Promise(resolve =>
+  setTimeout(() => resolve('Data from API 2'), 2000)
+);
```
mcollina: Please avoid recommending this pattern. use the equivalent from `node:timers/promises`

apps/site/pages/en/learn/asynchronous-work/promises-in-nodejs.md
```diff
@@ -0,0 +1,264 @@
...
+
+```js
+const promise1 = Promise.resolve('Success');
+const promise2 = Promise.reject('Failed');
+
+Promise.allSettled([promise1, promise2]).then(results => {
+  console.log(results);
+  // [ { status: 'fulfilled', value: 'Success' }, { status: 'rejected', reason: 'Failed' } ]
+});
+```
```
mcollina: What's the main difference in error handling between this and `Promise.all()`.

apps/site/pages/en/learn/asynchronous-work/promises-in-nodejs.md
```diff
@@ -0,0 +1,264 @@
...
+readFile();
+```
+
+In this example, `fs.promises.readFile()` returns a Promise, which we handle using `async/await` syntax to read the contents of a file asynchronously.
+
+## Advanced Promise Methods
+
+JavaScript's `Promise` global provides several powerful methods that help manage multiple asynchronous tasks more effectively:
+
+### **`Promise.all()`**:
```
mcollina: This should really mention the footgun that `Promise.all()` is in respect to memory handling.

apps/site/pages/en/learn/asynchronous-work/promises-in-nodejs.md
```diff
@@ -0,0 +1,264 @@
...
+  try {
+    const result = await myPromise;
+    console.log(result);
+  } catch (error) {
+    console.error(error); // Handles any errors
+  }
+}
+
+performTask();
+```
```
mcollina: How does finally fits?

mcollina (dismissed):
This should also cover the relationship between promises, queueMicrotask, process.nextTick and setImmediate.

avivkeller:
@mcollina I've attempted to resolve your reviews in 81801200

apps/site/pages/en/learn/asynchronous-work/promises-in-nodejs.md
avivkeller: @jasnell I've added your talk as a reference, just confirming that [we have permission](https://github.com/nodejs/nodejs.org/issues/7200#issuecomment-2564402143) to do so?

apps/site/pages/en/learn/asynchronous-work/promises-in-nodejs.md
```diff
@@ -0,0 +1,317 @@
+---
+title: Promises in Node.js
```
AugustinMauroy:
```suggestion
title: Discover Promises in Node.js
```

We can start with a slightly catchier title

AugustinMauroy (commented):
Nice but nothing about await on top level

apps/site/pages/en/learn/asynchronous-work/discover-promises-in-nodejs.md
```diff
@@ -0,0 +1,317 @@
...
+title: Discover Promises in Node.js
+layout: learn
+authors: avivkeller
+---
+
+# Discover Promises in Node.js
+
+A **Promise** is a special object in JavaScript that represents the eventual completion (or failure) of an asynchronous operation and its resulting value. Think of a Promise as a placeholder for a value that is not yet available but will be in the future.
+
+Think of a Promise like ordering a pizza: you don't get it right away, but the delivery person promises to bring it to you later. You don't know _exactly_ when, but you know the outcome will either be "pizza delivered" or "something went wrong."
```
jasnell, line 11: Nit... two sentences one after the other starting with "Think of a Promise..." is bit awkward.

apps/site/pages/en/learn/asynchronous-work/discover-promises-in-nodejs.md
```diff
@@ -0,0 +1,317 @@
...
+
+## Promise States
+
+A Promise can be in one of three states:
+
+- **Pending**: The initial state, where the asynchronous operation is still running.
+- **Fulfilled**: The operation completed successfully, and the Promise is now resolved with a value.
+- **Rejected**: The operation failed, and the Promise is settled with a reason (usually an error).
+
+For example, imagine you're waiting for an email from a friend. You're in the pending state. If your friend replies, you've entered the fulfilled state. If the email bounces back, you are in the rejected state.
```
jasnell: You use a pizza delivery example above then switch to an email example here. Likely better to just continue the pizza delivery theme.... When you order the pizza you're in the `Pending` state. When the pizza arrives you're in the `Fulfilled` state. If the Pizza delivery scooter breaks down on the way to your house and the pizza tragically falls to the ground, you're in the `Rejected` state... etc.

apps/site/pages/en/learn/asynchronous-work/discover-promises-in-nodejs.md
```diff
@@ -0,0 +1,317 @@
...
+
+## Basic Syntax of a Promise
+
+A Promise is created using the `new Promise()` constructor. The constructor takes a function with two parameters: `resolve` and `reject`. These functions are used to transition the Promise from the **pending** state to either **fulfilled** or **rejected**.
```
jasnell: This is one way of creating a promise but not the only.

apps/site/pages/en/learn/asynchronous-work/discover-promises-in-nodejs.md
```diff
@@ -0,0 +1,317 @@
...
+```js
+const myPromise = new Promise((resolve, reject) => {
+  let success = true;
+
+  if (success) {
+    resolve('Operation was successful!');
+  } else {
+    reject('Something went wrong.');
+  }
+});
```
jasnell, line 41: It's worth highlighting that the function passed to the `Promise` constructor is called immediately as part of constructing the `Promise`, and that any errors it may throw synchronously are caught and used to reject the promise.

It's also worth mentioning that this function should not return any value.

apps/site/pages/en/learn/asynchronous-work/discover-promises-in-nodejs.md
```diff
@@ -0,0 +1,317 @@
...
+
+myPromise
+  .then(result => {
+    console.log(result); // This will run if the Promise is fulfilled
+  })
+  .catch(error => {
+    console.error(error); // This will run if the Promise is rejected
+  });
+```
```
jasnell, line 78: Should mention `finally(...)` in here as well.

apps/site/pages/en/learn/asynchronous-work/discover-promises-in-nodejs.md
```diff
@@ -0,0 +1,317 @@
...
+
+## Chaining Promises
+
+One of the great features of Promises is that they allow you to chain multiple asynchronous operations together. When you chain Promises, each `.then()` block waits for the previous one to complete before it runs.
+
+```js
+const { setTimeout: delay } = require('node:timers/promises');
+
+const promise1 = delay(1000).then(() => 'First task completed');
+const promise2 = delay(1000).then(() => 'Second task completed');
```
jasnell: Hmm... these aren't chained. The promises are waiting concurrently with each other. The *continuations* are chained but that's not what I imagine was intended here. Chained async operations run serially to one another, not concurrently.

apps/site/pages/en/learn/asynchronous-work/discover-promises-in-nodejs.md
```diff
@@ -0,0 +1,317 @@
...
+    console.error(error); // If any Promise is rejected, catch the error
+  });
+```
+
+## Using Async/Await with Promises
+
+One of the best ways to work with Promises in modern JavaScript is using **async/await**. This allows you to write asynchronous code that looks synchronous, making it much easier to read and maintain.
+
+- `async` is used to define a function that returns a Promise.
+- `await` is used inside an `async` function to pause execution until a Promise settles. However, in [ECMAScript Modules](https://nodejs.org/api/esm.html), you can use [`await` at the top level](https://nodejs.org/api/esm.html#top-level-await) without needing an `async` function.
```
jasnell: The second sentence starting with `However, ...` is out of place here. Best to discuss top-level await in a separate section.

apps/site/pages/en/learn/asynchronous-work/discover-promises-in-nodejs.md
```diff
@@ -0,0 +1,317 @@
...
+    console.log(result2); // 'Second task completed'
+  } catch (error) {
+    console.error(error); // Catches any rejection or error
+  }
+}
+
+performTasks();
+```
+
+In the `performTasks` function, the `await` keyword pauses execution until each Promise is settled (resolved or rejected). This leads to a more linear and readable flow of asynchronous code.
```
jasnell: I'm not sure "pauses execution" is the right term to use here. async/await does not actually pause execution as much as it decomposes the function into a series of continuations. For instance, the example above logically decomposes into...

```js
promise1.then(function (result1) {
  console.log(result1);
  return promise2;
}).then(function (result2) {
  console.log(result2);
}).catch(function (error) {
  console.log(error);
});
```

When viewed this way you get a better sense of how this is operating.

apps/site/pages/en/learn/asynchronous-work/discover-promises-in-nodejs.md
```diff
@@ -0,0 +1,317 @@
...
+
+Async/await can be much more intricate than the simple examples provided. James Snell, a member of the Node.js Technical Steering Committee, has an [in-depth presentation](https://www.youtube.com/watch?v=XV-u_Ow47s0) that explores the complexities of Promises and async/await.
```
jasnell, line 153: ❤️

apps/site/pages/en/learn/asynchronous-work/discover-promises-in-nodejs.md
```diff
@@ -0,0 +1,317 @@
...
+
+In this example, `fs.promises.readFile()` returns a Promise, which we handle using `async/await` syntax to read the contents of a file asynchronously.
+
+## Advanced Promise Methods
+
+JavaScript's `Promise` global provides several powerful methods that help manage multiple asynchronous tasks more effectively:
+
+### **`Promise.all()`**:
+
+This method takes an array of Promises and resolves them all. It only resolves when all Promises are fulfilled. If any of the Promises is rejected, `Promise.all()` will reject immediately.
```
jasnell: Perhaps a better explanation: This method takes an array of Promises and returns another Promise that is fulfilled when they all resolve, or is rejected as soon as one is rejected.

apps/site/pages/en/learn/asynchronous-work/discover-promises-in-nodejs.md
```diff
@@ -0,0 +1,317 @@
...
+
+`Promise.all()` resolves when all Promises are completed. If you have a large number of Promises, especially in situations like batch processing, it could overwhelm the system's memory.
```
jasnell: The first sentence here is a bit repetitive of the previous paragraph.

apps/site/pages/en/learn/asynchronous-work/discover-promises-in-nodejs.md
```diff
@@ -0,0 +1,317 @@
...
+```js
+const { setTimeout: delay } = require('node:timers/promises');
+
+const task1 = delay(2000).then(() => 'Task 1 done');
+const task2 = delay(1000).then(() => 'Task 2 done');
+
+Promise.race([task1, task2]).then(result => {
+  console.log(result); // 'Task 2 done' (since task2 finishes first)
+});
+```
```
jasnell, line 232: Because so many people get this wrong, it's worth pointing out that while the `Promise.race(...)` is resolved or rejected when the first promise resolves/rejects, the other promises are still going... that is, the other tasks are not canceled when one promise wins the race. I've lost count of the number of bugs I've run into in the wild because people didn't realize this.

apps/site/pages/en/learn/asynchronous-work/discover-promises-in-nodejs.md
```diff
@@ -0,0 +1,317 @@
...
+
+### **`Promise.any()`**:
+
+This method resolves as soon as one of the Promises resolves. If all promises are rejected, it will reject with an `AggregateError`.
+
+```js
+const { setTimeout: delay } = require('node:timers/promises');
+
+const api1 = delay(2000).then(() => 'API 1 failed');
+const api2 = delay(1000).then(() => 'API 2 success');
```
jasnell, line 242: This example is probably better with a third promise in the mix.

apps/site/pages/en/learn/asynchronous-work/discover-promises-in-nodejs.md
```diff
@@ -0,0 +1,317 @@
...
+
+### **`Promise.reject()` and `Promise.resolve()`**:
+
+These methods create a rejected or resolved Promise directly.
+
+```js
+Promise.resolve('Resolved immediately').then(result => {
+  console.log(result); // 'Resolved immediately'
+});
+```
```
jasnell, line 262: `Promise.try(...)` was also recently added and is worth mentioning.

apps/site/pages/en/learn/asynchronous-work/discover-promises-in-nodejs.md
```diff
@@ -0,0 +1,317 @@
...
+  console.log('Immediate callback');
+});
+
+console.log('Synchronous task executed');
+```
+
+### **When to Use Each**:
+
+- Use `queueMicrotask()` for tasks that need to run immediately after the current script and before any I/O or timer callbacks, typically for Promise resolutions.
+- Use `process.nextTick()` for tasks that should execute before any I/O events, often useful for deferring operations or handling errors synchronously.
```
jasnell, line 383: It's likely worth pointing out that error handling between `process.nextTick(...)` and `queueMicrotask(...)` is quite different.

apps/site/pages/en/learn/asynchronous-work/discover-promises-in-nodejs.md
```diff
@@ -0,0 +1,317 @@
...
+- Use `setImmediate()` for tasks that should run after I/O events but before timers.
+
+In short, the execution order is as follows:
```
jasnell: This is not always true. When loading an ESM that uses `process.nextTick(...)` and `queueMicrotask(...)` at the top-level, the queued microtasks will run *first* ... actually this is true in any case where these are used in a promise continuation.

image

the reason is that microtasks scheduled while draining the microtask queue will also be drained at the same time.

apps/site/pages/en/learn/asynchronous-work/discover-promises-in-nodejs.md
```diff
@@ -0,0 +1,317 @@
...
+
+1. **Synchronous code** (e.g., regular function calls and script code)
+2. **`process.nextTick()`** callbacks
+3. **Microtasks** (e.g., Promises, `queueMicrotask()`)
+4. **Timers** (e.g., `setTimeout()`, `setInterval()`)
+5. **I/O callbacks** (e.g., network and file system operations)
+6. **`setImmediate()`** callbacks
+7. **Close callbacks** (e.g., `socket.on('close')`)
```
jasnell: It's worth noting that Node.js will typically use `process.nextTick` to schedule events like `close` and `error`, etc. This might be a bit misleading as is. The reason a `close` event often occurs in this order is because an `I/O` callback schedules a `process.nextTick(...)` to emit the event.

apps/site/pages/en/learn/asynchronous-work/discover-promises-in-nodejs.md
```diff
@@ -0,0 +1,317 @@
...
+
+```js
+setImmediate(() => {
+  console.log('Immediate callback');
+});
+
+console.log('Synchronous task executed');
+```
+
+### **When to Use Each**:
```
jasnell: It is probably worth stressing that the specific ordering of these is runtime implementation specific... that is, everything about the timing here is very specific to Node.js. While other runtimes might get close, none of the Node.js compatible runtimes adhere to the same ordering on these. For instance, Cloudflare Workers implements `process.nextTick(...)` simply by deferring to `queueMicrotask(...)` (that is, `process.nextTick` is really just an alias for `queueMicrotask`. There are actually quite a few timing differences between the runtimes.

apps/site/pages/en/learn/asynchronous-work/discover-promises-in-nodejs.md
```diff
@@ -0,0 +1,317 @@
...
+process.nextTick(() => {
+  console.log('Next tick callback');
+});
+
+console.log('Synchronous task executed');
+```
+
+### **`setImmediate()`**:
+
+`setImmediate()` is used to execute a callback after the current event loop cycle finishes and all I/O events have been processed. This means that `setImmediate()` callbacks run after any I/O callbacks, but before timers and the next tick queue.
```
jasnell: The last part of this sentence `... and the next tick queue` is not correct. Draining the next tick queue is independent of the event loop, and the next tick queue can be drained any number of times at each phase of the event loop where user code ends up running.

apps/site/pages/en/learn/asynchronous-work/discover-promises-in-nodejs.md
```diff
@@ -0,0 +1,317 @@
...
+});
+
+console.log('Synchronous task is executed');
+```
+
+In the above example, "Microtask is executed" will be logged after "Synchronous task is executed," but before any I/O operations like timers.
+
+### **`process.nextTick()`**
+
+`process.nextTick()` is used to schedule a callback to be executed after the current operation completes, but before any other event loop phases, such as I/O events, timers, or Promises. This is useful for situations where you want to ensure that a callback is executed as soon as possible, but still after the current execution context.
```
jasnell: This is inaccurate. The draining of the next tick queue is independent of the event loop.

apps/site/pages/en/learn/asynchronous-work/discover-promises-in-nodejs.md
jasnell: Also, it is worth mentioning `Promise.withResolvers()` in here also.

apps/site/pages/en/learn/asynchronous-work/promises-in-nodejs.md
jasnell: No need to ask. Feel free to use any of my talks/write ups.

apps/site/pages/en/learn/asynchronous-work/discover-promises-in-nodejs.md
avivkeller: How so?

apps/site/pages/en/learn/asynchronous-work/discover-promises-in-nodejs.md
avivkeller: So, when looking at the execution order, `process.nextTick()` can execute at any time?

jasnell: If you consider this flow (image borrowed from one of platformatic's blog posts)... 
image

The nextTick and queueMicrotask queues are drained when the control flow returns from JavaScript back to the C++ block there. So basically, any time any callback is triggered by the event loop that causes JavaScript to be executed. That can happen many times at any stage of the event loop turn.

apps/site/pages/en/learn/asynchronous-work/discover-promises-in-nodejs.md
jasnell: Just for example.. 
image

apps/site/pages/en/learn/asynchronous-work/discover-promises-in-nodejs.md
```diff
@@ -0,0 +1,401 @@
...
+1. **Synchronous code** (e.g., regular function calls and script code)
+2. **`process.nextTick()`** callbacks
+3. **Microtasks** (e.g., Promises, `queueMicrotask()`)
+   If a microtask queues another microtask, the new one is also executed before moving on.
+4. **Timers** (e.g., `setTimeout()`, `setInterval()`)
+5. **I/O callbacks** (e.g., network and file system operations)
+   Some operations, like `close` and `error` events, may use `process.nextTick()` internally and are executed in the next tick.
+6. **`setImmediate()`** callbacks
+
+`process.nextTick()` can execute at any stage in the execution order, as it is not bound by the event loop.
```
jasnell: Sorry if I wasn't clear... this is *also* true of draining the microtask queue. We always drain both the next tick and microtask queue (in that order) at the same time.

avivkeller: I realized we have another article on the event loop, so I've referenced it instead

jasnell (approved):
Let's ask @mcollina to review as well.

avivkeller:
bump @mcollina (Previously requested changes) @nodejs/nodejs-website (Required) for reviews.

AugustinMauroy (approved):
LGMT !

avivkeller:
It's been a while since any objections, so I'm merging this PR. Any changes can be requested in a follow up.

PR #7650, opened 2025-04-15
fix(download): Devbox is not supported on Windows

avivkeller:
DevBox is not supported on Windows without WSL. If you are using WSL, you should refer to the Linux instructions

### DIFF
--- a/apps/site/util/downloadUtils.tsx
+++ b/apps/site/util/downloadUtils.tsx
@@ -189,7 +189,7 @@ export const INSTALL_METHODS: Array<
   {
     label: InstallationMethodLabel.DEVBOX,
     value: 'DEVBOX',
-    compatibility: { os: ['WIN', 'MAC', 'LINUX'] },
+    compatibility: { os: ['MAC', 'LINUX'] },
     iconImage: <InstallMethodIcons.Devbox width={16} height={16} />,
     url: 'https://jetify.com/devbox/',
     info: 'layouts.download.codeBox.platformInfo.devbox',


ljharb (approved):
it'd be nice to have a "WSL" option somehow prominent, so that windows users who have WSL can choose nvm, devbox, etc, but atm this seems good

avivkeller:
Users running WSL should refer to the Linux instructions, as their environment is a Linux one.

bjohansebas (approved):

MattIPv4:
Are there snippets that need removing as part of this change?

avivkeller:
No, DevBox uses the same snippet for every operation system

MattIPv4 (approved):

ljharb:
@avivkeller yes, but windows users may not realize WSL maps to linux, or may not realize WSL is an option.

MattIPv4:
We should move this discussion to an issue, but yeh, either we should update "Linux" to be "Linux (+ WSL)" or add a dedicated "WSL" option in the dropdown. Or, we could display a banner whenever Windows is selected noting that if the user is using WSL, they should refer to the Linux options.

avivkeller:
@ljharb It's in the name, Windows Subsystem for Linux. IMHO users should be expected to know that if they WSL, they can refer to the Linux instructions

MattIPv4:
> IMHO users should be expected to know that if they WSL, they can refer to the Linux instructions

The whole point of this page is to _not_ make assumptions, this page is for beginners that're new to Node.js (and often development generally). They may have heard that WSL is the thing to use for Windows dev, and not know anything beyond that.

ljharb:
Yes, but windows users *who are not using or aware of WSL* won't know that WSL even exists as an option in the first place such that they *could* use the things listed under Linux.

avivkeller:
Feel free to open an issue / PR with whatever change you'd like to see

MattIPv4:
https://github.com/nodejs/nodejs.org/issues/7651 👍

AugustinMauroy (approved):

avivkeller:
I'm fast tracking this, as we have several approvals, and it's unlikely we will get an objection

PR #7666, opened 2025-04-18
fix(chromatic): better job name

avivkeller:
Looks like the required status checks might confuse the `chromatic` job id with the `Tests` job name. Naming them both Chromatic is probably a better idea

### DIFF
--- a/.github/workflows/chromatic.yml
+++ b/.github/workflows/chromatic.yml
@@ -43,7 +43,7 @@ jobs:
         github.event.pull_request.head.ref != 'chore/crowdin'
       )
 
-    name: Tests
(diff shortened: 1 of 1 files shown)


AugustinMauroy (approved):

avivkeller:
@nodejs/web-infra 

apologies for the incorrect information earlier, please adjust the required check to have a capital C in Chromatic.

this’ll need a push directly to main, since the required check technically does not exist until this PR lands (as it corresponds to name, not ID)

avivkeller:
Actually, does this even need to be a required check? not every PR needs a chromatic deployment?

PR #7662, opened 2025-04-17
feat(infra): adopt `pnpm`

bmuenzenmeyer:
- [x] `next.config.js`, `next-intl`, i18n work — [commit](https://github.com/nodejs/nodejs.org/pull/7662/commits/c495fb18eae7260ac682fe21269b9c1bbf8d1104)
- [x] `pnpm build` — import errors resolved, packages cleaned up
	- [x] `pnpm build` when installed with `--prod` 
- [x] `pnpm dev`
	- [x] `pnpm dev` when installed with `--prod` 
- [x] `pnpm test` - subpath import work needed
- [x] `pnpm test:ci`
- [x] `pnpm lint`
- [x] `pnpm lint:fix`
- [x] `pnpm prettier`
- [x] `pnpm prettier:fix`
- [x] `pnpm format`
- [x] `pnpm storybook`
- [x] `pnpm storybook:build`
- [x] `pnpm check-types`
- [x] `pnpm scripts:release-post`
- [x] Vercel Deployments 
- [ ] ~~npm use cases~~

### Tasks from Vote
- [x] [Add `pnpm` to README and docs](https://github.com/nodejs/nodejs.org/discussions/5334#discussioncomment-12827848)
- [x] [Add `package-lock.json` to `.gitignore`](https://github.com/nodejs/nodejs.org/discussions/5334#discussioncomment-12827850)
- [x] [Use `pnpm` in CI](https://github.com/nodejs/nodejs.org/discussions/5334#discussioncomment-12827853)
- [x] [Check `pnpm-lock.yaml` in CI](https://github.com/nodejs/nodejs.org/discussions/5334#discussioncomment-12827854)
	- This is always done when installing from a CI environment, see https://pnpm.io/cli/install#--frozen-lockfile 
- [x] [Add `devEngines` and `packageManager` to `package.json`](https://github.com/nodejs/nodejs.org/discussions/5334#discussioncomment-12827878)

PR #7676, opened 2025-04-23
add package-lock=true to .npmrc

mcollina:
As titled. A user can override this with their global `.npmrc` setting.

Note that the project _does not build_ if the deps are not installed from the lockfile.

### DIFF
--- a/.npmrc
+++ b/.npmrc
@@ -1,3 +1,4 @@
 # See bug in peerdeps resolution: https://github.com/npm/cli/issues/2999
 legacy-peer-deps = false
-package-lock=true
+# TODO(@mcollina): Currently, the lockfile is needed to build, so we explicitly require it's use
+package-lock = true
--- a/.npmrc
+++ b/.npmrc
@@ -1,2 +1,3 @@
 # See bug in peerdeps resolution: https://github.com/npm/cli/issues/2999
 legacy-peer-deps = false
+package-lock=true


avivkeller:
Can you elaborate why this is needed?

What are the consequences of _not_ doing this?

MattIPv4:
Per the docs, https://docs.npmjs.com/cli/v11/using-npm/config#package-lock, this is set to true by default -- why do we need to explicitly set it?

mcollina:
Because quite a few high-profile maintainers (including me) set this to `false` by default as it's way too much work to keep this up-to-date over 500+ projects. Also, it doesn't hurt to add it.

Note it would not be needed if the project could build without that file - currently it doesn't . My guess is some dependencies are not really following semver.

avivkeller:
> Note it would not be needed if the project could build without that file - currently it doesn't . My guess is some dependencies are not really following semver.

We should look into that

MattIPv4:
Ah, I see, makes sense to me to set it then 👍 Would it be worth adding a comment above the line explaining why we're setting what is a default?

.npmrc
```diff
@@ -1,2 +1,3 @@
 # See bug in peerdeps resolution: https://github.com/npm/cli/issues/2999
 legacy-peer-deps = false
+package-lock=true
```
avivkeller:
```suggestion
# TODO(@mcollina): Currently, the lockfile is needed to build, so we explicitly require it's use
package-lock = true
```

avivkeller (approved):

avivkeller:
> Note it would not be needed if the project could build without that file - currently it doesn't . My guess is some dependencies are not really following semver.

I am able to build with a deleted `package-lock.json` file, can you show the error you face when trying to build without the file? It would certainly help with reproducing and resolving the error.

```bash
git clean -dfX # Remove all non-tracked files
rm package-lock.json
npm i
npm build
```

bjohansebas:
> Note it would not be needed if the project could build without that file - currently it doesn't . My guess is some dependencies are not really following semver.

It's a precaution to avoid breaking production and to ensure everyone has the same development environment. It's a good practice, and that's why deleting the package-lock isn't an option I see as viable

.npmrc
```diff
@@ -1,2 +1,4 @@
 # See bug in peerdeps resolution: https://github.com/npm/cli/issues/2999
 legacy-peer-deps = false
+# TODO(@mcollina): Currently, the lockfile is needed to build, so we explicitly require it's use
```
bjohansebas, line 3: As I mentioned in my previous comment, it's a way to prevent breaking production and to ensure we all share the same development environment. So a comment like that can be useful, and it’s worth mentioning that some developers disabled it for a specific reason.

avivkeller:
Closing due to #7662

mcollina:
pnpm suffers from the same issue, and it's 100% irrelevant to why this is needed.

MattIPv4:
Given we no longer have an npmrc file at all, could you not just have your own locally and add it to your local gitignore, if this is something your setup needs still?

Issue #7681, opened 2025-04-24
Clarify setImmediate not always run before setTimeout at main module in the learn Asynchronous Work doc


kekekuli:
### Affected URL(s)

https://nodejs.org/en/learn/asynchronous-work/understanding-processnexttick

### Description of the problem
In that doc, here is an example code to explain how the nextTick() API affect the execute order. It is right for the nextTick() part. But readers will quickly get messy when they see the output example saying that the setImmediate always have higher order than setTimeout , while both at top-level. The [other doc](https://nodejs.org/en/learn/asynchronous-work/event-loop-timers-and-nexttick#setimmediate-vs-settimeout) have pointed out that behavior.

So it is better to chang the setimmediate-vs-settimeout part in the example code in that doc.

kekekuli:
Hello, @avivkeller. I have made changes based on your advise, please let me know if you'd like any adjustments!

avivkeller:
For future reference, it's much easier to review if you leave comments regarding the PR on the PR, rather than on the issue. I've responded to your PR with additional comments.

Thank you!

PR #7682, opened 2025-04-24
fix(learn): improve the learn Asynchronous Work doc

kekekuli:
## Description
The ['Understanding process.nextTick()' doc](https://nodejs.org/en/learn/asynchronous-work/understanding-processnexttick) have one messy example code and example ouput caused by [setImmediate() vs setTimeout()](https://nodejs.org/en/learn/asynchronous-work/event-loop-timers-and-nexttick#setimmediate-vs-settimeout). Example code in that doc saying that setImmediate will always have higher order been executed, which may confused for the readers

## Validation
The example code has been reviewed to correctly reflect the behavior of the executed order for both setImmediate and setTimeout.
Reviewers should verify that the updated explanation is both technically correct and clearly communicates. No functional or visual changes—only documentation improvements.

## Related Issues
nodejs/nodejs.org#7681

### Check List

- [Yes] I have read the [Contributing Guidelines](https://github.com/nodejs/nodejs.org/blob/main/CONTRIBUTING.md) and made commit messages that follow the guideline.
- [Yes] I have run `npm run format` to ensure the code follows the style guide.
- [Yes] I have run `npm run test` to check if all tests are passing.
- [Yes] I have run `npx turbo build` to check if the website builds without errors.
- [Yes] I've covered new added functionality with unit tests if necessary.

### DIFF
--- a/apps/site/pages/en/learn/asynchronous-work/understanding-processnexttick.md
+++ b/apps/site/pages/en/learn/asynchronous-work/understanding-processnexttick.md
@@ -31,11 +31,11 @@ Use `nextTick()` when you want to make sure that in the next event loop iteratio
 console.log('Hello => number 1');
 
 setImmediate(() => {
-  console.log('Running before the timeout => number 3');
+  console.log('Running at order 3 or 4, setImmediate');
 });
 
 setTimeout(() => {
-  console.log('The timeout running last => number 4');
+  console.log('Running at order 3 or 4, setTimeout');
 }, 0);
 
 process.nextTick(() => {
@@ -43,13 +43,13 @@ process.nextTick(() => {
 });
 ```
 
-#### Example output:
+#### Example output(possible):
 
 ```bash
 Hello => number 1
 Running at next tick => number 2
-Running before the timeout => number 3
-The timeout running last => number 4
+Running at order 3 or 4, setImmediate
+Running at order 3 or 4, setTimeout
 ```
 
-The exact output may differ from run to run.
+The exact output may differ from run to run. [setImmediate vs setTimeout](https://nodejs.org/en/learn/asynchronous-work/event-loop-timers-and-nexttick#setimmediate-vs-settimeout) explain the reason.
--- a/apps/site/pages/en/learn/asynchronous-work/understanding-processnexttick.md
+++ b/apps/site/pages/en/learn/asynchronous-work/understanding-processnexttick.md
@@ -54,4 +54,4 @@ Running at order 3 or 4, setTimeout
 
 In that case, `console.log('Hello => number 1');` will first run because event loop only starts after call stack is cleared.The `nextTick` queue is processed before entering the next phase, which is why the `process.nextTick()` callback runs immediately after.
 
-The execution order of `serImmediate` and `setTimeout` can not be determined in the main module. For a detailed explanation, refer to [setImmediate vs setTimeout](https://nodejs.org/en/learn/asynchronous-work/event-loop-timers-and-nexttick#setimmediate-vs-settimeout).
+The execution order of `setImmediate()` and `setTimeout()` can very based on the execution contexts. For more information, refer to [`setImmediate()` vs `setTimeout()`](https://nodejs.org/en/learn/asynchronous-work/event-loop-timers-and-nexttick#setimmediate-vs-settimeout).
--- a/apps/site/pages/en/learn/asynchronous-work/understanding-processnexttick.md
+++ b/apps/site/pages/en/learn/asynchronous-work/understanding-processnexttick.md
@@ -25,4 +25,4 @@ Calling `setTimeout(() => {}, 0)` will execute the function at the end of next t
 
 Use `nextTick()` when you want to make sure that in the next event loop iteration that code is already executed.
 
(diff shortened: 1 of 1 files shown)


apps/site/pages/en/learn/asynchronous-work/understanding-processnexttick.md
```diff
@@ -31,25 +31,27 @@ Use `nextTick()` when you want to make sure that in the next event loop iteratio
 console.log('Hello => number 1');
```
avivkeller: To clarify, The entire "Order of events" section should probably be replaced with a link to the event loop documentation.

Like:
```
The execution order of `setImmediate()` and `setTimeout()` can very based on the execution contexts. For more information, refer to [`setImmediate()` vs `setTimeout()`](https://nodejs.org/en/learn/asynchronous-work/event-loop-timers-and-nexttick#setimmediate-vs-settimeout).
```

avivkeller (changes requested):
I don't think we should give an example, and then say "Hey, this isn't always the case"

I think we should replace the entire section with a link to where a user can find valid examples / use cases, as I said in my other comment.

kekekuli:
> I don't think we should give an example, and then say "Hey, this isn't always the case"
> 
> I think we should replace the entire section with a link to where a user can find valid examples / use cases, as I said in my other comment.

The problem is the behavior of setTimeout vs setImmediate is not the key in the section. May be we can choose one to leave, and choose one to keep.

kekekuli:
> > I don't think we should give an example, and then say "Hey, this isn't always the case"
> > I think we should replace the entire section with a link to where a user can find valid examples / use cases, as I said in my other comment.
> 
> The problem is the behavior of setTimeout vs setImmediate is not the key in the section. May be we can choose one to left.
After that, left one link that redirect to the 

> I don't think we should give an example, and then say "Hey, this isn't always the case"
> 
> I think we should replace the entire section with a link to where a user can find valid examples / use cases, as I said in my other comment.

> I don't think we should give an example, and then say "Hey, this isn't always the case"
> 
> I think we should replace the entire section with a link to where a user can find valid examples / use cases, as I said in my other comment.

Using the **event loop documentation** as an example to illustrate the behavior of nextTick()—do you think this might be too heavy?

kekekuli:
> > I don't think we should give an example, and then say "Hey, this isn't always the case"
> > I think we should replace the entire section with a link to where a user can find valid examples / use cases, as I said in my other comment.
> 
> The problem is the behavior of setTimeout vs setImmediate is not the key in the section. May be we can choose one to left.
After that, left one link that redirect to the 

> I don't think we should give an example, and then say "Hey, this isn't always the case"
> 
> I think we should replace the entire section with a link to where a user can find valid examples / use cases, as I said in my other comment.

> I don't think we should give an example, and then say "Hey, this isn't always the case"
> 
> I think we should replace the entire section with a link to where a user can find valid examples / use cases, as I said in my other comment.

Use the **event loop documentation**  as the example to show the behavior of **nextTick()**

avivkeller:
What I'm saying is:

I think we should remove the entire example. The example isn't accurate, as you said, and we have an entire article dedicated to explaining the order of execution.

Now, if we are removing the example, it would make sense to include a link to its successor, that is, the other article.

kekekuli:
> What I'm saying is:
> 
> I think we should remove the entire example. The example isn't accurate, as you said, and we have an entire article dedicated to explaining the order of execution.
> 
> Now, if we are removing the example, it would make sense to include a link to its successor, that is, the other article.

I've removed the example code and replaced it with a link. Let me know if you'd like any changes.

apps/site/pages/en/learn/asynchronous-work/understanding-processnexttick.md
```diff
@@ -25,31 +25,4 @@ Calling `setTimeout(() => {}, 0)` will execute the function at the end of next t
...
-
-```bash
-Hello => number 1
-Running at next tick => number 2
-Running before the timeout => number 3
-The timeout running last => number 4
-```
-
-The exact output may differ from run to run.
+For more details on the order of events, refer to [Event loop](https://nodejs.org/en/learn/asynchronous-work/event-loop-timers-and-nexttick).
```
avivkeller:
```suggestion
To learn more about the order of execution and how the event loop works, check out [the dedicated article](learn/asynchronous-work/event-loop-timers-and-nexttick).
```

avivkeller (approved):

apps/site/pages/en/learn/asynchronous-work/understanding-processnexttick.md
```diff
@@ -25,4 +25,4 @@ Calling `setTimeout(() => {}, 0)` will execute the function at the end of next t
 
 Use `nextTick()` when you want to make sure that in the next event loop iteration that code is already executed.
 
-For more details on the order of events, refer to [Event loop](https://nodejs.org/en/learn/asynchronous-work/event-loop-timers-and-nexttick).
+To learn more about the order of execution and how the event loop works, check out [the dedicated article](learn/asynchronous-work/event-loop-timers-and-nexttick)
```
avivkeller: `/learn`, not `learn`

kekekuli: oops,  seems I accidentally deleted something

bjohansebas (approved):

PR #7690, opened 2025-04-30
chore: update env file inclusion order to follow common practices

nvitaterna:
## Description

This change updates the order of the `--env-file` multiple inclusions example to follow standard practices. Typically the `.env` file includes local vars that override other env files included in a project.

## Validation

N/A

## Related Issues

N/A

### Check List

### DIFF
--- a/apps/site/pages/en/learn/command-line/how-to-read-environment-variables-from-nodejs.md
+++ b/apps/site/pages/en/learn/command-line/how-to-read-environment-variables-from-nodejs.md
@@ -53,7 +53,7 @@ This command loads all the environment variables from the `.env` file, making th
 Also, you can pass multiple `--env-file` arguments. Subsequent files override pre-existing variables defined in previous files.
 
 ```bash
-node --env-file=.env --env-file=.development.env app.js
+node --env-file=.development.env --env-file=.env app.js
 ```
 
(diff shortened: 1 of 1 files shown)


avivkeller:
I’m not convinced this change adds much value. The article already mentions that the order is important. I don’t think we need to enforce a specific order in our example, but I won't block.

AugustinMauroy:
+1 as aviv opinion

MattIPv4:
I would +1 that there isn't reason to change this based on order already being made clear as something that matters (and IMO, you tend to have env vars for all environments in `.env` and then developement-specific overrides in `.development.env`, so the existing order makes sense to me).

dario-piotrowicz:
+1 to Aviv and Matt opinions

This feels quite subjective to me, I'm not against the change specifically but I also don't see much point in it honestly 😕

nvitaterna:
Thanks for the feedback everyone. I agree and will close this.

nvitaterna:
Thanks for the feedback everyone. I agree with the comments here and will close this.

dario-piotrowicz:
> Thanks for the feedback everyone. I agree with the comments here and will close this.

Thanks, and thanks anyways for the pull request 🙂

PR #7735, opened 2025-05-08
chore: use subpath imports instead of path aliases

### LINKED ISSUE #7732: Move from `tsconfig.json` paths to Subpath imports
(Self-assigning as I am happy to take this on)

Currently, we’re using TypeScript’s paths option to handle module imports (which, until recently, was enforced by our linter).

While this setup has worked, it’s created several ongoing hurdles:
	•	Packages can't import modules with the same selector
	•	Our test runner needs a TypeScript transpiler that supports these aliases, limiting our ability to adopt faster alternatives
	•	Extra overhead in maintaining consistency between tsconfig.json in various places

To simplify things, I propose we move to Node.js native import aliases. This would involve changing our current `@` prefix to something like `#`, following the Node convention.

avivkeller:
## Description

This PR uses Node.js's native subpath imports in place of the `tsconfig.json` `paths`, since these are natively supported

## Validation

Nothing should change, everything should build and test correctly

## Related Issues

Fixes #7732

### Check List

### DIFF
--- a/COLLABORATOR_GUIDE.md
+++ b/COLLABORATOR_GUIDE.md
@@ -356,7 +356,7 @@ To add a new download installation method, follow these steps:
    }
    ```
 
-3. **Update `InstallationMethodLabel` and `InstallationMethod` in `@/types/release.ts`:**
+3. **Update `InstallationMethodLabel` and `InstallationMethod` in `apps/site/types/release.ts`:**
 
    - Add the new method to the `InstallationMethodLabel` and `InstallationMethod` types.
 
--- a/apps/site/app/[locale]/[...path]/page.tsx
+++ b/apps/site/app/[locale]/[...path]/page.tsx
@@ -7,13 +7,13 @@
  * dynamic params, which will lead on static export errors and other sort of issues.
  */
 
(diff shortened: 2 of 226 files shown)


MattIPv4:
Perhaps bikeshedding, but personally I would prefer `#/` rather than `#`, if that's an option?

avivkeller:
I just tried that, assuming it would work, and it looks like `/` can not immediately follow the `#`. However, something like `#@/` would work. (Or like `#site/` and `#ui/`, etc)

https://github.com/nodejs/node/blob/daced4ab98be82953ef2fa73e0f81e2b1967be8b/test/es-module/test-esm-imports.mjs#L67

avivkeller:
I went with `#site/` and `#ui/` for the different parts of the repo.

MattIPv4:
👍 That satisfies my brain -- I think I just didn't like the bare `#` on its own

dario-piotrowicz (approved):
Looks good to me :slightly_smiling_face:

bjohansebas (approved):
LGTM

AugustinMauroy (approved):

PR #7741, opened 2025-05-12
chore(pnpm): use installed dependencies for pre-commit hook

avivkeller:
Per #7738

### DIFF
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -1,5 +1,5 @@
 # lint and format staged files
-pnpm dlx lint-staged
+pnpm exec lint-staged
 
 # verify typescript staged files
-pnpm --package=typescript dlx tsc --build apps/site
+pnpm check-types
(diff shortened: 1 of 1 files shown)


.husky/pre-commit
```diff
@@ -1,5 +1,5 @@
 # lint and format staged files
-pnpm dlx lint-staged
+pnpm exec lint-staged
 
 # verify typescript staged files
```
MattIPv4:
```suggestion
# verify typescript fully
```
This is doing a full type-check, not only staged

avivkeller: I've updated the comment. For the record, this was an intentional change as the alternative would be adding TypeScript as a workspace root dependency (I think), so I figured it'd just be better to use our pre existing scripts.

MattIPv4 (approved):

araujogui (approved):

AugustinMauroy (approved):
LGMT !

aymen94 (approved):

bmuenzenmeyer (approved):

PR #7743, opened 2025-05-13
Add CI check for Cloudflare Worker build

dario-piotrowicz:
## Description

As [suggested](https://github.com/nodejs/nodejs.org/pull/7383#issuecomment-2863779557) by @avivkeller I'm adding a CI check to make sure that the Cloudflare OpenNext adapter build succeeds on PRs/pushes to main.

As part of this I am removing the pre scripts I had before in favour of turborepo `dependsOn` configs (I am not super familiar with turborepo but I thing/guess that `dependsOn` configs should be preferred and replace pre scripts? 🤔)

> [!Note]
> This is a temporary CI check (that's also one of the reasons why I created a new, very easy to delete, workflow file for it instead of adding this to the build workflow) that can help us get some confidence in the OpenNext Cloudflare adapter and in the fact that the adapter doesn't effect DX. 
> However I think that this will become obsolete/unnecessary as soon as we have some e2e tests that also validate the OpenNext build.

## Validation

I've validated this workflow in my nodejs.org fork see [action run](https://github.com/dario-piotrowicz/nodejs.org/actions/runs/14999151591/job/42141131986). 

### Check List

### DIFF
--- a/.github/workflows/cloudflare-open-next-build.yml
+++ b/.github/workflows/cloudflare-open-next-build.yml
@@ -10,7 +10,7 @@ on:
   push:
     branches:
       - main
-  pull_request_target:
+  pull_request:
     branches:
       - main
     types:
--- a/.github/workflows/cloudflare-open-next-build.yml
+++ b/.github/workflows/cloudflare-open-next-build.yml
@@ -61,4 +61,4 @@ jobs:
         run: pnpm install --frozen-lockfile
 
       - name: Build Cloudflare Worker
-        run: pnpm exec turbo run --filter=@node-core/website cloudflare:build:worker ${{ env.TURBO_ARGS }}
+        run: pnpm exec turbo run cloudflare:build:worker ${{ env.TURBO_ARGS }}
--- a/.github/workflows/cloudflare-open-next-build.yml
+++ b/.github/workflows/cloudflare-open-next-build.yml
@@ -13,8 +13,6 @@ on:
   pull_request:
     branches:
       - main
-    types:
-      - labeled
 
 defaults:
   run:
@@ -33,7 +31,7 @@ env:
 
 jobs:
   build-cloudflare-worker:
-    name: Build Cloudflare Worker (on ubuntu)
+    name: Build Cloudflare Worker
     runs-on: ubuntu-latest
 
     steps:
--- a/.github/workflows/cloudflare-open-next-build.yml
+++ b/.github/workflows/cloudflare-open-next-build.yml
@@ -1,0 +1,69 @@
+# Security Notes
+# Only selected Actions are allowed within this repository. Please refer to (https://github.com/nodejs/nodejs.org/settings/actions)
+# for the full list of available actions. If you want to add a new one, please reach out a maintainer with Admin permissions.
+# REVIEWERS, please always double-check security practices before merging a PR that contains Workflow changes!!
+# AUTHORS, please only use actions with explicit SHA references, and avoid using `@master` or `@main` references or `@version` tags.
+
+name: Cloudflare OpenNext Build
+
+on:
+  push:
+    branches:
+      - main
+  pull_request_target:
+    branches:
+      - main
+    types:
+      - labeled
+
+defaults:
+  run:
+    # This ensures that the working directory is the root of the repository
+    working-directory: ./
+
+permissions:
+  contents: read
+  actions: read
+
+env:
+  # See https://turbo.build/repo/docs/reference/command-line-reference/run#--cache-dir
+  TURBO_ARGS: --cache-dir=.turbo/cache
+  # See https://turbo.build/repo/docs/reference/command-line-reference/run#--force
+  TURBO_FORCE: true
+
+jobs:
+  build-cloudflare-worker:
+    name: Build Cloudflare Worker on ${{ matrix.os }}
+    runs-on: ${{ matrix.os }}
+
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-latest, windows-latest]
+
+    steps:
+      - name: Harden Runner
+        uses: step-security/harden-runner@0634a2670c59f64b4a01f0f96f84700a4088b9f0 # v2.12.0
+        with:
+          egress-policy: audit
+
+      - name: Git Checkout
+        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+
+      - name: Set up pnpm
+        uses: pnpm/action-setup@a7487c7e89a18df4991f7f222e4898a00d66ddda # v4.1.0
+        with:
+          cache: true
+
+      - name: Set up Node.js
+        uses: actions/setup-node@49933ea5288caeca8642d1e84afbd3f7d6820020 # v4.4.0
+        with:
+          # We want to ensure that the Node.js version running here respects our supported versions
+          node-version-file: '.nvmrc'
+          cache: 'pnpm'
+
+      - name: Install packages
+        run: pnpm install --frozen-lockfile
+
+      - name: Build Cloudflare Worker
+        run: pnpm turbo run --filter=@node-core/website cloudflare:build:worker
--- a/.github/workflows/cloudflare-open-next-build.yml
+++ b/.github/workflows/cloudflare-open-next-build.yml
@@ -61,4 +61,4 @@ jobs:
         run: pnpm install --frozen-lockfile
 
       - name: Build Cloudflare Worker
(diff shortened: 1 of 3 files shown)


.github/workflows/cloudflare-open-next-build.yml
```diff
@@ -0,0 +1,69 @@
...
+        with:
+          # We want to ensure that the Node.js version running here respects our supported versions
+          node-version-file: '.nvmrc'
+          cache: 'pnpm'
+
+      - name: Install packages
+        run: pnpm install --frozen-lockfile
+
+      - name: Build Cloudflare Worker
+        run: pnpm turbo run --filter=@node-core/website cloudflare:build:worker
```
avivkeller:
```suggestion
        run: pnpm exec turbo run --filter=@node-core/website cloudflare:build:worker ${{ env.TURBO_ARGS }}
```

.github/workflows/cloudflare-open-next-build.yml
```diff
@@ -0,0 +1,69 @@
...
+        os: [ubuntu-latest, windows-latest]
+
+    steps:
+      - name: Harden Runner
+        uses: step-security/harden-runner@0634a2670c59f64b4a01f0f96f84700a4088b9f0 # v2.12.0
+        with:
+          egress-policy: audit
+
+      - name: Git Checkout
+        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
```
avivkeller, line 51: Make sure to check out the PR

.github/workflows/cloudflare-open-next-build.yml
```diff
@@ -0,0 +1,69 @@
...
+  actions: read
+
+env:
+  # See https://turbo.build/repo/docs/reference/command-line-reference/run#--cache-dir
+  TURBO_ARGS: --cache-dir=.turbo/cache
+  # See https://turbo.build/repo/docs/reference/command-line-reference/run#--force
+  TURBO_FORCE: true
+
+jobs:
+  build-cloudflare-worker:
```
avivkeller, line 35: Check the label

.github/workflows/cloudflare-open-next-build.yml
```diff
@@ -0,0 +1,69 @@
...
+# REVIEWERS, please always double-check security practices before merging a PR that contains Workflow changes!!
+# AUTHORS, please only use actions with explicit SHA references, and avoid using `@master` or `@main` references or `@version` tags.
+
+name: Cloudflare OpenNext Build
+
+on:
+  push:
+    branches:
+      - main
+  pull_request_target:
```
avivkeller: We don't use secrets in this workflow, so it can run on pull_request

.github/workflows/cloudflare-open-next-build.yml
avivkeller: Also, do we need to --filter?

Can you run it in the apps/site working directory, or just run it with it filter (if that's possible?)

.github/workflows/cloudflare-open-next-build.yml
```diff
@@ -0,0 +1,64 @@
...
+
+      - name: Set up Node.js
+        uses: actions/setup-node@49933ea5288caeca8642d1e84afbd3f7d6820020 # v4.4.0
+        with:
+          # We want to ensure that the Node.js version running here respects our supported versions
+          node-version-file: '.nvmrc'
+          cache: 'pnpm'
+
+      - name: Install packages
+        run: pnpm install --frozen-lockfile
```
avivkeller, line 59:
```suggestion
        run: pnpm install --frozen-lockfile --prod
```

.github/workflows/cloudflare-open-next-build.yml
dario-piotrowicz: I can just remove the filter it's not actually needed 🙂  (I just thought it made things slightly clearer)

.github/workflows/cloudflare-open-next-build.yml
dario-piotrowicz: Will this work? `@opennextjs/cloudflare` is a dev dependency after all 🤔

.github/workflows/cloudflare-open-next-build.yml
dario-piotrowicz: you're right, sorry I just copy-pasted this part without noticing this 👍

.github/workflows/cloudflare-open-next-build.yml
dario-piotrowicz: I tested this locally and indeed it doesn't work:
Screenshot 2025-05-13 at 15 38 18

avivkeller: I suppose we can leave it as is, and cross that bridge when we get there

avivkeller (approved):

.github/workflows/cloudflare-open-next-build.yml
dario-piotrowicz: is there any issue with not including `--prod`? 🤔

avivkeller: I was under the impression that this _should_ build in production mode (similar to how next does)–but since we aren't relying on OpenNext for deployment, it's not really a problem.

dario-piotrowicz: Yes, it does build in production mode, but it does so using some non-production dependencies, so I _feel_ like this is fine 🤔

I am guessing that this works with Next because in a sense the next package is both a dev (it builts your app) and production dependency (you can import stuff from next)

flakey5 (approved):

.github/workflows/cloudflare-open-next-build.yml
dario-piotrowicz: Ah! actually! The OpenNext adapter is the same, you can import from it ([`getCloudflareContext`](https://opennext.js.org/cloudflare/bindings#how-to-access-bindings-in-your-nextjs-app)), Victor indeed did somewhat recently update the [OpenNext getting started doc](https://opennext.js.org/cloudflare/get-started#1-install-opennextjscloudflare) to instruct users to install the adapter as a proper dependency, if you prefer that I can just make the adapter a proper dependency 🙂 (and add here the `--prod` flag)

(even though right now we don't want/can't use `getCloudflareContext` from the package since that wouldn't work in the Vercel deployment)

avivkeller: Let’s keep as a dev dependency until it needs to be a production one, at least for now.

.github/workflows/cloudflare-open-next-build.yml
```diff
@@ -0,0 +1,64 @@
...
+
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+    branches:
+      - main
+    types:
+      - labeled
```
avivkeller:
```suggestion
```

apps/site/turbo.json
```diff
@@ -24,7 +25,7 @@
       ]
     },
     "build": {
-      "dependsOn": ["^build"],
+      "dependsOn": ["build-blog-data", "^build"],
```
avivkeller, line 28: `build-blog-data` is already in `prebuild`, will this cause it to execute twice?

.github/workflows/cloudflare-open-next-build.yml
```diff
@@ -0,0 +1,64 @@
...
+
+env:
+  # See https://turbo.build/repo/docs/reference/command-line-reference/run#--cache-dir
+  TURBO_ARGS: --cache-dir=.turbo/cache
+  # See https://turbo.build/repo/docs/reference/command-line-reference/run#--force
+  TURBO_FORCE: true
+
+jobs:
+  build-cloudflare-worker:
+    name: Build Cloudflare Worker (on ubuntu)
```
avivkeller:
```suggestion
    name: Build Cloudflare Worker
```

I don't think we need to specify it's ubuntu. Ubuntu is the de-facto

avivkeller:
OpenNext CI run: https://github.com/nodejs/nodejs.org/actions/runs/15002557692/job/42153040053?pr=7743 (In the future, [it won't need the label](https://github.com/nodejs/nodejs.org/pull/7743#discussion_r2087279875)

apps/site/turbo.json
dario-piotrowicz: surprisingly no it doesn't seem to be the case 🤔 I wonder if turborepo is smart enough to handle that on its own 🤔

avivkeller: Wonderful!

bjohansebas (approved):

aymen94 (approved):

PR #7744, opened 2025-05-13
fix: Ensure links maintain green color within bold text

### LINKED ISSUE #7742: Bold links are not green
For example, in https://nodejs.org/en/learn/getting-started/an-introduction-to-the-npm-package-manager, it's not immediately clear that "Yarn" and "pnpm" in the sentence *"Yarn and pnpm are alternatives to npm cli. You can check them out as well."* are actually hyperlinks.

shashankxrm:
## Description

This PR fixes an issue where links within bold text (using markdown `**text**` syntax) were inheriting the white color from the strong tags instead of maintaining their green color. The fix adds specific CSS selectors to ensure links maintain their green color regardless of whether they are inside or contain bold text.

## Validation

The fix ensures that:
1. Links within bold text maintain their green color
2. Links containing bold text maintain their green color
3. The styling is consistent in both light and dark modes
4. Hover states work correctly for all link variations

Before:
SCR-20250514-cpfb-2
SCR-20250514-cpxs-2

After:

SCR-20250514-cpmr-2
SCR-20250514-cpsz-2

## Related Issues

This PR addresses the issue #7742 (Bold Links are not Green)

### Check List

- [ ] I've covered new added functionality with unit tests if necessary. (NOT REQUIRED)

### DIFF
--- a/packages/ui-components/styles/markdown.css
+++ b/packages/ui-components/styles/markdown.css
@@ -40,22 +40,18 @@ main {
   h4,
   h5,
   h6 {
+    @apply font-semibold
+      text-neutral-900
+      dark:text-white;
+
     &[id] a {
       @apply text-neutral-900
         dark:text-white;
     }
   }
 
-  h1,
-  h2,
-  h3,
-  h4,
-  h5,
-  h6,
   strong {
-    @apply font-semibold
-      text-neutral-900
-      dark:text-white;
+    @apply font-semibold;
   }
 
   code {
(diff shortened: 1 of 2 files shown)


avivkeller:
Wouldn't it make more sense to just not change the color on strongs (or just strong links) rather than making the green more important?

Earlier in the file the "strong" styles are defined. Should they really be changing the colors in the first place?

packages/ui-components/styles/markdown.css
```diff
@@ -77,14 +77,17 @@ main {
...
   a:not(h1 > a):not(h2 > a):not(h3 > a):not(h4 > a):not(h5 > a):not(h6 > a),
-  .anchor {
+  .anchor,
+  a strong,
+  strong a,
+  p a {
     @apply max-xs:font-semibold
-      text-green-600
-      dark:text-green-400;
+      !text-green-600
```
bjohansebas: !important, should never be used, unless absolutely necessary. Isn't there any other way to solve this issue?

avivkeller (changes requested):
Formally blocking based on my, and @bjohansebas's, comments. We don't think !important is the way to go.

packages/ui-components/styles/markdown.css
shashankxrm: Yeah fixed it.

AugustinMauroy (changes requested):
We have some stories (storybook) for text could you try to cover these cases

shashankxrm:
> We have some stories (storybook) for text could you try to cover these cases

Sure, I'll look into it.

avivkeller (approved):

AugustinMauroy (approved):

aymen94 (approved):

canerakdas (approved):
LGTM

PR #7755, opened 2025-05-16
fix(crowdin): workflow improvements

avivkeller:
This PR fixes a few bugs in, and hardens, the Crowdin workflow.


avivkeller:
See https://github.com/nodejs/nodejs.org/actions/runs/15078110844/job/42389862525

.github/workflows/translations-sync.yml
```diff
@@ -109,27 +107,19 @@ jobs:
...
+          # according to the configuration in package.json or .lintstagedrc
+          pnpm lint-staged
 
-      - name: Push Changes back to Pull Request
-        uses: stefanzweifel/git-auto-commit-action@b863ae1933cb653a53c021fe36dbb774e1fb9403 # v5.2.0
-        with:
-          commit_options: '--no-verify --signoff'
-          commit_message: 'chore: automated format of translated files'
-          branch: 'chore/crowdin'
+          git commit -m "$COMMIT_MSG" --signoff --no-verify
```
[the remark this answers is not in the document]
avivkeller: This is not needed. There are _always_ changes. In the unlikely case that their aren't, this step will fail-but without any impact on the final outcome.

AugustinMauroy (commented):
why there are changes on translation file ?

avivkeller:
I combined this with the weekly translate run do demo the new, working script.

AugustinMauroy (approved):

avivkeller:
@nodejs/web-infra can I get some approvals on your end 😃? Thanks!

ovflowd (approved):
LGTM, but let's wait for @bmuenzenmeyer LGTM

.github/workflows/translations-sync.yml
```diff
@@ -3,27 +3,27 @@
 
 # See translations-upload.yml for automation to upload our source content
 # See translations-pr-lint.yml for quality control we conduct on ingress of new translations.
-name: Crowdin Download
+name: Crowdin Translations Sync
 
 on:
-  workflow_dispatch: # Allow running when we want to, for events such as urgent translation mistakes or 100% completed languages
+  workflow_dispatch: # Manual trigger for urgent fixes
```
bmuenzenmeyer: 100% completed languages would be out of band, off Friday deployments. If we want to remove that self-documenting use case, I suppose that's fine.

.github/workflows/translations-sync.yml
```diff
@@ -3,27 +3,27 @@
...
   schedule:
-    - cron: '0 5 * * 5' # At 05:00 on Fridays. This guarantees that we have the 72 hour weekend time to review translations.
```
bmuenzenmeyer: I feel like we are losing something here - perhaps we should link explicitly to the 72 hour policy, here and maybe in the PR. We have had a number of problems with that.

.github/workflows/translations-sync.yml
```diff
@@ -3,27 +3,27 @@
...
+    - cron: '0 5 * * 5' # Fridays at 05:00 UTC, allowing weekend review time
 
 # Cancel any runs on the same branch
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
   cancel-in-progress: true
 
-permissions:
-  contents: read
+permissions: {}
```
bmuenzenmeyer: is this more clear? i can assume it means none - but is that right?

.github/workflows/translations-sync.yml
```diff
@@ -3,27 +3,27 @@
...
+
+env:
+  COMMIT_MSG: 'chore: sync translations from crowdin'
```
bmuenzenmeyer: re-using this twice destroys our ability to see that a formatting occurred post-ingestion. This has been useful information in the past. I'd prefer if both commits here have different messages.

.github/workflows/translations-sync.yml
```diff
@@ -3,27 +3,27 @@
...
+  HEAD_REF: chore/crowdin
 
 jobs:
-  synchronize-with-crowdin:
+  sync-translations:
```
bmuenzenmeyer: I am not aware of a standard, so why change? 

see 
- https://github.com/nodejs/nodejs.org/blob/main/.github/workflows/translations-pr-lint.yml#L30
- https://github.com/nodejs/nodejs.org/blob/main/.github/workflows/pull-request-label.yml#L29
- https://github.com/nodejs/nodejs.org/blob/main/.github/workflows/pull-request-label.yml#L29

there are other kebab cases too

.github/workflows/translations-sync.yml
```diff
@@ -35,48 +35,46 @@ jobs:
...
-          ref: refs/pull/${{ needs.synchronize-with-crowdin.outputs.pull_request_number }}/head
+          ref: ${{ env.HEAD_REF }}
           token: ${{ secrets.CROWDIN_GITHUB_BOT_TOKEN }}
+          fetch-depth: 0
+
+      # Setup git user for commits
+      - name: Setup Git User
+        run: |
+          git config --global user.name "nodejs-crowdin"
+          git config --global user.email "nodejs-crowdin@users.noreply.github.com"
```
bmuenzenmeyer: unknown if this is better than the default behavior provided by stefanzweifel/git-auto-commit-action

.github/workflows/translations-sync.yml
```diff
@@ -109,27 +107,19 @@ jobs:
...
+          git reset --soft HEAD^
 
-      - name: Run `prettier --write`
-        # This runs a specific version of Prettier with only the Translation Pages Globbing
-        # This avoid that unrelated changes get prettied/modified within this PR
-        run: pnpm exec prettier "apps/site/{pages,i18n}/**/*.{json,md,mdx}" --check --write --cache --cache-strategy=metadata --cache-location=apps/site/.prettiercache
+          # Run lint-staged on the staged files
+          # This will run linters/formatters only on changed files
+          # according to the configuration in package.json or .lintstagedrc
+          pnpm lint-staged
```
bmuenzenmeyer: overkill using this, albeit cleaner and not without drift

https://github.com/nodejs/nodejs.org/blob/main/.husky/pre-commit

we dont need typechecking or the whole project linted https://github.com/nodejs/nodejs.org/blob/main/.lintstagedrc.json as crowdin is only changing a subset. (I know it only should work on changed files)

can we call https://github.com/nodejs/nodejs.org/blob/main/package.json#L25 instead?

.github/workflows/translations-sync.yml
bmuenzenmeyer: see my other comment about loss of information here

.github/workflows/translations-sync.yml
```diff
@@ -109,27 +107,19 @@ jobs:
...
-      - name: Save Lint Cache
-        uses: actions/cache/save@5a3ec84eff668545956fd18022155c47e93e2684 # v4.2.3
-        with:
-          path: |
-            apps/site/.eslintmdcache
-            apps/site/.prettiercache
-          key: cache-lint-${{ hashFiles('pnpm-lock.yaml') }}-${{ hashFiles('apps/site/.eslintmdcache') }}
+      # Push the changes back to the PR branch
+      - name: Push Changes
+        run: git push origin HEAD:$HEAD_REF --force
```
bmuenzenmeyer: a force push in a CI environment feels like a bad thing, IMO

.github/workflows/translations-sync.yml
```diff
@@ -109,27 +107,19 @@ jobs:
...
-        # This runs a specific version of Prettier with only the Translation Pages Globbing
-        # This avoid that unrelated changes get prettied/modified within this PR
-        run: pnpm exec prettier "apps/site/{pages,i18n}/**/*.{json,md,mdx}" --check --write --cache --cache-strategy=metadata --cache-location=apps/site/.prettiercache
+          # Run lint-staged on the staged files
+          # This will run linters/formatters only on changed files
+          # according to the configuration in package.json or .lintstagedrc
+          pnpm lint-staged
 
-      - name: Push Changes back to Pull Request
-        uses: stefanzweifel/git-auto-commit-action@b863ae1933cb653a53c021fe36dbb774e1fb9403 # v5.2.0
```
bmuenzenmeyer, line 123: I do like the potential less usage of a dependency - i think this is the only place

.github/workflows/translations-sync.yml
```diff
@@ -35,48 +35,46 @@ jobs:
...
 
-        # see all the options at https://github.com/crowdin/github-action
-      - name: Crowdin PR
+      # Downloads translations from Crowdin and creates a PR
+      # See all the options at https://github.com/crowdin/github-action
+      - name: Download Translations & Create PR
         uses: crowdin/github-action@b8012bd5491b8aa8578b73ab5b5f5e7c94aaa6e2 # v2.7.0
-        id: crowdin_pr
         with:
-          # do not upload anything - this is a one-way operation download
```
bmuenzenmeyer, line 43: this, and all the comments, in my opinion are REALLY important. We encountered a problem in earlier crowdin integration where the absence of this config (I think it was this one...) caused all translation progress to be deleted. 

I'd prefer we retain all config comments.

.github/workflows/translations-sync.yml
```diff
@@ -35,48 +35,46 @@ jobs:
...
-          pull_request_body: 'New Crowdin translations from the [Node.js Crowdin project](https://crowdin.com/project/nodejs-web)'
-          commit_message: 'chore: synced translations from crowdin'
+          pull_request_title: '[automated]: Crowdin Translations Sync'
+          pull_request_body: 'New translations from the [Node.js Crowdin project](https://crowdin.com/project/nodejs-web)'
+          commit_message: ${{ env.COMMIT_MSG }}
         env:
           GITHUB_TOKEN: ${{ secrets.CROWDIN_GITHUB_BOT_TOKEN }}
-          # A numeric ID, found at https://crowdin.com/project/nodejs-web/tools/api
           CROWDIN_PROJECT_ID: ${{ secrets.CROWDIN_PROJECT_ID }}
-          # Created from https://crowdin.com/settings#api-key logged in using nodejs-crowdin-bot
```
bmuenzenmeyer: this sorta stuff is important for continuity as the project ages through people

.github/workflows/translations-sync.yml
```diff
@@ -35,48 +35,46 @@ jobs:
...
+    needs: sync-translations
     runs-on: ubuntu-latest
-
-    permissions:
-      # This permission is required by `stefanzweifel/git-auto-commit-action`
-      contents: write
-
     steps:
       - name: Harden Runner
-        uses: step-security/harden-runner@0634a2670c59f64b4a01f0f96f84700a4088b9f0 # v2.12.0
```
bmuenzenmeyer: dependabot manages these versions - why delete them? the previous was retained

bmuenzenmeyer (changes requested):
All in all, this seems like a ton of bikeshedding atop the substantive change of using lint-staged, which is doing more than it needs to (linting, in addition to formatting), with the loss of many code comments essential for the running of the action.

The force push is concerning to me too.

avivkeller:
Sorry for all the bikeshedding–I shouldn've kept this on topic.

avivkeller:
I've changed the `lint-staged` to just use ESLint and prettier. While they'll run over files that haven't changed, the cache should prevent them from actually checking said files.

bmuenzenmeyer (approved):

PR #7762, opened 2025-05-20
chore(ui): move `ArticleLayout` to `ui-components`

avivkeller:
I'm moving this component into the `ui-components` library so it can be re-used by `api-docs-tooling` as part of the JSX generation.


AugustinMauroy (approved):
LGTM but could you try to add a story for this component because we nearly have on for each part of ui package

bmuenzenmeyer (approved):

aymen94 (approved):

PR #7763, opened 2025-05-20
learn(promises): update `fs/promises` example

groupsvkg:
## Description

Updated `[Discover Promises in Node.js](https://nodejs.org/en/learn/asynchronous-work/discover-promises-in-nodejs#discover-promises-in-nodejs)` page

### Check List

### DIFF
--- a/apps/site/pages/en/learn/asynchronous-work/discover-promises-in-nodejs.md
+++ b/apps/site/pages/en/learn/asynchronous-work/discover-promises-in-nodejs.md
@@ -159,13 +159,13 @@ Node.js provides **Promise-based versions** of many of its core APIs, especially
 For example, the `fs` (file system) module has a Promise-based API under `fs.promises`:
 
 ```js
-const fs = require('node:fs');
+const fs = require('node:fs').promises;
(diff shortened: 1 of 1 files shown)


avivkeller (changes requested):
Why is this better? It's the exact same thing, except the previous example worked interchangeably with the code comment, and this would break if you replace the imported version with the promise version.

groupsvkg:
> Why is this better? It's the exact same thing

its better because the text below the code snippet says `fs.promises.readFile()`.

image

AugustinMauroy (approved):
IMO it's cool and more readable

avivkeller:
I think we should update the text to match the example. This example is purposely formatted in such a way where that time can be replaced with an `fs/promises` import and be used the same

groupsvkg:
> I think we should update the text to match the example. This example is purposely formatted in such a way where that time can be replaced with an `fs/promises` import and be used the same

Yes it makes sense.

avivkeller (approved):

aymen94 (approved):

PR #7771, opened 2025-05-21
chore(ui): don't depend on pnpm syntax

avivkeller:
When we import `@node-core/ui-components` from `api-docs-tooling`, it can't contain non-npm syntax in it's `package.json`. Therefore, this PR removes the dependency on the `workspace:*` syntax.

### DIFF
--- a/packages/ui-components/package.json
+++ b/packages/ui-components/package.json
@@ -48,7 +48,6 @@
     "tailwindcss": "~4.0.17"
   },
   "devDependencies": {
-    "@node-core/website-i18n": "workspace:*",
     "@storybook/addon-controls": "^8.6.12",
     "@storybook/addon-interactions": "^8.6.12",
     "@storybook/addon-styling-webpack": "^1.0.1",
--- a/packages/ui-components/types.ts
+++ b/packages/ui-components/types.ts
@@ -1,4 +1,3 @@
-import type { LocaleConfig } from '@node-core/website-i18n/types';
 import type {
   SVGProps,
   AnchorHTMLAttributes,
@@ -19,7 +18,8 @@ export type FormattedMessage =
   | ReactElement<HTMLElement, string | JSXElementConstructor<HTMLElement>>
   | ReadonlyArray<ReactNode>;
 
-export type SimpleLocaleConfig = Pick<
-  LocaleConfig,
-  'name' | 'code' | 'localName'
->;
+export type SimpleLocaleConfig = {
+  code: string;
+  localName: string;
+  name: string;
+};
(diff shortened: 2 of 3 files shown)


AugustinMauroy (approved):

bmuenzenmeyer (approved):
sometimes DRYness is the enemy

canerakdas (approved):

dario-piotrowicz:
@avivkeller sorry for the silly question, but could you clarify how `api-docs-tooling` imports the package? 🙏

avivkeller:
It's not a silly question, we are still trying to figure that out in https://github.com/nodejs/api-docs-tooling/issues/236. Currently, in https://github.com/nodejs/api-docs-tooling/pull/273, I'm using [a package I made for us](https://github.com/avivkeller/git-deps) to [import `@node-core/rehype-shiki`](https://github.com/nodejs/api-docs-tooling/pull/273/files#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R73-R74)

aymen94 (approved):

dario-piotrowicz:
> It's not a silly question, we are still trying to figure that out in [nodejs/api-docs-tooling#236](https://github.com/nodejs/api-docs-tooling/issues/236). Currently, in [nodejs/api-docs-tooling#273](https://github.com/nodejs/api-docs-tooling/pull/273), I'm using [a package I made for us](https://github.com/avivkeller/git-deps) to [import `@node-core/rehype-shiki`](https://github.com/nodejs/api-docs-tooling/pull/273/files#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R73-R74)

I see, thanks a lot Aviv 🙂

Personally I would consider the option to publish the package as @AugustinMauroy [suggested](https://github.com/nodejs/api-docs-tooling/issues/236#issuecomment-2748364665).

Having to remove pnpm `workspace:*` dependencies from the package.json seems like a hack to me 🤔
(besides the extra complexity and maintenance burden, also as far as I can tell if non-permalink repository links are used there isn't currently any check to accidentally prevent the re-introduction of `workspace:*` dependencies and similar ones that would break dependent packages)

So all things considered I think I am not a huge fan of the approach 🤔, however I can live with it if everyone else is happy with it 🙂  (also considering that you're already using it for rehype-shiki)

avivkeller:
I recommend you raise your concerns in the dedicated issue.


This PR is intended to work for npm installations of this package, which was one of the goals we had in mind when migrating to the monorepo. That's always something we can change, however, I think that's a discussion for the other issue.

dario-piotrowicz:
> I recommend you raise your concerns in the dedicated issue.
> 
> This PR is intended to work for npm installations of this package, which was one of the goals we had in mind when migrating to the monorepo. That's always something we can change, however, I think that's a discussion for the other issue.

I see, sorry for the extra noise then :sweat:

PR #7781, opened 2025-05-26
chore(i18n): improve type safety

avivkeller:
## Description

This PR updates our types to better reflect i18n. For example, `t("...")` will now only accept valid i18n keys.

### Check List

### DIFF
--- a/apps/site/components/Downloads/Release/ReleaseCodeBox.tsx
+++ b/apps/site/components/Downloads/Release/ReleaseCodeBox.tsx
@@ -15,7 +15,6 @@ import {
   ReleaseContext,
   ReleasesContext,
 } from '#site/providers/releaseProvider';
-import type { IntlMessageKeys } from '#site/types/i18n';
 import type { ReleaseContextType } from '#site/types/release';
 import { INSTALL_METHODS } from '#site/util/downloadUtils';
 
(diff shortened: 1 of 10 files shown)


avivkeller:
image

apps/site/components/Downloads/Release/ReleaseCodeBox.tsx
```diff
@@ -144,7 +145,7 @@ const ReleaseCodeBox: FC = () => {
 
       <span className="text-center text-xs text-neutral-800 dark:text-neutral-200">
         <Skeleton loading={renderSkeleton} hide={!currentPlatform}>
-          {t(info, { platform: label })}{' '}
+          {t(info as IntlMessageKeys, { platform: label })}{' '}
```
ovflowd: Why are we doing casting all over the files?

AugustinMauroy:
aviv could you apply thing such as proposed here it's look little be simpler 
https://next-intl.dev/docs/workflows/typescript

avivkeller:
> aviv could you apply thing such as proposed here it's look little be simpler 
> 
> https://next-intl.dev/docs/workflows/typescript

That's what I did, see the `global.d.ts` file

AugustinMauroy:
> That's what I did, see the global.d.ts file

But I didn't get why you need a `as IntlMessageKeys`

avivkeller:
We cast in the download utilities because they are from JSON, so the JSON strings must be casted to our type.

apps/site/components/Downloads/Release/ReleaseCodeBox.tsx
avivkeller: See https://github.com/nodejs/nodejs.org/pull/7781#issuecomment-2912259694 + 27f26312

AugustinMauroy (approved):

PR #7783, opened 2025-05-27
chore(ci): only run when certain files are changed

avivkeller:
Using the push event helps prevent this workflow from running unnecessarily. It won’t change the publishing frequency, but ensures the workflow only runs when there’s actually something to publish: rather than empty matrix generation and related errors.

### DIFF
--- a/.github/workflows/publish-packages.yml
+++ b/.github/workflows/publish-packages.yml
@@ -1,13 +1,11 @@
 name: Publish Packages
 
-# This workflow publishes packages to NPM when changes are merged to main branch or when manually triggered.
-# It runs automatically after successful tests or can be run manually for specific packages.
+# This workflow publishes packages to npm when changes are merged to main branch or when manually triggered.
 
 on:
-  workflow_run:
-    # Only run after linting and tests have passed on main branch
-    workflows: ['Linting and Tests']
-    types: [completed]
+  push:
+    paths:
+      - 'packages/**'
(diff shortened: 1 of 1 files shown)


.github/workflows/publish-packages.yml
```diff
@@ -1,13 +1,11 @@
 name: Publish Packages
 
-# This workflow publishes packages to NPM when changes are merged to main branch or when manually triggered.
-# It runs automatically after successful tests or can be run manually for specific packages.
+# This workflow publishes packages to npm when changes are merged to main branch or when manually triggered.
 
 on:
-  workflow_run:
-    # Only run after linting and tests have passed on main branch
-    workflows: ['Linting and Tests']
```
AugustinMauroy, line 9: now if test fails is it publishing ?

avivkeller: No, the tests still run on `merge_queue`, which the commits _have_ to use in order to publish. Verifying them here was a bit redudant actually.

AugustinMauroy: Okay got it maybe add comment

avivkeller: There's a comment explaining that merge queue is enforced. IMO that implies that merge queue check are as well.

AugustinMauroy (approved):
LGMT !

avivkeller:
@nodejs/web-infra PTAL

ovflowd (approved):

PR #7797, opened 2025-05-31
feat(ui): add `DataTag`

avivkeller:
This is the `CircularIcon` component, based on the design shown below:

Design reference

Storybook preview:

Storybook preview

### DIFF
--- a/packages/ui-components/Common/CircularIcon/index.module.css
+++ b/packages/ui-components/Common/CircularIcon/index.module.css
@@ -9,19 +9,16 @@
     text-white;
 
   &.lg {
-    @apply h-12
-      w-12
+    @apply size-12
       text-2xl;
   }
 
   &.md {
-    @apply h-10
-      w-10
+    @apply size-10
       text-xl;
   }
 
   &.sm {
-    @apply h-8
-      w-8;
+    @apply size-8
   }
 }
--- a/packages/ui-components/Common/CircularIcon/index.module.css
+++ b/packages/ui-components/Common/CircularIcon/index.module.css
@@ -1,0 +1,27 @@
+@reference "../../styles/index.css";
+
+.icon {
+  @apply flex
+    items-center
+    justify-center
+    rounded-full
+    font-semibold
+    text-white;
+
+  &.lg {
+    @apply h-12
+      w-12
+      text-2xl;
+  }
+
+  &.md {
+    @apply h-10
+      w-10
+      text-xl;
+  }
+
+  &.sm {
+    @apply h-8
+      w-8;
+  }
+}
--- a/packages/ui-components/Common/CircularIcon/index.module.css
+++ b/packages/ui-components/Common/CircularIcon/index.module.css
@@ -19,6 +19,34 @@
   }
 
   &.sm {
-    @apply size-8
+    @apply size-8;
+  }
+
+  &.event {
+    @apply bg-purple-600;
+  }
+
+  &.method {
+    @apply bg-info-600;
+  }
+
+  &.property {
+    @apply bg-green-600;
+  }
+
+  &.class {
+    @apply bg-warning-600;
+  }
+
+  &.module {
+    @apply bg-red-600;
+  }
+
+  &.classMethod {
+    @apply bg-blue-600;
+  }
+
+  &.ctor {
+    @apply bg-pink-600;
   }
 }
--- a/packages/ui-components/Common/CircularIcon/index.stories.tsx
+++ b/packages/ui-components/Common/CircularIcon/index.stories.tsx
@@ -1,0 +1,18 @@
+import type { Meta as MetaObj, StoryObj } from '@storybook/react';
+
+import CircularIcon from '#ui/Common/CircularIcon';
+
+type Story = StoryObj<typeof CircularIcon>;
+type Meta = MetaObj<typeof CircularIcon>;
+
+export const Icons: Story = {
+  render: () => (
+    <div style={{ display: 'flex', gap: '16px', alignItems: 'center' }}>
+      <CircularIcon symbol="B" color="#3b82f6" size="sm" />
+      <CircularIcon symbol="G" color="#10b981" size="md" />
+      <CircularIcon symbol="R" color="#ef4444" size="lg" />
+    </div>
+  ),
+};
+
+export default { component: CircularIcon } as Meta;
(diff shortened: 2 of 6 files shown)


packages/ui-components/Common/CircularIcon/index.stories.tsx
```diff
@@ -0,0 +1,18 @@
+import type { Meta as MetaObj, StoryObj } from '@storybook/react';
+
+import CircularIcon from '#ui/Common/CircularIcon';
+
+type Story = StoryObj<typeof CircularIcon>;
+type Meta = MetaObj<typeof CircularIcon>;
+
+export const Icons: Story = {
+  render: () => (
+    <div style={{ display: 'flex', gap: '16px', alignItems: 'center' }}>
```
AugustinMauroy: This should be handled by tailwind like the rest of your stories

packages/ui-components/Common/CircularIcon/index.tsx
```diff
@@ -0,0 +1,24 @@
+import classNames from 'classnames';
+
+import styles from './index.module.css';
+
+interface CircularIconProps {
+  symbol: string;
+  color: string;
```
AugustinMauroy: What if we have something like kind: "sting" | "object" ....

avivkeller: WDYM? What would the object be of?

packages/ui-components/Common/CircularIcon/index.module.css
```diff
@@ -0,0 +1,27 @@
...
+  @apply flex
+    items-center
+    justify-center
+    rounded-full
+    font-semibold
+    text-white;
+
+  &.lg {
+    @apply h-12
+      w-12
```
canerakdas:
```suggestion
    @apply size-12
```

packages/ui-components/Common/CircularIcon/index.module.css
```diff
@@ -0,0 +1,27 @@
...
+      text-2xl;
+  }
+
+  &.md {
+    @apply h-10
+      w-10
```
canerakdas:
```suggestion
    @apply size-10
```

packages/ui-components/Common/CircularIcon/index.module.css
```diff
@@ -0,0 +1,27 @@
...
+      text-xl;
+  }
+
+  &.sm {
+    @apply h-8
+      w-8;
```
canerakdas:
```suggestion
    @apply size-8
```

packages/ui-components/Common/CircularIcon/index.tsx
canerakdas: I think we should not pass the color code as props, I am in favor of adding the possibilities here as kind and defining the color set here. For example, you can look at the badge component;

https://github.com/nodejs/nodejs.org/blob/main/packages/ui-components/Common/Badge/index.tsx#L6

https://github.com/nodejs/nodejs.org/blob/main/packages/ui-components/Common/Badge/index.module.css#L21-L37

This allows us to choose a background with the right contrast ratio with the background colors and not go beyond our color system

AugustinMauroy: exactly !

avivkeller: ahh got it.

packages/ui-components/Common/CircularIcon/index.tsx
```diff
@@ -0,0 +1,33 @@
...
+
+export type CircularIconProps = {
+  kind:
+    | 'event'
+    | 'method'
+    | 'property'
+    | 'class'
+    | 'module'
+    | 'classMethod'
+    | 'ctor';
```
avivkeller: These kinds link up directly with the types in `HeadingMetadataEntry` of `api-docs-tooling`.

AugustinMauroy: maybe add jsdoc for that

packages/ui-components/Common/CircularIcon/index.module.css
```diff
@@ -0,0 +1,52 @@
...
+    @apply size-10
+      text-xl;
+  }
+
+  &.sm {
+    @apply size-8;
+  }
+
+  &.event {
+    @apply bg-purple-600;
```
AugustinMauroy:
```suggestion
    @apply bg-accent1-600;
```

packages/ui-components/Common/CircularIcon/index.module.css
```diff
@@ -0,0 +1,52 @@
...
+  &.module {
+    @apply bg-red-600;
+  }
+
+  &.classMethod {
+    @apply bg-blue-600;
+  }
+
+  &.ctor {
+    @apply bg-pink-600;
```
AugustinMauroy:
```suggestion
    @apply bg-accent2-600;
```

packages/ui-components/Common/CircularIcon/index.tsx
AugustinMauroy, line 1: call it dataTag (Figma name and I like it)

packages/ui-components/Common/CircularIcon/index.tsx
```diff
@@ -0,0 +1,33 @@
...
+  property: 'P',
+  class: 'C',
+  module: 'M',
+  classMethod: 'S',
+  ctor: 'C',
+} as const;
+
+export default function CircularIcon({ kind, size = 'md' }: CircularIconProps) {
+  return (
+    <div className={classNames(styles.icon, styles[size], styles[kind])}>
```
AugustinMauroy:
```suggestion
    <div className={classNames(styles.icon, styles[size], styles[kind])} title={kind}>
```

it's will allow the user to know what doesn't it mean by hovering

AugustinMauroy (approved):
LGTM ! That's clean

packages/ui-components/Common/DataTag/index.tsx
```diff
@@ -0,0 +1,36 @@
...
+  ctor: 'C',
+} as const;
+
+export default function DataTag({ kind, size = 'md' }: DataTagProps) {
+  return (
+    <div className={classNames(styles.dataTag, styles[size], styles[kind])}>
+      <span>{symbolMap[kind]}</span>
+    </div>
+  );
+}
```
canerakdas:
```tsx
const DataTag: FC<DataTagProps> = ({ kind, size = 'md' }) => (
  <div className={classNames(styles.dataTag, styles[size], styles[kind])}>
    <span>{symbolMap[kind]}</span>
  </div>
);

export default DataTag;
```
As a [best practice](https://github.com/nodejs/nodejs.org/blob/main/COLLABORATOR_GUIDE.md#how-a-new-component-should-look-like-when-freshly-created) you can use arrow function, also no need to return

avivkeller:
Fast-tracking so I can add this to the web generator. If you approve, merge when ready.

ovflowd (approved):

PR #7827, opened 2025-06-02
fix(icr): workflow changes

avivkeller:
Per the discussions

### DIFF
--- a/.github/scripts/report-inactive-collaborators.mjs
+++ b/.github/scripts/report-inactive-collaborators.mjs
@@ -15,6 +15,20 @@ const getDateMonthsAgo = (months = CONFIG.INACTIVE_MONTHS) => {
   return date.toISOString().split('T')[0];
 };
 
+// Check if there's already an open issue
+async function hasOpenIssue(github, context) {
+  const { owner, repo } = context.repo;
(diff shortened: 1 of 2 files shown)


bmuenzenmeyer (approved):

AugustinMauroy (approved):
LGMT !

ovflowd (approved):

Harkunwar (approved):

bjohansebas (approved):

mikeesto (commented):
Just thinking about @aymen94 's https://github.com/nodejs/nodejs.org/issues/7826#issuecomment-2930439558 . This addition looks like it is checking whether someone has authored an issue or PR, but not whether they have reviewed. It might be nice to capture this too. Actually, I would say it's more useful to capture this engagement instead of whether they have opened an issue

It could be done by forming a query like this:

```js
const query = `type:pr repo:nodejs/nodejs.org reviewed-by:${username} updated:>=${cutoffDate}`;
 
 const { data: searchResult } = await github.request("GET /search/issues", {
      q: query,
      per_page: 1,
      advanced_search: true,
    });
```

mikeesto (approved):
LGTM, thanks

PR #7794, opened 2025-05-31
feat: Introduce Downloads Archive page

canerakdas:
## Description

This PR introduces a new statically generated page that displays simplified download options by major version:

image

Also includes a download table of other minor versions of the major version;
image

Since I included the minor versions as a table, the content became quite lengthy. To prevent this, I used the HTML `details` element. The main reason I didn't use Radix Primitive Accordion is that we want this page to be fully usable even for users who have JavaScript disabled

I am open to suggestions for to add / remove content 🙇 

## Validation

The links below are accessible in the preview;

- /download/{major}
- /download/archive

## Related Issues

Addresses [#7443](https://github.com/nodejs/nodejs.org/issues/7443)

### Check List

- [x] I have read the [Contributing Guidelines](https://github.com/nodejs/nodejs.org/blob/main/CONTRIBUTING.md) and made commit messages that follow the guideline.
- [x] I have run `pnpm format` to ensure the code follows the style guide.
- [x] I have run `pnpm test` to check if all tests are passing.
- [x] I have run `pnpm build` to check if the website builds without errors.
- [ ] I've covered new added functionality with unit tests if necessary.

PR #7851, opened 2025-06-08
chore: enforce consistent object-shorthand

canerakdas:
## Description

We intend to add this as a rule based on the [feedback received during code reviews](https://github.com/nodejs/nodejs.org/pull/7794/files/a92ad35e369621142dfdb837ee0e1aa6c053374f#r2117785133), so it applies consistently across the website for all contributors and helps maintain the code style.

See; (https://eslint.org/docs/latest/rules/object-shorthand)

Additionally, cleaned up the `importX` imports that were given warnings in the project


AugustinMauroy (approved):

eslint.config.js
```diff
@@ -44,6 +44,7 @@ export default [
           },
         },
       ],
+      'object-shorthand': ['error', 'consistent-as-needed'],
```
araujogui, line 48: I don't know if `consistent-as-needed` is needed, I don't see a problem mixing shorthand and non-shorthand properties.

canerakdas: I think it's better to define the code style clearly from the beginning for future contributors. If mixing shorthand properties isn't an issue for everyone, it's fine by me too

bmuenzenmeyer (approved):
I'm in favor of this change, overall. just needs rebasing

aymen94 (approved):

Issue #7867, opened 2025-06-15
Improve username filtering regex in profiling example for better input handling


Dongnyoung:
### Enter your suggestions in details:

Hello,

I understand that this is a learning example for profiling,
but the current username special character filtering is quite limited.
Even for educational purposes, I believe it would be better to use a broader filtering pattern.

In this file:
[nodejs.org/apps/site/pages/en/learn/getting-started/profiling.md](url)

The relevant code snippet is:

```js
  app.get('/newUser', (req, res) => {
  let username = req.query.username || '';
  const password = req.query.password || '';

  username = username.replace(/[!@#$%^&*]/g, ''); //<-- this part

  if (!username || !password || users[username]) {
    return res.sendStatus(400);
  }

  const salt = crypto.randomBytes(128).toString('base64');
  const hash = crypto.pbkdf2Sync(password, salt, 10000, 512, 'sha512');

  users[username] = { salt, hash };

  res.sendStatus(200);
});
```
I suggest changing the filtering line to:
```js
username = username.replace(/[^a-zA-Z0-9]/g, '');
```

to allow only alphanumeric characters would make the input handling cleaner
and serve as a better example for learners.

I’d appreciate your thoughts on this suggestion!

PR #7868, opened 2025-06-15
feat: improve username sanitization in profiling example

Dongnyoung:
Hello,

This commit updates the username sanitization logic in the profiling
example code to allow only alphanumeric characters. The original
pattern (/[!@#$%^&*]/g) was limited and could lead to inconsistent
behavior depending on input.

Changed:
- From: username = username.replace(/[!@#$%^&*]/g, '');
- To:   username = username.replace(/[^a-zA-Z0-9]/g, '');

This change makes the input handling cleaner and more appropriate
for educational purposes, aligning better with common sanitization
practices.

Note: This change was initially merged into my forked repository by mistake during development. This PR now reflects the correct version intended for the original repository.

Relates to: #7867

Thank you for reviewing.

### DIFF
--- a/apps/site/pages/en/learn/getting-started/profiling.md
+++ b/apps/site/pages/en/learn/getting-started/profiling.md
@@ -41,7 +41,7 @@ app.get('/newUser', (req, res) => {
   let username = req.query.username || '';
   const password = req.query.password || '';
 
-  username = username.replace(/[!@#$%^&*]/g, '');
+  username = username.replace(/[^a-zA-Z0-9]/g, '');
 
   if (!username || !password || users[username]) {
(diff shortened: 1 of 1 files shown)


AugustinMauroy (approved):
In theory it's good. But I don't know what is the value behind this example

Dongnyoung:
Thanks for the feedback!

I suggested this change to make the example more aligned with common input sanitization practices. The original regex was quite limited, so I thought a more realistic pattern might help learners better understand typical use cases.

Happy to adjust further if needed!

avivkeller (approved):

Harkunwar (approved):

PR #7869, opened 2025-06-15
feat(ui-components): add `ChangeHistory`

Harkunwar:
## Description

Adding a stateless version of ChangeHistory component which can be potentially be used without JavaScript enabled.
image

## Validation

Tested by rendering a few different versions in Storybook

## Related Issues

Related to https://github.com/nodejs/api-docs-tooling/issues/10#issuecomment-2914209117

### Check List

### DIFF
--- a/packages/ui-components/Common/ChangeHistory/index.module.css
+++ b/packages/ui-components/Common/ChangeHistory/index.module.css
@@ -43,6 +43,12 @@
     dark:bg-neutral-950;
 }
 
+.dropdownContentInner {
+  @apply max-h-80
+    w-52
+    overflow-y-auto;
+}
+
 .dropdownItem {
   @apply outline-hidden
     block
@@ -60,4 +66,18 @@
     @apply bg-green-600
       text-white;
   }
+}
+
+.dropdownLabel {
+  @apply block
+    text-sm
+    font-medium
+    leading-tight;
+}
+
+.dropdownVersions {
+  @apply block
+    text-xs
+    leading-tight
+    opacity-75;
 }
--- a/packages/ui-components/Common/ChangeHistory/index.module.css
+++ b/packages/ui-components/Common/ChangeHistory/index.module.css
@@ -1,10 +1,5 @@
 @reference "../../styles/index.css";
 
-.changeHistory {
-  @apply relative
-    inline-block;
-}
-
 .summary {
   @apply outline-hidden
     flex
@@ -28,33 +23,11 @@
     @apply bg-neutral-100
       dark:bg-neutral-900;
   }
-
-  &.disabled {
-    @apply cursor-not-allowed
-      opacity-50;
-
-    &:hover,
-    &:focus-visible {
-      @apply bg-transparent
-        dark:bg-transparent;
-    }
-  }
 }
 
-.icon {
-  @apply h-4
-    w-4;
-}
-
-.chevron {
-  @apply h-3
-    w-3
-    group-open:rotate-180
-    motion-safe:transition-transform;
-}
-
-.dropdownContent {
+.dropdownContentWrapper {
   @apply absolute
+    right-0
     top-full
     z-50
     mt-1
@@ -68,20 +41,6 @@
     shadow-lg
     dark:border-neutral-900
     dark:bg-neutral-950;
-}
-
-.dropdownContent.alignLeft {
-  @apply left-0;
-}
-
-.dropdownContent.alignRight {
-  @apply right-0;
-}
-
-.dropdownContentInner {
-  @apply max-h-80
-    w-52
-    overflow-y-auto;
 }
 
 .dropdownItem {
@@ -101,18 +60,4 @@
     @apply bg-green-600
       text-white;
   }
-}
-
-.dropdownLabel {
-  @apply block
-    text-sm
-    font-medium
-    leading-tight;
-}
-
-.dropdownVersions {
-  @apply block
-    text-xs
-    leading-tight
-    opacity-75;
 }
--- a/packages/ui-components/Common/ChangeHistory/index.module.css
+++ b/packages/ui-components/Common/ChangeHistory/index.module.css
@@ -1,0 +1,128 @@
+@reference "../../styles/index.css";
+
+.changeHistory {
+  @apply relative
+    inline-block;
+}
+
+.summary {
+  @apply outline-hidden
+    flex
+    h-9
+    cursor-pointer
+    select-none
+    items-center
+    gap-2
+    rounded-md
+    border
+    border-neutral-200
+    p-2
+    text-sm
+    text-neutral-700
+    motion-safe:transition-colors
+    dark:border-neutral-900
+    dark:text-neutral-300;
+
+  &:hover,
+  &:focus-visible {
+    @apply bg-neutral-100
+      dark:bg-neutral-900;
+  }
+
+  &.disabled {
+    @apply cursor-not-allowed
+      opacity-50;
+
+    &:hover,
+    &:focus-visible {
+      @apply bg-transparent
+        dark:bg-transparent;
+    }
+  }
+}
+
+.icon {
+  @apply h-4
+    w-4;
+}
+
+.chevron {
+  @apply h-3
+    w-3
+    group-open:rotate-180
+    motion-safe:transition-transform;
+}
+
+.dropdownContent {
+  @apply absolute
+    top-full
+    z-50
+    mt-1
+    max-h-80
+    w-52
+    overflow-hidden
+    rounded-sm
+    border
+    border-neutral-200
+    bg-white
+    shadow-lg
+    dark:border-neutral-900
+    dark:bg-neutral-950;
+}
+
+.dropdownContent.alignLeft {
+  @apply left-0;
+}
+
+.dropdownContent.alignRight {
+  @apply right-0;
+}
+
+.dropdownContentInner {
+  @apply max-h-80
+    w-52
+    overflow-y-auto;
+}
+
+.dropdownItem {
+  @apply outline-hidden
+    block
+    px-2.5
+    py-1.5
+    text-sm
+    font-medium
+    text-neutral-800
+    no-underline
+    motion-safe:transition-colors
+    dark:text-white;
+
+  &:hover,
+  &:focus-visible {
+    @apply bg-green-600
+      text-white;
+  }
+}
+
+.dropdownLabel {
+  @apply block
+    text-sm
+    font-medium
+    leading-tight;
+}
+
+.dropdownVersions {
+  @apply block
+    text-xs
+    leading-tight
+    opacity-75;
+}
+
+.emptyState {
+  @apply px-2.5
+    py-1.5
+    text-sm
+    font-medium
+    text-neutral-800
+    opacity-75
+    dark:text-white;
+}
--- a/packages/ui-components/Common/ChangeHistory/index.module.css
+++ b/packages/ui-components/Common/ChangeHistory/index.module.css
@@ -116,13 +116,3 @@
     leading-tight
     opacity-75;
 }
-
-.emptyState {
-  @apply px-2.5
-    py-1.5
-    text-sm
-    font-medium
-    text-neutral-800
-    opacity-75
-    dark:text-white;
-}
--- a/packages/ui-components/Common/ChangeHistory/index.tsx
+++ b/packages/ui-components/Common/ChangeHistory/index.tsx
@@ -24,10 +24,10 @@ const ChangeHistory: FC<ChangeHistoryProps> = ({
   <div className={`${styles.changeHistory} ${className}`} {...props}>
     {changes.length > 0 ? (
       <details className="group">
-        <summary className={styles.summary} role="button" aria-haspopup="true">
-          <ClockIcon className={styles.icon} aria-hidden="true" />
+        <summary className={styles.summary} role="button" aria-haspopup="menu">
+          <ClockIcon className={styles.icon} />
           <span>{label}</span>
-          <ChevronDownIcon className={styles.chevron} aria-hidden="true" />
+          <ChevronDownIcon className={styles.chevron} />
         </summary>
         <div
           className={`${styles.dropdownContent} ${
@@ -75,7 +75,7 @@ const ChangeHistory: FC<ChangeHistoryProps> = ({
         aria-disabled="true"
       >
         <ClockIcon className={styles.icon} aria-hidden="true" />
-        <span>History</span>
+        <span>{label}</span>
         <ChevronDownIcon className={styles.chevron} aria-hidden="true" />
       </div>
     )}
--- a/packages/ui-components/Common/ChangeHistory/index.tsx
+++ b/packages/ui-components/Common/ChangeHistory/index.tsx
@@ -1,0 +1,80 @@
+import { ChevronDownIcon, ClockIcon } from '@heroicons/react/24/outline';
+import type { FC } from 'react';
+
+import styles from './index.module.css';
+
+type ChangeHistoryProps = React.JSX.IntrinsicElements['div'] & {
+  label: string;
+  changes: Array<{
+    versions: Array<string>;
+    label: string;
+    url?: string;
+  }>;
+  align?: 'left' | 'right';
+};
+
+const ChangeHistory: FC<ChangeHistoryProps> = ({
+  label = 'History',
+  changes = [],
+  align = 'right',
+  className,
+  'aria-label': ariaLabel = label,
+  ...props
+}) => (
+  <div className={`${styles.changeHistory} ${className}`} {...props}>
+    {changes.length > 0 ? (
+      <details className="group">
+        <summary className={styles.summary} role="button" aria-haspopup="true">
+          <ClockIcon className={styles.icon} aria-hidden="true" />
+          <span>{label}</span>
+          <ChevronDownIcon className={styles.chevron} aria-hidden="true" />
+        </summary>
+        <div
+          className={`${styles.dropdownContent} ${
+            align === 'left' ? styles.alignLeft : styles.alignRight
+          }`}
+          role="menu"
+          aria-label={ariaLabel}
+        >
+          <div className={styles.dropdownContentInner}>
+            {changes.map((change, index) => {
+              const content = (
+                <>
+                  <div className={styles.dropdownLabel}>{change.label}</div>
+                  <div className={styles.dropdownVersions}>
+                    {change.versions.join(', ')}
+                  </div>
+                </>
+              );
+
+              return (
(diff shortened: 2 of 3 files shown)


packages/ui-components/Common/ChangeHistory/index.tsx
```diff
@@ -0,0 +1,85 @@
...
+        role="button"
+        aria-disabled="true"
+      >
+        <ClockIcon className={styles.icon} aria-hidden="true" />
+        <span>{label}</span>
+        <ChevronDownIcon className={styles.chevron} aria-hidden="true" />
+      </div>
+    )}
+  </div>
+);
```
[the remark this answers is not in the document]
Harkunwar: Should we try to implement this if we want our components to work without Javascript. This above change only affects aria labels so it shouldn't block functionality any way.

packages/ui-components/Common/ChangeHistory/index.tsx
```diff
@@ -0,0 +1,85 @@
...
+                  <div className={styles.dropdownVersions}>
+                    {change.versions.join(', ')}
+                  </div>
+                </>
+              );
+
+              const ariaLabel = `${change.label}: ${change.versions.join(', ')}`;
+
+              const itemProps = {
+                key: ariaLabel,
```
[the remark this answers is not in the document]
Harkunwar: I believe the above combination should be unique enough, does anyone disagree?

packages/ui-components/Common/ChangeHistory/index.tsx
```diff
@@ -0,0 +1,85 @@
...
+
+const ChangeHistory: FC<ChangeHistoryProps> = ({
+  label = 'History',
+  changes = [],
+  align = 'right',
+  className,
+  'aria-label': ariaLabel = label,
+  ...props
+}) => (
+  <div className={`${styles.changeHistory} ${className ?? ''}`} {...props}>
```
avivkeller: Use `classNames`

packages/ui-components/Common/ChangeHistory/index.tsx
```diff
@@ -0,0 +1,85 @@
...
+                ['aria-label']: ariaLabel,
+              };
+
+              return change.url ? (
+                <a {...itemProps} href={change.url}>
+                  {content}
+                </a>
+              ) : (
+                <div {...itemProps}>{content}</div>
+              );
```
avivkeller: This should use an "As" prop for the component, see other components for how that works.

packages/ui-components/Common/ChangeHistory/index.tsx
```diff
@@ -0,0 +1,85 @@
...
+      <details className="group">
+        <summary className={styles.summary} role="button" aria-haspopup="menu">
+          <ClockIcon className={styles.icon} />
+          <span>{label}</span>
+          <ChevronDownIcon className={styles.chevron} />
+        </summary>
+        <div
+          className={`${styles.dropdownContent} ${
+            align === 'left' ? styles.alignLeft : styles.alignRight
+          }`}
```
avivkeller: Ditto

packages/ui-components/Common/ChangeHistory/index.tsx
avivkeller: The key can just be the index, right?

packages/ui-components/Common/ChangeHistory/index.tsx
```diff
@@ -0,0 +1,85 @@
...
+const ChangeHistory: FC<ChangeHistoryProps> = ({
+  label = 'History',
+  changes = [],
+  align = 'right',
+  className,
+  'aria-label': ariaLabel = label,
+  ...props
+}) => (
+  <div className={`${styles.changeHistory} ${className ?? ''}`} {...props}>
+    {changes.length > 0 ? (
```
avivkeller: IMO we don't need a scenario where there are no changes. In that case, the element wouldn't be there as all.

packages/ui-components/Common/ChangeHistory/index.tsx
```diff
@@ -0,0 +1,85 @@
...
+
+import styles from './index.module.css';
+
+type ChangeHistoryProps = React.JSX.IntrinsicElements['div'] & {
+  label: string;
+  changes: Array<{
+    versions: Array<string>;
+    label: string;
+    url?: string;
+  }>;
```
avivkeller: Can you export a Change type, so we can import the type into the tooling?

packages/ui-components/Common/ChangeHistory/index.tsx
```diff
@@ -0,0 +1,85 @@
...
+  align?: 'left' | 'right';
```
avivkeller: Do we have a use case for both alignments?

packages/ui-components/Common/ChangeHistory/index.module.css
```diff
@@ -0,0 +1,118 @@
...
+
+.dropdownContent.alignRight {
+  @apply right-0;
+}
+
+.dropdownContentInner {
+  @apply max-h-80
+    w-52
+    overflow-y-auto;
+}
```
avivkeller: Let's use nesting
+ we are only applying a single class, can't we just use tailwind in the JSX

packages/ui-components/Common/ChangeHistory/index.tsx
```diff
@@ -0,0 +1,85 @@
+import { ChevronDownIcon, ClockIcon } from '@heroicons/react/24/outline';
+import type { FC } from 'react';
+
+import styles from './index.module.css';
+
+type ChangeHistoryProps = React.JSX.IntrinsicElements['div'] & {
```
avivkeller: Let's import the type directly

packages/ui-components/Common/ChangeHistory/index.tsx
```diff
@@ -0,0 +1,85 @@
...
+  }>;
+  align?: 'left' | 'right';
+};
+
+const ChangeHistory: FC<ChangeHistoryProps> = ({
+  label = 'History',
+  changes = [],
+  align = 'right',
+  className,
+  'aria-label': ariaLabel = label,
```
avivkeller, line 26: Will these ever need to be different than the label?

Harkunwar: I was wondering in case you want to say "Change History" in the aria label

avivkeller: Can't we already do that via the {...props} to the element?

packages/ui-components/Common/ChangeHistory/index.tsx
Harkunwar: I used an index before, copilot complained. I'll switch it back to index.

packages/ui-components/Common/ChangeHistory/index.tsx
Harkunwar: Just added it if we needed to use them in different locations.

packages/ui-components/Common/ChangeHistory/index.tsx
Harkunwar: I'll fix this

packages/ui-components/Common/ChangeHistory/index.tsx
avivkeller: Copilot _loves_ to make bogus complaints

packages/ui-components/Common/ChangeHistory/index.tsx
avivkeller: I feel like the alignment can be inherited from the parent, if that's possible? Otherwise, as is works.

packages/ui-components/Common/ChangeHistory/index.tsx
Harkunwar: Yep, but it's being applied to `dropdownContentWrapper` so it's not where `{ ...props }` are being used

packages/ui-components/Common/ChangeHistory/index.module.css
Harkunwar: I'll move some of these to JSX

packages/ui-components/Common/ChangeHistory/index.tsx
Harkunwar: Sure, that makes sense. I'll see if I can get that to work

packages/ui-components/Common/ChangeHistory/index.tsx
Harkunwar: Yep, I'll export the type

packages/ui-components/Common/ChangeHistory/index.tsx
Harkunwar: Okay, I'll just return `null` when changes.length is 0 then

packages/ui-components/Common/ChangeHistory/index.tsx
Harkunwar: I'll use them

packages/ui-components/Common/ChangeHistory/index.tsx
Harkunwar: Opted for `const MenuItem = change.url ? 'a' : 'div';` instead.

packages/ui-components/Common/ChangeHistory/index.tsx
Harkunwar: Just defaulted it to right at the moment. This alignment isn't for text but if the dropdown container should grow towards left of the button or the right

packages/ui-components/Common/ChangeHistory/index.tsx
Harkunwar: Fixed

packages/ui-components/Common/ChangeHistory/index.tsx
Harkunwar: Updated

packages/ui-components/Common/ChangeHistory/index.tsx
```diff
@@ -0,0 +1,78 @@
...
+const ChangeHistory: FC<ChangeHistoryProps> = ({
+  label = 'History',
+  changes = [],
+  className,
+  'aria-label': ariaLabel = label,
+  ...props
+}) => {
+  if (changes.length === 0) {
+    return null;
+  }
```
avivkeller: Let's assume that the consumer has logic to handle when changes aren't there.
(So remove the if condition)

packages/ui-components/Common/ChangeHistory/index.tsx
```diff
@@ -0,0 +1,78 @@
...
+                  <div className="block text-sm font-medium leading-tight">
+                    {change.label}
+                  </div>
+                  <div className="block text-xs leading-tight opacity-75">
+                    {change.versions.join(', ')}
+                  </div>
+                </>
+              );
+
+              const MenuItem = change.url ? 'a' : 'div';
```
avivkeller: Let's have an `As` prop, defaulting to `a`, so the user can (if they want) pass a custom link component.

MenuItem = ... ? As : div

packages/ui-components/Common/ChangeHistory/index.tsx
```diff
@@ -0,0 +1,78 @@
...
+              return (
+                <MenuItem
+                  key={index}
+                  className={styles.dropdownItem}
+                  role={'menuitem'}
+                  tabIndex={0}
+                  aria-label={`${change.label}:${change.versions.join(', ')}`}
```
avivkeller:
```suggestion
                  aria-label={`${change.label}: ${change.versions.join(', ')}`}
```

packages/ui-components/Common/ChangeHistory/index.tsx
```diff
@@ -0,0 +1,78 @@
...
+              const content = (
+                <>
+                  <div className="block text-sm font-medium leading-tight">
+                    {change.label}
+                  </div>
+                  <div className="block text-xs leading-tight opacity-75">
+                    {change.versions.join(', ')}
+                  </div>
+                </>
+              );
```
avivkeller: Can we put this directly inline?

avivkeller: nit: let's use the css file for these classes, since they are more complex

avivkeller (commented):
Almost there!

packages/ui-components/Common/ChangeHistory/index.tsx
Harkunwar: I'll remove this

packages/ui-components/Common/ChangeHistory/index.tsx
Harkunwar: Okay, I'll do that.

avivkeller (approved):

packages/ui-components/Common/ChangeHistory/index.tsx
```diff
@@ -0,0 +1,66 @@
...
+  changes = [],
+  className,
+  as: As = 'a',
+  'aria-label': ariaLabel = label,
+  ...props
+}) => (
+  <div className={classNames('relative', 'inline-block', className)} {...props}>
+    <details className="group">
+      <summary className={styles.summary} role="button" aria-haspopup="menu">
+        <ClockIcon className="h-4 w-4" />
```
AugustinMauroy:
```suggestion
        <ClockIcon className="size-4" />
```

packages/ui-components/Common/ChangeHistory/index.tsx
```diff
@@ -0,0 +1,66 @@
...
+        <span>{label}</span>
+        <ChevronDownIcon className="h-3 w-3 group-open:rotate-180 motion-safe:transition-transform" />
```
AugustinMauroy:
```suggestion
        <ChevronDownIcon className="size-3 group-open:rotate-180 motion-safe:transition-transform" />
```

AugustinMauroy (approved):

aymen94 (approved):

packages/ui-components/Common/ChangeHistory/index.tsx
```diff
@@ -0,0 +1,66 @@
...
+  as: As = 'a',
+  'aria-label': ariaLabel = label,
+  ...props
+}) => (
+  <div className={classNames('relative', 'inline-block', className)} {...props}>
+    <details className="group">
+      <summary className={styles.summary} role="button" aria-haspopup="menu">
+        <ClockIcon className="size-4" />
+        <span>{label}</span>
+        <ChevronDownIcon className="size-3 group-open:rotate-180 motion-safe:transition-transform" />
```
canerakdas, line 34:
```suggestion
        {children}
```

Instead of defining the icons and content within the component itself, it might be better to pass them as `children` from the outside. This way, we can reuse it not only in `ChangeHistory` but also in many other places

packages/ui-components/Common/ChangeHistory/index.tsx
```diff
@@ -0,0 +1,66 @@
...
+        className={styles.dropdownContentWrapper}
+        role="menu"
+        aria-label={ariaLabel}
+      >
+        <div className={styles.dropdownContentInner}>
+          {changes.map((change, index) => {
+            const MenuItem = change.url ? As : 'div';
+            return (
+              <MenuItem
+                key={index}
```
canerakdas, line 47: Instead of using only the `index` as the key, using a combination of `label` and `change.label` would be more reliable. When there are multiple instances of this component on the same page, using just the index might cause issues. See; https://react.dev/learn/rendering-lists#why-does-react-need-keys

packages/ui-components/Common/ChangeHistory/index.tsx
```diff
@@ -0,0 +1,66 @@
...
+                className={styles.dropdownItem}
+                role={'menuitem'}
```
canerakdas:
```suggestion
                role="menuitem"
```

packages/ui-components/Common/ChangeHistory/index.tsx
```diff
@@ -0,0 +1,66 @@
...
+  url?: string;
+};
+
+type ChangeHistoryProps = ComponentProps<'div'> & {
+  label: string;
+  changes: Array<HistoryChange>;
+  as?: LinkLike;
+};
+
+const ChangeHistory: FC<ChangeHistoryProps> = ({
```
canerakdas, line 21: Since we'll be using this component not only in Docs pages but also in places where JavaScript is disabled, it might be better to give it a more generic name for example, `DropdownMenu`

packages/ui-components/Common/ChangeHistory/index.tsx
```diff
@@ -0,0 +1,66 @@
...
+      </summary>
+      <div
+        className={styles.dropdownContentWrapper}
+        role="menu"
+        aria-label={ariaLabel}
+      >
+        <div className={styles.dropdownContentInner}>
+          {changes.map((change, index) => {
+            const MenuItem = change.url ? As : 'div';
+            return (
```
canerakdas:
```suggestion
            const MenuItem = change.url ? As : 'div';

            return (
```

nitpick: Adding an extra space makes it more readable IMO

packages/ui-components/Common/ChangeHistory/index.tsx
Harkunwar: I'm planning to refactor the contents out of this component into a dropdown component in my next PR. I believe we wanted a specific implementation at the moment.

packages/ui-components/Common/ChangeHistory/index.tsx
Harkunwar: I was using the ariaLabel as a key before and switched to index after a previous review. cc @avivkeller 

Secondly, I agree with him, I don't think we have a guarantee for the label being distinct for this component. There could easily be two entries with the same label. 

Using index as a key is only an issue if you're updating the list. For this specific component, I don't believe we plan to modify our array list of history, and so being a static component, index might actually be the best key here with a guarantee to be distinct.

packages/ui-components/Common/ChangeHistory/index.tsx
Harkunwar: I guess I already answered this above. I plan to do this in my next PR while keeping this component distinct.

packages/ui-components/Common/ChangeHistory/index.tsx
canerakdas: If we sure it won't cause any issues, it's fine by me as well

canerakdas (approved):

packages/ui-components/Common/ChangeHistory/index.tsx
avivkeller: Either works. Ultimately the tooling will integrate with whatever is added here

PR #7884, opened 2025-06-22
Revert ESP related changes

anonrig:
[This PR landed](https://github.com/nodejs/nodejs.org/pull/7875) while there is an active block by a TSC member and a collaborator. This PR reverts it. https://github.com/nodejs/nodejs.org/issues/7773

cc @nodejs/tsc

### DIFF
--- a/apps/site/components/Downloads/Release/ReleaseCodeBox.tsx
+++ b/apps/site/components/Downloads/Release/ReleaseCodeBox.tsx
@@ -38,7 +38,6 @@ const parseSnippet = (s: string, releaseContext: ReleaseContextType) => {
 
 const ReleaseCodeBox: FC = () => {
   const { snippets } = useContext(ReleasesContext);
-
   const { installMethod, os, packageManager, release } =
     useContext(ReleaseContext);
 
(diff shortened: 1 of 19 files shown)


jasnell (dismissed):
I'd like to give it a few days before taking action on this, per discussion in other channels. Escalating this to TSC agenda

jasnell:
I'll also clarify, there was no block on the PR that landed itself. The folks that landed the PR were under the impression that all objections had been cleared and that by all evidence the appropriate governance process was followed, even tho there does appear to have been a breakdown in communication following.

anonrig:
@jasnell There is a gap in our governance. This is why I propose https://github.com/nodejs/node/pull/58796

avivkeller:
FYI the esp/revert branch contains a revert as well, could you delete that branch if you intend to continue here?

MattIPv4:
Not a hard block on this, but instead of outright reverting the whole set of changes (including the other design changes to the homepage and downloads components), I think we'd be better off with a branch that just removes the ESP button (if we go down this route).

RaisinTen:
I'm +1 about removing the ESP button. Yagiz's -1 was unintentionally overlooked, so landing the change was a mistake. Removing it brings us back to stage 1, where we can move forward with whatever plan we decide to follow. I'm having trouble understanding the argument for blocking an attempt to undo a clear mistake.

If there are arguments for urgency, the reasoning and consequences should be communicated publicly to minimize misunderstandings.

mcollina:
@RaisinTen the reasons for urgency were clearly communicated to you and the rest of the TSC privately (both in a meeting and in the slack channel) - because they include some sensitive (financial) information. They were clearly communicated as such. We cannot disclose them at this point.

We cannot debate another month or just weeks if or not we want to gather the benefits of the ESP program.

mcollina (changes requested):
I’m -1 in reverting due to the explained urgency.

AugustinMauroy:
I agree with @RaisinTen that may allow us to return to stage 0 and reduce pressure. But apparently we cannot dude to contract restrictions. 
So IMO we should give effort to land https://github.com/nodejs/nodejs.org/pull/7883

avivkeller:
@anonrig https://github.com/nodejs/nodejs.org/pull/7884/commits/70cbf5f6622a278f51f63d4de76ce6db1444ab47 links the wrong issue (#7880 rather than #7881)

MattIPv4:
Closing as #7883 has landed instead per TSC consensus.

PR #7890, opened 2025-06-24
fix(AlertBox): override code styles in a markdown context

avivkeller:
## Description

When the inherited text is white, the `code` element does not have a readable contrast, as shown below:
image

## Validation

Everything on the website should look identical, but this should make a readable contrast on `api-docs-tooling`

### Check List

### DIFF
--- a/packages/ui-components/Common/AlertBox/index.module.css
+++ b/packages/ui-components/Common/AlertBox/index.module.css
@@ -73,4 +73,11 @@
       @apply bg-danger-700;
     }
   }
+
+  code {
+    /* Ignore the styles from `markdown.css` */
+    all: unset;
+
+    @apply font-mono;
+  }
 }
--- a/packages/ui-components/Common/AlertBox/index.module.css
+++ b/packages/ui-components/Common/AlertBox/index.module.css
@@ -78,6 +78,6 @@
     /* Ignore the styles from `markdown.css` */
     all: unset;
 
-    @apply font-mono;
+    @apply font-ibm-plex-mono;
   }
 }
--- a/packages/ui-components/Common/AlertBox/index.stories.tsx
+++ b/packages/ui-components/Common/AlertBox/index.stories.tsx
@@ -6,6 +6,12 @@ import AlertBox from '#ui/Common/AlertBox';
 type Story = StoryObj<typeof AlertBox>;
 type Meta = MetaObj<typeof AlertBox>;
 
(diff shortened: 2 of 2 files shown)


AugustinMauroy (approved):

packages/ui-components/styles/markdown.css
```diff
@@ -90,10 +90,6 @@ main {
     &:has(code) {
       @apply max-xs:decoration-neutral-800
         dark:max-xs:decoration-neutral-200;
-
```
canerakdas: IMO this should be made within the relevant component instead of `markdown.css`. For example, code links within the markdown content have also been unintentionally affected by this change;

Before:
image

Preview:
image

Source;
`/en/learn/getting-started/introduction-to-nodejs#an-example-nodejs-application`

avivkeller: I see, i'll update `AlertBox` (the affected component)

avivkeller: Looking at the css again, it seems it's supposed to _only_ apply that to the `a` elements, so maybe there's a selector issue, i'll investigate

packages/ui-components/styles/markdown.css
```diff
@@ -90,10 +90,6 @@ main {
...
-      code {
-        @apply text-inherit;
```
ovflowd: But links with code should have green colour btw.

ovflowd: (or as far as I understand)

avivkeller: I know, I'm pushing a fix now

avivkeller:
This has a slight visual difference:
image

ovflowd (approved):
LGTM!

ovflowd:
@avivkeller https://www.chromatic.com/review?appId=64c7d71358830e9105808652&number=7890&type=linked&activeElementId=comparison-test-key-685ad65269ceefc416a0df49-dark%20desktop&view=changes these look weird no?

avivkeller:
Would you prefer?:
image

---

So basically ignore the `markdown.css` styles in an AlertBox when they conflict?

ovflowd:
> Would you prefer?: image
> 
> So basically ignore the `markdown.css` styles in an AlertBox when they conflict?

I think that'd make more sense to respect the style of the component.

avivkeller:
In the Figma, there isn't a font change in `<code/>`, but for this, I used `ibm-plex-mono`, which is used in the other code blocks.

PR #7896, opened 2025-06-25
feat(shiki): use wasm engine

avivkeller:
Using the WASM shiki on the server, and the JavaScript shiki on the client, should result in the best mix of speed and size for our use-case.

See https://github.com/nodejs/api-docs-tooling/pull/320.

Rouhgly a 50% speed improvement on the server, per my testing.

### DIFF
--- a/packages/rehype-shiki/package.json
+++ b/packages/rehype-shiki/package.json
@@ -15,6 +15,7 @@
   "dependencies": {
     "@shikijs/core": "^3.4.2",
     "@shikijs/engine-javascript": "^3.4.2",
+    "@shikijs/engine-oniguruma": "^3.7.0",
     "classnames": "~2.5.1",
     "hast-util-to-string": "^3.0.1",
     "shiki": "~3.4.2",
--- a/packages/rehype-shiki/src/index.mjs
+++ b/packages/rehype-shiki/src/index.mjs
@@ -1,3 +1,5 @@
+import { createJavaScriptRegexEngine } from '@shikijs/engine-javascript';
+import { createOnigurumaEngine } from '@shikijs/engine-oniguruma';
 import cLanguage from 'shiki/langs/c.mjs';
 import coffeeScriptLanguage from 'shiki/langs/coffeescript.mjs';
 import cPlusPlusLanguage from 'shiki/langs/cpp.mjs';
@@ -17,6 +19,21 @@ import { createHighlighter } from './highlighter.mjs';
 
 const { shiki, getLanguageDisplayName, highlightToHast, highlightToHtml } =
   createHighlighter({
+    // We use the faster WASM engine on the server instead of the web-optimized version.
+    //
+    // Currently we fall back to the JavaScript RegEx engine
+    // on Cloudflare workers because `shiki/wasm` requires loading via
+    // `WebAssembly.instantiate` with custom imports, which Cloudflare doesn't support
+    // for security reasons.
+    //
+    // TODO(@avivkeller): When available, use `OPEN_NEXT_CLOUDFLARE` environmen
+    // variable for detection instead of current method, which will enable better
+    // tree-shaking.
+    // Reference: https://github.com/nodejs/nodejs.org/pull/7896#issuecomment-3009480615
+    engine:
+      'Cloudflare' in globalThis
+        ? createJavaScriptRegexEngine()
+        : await createOnigurumaEngine(import('shiki/wasm')),
     langs: [
       ...cLanguage,
       ...coffeeScriptLanguage,
--- a/packages/rehype-shiki/src/minimal.mjs
+++ b/packages/rehype-shiki/src/minimal.mjs
@@ -1,3 +1,4 @@
+import { createJavaScriptRegexEngine } from '@shikijs/engine-javascript';
 import powershellLanguage from 'shiki/langs/powershell.mjs';
 import shellScriptLanguage from 'shiki/langs/shellscript.mjs';
 
@@ -5,6 +6,9 @@ import { createHighlighter } from './highlighter.mjs';
 
 const { shiki, getLanguageDisplayName, highlightToHast, highlightToHtml } =
   createHighlighter({
+    // For the minimal (web) Shiki, we want to use the simpler,
+    // JavaScript based engine.
+    engine: createJavaScriptRegexEngine(),
     langs: [...powershellLanguage, ...shellScriptLanguage],
   });
 
--- a/apps/site/next.config.mjs
+++ b/apps/site/next.config.mjs
@@ -95,6 +95,11 @@ const nextConfig = {
       'shiki',
     ],
   },
+
+  webpack: config => ({
+    ...config,
+    experiments: { ...config.experiments, asyncWebAssembly: true },
+  }),
 };
 
 const withNextIntl = createNextIntlPlugin('./i18n.tsx');
--- a/apps/site/next.config.mjs
+++ b/apps/site/next.config.mjs
@@ -95,11 +95,6 @@ const nextConfig = {
       'shiki',
     ],
   },
-
-  webpack: config => ({
-    ...config,
-    experiments: { ...config.experiments, asyncWebAssembly: true },
-  }),
 };
 
 const withNextIntl = createNextIntlPlugin('./i18n.tsx');
--- a/packages/rehype-shiki/eslint.config.js
+++ b/packages/rehype-shiki/eslint.config.js
@@ -5,8 +5,12 @@ export default [
   {
     languageOptions: {
       parserOptions: {
-        // Allow nullish syntax (i.e. "?." or "??")
-        ecmaVersion: 2020,
+        // Allow nullish syntax (i.e. "?." or "??"),
+        // and top-level await
+        ecmaVersion: 'latest',
+      },
+      globals: {
+        globalThis: 'readonly',
       },
     },
     rules: {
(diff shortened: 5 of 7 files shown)


packages/rehype-shiki/src/index.mjs
```diff
@@ -15,8 +16,13 @@ import yamlLanguage from 'shiki/langs/yaml.mjs';
 
 import { createHighlighter } from './highlighter.mjs';
 
+// For the larger, server-based highlighter, we import the faster
+// WASM engine.
+const engine = await createOnigurumaEngine(import('shiki/wasm'));
```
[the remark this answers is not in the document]
AugustinMauroy: -1

packages/rehype-shiki/package.json
```diff
@@ -15,6 +15,7 @@
   "dependencies": {
     "@shikijs/core": "^3.4.2",
     "@shikijs/engine-javascript": "^3.4.2",
+    "@shikijs/engine-oniguruma": "^3.7.0",
```
AugustinMauroy, line 18: do we need to modify dependabot ?

avivkeller: No, Dependabot checks `@shikijs/*`.

AugustinMauroy: nice !

AugustinMauroy (approved):
That's super !

avivkeller:
@dario-piotrowicz https://github.com/nodejs/nodejs.org/actions/runs/15880572422/job/44779523238?pr=7896

Does OpenNext support WASM?

dario-piotrowicz:
> @dario-piotrowicz https://github.com/nodejs/nodejs.org/actions/runs/15880572422/job/44779523238?pr=7896
> 
> Does OpenNext support WASM?

sorry, not sure... this [code comment](https://github.com/opennextjs/opennextjs-cloudflare/blob/1868c99071f10f30988844ee9070dc5276ca1f95/packages/cloudflare/src/cli/build/patches/plugins/wrangler-external.ts#L4-L5) suggests that it should work... I'll check with @vicb tomorrow, but I guess that we might be hitting some open-next bug here 🤔

vicb:
> > @dario-piotrowicz https://github.com/nodejs/nodejs.org/actions/runs/15880572422/job/44779523238?pr=7896
> > Does OpenNext support WASM?
> 
> sorry, not sure... this [code comment](https://github.com/opennextjs/opennextjs-cloudflare/blob/1868c99071f10f30988844ee9070dc5276ca1f95/packages/cloudflare/src/cli/build/patches/plugins/wrangler-external.ts#L4-L5) suggests that it should work... I'll check with @vicb tomorrow, but I guess that we might be hitting some open-next bug here 🤔

It looks like the code is trying to create the module from a `UInt8Array`:

image

That is not supported on Workers - they need the wasm to be loaded from a `.wasm`. Similar to why `eval(...)` doesn't work.

ovflowd:
Can we store the Uint8Array as a file? The contents of it, or is there some way we can convert this to a wasm file. My understanding is we can pass the wasm instance to Shiki and we don't need to import this module....

vicb:
> Can we store the Uint8Array as a file? The contents of it, or is there some way we can convert this to a wasm file. My understanding is we can pass the wasm instance to Shiki and we don't need to import this module....

It should be a `.wasm` file.
I guess the `Uint8Array` is generated from some `.wasm` file so that file should be used instead - it doesn't look like it's part of `@shikijs/engine-oniguruma`. Maybe then a PR upstream.

vicb:
Looking at the [code](https://github.com/shikijs/shiki/blob/d25afc245b56c5c7811f39c01f0de1eafffa0547/packages/engine-oniguruma/src/wasm-inlined.ts#L3C21-L3C37) it looks like the wasm file is in [vscode-oniguruma](https://www.npmjs.com/package/vscode-oniguruma) - shiki uses [v1.7.0](https://github.com/shikijs/shiki/blob/d25afc245b56c5c7811f39c01f0de1eafffa0547/pnpm-workspace.yaml#L64)

apps/site/wrangler.jsonc
```diff
@@ -6,6 +6,7 @@
   "account_id": "8ed4d03ac99f77561d0e8c9cbcc76cb6",
   "minify": true,
   "keep_names": false,
+  "find_additional_modules": true,
```
vicb: This should not be needed as .wasm file will be added by default.
See https://developers.cloudflare.com/workers/wrangler/bundling/#including-non-javascript-modules

avivkeller:
@vicb is there any way OpenNext can support loading WASM files via `WebAssembly.instantiate`-ing the content, like is done here?

```
 CompileError: WebAssembly.instantiate(): Wasm code generation disallowed by embedder
```

Shiki supplies custom imports (`env` and `wasi_snapshot_preview1`) via a custom `importObject` passed to `WebAssembly.instantiate`.

Hypothetically, we might be able to do some custom import resolving, but that's a lot of boilerplate. We could also conditionally use the JavaScript engine in OpenNext, and the WASM one otherwise? Does OpenNext have any specific environment variable we can use to identify when it's in use?

avivkeller:
Ah, I see https://github.com/cloudflare/cloudflare-docs/issues/13469 documents that this is not allowed. Is there a way to enable it on our end, knowing the security concerns?

If not, I think the conditional JS Engine is the way to go.

vicb:
> @vicb is there any way OpenNext can support loading WASM files via `WebAssembly.instantiate`-ing the content, like is done here?

Yes you can [`WebAssembly.instantiate(module)` (a .wasm file uploaded along your source code)](https://developers.cloudflare.com/workers/runtime-apis/webassembly/), just not [`WebAssembly.instantiate(buffer)`](https://developers.cloudflare.com/workers/runtime-apis/web-standards/#javascript-standards)

> Ah, I see https://github.com/cloudflare/cloudflare-docs/issues/13469 documents that this is not allowed. Is there a way to enable it on our end, knowing the security concerns?

The security concerns are for Cloudflare too - that's why there is no way to instantiate from a buffer.
But instantiating from a module should work.
Using the JS engine is a good solution as well.

avivkeller:
Thank you for all your help! I'll investigate this!

vicb:
> Thank you for all your help! I'll investigate this!

I pointed to the .wasm file to use in an [earlier comment](https://github.com/nodejs/nodejs.org/pull/7896#issuecomment-3007595613)

avivkeller:
In case there isn't a viable solution (which, hopefully, it seems there is), is there a way to tell when our code in running in Cloudflare, versus everything else, so we can go with the conditional JS engine?

vicb:
> In case there isn't a viable solution (which, hopefully, it seems there is), is there a way to tell when our code in running in Cloudflare, versus everything else, so we can go with the conditional JS engine?

You can use `"Cloudflare" in globalThis` for testing purpose as a runtime check.
If you go down that road, we could define `process.env.OPEN_NEXT_CLOUDFLARE=1` so that the code can be tree shaken during the build.

avivkeller:
Thank you again for your help! We will, unfortunately, need to conditionally use the JavaScript RegExp engine when we are on Cloudflare Workers.

If we were to import the WASM file like:
```js
// Also available at "vscode-oniguruma/release/onig.wasm"
import wasm from "shiki/onig.wasm";
await import("shiki/onig.wasm");
```

It would result in errors like, since the imports need to be defined _before_ we import the WASM file:
```
Module not found: Can't resolve 'wasi_snapshot_preview1'
Module not found: Can't resolve 'env'
```

packages/rehype-shiki/src/index.mjs
```diff
@@ -17,6 +19,21 @@ import { createHighlighter } from './highlighter.mjs';
...
+    // on Cloudflare workers because `shiki/wasm` requires loading via
+    // `WebAssembly.instantiate` with custom imports, which Cloudflare doesn't support
+    // for security reasons.
+    //
+    // TODO(@avivkeller): When available, use `OPEN_NEXT_CLOUDFLARE` environmen
+    // variable for detection instead of current method, which will enable better
+    // tree-shaking.
+    // Reference: https://github.com/nodejs/nodejs.org/pull/7896#issuecomment-3009480615
+    engine:
+      'Cloudflare' in globalThis
```
ovflowd, line 34: Is this the best way of verifying this? But guess it makes sense

packages/rehype-shiki/src/minimal.mjs
```diff
@@ -1,10 +1,14 @@
...
 import powershellLanguage from 'shiki/langs/powershell.mjs';
 import shellScriptLanguage from 'shiki/langs/shellscript.mjs';
 
 import { createHighlighter } from './highlighter.mjs';
 
 const { shiki, getLanguageDisplayName, highlightToHast, highlightToHtml } =
   createHighlighter({
+    // For the minimal (web) Shiki, we want to use the simpler,
+    // JavaScript based engine.
+    engine: createJavaScriptRegexEngine(),
```
ovflowd, line 11: Where is this one used?

ovflowd (approved):
LGTM!

packages/rehype-shiki/src/index.mjs
avivkeller: No, see the TODO comment above

packages/rehype-shiki/src/minimal.mjs
avivkeller: This is used on the download snippets, since they are highlighted on the client side.

PR #7904, opened 2025-06-27
chore(actions): use gov repo for reused actions

avivkeller:
🎉 We now have `nodejs/web-team`, which'll contain our re-used actions.

This should make our lives a bit easier, as we don't need to update a thousand workflows whenever something changes.


.github/workflows/publish-packages.yml
```diff
@@ -95,23 +95,11 @@ jobs:
...
-
-      - name: Checkout repository
-        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
-
-      - name: Set up pnpm
-        uses: pnpm/action-setup@a7487c7e89a18df4991f7f222e4898a00d66ddda # v4.1.0
-
-      - name: Setup Node.js
-        uses: actions/setup-node@49933ea5288caeca8642d1e84afbd3f7d6820020 # v4.4.0
+      - uses: nodejs/web-team/actions/setup-environment@main
```
ovflowd: Shouldn't we use hashes here or since it is ours it is safe enough?

ovflowd (approved):

.github/workflows/publish-packages.yml
avivkeller: I feel like, since we control that repository, we can assume that it's safe. The only people with `write` permissions to it also have `write` permissions here.

ovflowd: We need to ensure then that also there are branch protection rules and the same rules for only allowing certain actions rules, etc.

ovflowd: cc @bmuenzenmeyer

bmuenzenmeyer (changes requested):
we cannot assume another repo is safe. this is a huge vulnerability

- malicious upstream
- disgruntled contributor
- compromised developer machine

https://blog.rafaelgss.dev/why-you-should-pin-actions-by-commit-hash

avivkeller:
Typically, that's the case, but I feel like since we are scoping the permissions almost identically to the ones here, the risk of compromising that repository is identical to the risk of compromising this repository.

If an attacker somehow gained permissions on that repo via a leaked access token, for example, they would already have access in this repository.

bmuenzenmeyer:
> Typically, that's the case, but I feel like since we are scoping the permissions almost identically to the ones here, the risk of compromising that repository is identical to the risk of compromising this repository.
> 
> If an attacker somehow gained permissions on that repo via a leaked access token, for example, they would already have access in this repository.

points taken, but that doesn't necessarily make me feel much better. the indirection of this repo and setup in the name of simplification and reuse only serves to make our environment harder to understand, and IMO, less secure in the long run. but now i am dredging up past disagreements and should self-moderate. 

i will lift my block if the majority approves of this approach...

@RafaelGSS - I invoked your blog post before - curious if you have any wisdom that could help us here

avivkeller:
I'll add the explicit commits. We can always change it the future.

GOVERNANCE.md
bjohansebas, line 1: I’d recommend leaving this as a fallback and mentioning in this file that the governance file has been moved elsewhere, to avoid a 404 error.

bjohansebas (approved):
LGTM

GOVERNANCE.md
avivkeller: I don't think this file has many (any?) backlinks

avivkeller:
@bmuenzenmeyer Are you still blocking?

RafaelGSS (commented):
Sorry for missing your notification, @bmuenzenmeyer.

I don’t think this change should be merged. I will not block this, but defer the final decision to the Website team.

Reusing these actions implicitly trusts the target repository’s safety. Anyone with write access to `nodejs/web-team/actions` could introduce malicious changes into the shared workflows, and those changes would be harder to spot than modifications to the `nodejs.org` codebase. Only people closely monitoring `setup-environment` would catch such alterations, which I consider too risky. While reducing duplication is valuable, keeping each workflow explicit is far safer.

avivkeller:
I've changed to use explicit commits, so they will need a double manual review to merge (1, in web-team, and 2, here)

MattIPv4:
👍 Let's pin these to specific hashes, Dependabot should still be able to bump them just fine without tags being created. We can always look at having Dependabot run more frequently too for these actions if needed.

RafaelGSS:
Pinning should fix that (if people review the changes in the repo)

bmuenzenmeyer (approved):

bmuenzenmeyer:
thanks @RafaelGSS for your thoughts here

avivkeller:
We do however, need a TSC CODEOWNER review, since this modifies (moves) the GOVERNANCE file.

avivkeller:
Bump @nodejs/tsc to approve. This doesn't change any governance, it only moves it.

.github/workflows/codeql.yml
```diff
@@ -25,54 +13,4 @@ permissions:
...
-
-      # - run: |
-      #   echo "Run, Build Application using script"
-      #   ./location_of_script_within_repo/buildscript.sh
-
-      - name: Perform CodeQL Analysis
-        uses: github/codeql-action/analyze@181d5eefc20863364f96762470ba6f862bdef56b # v3.29.2
-        with:
-          category: '/language:${{matrix.language}}'
+    uses: nodejs/web-team/.github/workflows/codeql.yml
```
MattIPv4: Should this be sha-pinned?

.github/workflows/dependency-review.yml
```diff
@@ -23,15 +15,4 @@ permissions:
...
-        uses: step-security/harden-runner@6c439dc8bdf85cadbbce9ed30d1c7b959517bc49 # v2.12.2
-        with:
-          egress-policy: audit
-
-      - name: Git Checkout
-        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
-
-      - name: Review Dependencies
-        uses: actions/dependency-review-action@da24556b548a50705dd671f47852072ea4c105d9 # v4.7.1
+    uses: nodejs/web-team/.github/workflows/dependency-review.yml
```
MattIPv4: Should this be sha-pinned?

.github/workflows/pull-request-label.yml
```diff
@@ -1,41 +1,26 @@
...
-          labels: github_actions:pull-request
+  analysis:
+    permissions:
+      # Needed to upload the results to code-scanning dashboard.
+      security-events: write
+      # Needed to publish results and get a badge (see publish_results below).
+      id-token: write
+      contents: read
+      actions: read
+    uses: nodejs/web-team/.github/workflows/scorecard.yml
```
MattIPv4: Should this be sha-pinned?

MattIPv4 (approved):

AugustinMauroy (approved):

bmuenzenmeyer:
conflicts

avivkeller:
Whoops! Git rebase does not like me today.

avivkeller:
image

avivkeller:
Since the TSC hasn't approved this, adding to their agenda for visibility: We just need one approval for this to land, thanks!

BridgeAR (commented):
I just read through the topic and I am not sure in what way things where changed since the beginning. If I am not mistaken, the current state is considered fine by the overall web team? I didn't look into the change closely, and I would handle it as @RafaelGSS.
If you do not have any concerns left, I am fine to approve for the CODEOWNERS file change.

@avivkeller when marking something for the TSC, please add a brief summary of what we should have a look at in particular.

ovflowd:
@avivkeller have you added a summary for the TSC? Is this still relevant for TSC's agenda?

avivkeller:
It's really just for visibility. We just need them to approve this PR, that's all

avivkeller:
@ovflowd can you bypass the merge queue (and the TSC approval)

ovflowd:
@avivkeller apologies, missed this comment. Can you rebase?

ovflowd:
@avivkeller could you rebase again? 😭 feel free to merge right after.

ovflowd:
@avivkeller there are still conflicts 🥲

aymen94 (approved):

ovflowd:
@avivkeller how there are so many conflicts ehre? O.o

avivkeller:
Ugh! I'll get these up-to-date right before our meeting, so we can merge them quickly

ovflowd:
> Ugh! I'll get these up-to-date right before our meeting, so we can merge them quickly

Bump!

avivkeller:
Waiting till the meeting to ensure this can land smoothly

ovflowd:
> Waiting till the meeting to ensure this can land smoothly

Ready?

avivkeller:
FYI @ovflowd

PR #7933, opened 2025-07-02
fix/The toggle navigation menu button is not keyboard accessible

### LINKED ISSUE #7895: The toggle navigation menu button is not keyboard accessible
The toggle navigation menu button when present (on smaller screens) is not keyboard accessible (it can be interacted via tab navigation + enter):
Image

___

Original context for the issue: https://github.com/nodejs/api-docs-tooling/pull/285#issuecomment-2998307585

anuj123upadhyay:
## Description

This PR improves accessibility by making the `mobile navigation toggle icon (navInteractionIcon) keyboard accessible.` 
Previously, the icon could only be toggled via mouse click. With this update, users can now navigate to the toggle using the `Tab key and activate it using Enter or Space.`

## Validation

https://github.com/user-attachments/assets/896ad9af-ee22-46d3-960d-64780f823734

## Related Issues

Fixes #7895

### Check List

- [✅] I have read the [Contributing Guidelines](https://github.com/nodejs/nodejs.org/blob/main/CONTRIBUTING.md) and made commit messages that follow the guideline.
- [✅] I have run `pnpm format` to ensure the code follows the style guide.
- [✅] I have run `pnpm test` to check if all tests are passing.
- [✅] I have run `pnpm build` to check if the website builds without errors.

### DIFF
--- a/packages/ui-components/Containers/NavBar/index.tsx
+++ b/packages/ui-components/Containers/NavBar/index.tsx
@@ -55,7 +55,17 @@ const NavBar: FC<PropsWithChildren<NavbarProps>> = ({
           className={style.sidebarItemTogglerLabel}
           htmlFor="sidebarItemToggler"
           role="button"
+          tabIndex={0}
           aria-label={sidebarItemTogglerAriaLabel}
+          onKeyDown={e => {
+            if (e.key === 'Enter' || e.key === ' ') {
+              e.preventDefault();
+              const input = document.getElementById(
+                'sidebarItemToggler'
+              ) as HTMLInputElement;
+              input?.click(); // Triggers input toggle
+            }
+          }}
         >
           {navInteractionIcons[isMenuOpen ? 'close' : 'show']}
         </Label.Root>


packages/ui-components/Containers/NavBar/index.tsx
```diff
@@ -55,7 +55,17 @@ const NavBar: FC<PropsWithChildren<NavbarProps>> = ({
           className={style.sidebarItemTogglerLabel}
           htmlFor="sidebarItemToggler"
           role="button"
+          tabIndex={0}
           aria-label={sidebarItemTogglerAriaLabel}
+          onKeyDown={e => {
```
avivkeller, line 60: Why do we need an onKeyDown? It worked fine without one

avivkeller (changes requested):

packages/ui-components/Containers/NavBar/index.tsx
anuj123upadhyay: onKeyDown will open the navInteractionIcon by pressing enter or space when the navInteractionIcon is in focused state.
like displayed in the video provided above.

anuj123upadhyay: if you want me to add or remove something then please let me know.

avivkeller: We don't have an `onClick` handler, so I don't think we need an `onKeyDown` handler, is there a way  to do this without?

anuj123upadhyay: i don't think so as we have to mimic the click event by pressing tab and enter on the navInteractionIcon,
if you can suggest , please let me know i am ready to for that.

ovflowd (changes requested):
In which circumstances will someone need to use the keyboard on a **touch device** such as a phone? Since this view is only rendered for small screens, supposedly touch? I feel we're trying to fix a non-existent issue.

I also don't think we should go the JavaScript route, this is 100% doable without JavaScript as @avivkeller mentioned. Check https://stackoverflow.com/questions/53753223/how-to-allow-for-users-to-keyboard-tab-over-and-select-invisible-checkboxes-with for example.

ShubhamOulkar:
> I feel we're trying to fix a non-existent issue.

Dev 😃 issue

avivkeller:
This seems to be stale, however, @anuj123upadhyay if you disagree, let me know

PR #7954, opened 2025-07-05
feat(get-involved): add `r/node`

avivkeller:
This PR adds `r/node` (The subreddit) to our "Unofficial Discussion Areas", since it's a well-known (310k users) unofficial discussion area.


avivkeller:
(Maybe we should move the OpenJSF Slack to an official discussion area–it's about as official as it gets, no?)

AugustinMauroy (approved):
LGMT !

AugustinMauroy:
> (Maybe we should move the OpenJSF Slack to an official discussion area–it's about as official as it gets, no?)

if I recall correctly, last time we talk about that @ovflowd was in favor of leaving as community since it isn't maintained directly by node or something like that

bjohansebas (approved):

aymen94 (approved):

bmuenzenmeyer (approved):

bmuenzenmeyer:
see the ongoing rate limit problems being tracked elsewhere

PR #7956, opened 2025-07-06
chore(logo): update logos, remove unused

avivkeller:
Ref #7880.

This PR updates the logo variants to use the newer icon.


avivkeller:
If you'd like me to also update the stacked variants in this PR, draft this PR and let me know. (This is separate right now, since updating the stacked logos requires _a lot_ more math.

MattIPv4:
Is there a reason ui-components doesn't just use the same SVGs, rather than a custom component that requires complex maintenance to combine two different SVGs into one?

MattIPv4:
And what's the reasoning for not updating the stacked versions?

avivkeller:
> And what's the reasoning for not updating the stacked versions?

I'm more than happy to, it's just exponentially more math, so I figured it'd be better to get these out, since they are done

AugustinMauroy (approved):
LGMT but what about the component one ?

avivkeller:
> LGMT but what about the component one ?

See the PR description

AugustinMauroy:
> Is there a reason ui-components doesn't just use the same SVGs, rather than a custom component that requires complex maintenance to combine two different SVGs into one?

/cc @ovflowd you was requiring during 2024 pride month

avivkeller:
@MattIPv4 @ovflowd Before I merge, just confirming this is okay to land on its own?

MattIPv4:
👍 No objections, but lets make sure we get all the other variants updated too.

MattIPv4:
> I'm more than happy to, it's just exponentially more math

Can you expand on this? Is it not just a case of updating the assets in Figma with the cleaned up core Node.js icon, and then re-exporting them?

The Pride variant should be roughly the same process, it uses the core Node.js icon at its core with blend modes applied.

avivkeller:
> Can you expand on this? Is it not just a case of updating the assets in Figma with the cleaned up core Node.js icon, and then re-exporting them?

For these logos, I didn't use Figma, I manually edited the SVGs, as to avoid re-exporting.

avivkeller:
I’ll draft this and try and get _all_ public assets in a single PR, as to avoid making branding seem unaligned

avivkeller:
@MattIPv4 Would you be so kind as to handle the pride logo changes in this PR? Feel free to push directly to this branch. All the other logos have been changed. TYSM.

I'm not really an expert UI designer, so I don't want to mis-export your vision.

I'm in the process of updating all the logos (except Pride) in the Figma.

MattIPv4:
👍 Updated them by hand based on the new logos you'd pushed. You should be able to generate a nice diff comparing them against the non-pride versions.

avivkeller:
Thank you! I'll add the finishing touches (ui-components) and get this to finish line today!

avivkeller:
@nodejs/nodejs-website Please re-review, and merge

PR #7968, opened 2025-07-09
fix: tr cannot be a child of table

btea:
## Description
image

### Check List


avivkeller (changes requested):
Can you instead change this in Crowdin?

btea:
It seems that it is correct on Crowdin, but it is not synchronized in the repository.

image

avivkeller:
In that case, it will sync at the next Crowdin download.

PR #7975, opened 2025-07-12
learn: fix net module import in server and client example

9saurabh9:
Replaced incorrect destructured import with default import for the 'net' module.

Old:
  import { net } from 'node:net'

Corrected to:
  import net from 'node:net'

This fix applies to the server and client example for mjs on this page :- 
https://nodejs.org/en/learn/modules/how-to-use-streams#key-methods-and-events-in-duplex-streams


avivkeller (approved):

AugustinMauroy (approved):
LGTM

PR #7984, opened 2025-07-15
feat: translate download of zh-cn

leedom92:


avivkeller (changes requested):
Hi! Thank you! We actually use Crowdin for our translations, so can you update this there?

Issue #7969, opened 2025-07-09
Download option list "using" too short


MikeMcC399:
### URL:

https://nodejs.org/en/download

### Browser Name:

Google Chrome

### Browser Version:

Version 138.0.7204.100 (Official Build) (64-bit)

### Operating System:

Ubuntu 24.04.2 LTS

### How to reproduce the issue:

- Open https://nodejs.org/en/download in Chrome
- Click the arrow in the dropdown list to the right of "using" and note that for Linux it shows three options under the heading "Recommended (Official)" and shows the heading "Community (Unofficial)" without displaying any of these options

<img width="826" height="500" alt="Image" src="https://github.com/user-attachments/assets/fe1e47ee-c0c1-492d-a294-5304049f8158" />

- Hover over the down-arrow in the list and it automatically scrolls to the end of the list without showing the title "Community (Unofficial)" anymore

<img width="833" height="544" alt="Image" src="https://github.com/user-attachments/assets/33d8d22f-724e-43a3-9ad3-e435b7089256" />

- Use the mouse to scroll up the list and check if there were some options under "Community (Unofficial)" that have been scrolled out of view. Using this scrolling list feels awkward at the moment.
- I would like the displayed list to be longer, so that when community options are displayed, that the title "Community (Unofficial)" is also displayed and that it is immediately clear whether all options are being shown or just a subset of options.

abinadiS:
I can assist with that.

avivkeller:
Be our guest! Thanks!

MikeMcC399:
- See #7989

PR #7989, opened 2025-07-16
fix(InstallationMethodDropdown): increase height

abinadiS:
## Description

 Adjust dropdown max-height from 48 to 56
https://github.com/nodejs/nodejs.org/issues/7969

### Check List

### DIFF
--- a/packages/ui-components/src/Common/Select/index.tsx
+++ b/packages/ui-components/src/Common/Select/index.tsx
@@ -39,6 +39,12 @@ type SelectProps<T extends string> = {
   inline?: boolean;
   onChange?: (value: T) => void;
   className?: string;
+  /**
+   * Allows passing custom CSS classes to the dropdown container element.
+   * This is useful for overriding default styles, such as adjusting `max-height`.
+   * The dropdown is rendered within a `Portal`.
+   */
+  dropdownClassName?: string;
   ariaLabel?: string;
   loading?: boolean;
   disabled?: boolean;
@@ -52,6 +58,7 @@ const Select = <T extends string>({
   inline,
   onChange,
   className,
+  dropdownClassName,
   ariaLabel,
   loading = false,
   disabled = false,
@@ -166,9 +173,11 @@ const Select = <T extends string>({
           <SelectPrimitive.Portal>
             <SelectPrimitive.Content
               position={inline ? 'popper' : 'item-aligned'}
-              className={classNames(styles.dropdown, {
-                [styles.inline]: inline,
-              })}
+              className={classNames(
+                styles.dropdown,
+                { [styles.inline]: inline },
+                dropdownClassName
+              )}
             >
               <SelectPrimitive.ScrollUpButton>
                 <ChevronUpIcon className={styles.scrollIcon} />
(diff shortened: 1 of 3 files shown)


AugustinMauroy (commented):
IDK what should we do. Like if we add new entry in download select do we will increase size ?

MikeMcC399:
@abinadiS 

I see that your PR is still in draft, so maybe you are not ready for comment yet, but anyway I will ask:

- What happens if more items are added to the list. Will this still work?
- Is it necessary to have the maximum height restriction or could this be removed completely? So it would look like this:
image


- If maximum height is removed would it adversely affect other dropdowns on the website? Edit: answer - yes, as the same class is also used for each of the dropdowns, including the version list, which looks very untidy if all the versions are displayed.

I wonder if the dropdown for `Release.InstallationMethodDropdown` needs its own class so that the list length for this can be handled differently?

Also your commit message would need to comply with https://github.com/nodejs/nodejs.org/blob/main/docs/code-style.md#commit-guidelines so, you would need to change to `fix: height dropdown`, for example.

Finally, if you write `closes #7969` in your original post, it will automatically link to the issue.

abinadiS:
Yes, it's because it shares the same class, and that also affects the versions dropdown menu. For it to only affect that specific element, it would have to be done with a 'prop' (property), and that would involve more changes. I just tried to make it as simple as possible and, seeing that it didn't significantly affect the versions list, I decided to submit it like that. I can adjust it so that everything displays correctly, even if more items are added to that list.

MikeMcC399:
@abinadiS 

Thanks for explaining! It's definitely an improvement from the current state. I'm just a user of this site, not a front-end developer or a site owner, so I think the site maintainers will need to give their opinion on the change. 

If you are ready for the PR to be reviewed, you can take it out of draft state.

abinadiS:
@MikeMcC399 is very important user opinions, thanks for your comment, I can test what you say to find the easiest way and expect approval.

abinadiS:
@MikeMcC399  
Captura de pantalla 2025-07-16 a la(s) 5 59 54 p m

Thanks for your comment, I made some adjustments.

MikeMcC399:
@abinadiS 

This is an excellent improvement from a user perspective! 🎉  Thank you very much!

packages/ui-components/src/Common/Select/index.tsx
```diff
@@ -39,6 +39,12 @@ type SelectProps<T extends string> = {
   inline?: boolean;
   onChange?: (value: T) => void;
   className?: string;
+  /**
+   * Allows passing custom CSS classes to the dropdown container element.
+   * This is useful for overriding default styles, such as adjusting `max-height`.
+   * The dropdown is rendered within a `Portal`.
+   */
```
avivkeller, line 46:
```suggestion
```

Do we need this code comment, we don't have any other comments on this type?

avivkeller (approved):

packages/ui-components/src/Common/Select/index.tsx
abinadiS: It's not necessary.

avivkeller:
@nodejs/nodejs-website what if we just made them all bigger

MikeMcC399:
@avivkeller 

> what if we just made them all bigger

That would be a step back in the evolution of this PR. I tried doing that and as I mentioned in https://github.com/nodejs/nodejs.org/pull/7989#issuecomment-3078343932 the dropdown list for versions becomes very long and untidy, displaying additional end-of-life versions to download.

This is off-topic for this PR, but I'll just mention it here: as an enhancement, the version list could display only supported versions by default and have a toggle to show all versions.

bjohansebas (approved):
LGTM

PR #7994, opened 2025-07-19
fix(CodeTabs): name overrides, duplicate name titles

avivkeller:
Ref: https://github.com/nodejs/node/issues/52343#issuecomment-3092432939
Ref: https://github.com/nodejs/node/issues/52343#issuecomment-3092438900

This PR renames `MJS` to `ESM`, and adds a `(1)` (etc) when multiple identically titled codetabs are used.
image


AugustinMauroy (approved):
LGMT ! Do we have test for this component ?

araujogui (approved):

PR #8005, opened 2025-07-21
chore: bump minimum required node version to 22.14.0

dario-piotrowicz:
## Description

Very minor, but I just noticed that nvm is suggesting node 22, when trying to use a version of node >= 22 but < 22.14.0 results in errors when running commands such as `pnpm dev` and `pnpm build` due to the fact that that the generate script uses the glob's exclude option as a pattern, but that's only available in >= 22.14.0 (https://nodejs.org/docs/latest-v22.x/api/fs.html#fspromisesglobpattern-options). So I figured it'd be more correct to bump the required node version 🙂 

## Validation

I've manually validated that on my machine

### Check List


AugustinMauroy (approved):

MattIPv4 (approved):
I wonder if .nvmrc should have web-infra as a code owner

mikeesto (approved):
Thanks @dario-piotrowicz

PR #8054, opened 2025-07-30
chore: apply useless change just to see if the open-next CI e2e workflow is flaky

dario-piotrowicz:
Please ignore this PR 🙂


dario-piotrowicz:
Closing since there doesn't seem to be any flakiness, also @avivkeller looked into the issue and it sounds like there was an actually legitimate bug (unrelated to open-next, which has since been solved) that caused the flakiness

PR #8057, opened 2025-08-01
feat(remark-lint): add

### LINKED ISSUE #350: `doc-kit` vs `remark-preset-lint-node`
In addition to the linting defined in this repository, we also rely on [`remark-preset-lint-node`](https://github.com/nodejs/remark-preset-lint-node) for shared linting rules.

How do we decide which linting rules belong in this repository vs. in `remark-preset-lint-node`?

Hypothetically, we _could_ perform all the linting as (and move `remark-present-lint-node` into) a new remark plugin (`@node-core/remark-lint` in https://github.com/nodejs/nodejs.org, or something)

avivkeller:
Fixes https://github.com/nodejs/doc-kit/issues/350

This PR updates the monorepo to include a rewrite of `remark-preset-lint-node`. Now that we've migrated to the new generators, it makes sense to centralize our linting rules in one place, maintained by the same team. This ensures the generator syntax aligns exactly with the linting configuration.

The updated preset includes all existing rules plus the custom ones currently defined in `doc-kit`. With a follow-up in `doc-kit`, this will also address https://github.com/nodejs/doc-kit/issues/328.

Once the migration to `doc-kit` is finished, we can follow up in node core to replace `remark-preset-lint-node` with `@nodejs/remark-lint/api`.

Our current setup is fragmented:

* `remark-preset-lint-node`
* Custom rules in node core
* More rules in `doc-kit`
* No clear separation in `remark-preset-lint-node` between general and API-specific rules

By consolidating everything into a single package, we’ll simplify future maintenance.

> [!TIP]
> This package is _versioned_, and will only publish when the `version` package.json field is bumped (or via `workflow_dispatch`).

cc @nodejs/web-infra
cc @nodejs/linting
cc @Trott

~~**Blocked**: `NPM_TOKEN` secret needs to be updated to allow publishing `@node-core/remark-lint`~~


avivkeller:
The only lint changes are escaping characters (i.e `_` because `\_`)

araujogui (commented):
Where's the llm_description lint rules?

avivkeller:
> Where's the llm_description lint rules?

# 🤦

I'll add them today, whoops

Trott:
Is the idea that this is what will also be used to lint non-documentation .md files in the nodejs/node and nodejs/TSC repositories? Or will those continue to use remark-lint-preset-node and this is for website linting only?

avivkeller:
> Is the idea that this is what will also be used to lint non-documentation .md files in the nodejs/node and nodejs/TSC repositories? Or will those continue to use remark-lint-preset-node and this is for website linting only?

The idea is for this to be the linter associated with `doc-kit`, the new generator for nodejs/node.

So, the web-infra and website teams will maintain a linter that complies with the generators we wrote (to ensure that they integrate well together).

Yes, it'll work on the other repositories, but there's no obligation to use it there.

avivkeller:
It's going to have two presets. `index` (just the default import), and `api`.

`api` will include all of the rules that are (as I see it) all over the place. (Some in doc kit, some in core, etc).

`index` will work for website linting, but could work anywhere in the organization.

bmuenzenmeyer (commented):
Am I missing a place or decision for us to have/not have READMEs in our packages? Seems like some of the context of this PR and the discussion already would be great to document as intents.

I noticed that the other packages don't have READMEs either. I don't intent to scope creep you but this seems like a good idea, no?

More review coming.

ovflowd:
> **Blocked**: `NPM_TOKEN` secret needs to be updated to allow publishing `@node-core/remark-lint`

Issue on /admin repo?

packages/remark-lint/src/index.mjs
ovflowd, line 1: In order to avoid self-referenciating rules, can you rename this to base.mjs, and then simply export it also on index.mjs? Then api.mjs imports base.mjs

packages/remark-lint/src/rules/required-metadata.mjs
```diff
@@ -0,0 +1,38 @@
+import { lintRule } from 'unified-lint-rule';
+import { visit } from 'unist-util-visit';
+
+const requiredKeys = ['introduced_in', 'llm_description'];
+const checks = requiredKeys.map(key => new RegExp(`<!--\\s*${key}=.*?-->`));
+
+const hasRequiredMetadata = (tree, vfile) => {
```
ovflowd, line 14: Can you clean up this a bit? I assume this function will replace the rule on doc-kit?

avivkeller:
> > **Blocked**: `NPM_TOKEN` secret needs to be updated to allow publishing `@node-core/remark-lint`
> 
> 
> 
> Issue on /admin repo?

Yes, I'll open one, once this gets some approvals

packages/remark-lint/src/index.mjs
avivkeller: I don't think there is a difference, right? Am I missing something?

avivkeller:
Can someone tell me why Vercel failed? @nodejs/web-infra

MattIPv4:
```
@node-core/website:build: Error: Cannot apply utility class `after:md:-left-1/2` because the `md` variant does not exist.
```

```
An unexpected error happened while finalizing this build. The real error may be shown somewhere else in the logs. We have been notified of the problem. If you have any questions, please contact Vercel Support https://vercel.com/help
```

avivkeller:
> ```
> @node-core/website:build: Error: Cannot apply utility class `after:md:-left-1/2` because the `md` variant does not exist.
> ```

That's a warning disguised as an error, but has since been fixed.

> ```
> An unexpected error happened while finalizing this build. The real error may be shown somewhere else in the logs. We have been notified of the problem. If you have any questions, please contact Vercel Support https://vercel.com/help
> ```

🤔 I'm going to rebase with `main` and see if it fixes itself.

canerakdas:
@avivkeller The text of the links seems to be corrupted;

image


Before:
https://nodejs.org/en/blog/release/v0.10.33

After:
https://nodejs-kgcow7iem-openjs.vercel.app/en/blog/release/v0.10.33

avivkeller:
> @avivkeller The text of the links seems to be corrupted;

I'll fix them, thank you! Release blogs before the redesign don't properly escape

ovflowd:
Failing?

avivkeller:
> Failing?

https://github.com/nodejs/nodejs.org/pull/8057/commits/e754df856f6a1179917dce0843265ffb856279fc#diff-576a7d7334476f79303d7d33d6bc1493d367708909ee73d3b9da80f0f24caf28R49 caused a dependency to be unable to determine the type of node that it was processing, leading it failing, fixed.

ovflowd:
Is the blocked tag still relevant? Or is this PR ready?

avivkeller:
> Is the blocked tag still relevant? Or is this PR ready?

It's ready for review, but cannot land until our NPM_TOKEN's scope is expanded (I have an open issue in admin)

packages/remark-lint/src/rules/yaml/validate-versions.mjs
```diff
@@ -0,0 +1,61 @@
+import { valid, parse, gt } from 'semver';
+
+const NODE_RELEASED_VERSIONS =
+  process.env.NODE_RELEASED_VERSIONS?.split(',') ?? [];
```
ovflowd: Should the linter do direct process.env access? I think this should be an option for the linter itself, as on ESlint rule options.

packages/remark-lint/src/rules/yaml/validate-versions.mjs
ovflowd, line 1: I feel this file could be cleaned up a bit

packages/remark-lint/src/rules/yaml/validate-changes.mjs
```diff
@@ -0,0 +1,62 @@
...
+import { validateVersion } from './validate-versions.mjs';
+
+const VALID_KEYS = ['version', 'pr-url', 'description'];
+const VALID_PR_URL =
+  /^https:\/\/github.com\/nodejs(?:-private)?\/node(?:-private)?\/pull\/\d+$/;
+
+const isSecurityRelated = change =>
+  typeof change['pr-url'] === 'string' &&
+  change['pr-url']?.startsWith(
+    'https://github.com/nodejs-private/node-private/pull/'
```
ovflowd: Move to constant file

ovflowd (approved):
Mostly LGTM, but I feel our YAML rules could receive a bit of cleanup.

packages/remark-lint/src/rules/yaml/validate-changes.mjs
avivkeller: I've moved it to constant, but not a constant file. It seems, to me, like adding a constants files is redundant, since, all the constants are only used by the files that define them

packages/remark-lint/src/rules/yaml/validate-versions.mjs
avivkeller: Ultimately, the consumers (node core) will control this through the `NODE_RELEASED_VERSIONS` environment variable. Why require them to update their configs to move it from the environment to an option when it can instead be handled internally as it is now?

.github/workflows/publish-packages.yml
```diff
@@ -120,8 +132,12 @@ jobs:
         run: |
           # Install deps
           pnpm install --frozen-lockfile
-          # Create a unique version using the commit SHA as a prerelease identifier
-          npm version --no-git-tag-version 1.0.1-$COMMIT_SHA
+
+          HAS_VERSION=$(jq 'has("version")' package.json)
+          if [ "$HAS_VERSION" == "false" ]; then
+            # Only bump version if package has no version field
+            npm version --no-git-tag-version 1.0.1-$COMMIT_SHA
```
Ethan-Arrowood, line 144: is the 1.0.1 meant to be hardcoded?

packages/remark-lint/README.md
```diff
@@ -0,0 +1,39 @@
...
+```json
+{
+  "plugins": ["@node-core/remark-lint"]
+}
+```
+
+You can then run `remark` over your markdown files:
+
+```bash
+npx remark . --frail
```
Ethan-Arrowood, line 24: What is the purpose of `--frail` ?

packages/remark-lint/README.md
```diff
@@ -0,0 +1,39 @@
...
+```
+
+## Settings
+
+### `NODE_RELEASED_VERSIONS`
+
+Some lint rules (such as `node-core:yaml-comments`) require knowledge of released Node.js versions to validate version references.
+
+You can provide these using the `NODE_RELEASED_VERSIONS` environment variable:
```
Ethan-Arrowood: How do you provide the complete set of released versions? Is there a script or something to do so?

packages/remark-lint/package.json
```diff
@@ -0,0 +1,59 @@
...
+  "scripts": {
+    "lint": "node --run lint:js",
+    "lint:fix": "node --run lint:js:fix",
+    "lint:js": "eslint \"**/*.mjs\"",
+    "lint:js:fix": "node --run lint:js -- --fix",
+    "test": "node --run test:unit",
+    "test:unit": "cross-env NODE_NO_WARNINGS=1 node --experimental-test-coverage --test \"**/*.test.mjs\""
+  },
+  "dependencies": {
+    "remark-gfm": "~4.0.1",
```
Ethan-Arrowood: why are some deps using different version specifiers? ~ v ^

packages/remark-lint/src/api.mjs
```diff
@@ -0,0 +1,57 @@
...
+        { yes: 'Node.js' },
+        { no: 'Node[Jj][Ss]', yes: 'Node.js' },
+        { no: "Node\\.js's?", yes: 'the Node.js' },
+        { no: '[Nn]ote that', yes: '<nothing>' },
+        { yes: 'RFC' },
+        { no: '[Rr][Ff][Cc]\\d+', yes: 'RFC <number>' },
+        { yes: 'TypeScript' },
+        { yes: 'Unix' },
+        { yes: 'Valgrind' },
+        { yes: 'V8' },
```
Ethan-Arrowood, line 63: This is cool! Would it be valuable to add npm in here somehow? 
```suggestion
        { yes: 'V8' },
        { yes: 'npm' },
```

packages/remark-lint/src/rules/required-metadata.mjs
Ethan-Arrowood, line 1: it may be valuable to provide correct/incorrect usage docs and examples for all of these rules

packages/remark-lint/src/rules/yaml/validate-versions.mjs
Ethan-Arrowood: My thinking is that that list would generally only be additive and I'd want it to be shared amongst my contributors. Like it'd be valuable to store that list in my config file, commit it to source, then over time add to it and share that with the team. This way not everyone has to maintain their own list of versions. 

Simultaneously, I agree this should be configurable so that I could potentially use this with non-node projects (like undici). So maybe the option should just be called "release versions" instead?

Ethan-Arrowood (approved):
All in all lgtm! I have some non blocking questions but otherwise the tests look great and I'm excited for this to land.

.github/workflows/publish-packages.yml
avivkeller: Yes. This is our commit-based versioning system. See #7777.

packages/remark-lint/README.md
avivkeller: `--frail` causes Remark to treat warnings as errors

packages/remark-lint/README.md
avivkeller: https://github.com/nodejs/node/blob/main/tools/lint-md/list-released-versions-from-changelogs.mjs

packages/remark-lint/src/api.mjs
avivkeller: I can do that as a follow-up, this list was taken directly from the currently used list

packages/remark-lint/README.md
Ethan-Arrowood: Okay cool so its more strict. I like it!

packages/remark-lint/README.md
Ethan-Arrowood: Maybe linking to this in the doc would be beneficial?

packages/remark-lint/src/rules/yaml/validate-versions.mjs
avivkeller: Noted. I've updated it to use a `releasedVersions` config, specifiable via passing options

PR #8073, opened 2025-08-13
fix: RSS button navigate to 404 page if not with en locale

### LINKED ISSUE #8070: RSS feed button on non-English pages leads to 404 error page
### URL:

https://nodejs.org/ja/feed/blog.xml (Or blog page with other locales except for en)

### Browser Name:

n/a

### Browser Version:

n/a

### Operating System:

n/a

## Bug Description
The RSS feed button appears on all language versions of the website, but clicking it on non-English pages results in a 404 error because RSS feeds only exist for English content.

## Steps to Reproduce
1. Go to any non-English language blog page (e.g., https://nodejs.org/ja/blog for Japanese)
2. Click the RSS feed button in the blog header
3. Observe that the URL shows `/ja/feed/blog.xml`
4. Page returns 404 Not Found

## Expected Behavior
IMO, one of the following should happen:
- RSS button should be hidden on non-English pages
- RSS button should redirect to the English RSS feed (`/en/feed/blog.xml`)
- Show a clear message explaining RSS is only available in English

(If the team think this is a bug and have a preferred way for fixing it, I can help we that)

## Actual Behavior
- RSS button is visible on all language pages
- Clicking leads to 404 error with no explanation
- Users get confused about broken functionality

kylemocode:
## Description

close #8070 

RSS feed links on non-English pages redirect to English feeds because RSS feeds only exist in English.
It is more user friendly instead of showing 404 page without any explanation to the user.

## Validation

 To test this fix:
 1. Start development server: `pnpm dev`
 2. Navigate to a non-English blog page (e.g.,`/ja/blog`)
 3. Click the RSS icon in the blog header
 4. Verify it redirects to the English RSS feed (`/en/feed/blog.xml` or `/en/feed/${file}.xml` if you are in certain feed type) instead of 404 not found page

### Check List

### DIFF
--- a/apps/site/components/Blog/BlogHeader/index.tsx
+++ b/apps/site/components/Blog/BlogHeader/index.tsx
@@ -21,7 +21,6 @@ const BlogHeader: FC<BlogHeaderProps> = ({ category }) => {
         {t('layouts.blog.title')}
         <Link
           href={`/feed/${currentFile}`}
-          {...{ locale: 'en' }} // RSS feeds only exist in English
           aria-label={t('components.blog.blogHeader.rssLink')}
         >
           <RssIcon />
--- a/apps/site/components/Blog/BlogHeader/index.tsx
+++ b/apps/site/components/Blog/BlogHeader/index.tsx
@@ -21,7 +21,7 @@ const BlogHeader: FC<BlogHeaderProps> = ({ category }) => {
         {t('layouts.blog.title')}
         <Link
           href={`/feed/${currentFile}`}
-          {...{ locale: 'en' }} // RSS feeds only exist in English
+          locale="en"
           aria-label={t('components.blog.blogHeader.rssLink')}
         >
           <RssIcon />
--- a/apps/site/components/Blog/BlogHeader/index.tsx
+++ b/apps/site/components/Blog/BlogHeader/index.tsx
@@ -21,6 +21,7 @@ const BlogHeader: FC<BlogHeaderProps> = ({ category }) => {
         {t('layouts.blog.title')}
         <Link
           href={`/feed/${currentFile}`}
+          {...{ locale: 'en' }} // RSS feeds only exist in English
           aria-label={t('components.blog.blogHeader.rssLink')}
         >
           <RssIcon />
--- a/apps/site/app/[locale]/feed/[feed]/route.ts
+++ b/apps/site/app/[locale]/feed/[feed]/route.ts
@@ -2,7 +2,7 @@ import { NextResponse } from 'next/server';
 
 import provideWebsiteFeeds from '#site/next-data/providers/websiteFeeds';
 import { siteConfig } from '#site/next.json.mjs';
-import { availableLocaleCodes } from '#site/next.locales.mjs';
+import { defaultLocale } from '#site/next.locales.mjs';
 
 type DynamicStaticPaths = { locale: string; feed: string };
 type StaticParams = { params: Promise<DynamicStaticPaths> };
@@ -26,12 +26,10 @@ export const GET = async (_: Request, props: StaticParams) => {
 // `[locale]/feeds/[feed]` and returns an array of all available static paths
 // This is used for ISR static validation and generation
 export const generateStaticParams = async () =>
-  availableLocaleCodes.flatMap(locale =>
-    siteConfig.rssFeeds.map(feed => ({
-      locale: locale,
-      feed: feed.file,
-    }))
-  );
+  siteConfig.rssFeeds.map(feed => ({
+    locale: defaultLocale.code,
+    feed: feed.file,
+  }));
 
 // Enforces that only the paths from `generateStaticParams` are allowed, giving 404 on the contrary
 // @see https://nextjs.org/docs/app/api-reference/file-conventions/route-segment-config#dynamicparams
--- a/apps/site/app/[locale]/feed/[feed]/route.ts
+++ b/apps/site/app/[locale]/feed/[feed]/route.ts
@@ -2,7 +2,7 @@ import { NextResponse } from 'next/server';
 
 import provideWebsiteFeeds from '#site/next-data/providers/websiteFeeds';
 import { siteConfig } from '#site/next.json.mjs';
-import { defaultLocale } from '#site/next.locales.mjs';
+import { availableLocaleCodes } from '#site/next.locales.mjs';
 
 type DynamicStaticPaths = { locale: string; feed: string };
 type StaticParams = { params: Promise<DynamicStaticPaths> };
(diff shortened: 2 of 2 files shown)


AugustinMauroy (changes requested):
The fix should be applied on link but on generation. Because the feed actually exist on multiple languages but actually the content isn't translated.

https://github.com/nodejs/nodejs.org/tree/main/apps/site/app/%5Blocale%5D/feed/%5Bfeed%5D

apps/site/components/Blog/BlogHeader/index.tsx
```diff
@@ -21,6 +21,7 @@ const BlogHeader: FC<BlogHeaderProps> = ({ category }) => {
         {t('layouts.blog.title')}
         <Link
           href={`/feed/${currentFile}`}
+          {...{ locale: 'en' }} // RSS feeds only exist in English
```
avivkeller:
```suggestion
          locale='en' // RSS feeds only exist in English
```

avivkeller (approved):

kylemocode:
@AugustinMauroy @avivkeller 
Thanks for the quick reply. I change the fix on route generation stage instead of navigate to en feed directly.
Now we can see feeds under all locales as prefixes (The content isn't translated, which I believe will be future enhancement)

AugustinMauroy:
> The content isn't translated, which I believe will be future enhancement

We have added feeds with translated links in case anyone wants to translate the blog posts. At the moment, this is not being done because we are struggling to get a decent translation for the limited amount of content

kylemocode:
@AugustinMauroy 
I see, thank you for the detailed explanation.
The current fix should provide a basic user experience, allowing users from different regions to still use the RSS feed feature.

AugustinMauroy (approved):
LGMT !

kylemocode:
@AugustinMauroy 
For the pending CI checks, only maintainers can trigger the action ? Could you help with that ?

kylemocode:
@AugustinMauroy 
Playwright check failed, but I don't think it is relative to the change

AugustinMauroy:
> @AugustinMauroy Playwright check failed, but I don't think it is relative to the change

Don't worry about that

kylemocode:
@ovflowd @mikeesto @avivkeller 
Get other core team members to review this idea

avivkeller:
> The fix should be applied on link but on generation. Because the feed actually exist on multiple languages but actually the content isn't translated.
> 
> 
> 
> https://github.com/nodejs/nodejs.org/tree/main/apps/site/app/%5Blocale%5D/feed/%5Bfeed%5D

I don't have a _huge_ preference either way, but I'd prefer to, contrary to what Augustin said, not do this in the routing side, but rather on the linking side, since we only need to adjust a broken link, not add several more files to the site.

We only have an English RSS feed, not one for each language.

Opinions, @nodejs/nodejs-website?

kylemocode:
I think this also depends on whether you plan to support multiple languages for the RSS feed in the future.
If there’s no such plan, then making the change on the link side might be the more straightforward approach.

From a user’s perspective, if we see that the URL for a feed is in the `jp` locale, for example, but the actual content is in English, some people might think it’s a bug.

However, if there is a plan to support multiple languages for the RSS feed, then making the change on the routing side seems reasonable, as it would prepare for future translation work.

ovflowd (changes requested):
This was intentional. Blog posts are not translated.

kylemocode:
@avivkeller @ovflowd @AugustinMauroy 
Based on your feedbacks, I rollback the change to the initial solution, which is on the linking side.
Now users from different locales still can see the RSS feed instead of navigating to 404 page without explanation.
Please help to review again.

apps/site/components/Blog/BlogHeader/index.tsx
kylemocode: @avivkeller 
I used object spread here to pass locale because our custom Link component’s prop types are based on `HTMLAnchorElement` and don’t include locale, thus will cause type error. This approach avoids a larger refactor while keeping the behavior consistent with how the locale prop would be passed in a properly typed Next.js Link. It also ensures we don’t introduce unnecessary complexity just to satisfy type checking for a value that is already fixed (`en`).

截圖 2025-08-15 晚上9 06 08

mikeesto (approved):
I think linking to the English feed here instead of a 404 is an improvement - LGTM

apps/site/components/Blog/BlogHeader/index.tsx
avivkeller: I'd rather perform the larger change than have a hacky solution.

avivkeller: I'll fix it in a PR right now

kylemocode: I can also do that change in my day time tomorrow.
I will see if there is an updated one tomorrow, or else I can help the change.
Thanks for the feedback @avivkeller

avivkeller: See #8081

kylemocode:
Blocked by #8081

apps/site/components/Blog/BlogHeader/index.tsx
avivkeller: @kylemocode Can you rebase and accept this suggestion, after that, LGTM

avivkeller (approved):

kylemocode:
@avivkeller @ovflowd 
I changed based on the type change from #8081 
Please help to review and merge, thanks again for all the feedbacks and supports 🙏

avivkeller:
@ovflowd can you approve?

kylemocode:
@ovflowd 
I think after your approval, we are go to go 🚀
This PR is one line change and we had conclusion about the issue in above threads

kylemocode:
@ovflowd 
Could you help ?

apps/site/components/Blog/BlogHeader/index.tsx
```diff
@@ -21,6 +21,7 @@ const BlogHeader: FC<BlogHeaderProps> = ({ category }) => {
         {t('layouts.blog.title')}
         <Link
           href={`/feed/${currentFile}`}
+          locale="en"
```
ovflowd, line 24: Is this the only place that needs update? I also wonder, should we really redirect people out of their current locale to the "en" one? Although yes, the content is in English, the rest of the site is still localized... So I'm unsure if this is a warranted solution that makes sense.

cc @nodejs/nodejs-website

kylemocode: @ovflowd 
We had a discussion before
https://github.com/nodejs/nodejs.org/pull/8073#issuecomment-3190635849

BTW, the user will still stay in their current locale even though they click the RSS icon.
And it should be a better UX compare with the original one: go to 404 not found page without any explanation to the user.

@ovflowd Do you have a better solution in your mind ? we can have further discussion on that

ovflowd: I'm unsure whatt his locale flag is really doing. I'll defer to @bmuenzenmeyer or the @nodejs/nodejs-website team to make an educated judgement here. Feel free to dismiss my review.

kylemocode: `Previous behavio`r: For any locale other than en, when users clicked the RSS feed link, they were redirected to a 404 page. This was because, during the RSS feed route generation, we deliberately only generated the en route. (There was a commit [2c727554e4ba2facf8fd815fdeba9500c7297116] that attempted to change this behavior, but feedback suggested it should instead be handled via the link.)

`The new approach`: Regardless of the locale, clicking the RSS feed link will now direct users to the en feed. Since the team currently has no plans to translate the feed, this approach is better than the previous one where users would just encounter a 404 with no explanation.

@avivkeller Do you have the permission to bypass the request change ? Or what is the next step ?

ovflowd: I see now, this is just a link to the RSS feed, not the blog post.

ovflowd (approved):

PR #8105, opened 2025-08-22
chore(publishing): version packages

### LINKED ISSUE #7777: Version `@node-core/*` packages

avivkeller:
Fixes #7777 by adding a `"version"` field to all current `package.json` files. A new version will be released whenever a commit is pushed that bumps this field.

On crowdin syncs, the `@node-core/website-i18n` versions is bumped automatically

### DIFF
--- a/.github/workflows/publish-packages.yml
+++ b/.github/workflows/publish-packages.yml
@@ -81,17 +81,10 @@ jobs:
 
               # Determine if the package has changed (or include all on manual trigger)
               if [ "$EVENT_NAME" == "workflow_dispatch" ] || ! git diff --quiet $COMMIT_SHA~1 $COMMIT_SHA -- "$pkg/"; then
-                HAS_VERSION=$(jq 'has("version")' "$PKG_JSON")
-                if [ "$HAS_VERSION" == "false" ]; then
-                  # Include packages without version field
+                OLD_VERSION=$(git show $COMMIT_SHA~1:$PKG_JSON | jq -r '.version')
+                NEW_VERSION=$(jq -r '.version' "$PKG_JSON")
+                if [ "$OLD_VERSION" != "$NEW_VERSION" ]; then
                   CHANGED_PACKAGES+=("$PKG_NAME")
-                else
-                  # For packages with version field, include only if version changed
-                  OLD_VERSION=$(git show $COMMIT_SHA~1:$PKG_JSON | jq -r '.version')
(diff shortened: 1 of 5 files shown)


.github/workflows/publish-packages.yml
MattIPv4, line 1: What does the `workflow_dispatch` trigger do at this point -- presumably it'll try to publish versions that already exist, which seems problematic?

.github/workflows/translations-sync.yml
```diff
@@ -75,6 +75,7 @@ jobs:
         with:
           ref: ${{ env.BRANCH_NAME }}
           token: ${{ secrets.CROWDIN_GITHUB_BOT_TOKEN }}
+          fetch-depth: 2
```
MattIPv4, line 78: Why?

avivkeller: The fetch depth allows us to later compare the main branch to that of Crowdin, so, if the translation keys changed, the version can be bumped

.github/workflows/publish-packages.yml
avivkeller: Suppose a publish fails due to an expired token. The dispatch trigger will allow a manual publish once a new token is issued.

.github/workflows/translations-sync.yml
MattIPv4: Wouldn't the default fetch depth of 1 also let you compare against HEAD?

.github/workflows/publish-packages.yml
MattIPv4: Ah I see

.github/workflows/translations-sync.yml
avivkeller: We don't want to compare 1 commit above. We want to compare 2 commits above.

The first (most recent) commit will be the translation changes.

The second commit will be the base branch.

avivkeller:
Bump @nodejs/web-infra for reviews, I'd like to version these on the sooner side, so we can remove the hacky "update internal dependencies" script from doc-kit

flakey5 (approved):

MattIPv4 (approved):

ovflowd (approved):

aymen94 (approved):

AugustinMauroy (approved):

PR #8106, opened 2025-08-22
feat(remark-lint): add strict rules

avivkeller:
Follow-up #8057.

Now that we have a basic package, this PR modifies the rules to use `@nodejs/doc-kit`'s functionality, ensuring that _any_ changes their get reflected in the linter.

**Why was this separate from #8057?** I wanted a version (`1.0.0`) which is almost a drop-in replacement for the original linter, without relying on `@nodejs/doc-kit`, in case something broke.

### DIFF
--- a/packages/remark-lint/README.md
+++ b/packages/remark-lint/README.md
@@ -65,6 +65,22 @@ Prevents redundant stability markers in nested sections.
 > Stability: 2 - Stable <!-- Redundant! -->
 ```
 
+### `node-core:invalid-type-reference`
+
+Ensures that all `{type}` references are valid types and formatted correctly.
+
+**Allowed:**
+
+```markdown
+This is usually a {boolean}, but it could also be a {string|number}.
+```
+
+**Not allowed:**
+
+```markdown
+This is an {invalid} type, and so is {string | number} because there should **not** be whitespace around the `|`.
+```
+
 ### `node-core:hashed-self-reference`
 
 Ensures self-references use fragment-only links.
--- a/packages/remark-lint/src/api.mjs
+++ b/packages/remark-lint/src/api.mjs
@@ -7,6 +7,7 @@ import remarkLintUnorderedListMarkerStyle from 'remark-lint-unordered-list-marke
 import basePreset from './index.mjs';
 import duplicateStabilityNodes from './rules/duplicate-stability-nodes.mjs';
 import hashedSelfReference from './rules/hashed-self-reference.mjs';
+import invalidTypeReference from './rules/invalid-type-reference.mjs';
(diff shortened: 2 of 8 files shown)


packages/remark-lint/src/api.mjs
```diff
@@ -61,6 +63,7 @@ export default (options = {}) => ({
         { yes: 'Unix' },
         { yes: 'Valgrind' },
         { yes: 'V8' },
+        { yes: 'npm' },
```
avivkeller, line 66: @Ethan-Arrowood is this what you had in mind?

avivkeller:
@nodejs/web what other lint rules are needed?

I'm thinking:
- Typed lists syntax
- Stability node syntax
- Alphabetical typed headings

What else?

ovflowd (commented):
OOC, when we add new rules, in which repositories are they being applied? doc-kit? So it means Node's core API docs? Shouldn't then any PR to remark-lint actually require nodejs/collaborators as code owners?

avivkeller:
> OOC, when we add new rules, in which repositories are they being applied? doc-kit? So it means Node's core API docs? Shouldn't then any PR to remark-lint actually require nodejs/collaborators as code owners?

My plan is for these to be applied to `nodejs/node` once the new tooling is adopted (but it is not dependent on, and could land beforehand).

The previous implementation did not have any CODEOWNERS, thus, this is the same, however, that can change

packages/remark-lint/README.md
```diff
@@ -106,6 +106,22 @@ Metadata can be provided in comments:
...
+**Allowed:**
+
+```markdown
+This is usually a {boolean}, but it could also be a {string|number}.
+```
+
+**Not allowed:**
+
+```markdown
+This is an {invalid} type, and so is {string | number}.
```
Ethan-Arrowood:
```suggestion
This is an {invalid} type, and so is {string | number} because there should **not** be whitespace around the `|`.
```

Ethan-Arrowood (approved):
lgtm

AugustinMauroy (approved):
LGMT !

aymen94 (approved):

avivkeller:
I'm going to rebase this, as there have been no objections from core collaborators

avivkeller:
There seems to be some issues with the package and the windows CI 🤔 Perhaps the linting shouldn't import it at this stage?

avivkeller:
The issue is the `prepare` script is running from the dependency, since `npm install` is ran on it, and husky isn't installed, since it's only a devDep

avivkeller:
Bump @nodejs/web-infra for approval

flakey5 (approved):

PR #8118, opened 2025-08-29
fix(crowdin): improve formatting workflow

avivkeller:
I had the wrong cwd when making changes to this workflow, `i18n` over `website-i18n`.

### DIFF
--- a/.github/workflows/translations-sync.yml
+++ b/.github/workflows/translations-sync.yml
@@ -106,6 +106,14 @@ jobs:
       - name: Install packages
         run: pnpm install --frozen-lockfile
 
+      - name: Patch version if the files changed
+        working-directory: packages/i18n
+        run: |
+          CHANGED_FILES=$(git diff --name-only HEAD^1 HEAD)
+          if [ -n "$CHANGED_FILES" ]; then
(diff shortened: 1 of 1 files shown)


avivkeller:
I don't know _why_ it's not patching the version in the `package.json`, this workflow _seems_ correct.

.github/workflows/translations-sync.yml
```diff
@@ -115,18 +115,18 @@ jobs:
 
       - name: Patch version if the files changed
         run: |
-          CHANGED_FILES=$(git diff --name-only origin/main HEAD -- packages/website-i18n)
+          CHANGED_FILES=$(git diff --name-only $GITHUB_SHA HEAD -- packages/website-i18n)
```
MattIPv4:
```suggestion
          CHANGED_FILES=$(git diff --name-only "$GITHUB_SHA" HEAD -- packages/website-i18n)
```
I don't think this could be abused, but quotes for safety

avivkeller:
> I don't know _why_ it's not patching the version in the `package.json`, this workflow _seems_ correct.

OMG I had the wrong cwd 🤦

MattIPv4 (approved):

PR #8120, opened 2025-08-29
[automated]: crowdin sync

nodejs-crowdin:
New Crowdin translations from the [Node.js Crowdin project](https://crowdin.com/project/nodejs-web)


.github/workflows/translations-sync.yml
```diff
@@ -115,18 +115,18 @@ jobs:
 
       - name: Patch version if the files changed
         run: |
-          CHANGED_FILES=$(git diff --name-only origin/main HEAD -- packages/website-i18n)
+          CHANGED_FILES=$(git diff --name-only $GITHUB_SHA HEAD -- packages/website-i18n)
```
MattIPv4, line 118:
```suggestion
          CHANGED_FILES=$(git diff --name-only "$GITHUB_SHA" HEAD -- packages/website-i18n)
```
I don't think this could be abused, but quotes for safety

avivkeller: Can you make this suggestion on #8118?

PR #8122, opened 2025-08-29
feat(linting): improve eslint

avivkeller:
Improves our ESLint configuration as follows;
- Performs the TypeScript ESLint and globals setup at root (as multiple packages use this)
- Matches `doc-kit` for some root rules
- Enables stylistic TypeScript rules, to maintain consistency
- Stops using Legacy ESLint syntax
- Uses `eslint-react` over `eslint-plugin-react`, since it has better TS support, and claims to be 7x faster than `eslint-plugin-react`

**Note**: pnpm removed the comments from the workspace file. Do we want to re-add these, or not?, since pnpm may remove them again.


ovflowd (approved):
LGTM! Good effort done here ✨

PR #8132, opened 2025-09-01
feat(shiki): add twoslash support

### LINKED ISSUE #7932: Type Info on the Docs & Guides

Provide type info for code snippets on the docs and guides. For reference, see Hono's docs: https://hono.dev/docs/concepts/middleware.

This is very much so in the backlog until we have less things to do

(slack convo https://openjs-foundation.slack.com/archives/CVAMEJ4UV/p1751225431726929)

avivkeller:
Fixes #7932

---

Introduces `@shikijs/twoslash` support in `@node-core/rehype-shiki`. Since `twoslash` depends on TypeScript declaration files, it cannot be bundled and must instead be listed under `serverExternalPackages`.

Because TypeScript declarations aren’t imported like regular files (they’re loaded through a TSVFS), they need to be explicitly included in the output bundle.

Because Cloudflare doesn't support WASM or Top-Level awaits, rehype-shiki has been rewritten to not rely on them.


AugustinMauroy:
Hype 😁 !

avivkeller:
Locally:
image

jakebailey:
Is this all server side?

I'm asking because given the TS port to Go, I'm not sure what the deal will be for shiki on the web; probably having to include a large Go binary in Wasm to work with all of the fancy stuff if not prerendered...

avivkeller:
> Is this all server side?
> 
> I'm asking because given the TS port to Go, I'm not sure what the deal will be for shiki on the web; probably having to include a large Go binary in Wasm to work with all of the fancy stuff if not prerendered...

All server-side, yes. We don't pass the complicated WASM binaries and Shiki configurations to the client. The client gets a basic syntax highlighter w/ Shiki's RegEx engine.

avivkeller:
@dario-piotrowicz Does this setup not support OpenNext? If not, we can make it conditional, like we do for the WASM engine

AugustinMauroy:
image

The box have a light mode but it's shouldn't since the shiki theme is always dark so the contrast isn't good

avivkeller:
@AugustinMauroy fixed

apps/site/pages/en/learn/typescript/transpile.md
```diff
@@ -88,31 +89,4 @@ const justine: User = {
 const isJustineAnAdult: string = isAdult(justine, "I shouldn't be here!");
 ```
 
-And this is what TypeScript has to say about this:
```
AugustinMauroy, line 91: This is kind of useful because it's the output that the reader/user will see in this terminal

packages/rehype-shiki/src/twoslash.css
AugustinMauroy, line 1: Is this file is minimified during publish process ?

Also should add docs somewhere about why here we don't use tailwind.

AugustinMauroy (commented):
Wow soo cool

canerakdas:
On the line where the error messages are displayed, the scrolled section looks cut off. Additionally, the text and background color used here don’t seem to have sufficient contrast ratio
image

The popover that’s supposed to display the types on horizontally scrollable sections doesn’t open in the correct position
image

For some colors, the contrast ratio doesn’t meet the requirement. The simplest solution might be to use the same background color as the codebox
image

Copy to clipboard also copies the types into the clipboard, for example;
```mjs
// server.mjs
import { function createServer<Request extends typeof IncomingMessage = typeof IncomingMessage, Response extends typeof ServerResponse = typeof ServerResponse>(requestListener?: RequestListener<Request, Response>): Server<Request, Response> (+1 overload)createServer } from 'node:http';

const const server: Server<typeof IncomingMessage, typeof ServerResponse>server = createServer<typeof IncomingMessage, typeof ServerResponse>(requestListener?: RequestListener<typeof IncomingMessage, typeof ServerResponse> | undefined): Server<...> (+1 overload)createServer((req: IncomingMessagereq, res: ServerResponse<IncomingMessage> & {
    req: IncomingMessage;
}res) => {
  res: ServerResponse<IncomingMessage> & {
    req: IncomingMessage;
}res.ServerResponse<IncomingMessage>.writeHead(statusCode: number, headers?: OutgoingHttpHeaders | OutgoingHttpHeader[]): ServerResponse<IncomingMessage> & {
    ...;
} (+1 overload)writeHead(200, { 'Content-Type': 'text/plain' });
  res: ServerResponse<IncomingMessage> & {
    req: IncomingMessage;
}res.Stream.Writable.end(chunk: any, cb?: () => void): ServerResponse<IncomingMessage> & {
    req: IncomingMessage;
} (+2 overloads)end('Hello World!\n');
});

// starts a simple http server locally on port 3000
const server: Server<typeof IncomingMessage, typeof ServerResponse>server.Server.listen(port?: number, hostname?: string, listeningListener?: () => void): Server<typeof IncomingMessage, typeof ServerResponse> (+8 overloads)listen(3000, '127.0.0.1', () => {
  var console: Consoleconsole.Console.log(message?: any, ...optionalParams: any[]): void (+1 overload)log('Listening on 127.0.0.1:3000');
});

// run with \`node server.mjs\`

```

Lastly, we also need to look at how this impacts page performance.
When checking the output HTML file sizes for our home page, I see:

Nodejs.org: `175.73 KB`
Preview: `228.79 KB`

It might be good to run a comparison with `@next/bundle-analyzer` to see how much load this adds to the pages 🤔

avivkeller:
> Lastly, we also need to look at how this impacts page performance.
> 
> When checking the output HTML file sizes for our home page, I see:
> 
> 
> 
> Nodejs.org: `175.73 KB`
> 
> Preview: `228.79 KB`
> 
> 
> 
> It might be good to run a comparison with `@next/bundle-analyzer` to see how much load this adds to the pages 🤔 

It's not _supposed_ to be passed to the client at all, so we should definitely investigate that

avivkeller:
> The popover that’s supposed to display the types on horizontally scrollable sections doesn’t open in the correct position

> Copy to clipboard also copies the types into the clipboard

Fixed in the latest commit, which introduced some styling issues (regarding wrapping).

packages/rehype-shiki/src/twoslash.css
avivkeller: There's no point in transpiling this CSS file, it's extremely simple

apps/site/pages/en/learn/typescript/transpile.md
avivkeller: I don't think the original version was very clear, the inline errors are clearer imo

avivkeller:
> On the line where the error messages are displayed, the scrolled section looks cut off. Additionally, the text and background color used here don’t seem to have sufficient contrast ratio

Is that even possible to fix? iiuc CSS doesn't allow us to set this element to width + overflow?

avivkeller:
The size increase comes from the additional HTML needed for these tooltips

avivkeller:
I don't like that it's all chunked on the main page, this would be better as an imported script, but that would need to be changed as part of our Markdown to JSX

dario-piotrowicz:
> @dario-piotrowicz Does this setup not support OpenNext? If not, we can make it conditional, like we do for the WASM engine

Sorry @avivkeller, I only now saw your message 🙇 

Tomorrow I'll have a closer look 🙇 , however by quickly skimming though your changes I didn't spot anything that could be problematic, on the contrary, it seems like you are conditionally using the wasm implementation only if `Cloudflare` is not in `global` 🤔 

Do you have any suspicion on what might be causing the Cloudflare incompatibility? 🙂

dario-piotrowicz:
@avivkeller I'm looking into the issue, for now what I've seen is that this problematic `WebAssembly.instantiate` call is being made by shiki:
https://github.com/shikijs/shiki/blob/e46261819005a7519c64b54679a504d783dfdcbc/packages/engine-oniguruma/src/wasm-inlined.ts#L8

but I thought that in our setup shiki would not use wasm? 🤔

avivkeller:
Our Shiki setup doesn't use WASM when `"Cloudflare" in global`, can you point me to how you know this is the cause of the issue, so I can investigate why it's occurring?

dario-piotrowicz:
> Our Shiki setup doesn't use WASM when `"Cloudflare" in global`

Yes, that's exactly what I saw/thought, but definitely shiki seems to be trying to instantiate wasm anyways here 😕 

(Also via the debugger I can indeed see that `Cloudflare` is in `global` when the error is thrown 😕)

> can you point me to how you know this is the cause of the issue, so I can investigate why it's occurring?

In the `site` directory you can build the worker via: `pnpm cloudflare:build:worker` then you spin it up with `pnpm cloudflare:preview` when you do that, before doing anything press `d` that will open your chrome devtools, there make sure to have ticked the `pause on caught exceptions` checkbox and then simply visit the site, then in the devtools keep resuming the script until it hits the wasm issue:
Screenshot 2025-09-11 at 13 47 30

<details><summary>Minification</summary>
<p>

To make the code a bit more decipherable I've disable minification in the next config:
Screenshot 2025-09-11 at 13 52 49

</p>
</details>

packages/rehype-shiki/src/index.mjs
```diff
@@ -15,46 +14,85 @@ import shellSessionLanguage from 'shiki/langs/shellsession.mjs';
...
+      '@shikijs/engine-javascript'
+    );
+    return createJavaScriptRegexEngine();
+  }
+
+  const { createOnigurumaEngine } = await import('@shikijs/engine-oniguruma');
+  return createOnigurumaEngine(
+    typeof wasm === 'boolean' ? await import('shiki/wasm') : wasm
+  );
+}
```
avivkeller, line 42: @dario-piotrowicz i see the problem... this check is backwards lol

dario-piotrowicz: lol 🤦 

good find! 🚀

avivkeller:
FYI @dario-piotrowicz, still not working

avivkeller:
Bump @dario-piotrowicz

avivkeller:
@nodejs/nodejs-website @nodejs/web-infra for reviews. Currently, this is disabled on CF, but there's a TODO to enable it.

apps/site/pages/en/learn/typescript/introduction.md
```diff
@@ -70,16 +70,15 @@ npm add --save-dev @types/node
 
 These type definitions allow TypeScript to understand Node.js APIs and provide proper type checking and autocompletion when you use functions like `fs.readFile` or `http.createServer`. For example:
 
-```js
```
AugustinMauroy, line 73: This is supper weird. I see the idea of showing auto completion.
But it's strange 



When I was looking at this, I was expecting an interactive thing.

AugustinMauroy (commented):
Light/dark theme didn't work well but I didn't see why
Capture d’écran 2025-09-25 à 08 53 20
Capture d’écran 2025-09-25 à 08 53 35

avivkeller:
I'll get to Augustin's review today, @nodejs/web any other comments / concerns

apps/site/pages/en/learn/typescript/introduction.md
avivkeller: How so? I can change it if it's not good

flakey5 (approved):
thank you!

apps/site/pages/en/learn/typescript/introduction.md
AugustinMauroy: 1. (big) solution having "editor" that user can use to learn such as a complet learning platform. IMO not good idea.
2. just don't use this feature

avivkeller: I think it makes sense, as we are discussing auto-complete, cc @nodejs/nodejs-website @nodejs/typescript for opinions

avivkeller:
I need an @nodejs/nodejs-website approval, and then I'll rebase

apps/site/pages/en/learn/typescript/introduction.md
ovflowd: Why would we want autocomplete here? Aren't these read-only code boxes?

avivkeller: The caption discusses auto complete, so I modified the code box to show what an editor would show. I can also revert to the previous iteration

packages/rehype-shiki/src/index.mjs
```diff
@@ -15,46 +14,85 @@ import shellSessionLanguage from 'shiki/langs/shellsession.mjs';
...
+
+/**
+ * Creates the appropriate regex engine based on configuration
+ * @param {HighlighterOptions} options - Configuration options
+ */
+async function getEngine({ wasm = false }) {
+  if (wasm) {
+    const { createOnigurumaEngine } = await import('@shikijs/engine-oniguruma');
+    return createOnigurumaEngine(
+      typeof wasm === 'boolean' ? await import('shiki/wasm') : wasm
```
ovflowd: What are the situations a custom wasm engine is passed?

avivkeller: None, at the moment, but I want the ability to pass one in the future, if a WASM engine that supports CF workers is released

packages/ui-components/src/MDX/Tooltip.tsx
```diff
@@ -0,0 +1,33 @@
+import * as TooltipPrimitive from '@radix-ui/react-tooltip';
+import classNames from 'classnames';
+import type { FC, HTMLAttributes, PropsWithChildren } from 'react';
+
+import styles from '../Common/Tooltip/index.module.css';
```
ovflowd: @ imports don't work here?

packages/rehype-shiki/src/plugin.mjs
```diff
@@ -162,30 +160,35 @@ export default function rehypeShikiji() {
...
-      const showCopyButton = getMetaParameter(
-        preElement.data?.meta,
-        'showCopyButton'
-      );
-
       // Adds a Copy Button to the CodeBox if requested as an additional parameter
       // And avoids setting the property (overriding) if undefined or invalid value
-      if (showCopyButton && ['true', 'false'].includes(showCopyButton)) {
-        children[0].properties.showCopyButton = showCopyButton;
+      if (
```
ovflowd, line 187: This is some odd sort of comparison.

packages/rehype-shiki/src/plugin.mjs
```diff
@@ -4,34 +4,34 @@ import classNames from 'classnames';
...
   }
 
-  const matches = meta.match(new RegExp(`${key}="(?<parameter>[^"]*)"`));
-  const parameter = matches?.groups.parameter;
+  let match;
 
-  return parameter !== undefined && parameter.length > 0
-    ? parameter
-    : undefined;
+  while ((match = rMeta.exec(meta)) !== null) {
```
ovflowd, line 30: This feels like some voodoo wizardry and so prone of going wrong I guess

packages/rehype-shiki/src/index.mjs
```diff
@@ -15,46 +14,85 @@ import shellSessionLanguage from 'shiki/langs/shellsession.mjs';
...
+
+  if (options) {
+    const { twoslash } = await import('./transformers/twoslash/index.mjs');
+    transformers.push(twoslash(options));
+  }
+
+  return transformers;
+}
+
+export const LANGS = [
```
ovflowd, line 59: Does this need to be on the same file?

apps/site/next.config.mjs
```diff
@@ -51,6 +51,16 @@ const nextConfig = {
       },
     ],
   },
+  serverExternalPackages: ['twoslash'],
+  outputFileTracingIncludes: {
+    // Twoslash needs TypeScript declarations to function, and, by default,
+    // Next.js strips them for brevity. Therefore, they must be explicitly
+    // included.
```
ovflowd: this word could be on the previous line lol

packages/rehype-shiki/src/plugin.mjs
avivkeller: I didn't modify the check, it's from the previous implementation

packages/rehype-shiki/src/transformers/twoslash/index.mjs
```diff
@@ -0,0 +1,48 @@
...
+        completionPopup: {
+          tagName: 'MDXTooltipContent',
+          properties: {
+            align: 'start',
+          },
+        },
+        completionCompose: compose,
+      },
+    },
+    throws: false,
```
ovflowd, line 46: I feel you could just pass options and default to empty object on the args

packages/rehype-shiki/src/transformers/twoslash/index.mjs
```diff
@@ -0,0 +1,48 @@
...
+  {
+    type: 'element',
+    tagName: 'MDXTooltipTrigger',
+    children: [token || cursor],
+    properties: { className: ['twoslash-hover'] },
+  },
+  popup,
+];
+
+export const twoslash = options =>
```
ovflowd:
```suggestion
export const twoslash = (options = {}) =>
```

packages/rehype-shiki/src/index.mjs
avivkeller: Ideally yes, the `index.mjs` `LANGS` and the `minimal.mjs` `LANGS`

packages/rehype-shiki/src/transformers/twoslash/index.mjs
avivkeller: No, it's intentionally like this, since `options` may be a boolean

packages/rehype-shiki/src/transformers/twoslash/index.mjs
avivkeller: See https://github.com/nodejs/nodejs.org/pull/8132#discussion_r2417784626

packages/rehype-shiki/src/plugin.mjs
avivkeller: How so? The RegEx (AFAICT) is pretty robust.

apps/site/mdx/components.mjs
```diff
@@ -3,6 +3,11 @@
 import BadgeGroup from '@node-core/ui-components/Common/BadgeGroup';
 import Blockquote from '@node-core/ui-components/Common/Blockquote';
 import MDXCodeTabs from '@node-core/ui-components/MDX/CodeTabs';
+import {
```
ovflowd, line 6: These should all be separated files

ovflowd: Also are these client components? This file should only contain server components. There's another file for client components no?

packages/rehype-shiki/src/transformers/twoslash/index.mjs
ovflowd: Why would it also be a boolean? Can it only be an object please?

packages/rehype-shiki/src/transformers/twoslash/index.mjs
ovflowd: -1 it being a boolean and object. Keep types consistent.

packages/rehype-shiki/src/highlighter.mjs
```diff
@@ -3,45 +3,47 @@ import shikiNordTheme from 'shiki/themes/nord.mjs';
...
-  // @see https://www.w3.org/WAI/WCAG21/Understanding/contrast-minimum.html
+  // in the Codebox component do not comply with accessibility standards.
+  // See: https://www.w3.org/WAI/WCAG21/Understanding/contrast-minimum.html
   colorReplacements: { '#616e88': '#707e99' },
   ...shikiNordTheme,
 };
 
+export const getLanguageByName = (language, langs) =>
+  langs.find(
+    ({ name, aliases }) =>
```
ovflowd, line 16: nit: Way too many toLowerCase calls.

packages/rehype-shiki/src/index.mjs
ovflowd: Let's then only implement that when it becomes a need, let's keep it simple :)

ovflowd: As we have no idea if and when it would even become a thing. Although it sucks we cannot use the WASM engine which is much faster on Cloudflare.

packages/rehype-shiki/src/index.mjs
ovflowd: IDK, feels like this should have a dedicated langs.mjs file tho

packages/rehype-shiki/src/plugin.mjs
ovflowd: This regex pattern seems to be vulnerable to ReDoS.

ovflowd (approved):
I have a few reservations, would love some comments to be addressed and some benchmark images of build times / bundle sizes with this addition.

This is a cool visual addition but if it's to heavy, might not be worth it.

apps/site/mdx/components.mjs
avivkeller: No, we merged them into a single file in #8167

packages/rehype-shiki/src/index.mjs
avivkeller: I don't think so. The `index.mjs` exports all of the highlighting utilities needed for our server, and the `minimal.mjs` exports all for our client. A `langs.mjs` file for `index.mjs` doesn't make sense to me.

WDYT @nodejs/nodejs-website

packages/rehype-shiki/src/plugin.mjs
avivkeller: I don't think it is:
image
image
image

avivkeller:
> I have a few reservations, would love some comments to be addressed and some benchmark images of build times / bundle sizes with this addition.

The bundle is a bit bigger, since it needs to load the extra JS (See #8160). The most recent Twoslash build took ~1:33, almost identical to the most recent `main` build (~1:29)

ovflowd:
Bypassing ownership from typescript team as it is a false positive.

PR #8134, opened 2025-09-01
refactor: download archive page unnecessary checks

canerakdas:
## Description

I’ve made some small refactors on the download archive page, per related comments here;
https://github.com/nodejs/nodejs.org/pull/7794#discussion_r2302497142
https://github.com/nodejs/nodejs.org/pull/7794#discussion_r2302526703

In addition to this PR, I’ve also added the changelog of the displayed version to the section where we provide version-related information. Since this information is valuable, I believe placing the links higher up makes more sense 

### Check List


canerakdas:
I'll check why the `/en/download/archive` is returns to 500

apps/site/redirects.json
```diff
@@ -332,6 +332,10 @@
       "source": "/:locale/download/source-code/current",
       "destination": "/:locale/download/current"
     },
+    {
```
canerakdas: Since the file (`/pages/download/archive/index.mdx`) is under the pages directory, it maps to `/download/archive` (When we remove version control, it drops to 500 page because there is no version named `archive`). In fact, this file doesn’t really need to live under `pages` directory

- Store the dynamically generated markdown files in a different directory instead of `pages`
- Redirect `/download/archive` to `/download/archive/current`
- Use an existing version(eg: `v22.0.0.mdx`) as the filename instead of `index.mdx`

IDK wich one suits for us better, for a quick fix i added this redirect :shipit:

canerakdas: ah this will also cause errors at build time 🤔

canerakdas: IDK, instead of going for an overkill solution, I added back the release check for now

avivkeller (approved):

ovflowd (approved):
SGTM

AugustinMauroy (approved):

PR #8209, opened 2025-10-06
fix: download options alignment

### LINKED ISSUE #8208: Misalignment of download options
### URL:

### Browser Name:

Chrome

### Browser Version:

140.0.7339.128 (Official Build) (64-bit)

### Operating System:

Windows 10

Inline download behavior alignment changes based on the presence of an icon in the select/dropdown

Image

Image

Affected code is https://github.com/nodejs/nodejs.org/blob/main/packages/ui-components/src/Common/Select/index.module.css#L52 and a variant of it lower in the file when `.inline`

Note the source code / usage is markdown within https://github.com/nodejs/nodejs.org/blob/main/apps/site/pages/en/download/index.mdx?plain=1#L8 - so options to wrap are limited unless we get more intrusive.

canerakdas:
## Description

This PR aims to fix the height of the Skeleton used with Select and to adjust the alignment of the select on the download page

## Validation

Download select must be aligned

## Related Issues

Fixes #8208

### DIFF
--- a/packages/ui-components/src/Common/Select/index.module.css
+++ b/packages/ui-components/src/Common/Select/index.module.css
@@ -3,7 +3,8 @@
 .select {
   @apply inline-flex
     flex-col
-    gap-1.5;
+    gap-1.5
+    align-middle;
 
   .label {
     @apply block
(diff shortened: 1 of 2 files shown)


ovflowd (approved):

bmuenzenmeyer (approved):
works! 

i was messing with `items-center` in my debugging, and missed skeleton completely!

AugustinMauroy (approved):

aymen94 (approved):

bmuenzenmeyer:
looking for fast track 👍 to merge this bug fix

mikeesto (approved):

PR #8210, opened 2025-10-07
feat(orama): allow manual syncing

avivkeller:


justinemmanuel363-spec:
res.statusCode = 200;

AugustinMauroy:
just what is the use case of that ?

avivkeller:
To allow deploying Orama from non-default branches to test credentials, etc

flakey5 (approved):

avivkeller:
I'm fast tracking this so I can test the new Orama

PR #8227, opened 2025-10-15
feat(ci): Add workflow to flag future-dated blog posts

sozua:
## Description

Adds a GitHub Actions workflow that scans PR blog posts for future publish dates and keeps an automated review comment updated, with tests covering both the detector and formatter.

## Validation

- [Example with one blog post](https://github.com/sozua/nodejs.org/pull/7)
- [Example with multiple blog posts](https://github.com/sozua/nodejs.org/pull/6)
- [Example with a post that was fixed](https://github.com/sozua/nodejs.org/pull/8)
- The JS code also have unit tests with 100% code coverage

## Related Issues

- #7845 

### Check List

### DIFF
--- a/.github/workflows/check-post-publish-date.yml
+++ b/.github/workflows/check-post-publish-date.yml
@@ -1,10 +1,10 @@
 # Security Notes
-# This workflow uses `pull_request_target`, so will run against all PRs automatically (without approval), be careful with allowing any user-provided code to be run here
+# This workflow uses `pull_request`, so will run against PRs in their own context; be careful with allowing any user-provided code to be run here
 # Only selected Actions are allowed within this repository. Please refer to (https://github.com/nodejs/nodejs.org/settings/actions)
 # for the full list of available actions. If you want to add a new one, please reach out a maintainer with Admin permissions.
 # REVIEWERS, please always double-check security practices before merging a PR that contains workflow changes!!
 # AUTHORS, please only use actions with explicit SHA references, and avoid using `@master` or `@main` references or `@version` tags.
-# MERGE QUEUE NOTE: This workflow runs on both `pull_request_target` and `merge_group` to ensure blog post dates are validated before merging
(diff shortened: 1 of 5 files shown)


avivkeller (changes requested):
A few concerns:
- Firstly, I don't really think this is needed, but I won't block given that others may like this
- This is quite a large and complex PR for something that can probably be very simple. We can probably just use a RegEx / dependency to get the date, and check if it's in the future, and do all other processing in the workflow itself.
- Is there a lint plugin that'll do this for us? Can one be made instead?

MattIPv4 (changes requested):
++ to Aviv's concerns.

I'm not sure I understand why we particularly need this, has this been an issue where future dates posts have been merged in the past? And even if we do need this, over 1,000 LoC to check some dates seems very excessive and smells like it'll be a maintenance burden down the line?

avivkeller:
> has this been an issue where future dates posts have been merged in the past?

It happened once, but IMO it's the responsibility of the author to ensure that doesn't happen, not ours.

I've added the original issue to the agenda for discussion.

sozua:
Hi guys! First off, thanks for the review and for being so quick!

About the PR size/complexity: I agree - I think it can definitely be simplified into something much smaller since it isnt something that happens very often. I ended up misunderstanding the requirements from that discussion.

I liked the suggestions, and i think the idea of creating a plugin to trigger eslint warnings locally based on the front-matter metadata could work. If that sounds good, I can try work on this solution in the next hours and update this PR when done. What do you think, @avivkeller?

avivkeller:
That's definitely a step in the right direction, but I've added the original issue to our agenda to discuss whether this is really needed.

Nevertheless, you certainly put a lot of effort here, and we are extremely grateful.

sozua:
> That's definitely a step in the right direction, but I've added the original issue to our agenda to discuss whether this is really needed.
> 
> Nevertheless, you certainly put a lot of effort here, and we are extremely grateful.

Ok! I'll wait for the discussion and I'm excited to help if needed

PR #8228, opened 2025-10-15
fix(OpenNext): build blog data

avivkeller:
We need to build the blog data before deploying

### DIFF
--- a/.github/workflows/tmp-cloudflare-open-next-deploy.yml
+++ b/.github/workflows/tmp-cloudflare-open-next-deploy.yml
@@ -50,10 +50,6 @@ jobs:
       - name: Install packages
         run: pnpm install --frozen-lockfile
 
-      - name: Build open-next site
-        working-directory: apps/site
-        run: node --run cloudflare:build:worker
-
       - name: Deploy open-next site
         working-directory: apps/site
         run: node --run cloudflare:deploy
--- a/.github/workflows/tmp-cloudflare-open-next-deploy.yml
+++ b/.github/workflows/tmp-cloudflare-open-next-deploy.yml
@@ -54,6 +54,10 @@ jobs:
         working-directory: apps/site
         run: node --run build:blog-data
 
+      - name: Build open-next site
+        working-directory: apps/site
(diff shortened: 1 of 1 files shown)


MattIPv4:
Can we dispatch the workflow against this branch to confirm this fix works?

dario-piotrowicz:
I am not sure why this is necessary, wasn't the blog building done in a postinstall script?

dario-piotrowicz:
> I am not sure why this is necessary, wasn't the blog building done in a postinstall script?

Ah yeah there are `pre*` scripts that should be triggering that :thinking: 
https://github.com/nodejs/nodejs.org/blob/bd92fd2a5cc0fe5e6aa4c82cd61d779c79de9851/apps/site/package.json#L5

MattIPv4:
`node --run` does not run pre/post hooks, that is the one major difference vs. what `npm run` et al do

avivkeller:
> Can we dispatch the workflow against this branch to confirm this fix works?

I did when I opened the PR, apologies for not linking the run, https://github.com/nodejs/nodejs.org/actions/runs/18532159595

avivkeller:
> `node --run` does not run pre/post hooks, that is the one major difference vs. what `npm run` et al do

Even if it did, those lifecycle scripts wouldn't apply to the worker variants, right?

dario-piotrowicz:
> `node --run` does not run pre/post hooks, that is the one major difference vs. what `npm run` et al do

ah ok, thanks for the extra context :slightly_smiling_face:

dario-piotrowicz:
> > `node --run` does not run pre/post hooks, that is the one major difference vs. what `npm run` et al do
> 
> Even if it did, those lifecycle scripts wouldn't apply to the worker variants, right?

oh yeah you're right :confused:

dario-piotrowicz (approved):
LGTM :slightly_smiling_face:

avivkeller:
@nodejs/web-infra feel free to merge once approved, I'm requesting fast track

MattIPv4 (approved):

PR #8231, opened 2025-10-15
feat(cicd): automate the creation of blog posts

avivkeller:
cc @nodejs/releasers

### DIFF
--- a/.github/workflows/create-release-post.yml
+++ b/.github/workflows/create-release-post.yml
@@ -1,0 +1,50 @@
+# Security Notes
+# Only selected Actions are allowed within this repository. Please refer to (https://github.com/nodejs/nodejs.org/settings/actions)
+# for the full list of available actions. If you want to add a new one, please reach out a maintainer with Admin permissions.
+# REVIEWERS, please always double-check security practices before merging a PR that contains Workflow changes!!
+# AUTHORS, please only use actions with explicit SHA references, and avoid using `@master` or `@main` references or `@version` tags.
+
+name: Create Release Blog Post
+
+on:
+  workflow_dispatch:
+    inputs:
+      version:
+        type: string
+
+defaults:
+  run:
+    # This ensures that the working directory is the root of the repository
+    working-directory: ./
+
+permissions:
+  contents: write
+
+jobs:
+  create-post:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: nodejs/web-team/actions/setup-environment@2c2897a93eb99b4cdca270729100bc0887c758d9
+        with:
+          pnpm: true
+          use-version-file: true
+
+      - id: create-release-post
+        working-directory: apps/site
+        run: node --run scripts:release-post ${{ inputs.version }} >> $GITHUB_OUTPUT
+
+      - name: Open pull request
+        uses: gr2m/create-or-update-pull-request-action@b65137ca591da0b9f43bad7b24df13050ea45d1b # v1.10.1
+        # Creates a PR or update the Action's existing PR, or
+        # no-op if the base branch is already up-to-date.
+        env:
+          GITHUB_TOKEN: ${{ secrets.GH_USER_TOKEN }}
+        with:
+          author: Node.js GitHub Bot <github-bot@iojs.org>
+          branch: release-${{ steps.create-release-post.output.version }}
+          body: This PR was generated automatically.
+          commit-message: 'feat(blog): create post for ${{ steps.create-release-post.output.version }}'
+          labels: fast-track
+          title: 'feat(blog): create post for ${{ steps.create-release-post.output.version }}'
+          update-pull-request-title-and-body: true
--- a/.github/workflows/create-release-post.yml
+++ b/.github/workflows/create-release-post.yml
@@ -42,12 +42,12 @@ jobs:
           GITHUB_TOKEN: ${{ secrets.GH_USER_TOKEN }}
         with:
           author: Node.js GitHub Bot <github-bot@iojs.org>
-          branch: release-${{ steps.create-release-post.output.version }}
+          branch: release-${{ steps.create-release-post.outputs.version }}
           body: |
-            Creates a new blog post for ${{ steps.create-release-post.output.version }}.
+            Creates a new blog post for ${{ steps.create-release-post.outputs.version }}.
 
             Check this workflow's logs at ${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}.
-          commit-message: 'feat(blog): create post for ${{ steps.create-release-post.output.version }}'
+          commit-message: 'feat(blog): create post for ${{ steps.create-release-post.outputs.version }}'
           labels: fast-track
-          title: 'feat(blog): create post for ${{ steps.create-release-post.output.version }}'
+          title: 'feat(blog): create post for ${{ steps.create-release-post.outputs.version }}'
           update-pull-request-title-and-body: true
--- a/.github/workflows/create-release-post.yml
+++ b/.github/workflows/create-release-post.yml
@@ -45,7 +45,7 @@ jobs:
           branch: release-${{ steps.create-release-post.output.version }}
           body: |
             Creates a new blog post for ${{ steps.create-release-post.output.version }}.
-            
+
             Check this workflow's logs at ${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}.
           commit-message: 'feat(blog): create post for ${{ steps.create-release-post.output.version }}'
           labels: fast-track
--- a/.github/workflows/create-release-post.yml
+++ b/.github/workflows/create-release-post.yml
@@ -10,8 +10,9 @@ on:
   workflow_dispatch:
     inputs:
       version:
-        description: The version to generate a blog post for. If none provided, use the latest release.
+        description: The version to generate a blog post for.
         type: string
+        required: true
 
 defaults:
   run:
@@ -31,24 +32,22 @@ jobs:
           pnpm: true
           use-version-file: true
 
-      - id: create-release-post
-        working-directory: apps/site
-        run: node --run scripts:release-post ${{ inputs.version }}
+      - run: node --run scripts:release-post "$VERSION"
+        env:
+          VERSION: ${{ inputs.version }}
 
       - name: Open pull request
         uses: gr2m/create-or-update-pull-request-action@b65137ca591da0b9f43bad7b24df13050ea45d1b # v1.10.1
         # Creates a PR or update the Action's existing PR, or
         # no-op if the base branch is already up-to-date.
-        env:
-          GITHUB_TOKEN: ${{ secrets.GH_USER_TOKEN }}
         with:
-          author: Node.js GitHub Bot <github-bot@iojs.org>
-          branch: release-${{ steps.create-release-post.outputs.version }}
+          update-pull-request-title-and-body: true
+          branch: release-${{ inputs.version }}
           body: |
-            Creates a new blog post for ${{ steps.create-release-post.outputs.version }}.
+            Creates a new blog post for ${{ inputs.version }}
 
             Check this workflow's logs at ${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}.
-          commit-message: 'feat(blog): create post for ${{ steps.create-release-post.outputs.version }}'
+          commit-message: 'feat(blog): create post for ${{ inputs.version }}'
           labels: fast-track
-          title: 'feat(blog): create post for ${{ steps.create-release-post.outputs.version }}'
-          update-pull-request-title-and-body: true
+          title: 'feat(blog): create post for ${{ inputs.version }}'
+          draft: true
--- a/.github/workflows/create-release-post.yml
+++ b/.github/workflows/create-release-post.yml
@@ -10,6 +10,7 @@ on:
   workflow_dispatch:
     inputs:
       version:
+        description: The version to generate a blog post for. If none provided, use the latest release.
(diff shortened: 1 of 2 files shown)


.github/workflows/create-release-post.yml
```diff
@@ -0,0 +1,50 @@
...
+      - name: Open pull request
+        uses: gr2m/create-or-update-pull-request-action@b65137ca591da0b9f43bad7b24df13050ea45d1b # v1.10.1
+        # Creates a PR or update the Action's existing PR, or
+        # no-op if the base branch is already up-to-date.
+        env:
+          GITHUB_TOKEN: ${{ secrets.GH_USER_TOKEN }}
+        with:
+          author: Node.js GitHub Bot <github-bot@iojs.org>
+          branch: release-${{ steps.create-release-post.output.version }}
+          body: This PR was generated automatically.
```
richardlau: Could we link through to the run?
ref: https://docs.github.com/en/actions/reference/workflows-and-actions/variables#default-environment-variables

> If you need to use a workflow run's URL from within a job, you can combine these variables: 
> `$GITHUB_SERVER_URL/$GITHUB_REPOSITORY/actions/runs/$GITHUB_RUN_ID`

flakey5 (approved):

.github/workflows/create-release-post.yml
avivkeller: Fixed! (I know lint is failing, I'll fix it once I get back from JSConf)

avivkeller:
https://github.com/nodejs/nodejs.org/labels/blocked by https://github.com/nodejs/admin/pull/1016

ovflowd (approved):

.github/workflows/create-release-post.yml
```diff
@@ -0,0 +1,54 @@
...
+
+    steps:
+      - uses: nodejs/web-team/actions/setup-environment@2c2897a93eb99b4cdca270729100bc0887c758d9
+        with:
+          pnpm: true
+          use-version-file: true
+
+      - id: create-release-post
+        working-directory: apps/site
+        run: node --run scripts:release-post ${{ inputs.version }}
```
aduh95: to protect against shell injection
```suggestion
        run: node --run scripts:release-post "$VERSION"
        env:
          VERSION: ${{ inputs.version }}
```

.github/workflows/create-release-post.yml
```diff
@@ -0,0 +1,54 @@
...
+        with:
+          author: Node.js GitHub Bot <github-bot@iojs.org>
+          branch: release-${{ steps.create-release-post.outputs.version }}
+          body: |
+            Creates a new blog post for ${{ steps.create-release-post.outputs.version }}.
+
+            Check this workflow's logs at ${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}.
+          commit-message: 'feat(blog): create post for ${{ steps.create-release-post.outputs.version }}'
+          labels: fast-track
+          title: 'feat(blog): create post for ${{ steps.create-release-post.outputs.version }}'
```
aduh95:
```suggestion
          branch: release-${{ inputs.version }}
          body: |
            Creates a new blog post for ${{ inputs.version }}

            Check this workflow's logs at ${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}.
          commit-message: 'feat(blog): create post for ${{ inputs.version }}'
          labels: fast-track
          title: 'feat(blog): create post for ${{ inputs.version }}'
```

apps/site/scripts/release-post/index.mjs
```diff
@@ -263,7 +263,15 @@ if (import.meta.url.startsWith('file:')) {
...
-        filepath => console.log('Release post created:', filepath),
+        async ([filepath, { version }]) => {
+          console.log('Release post created:', filepath);
+          if (process.env.GITHUB_OUTPUT) {
+            await appendFile(
+              process.env.GITHUB_OUTPUT,
+              `version=v${version}\n`
+            );
+          }
+        },
```
aduh95:
```suggestion
        filepath => console.log('Release post created:', filepath),
```

.github/workflows/create-release-post.yml
avivkeller: It's possible that the input version is blank, and the latest version is used, that's why we output it in the script

aduh95: In any case, let's not put GHA specific code in `apps/site/scripts/release-post/index.mjs`, instead we should add a flag to output the version to stdout

.github/workflows/create-release-post.yml
```diff
@@ -0,0 +1,54 @@
...
+defaults:
+  run:
+    # This ensures that the working directory is the root of the repository
+    working-directory: ./
+
+permissions:
+  contents: write
+
+jobs:
+  create-post:
```
mcollina, line 26: Please tie this to a specific `Environment` in GitHub. While an imperfect solution, it would prevent people without that permission from accessing the token and mitigate security risks. In other terms, a normal PR would not have access to that token.

mcollina (changes requested):

mcollina:
@RafaelGSS @nodejs/security-wg please review this.

.github/workflows/create-release-post.yml
```diff
@@ -0,0 +1,54 @@
...
+# AUTHORS, please only use actions with explicit SHA references, and avoid using `@master` or `@main` references or `@version` tags.
+
+name: Create Release Blog Post
+
+on:
+  workflow_dispatch:
+    inputs:
+      version:
+        description: The version to generate a blog post for. If none provided, use the latest release.
+        type: string
```
RafaelGSS, line 14:
```suggestion
        type: string
        required: true
```

Let's always request a version string.

.github/workflows/create-release-post.yml
aduh95:
```suggestion
        run: echo "version=$(node --run scripts:release-post "$VERSION" --output-version)" >> "$GITHUB_OUTPUT"
        env:
          VERSION: ${{ inputs.version }}
```

.github/workflows/create-release-post.yml
```diff
@@ -0,0 +1,54 @@
...
+          GITHUB_TOKEN: ${{ secrets.GH_USER_TOKEN }}
+        with:
+          author: Node.js GitHub Bot <github-bot@iojs.org>
+          branch: release-${{ steps.create-release-post.outputs.version }}
+          body: |
+            Creates a new blog post for ${{ steps.create-release-post.outputs.version }}.
+
+            Check this workflow's logs at ${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}.
+          commit-message: 'feat(blog): create post for ${{ steps.create-release-post.outputs.version }}'
+          labels: fast-track
```
RafaelGSS, line 51: shouldn't it have the gh: pull_request label too?

.github/workflows/create-release-post.yml
```diff
@@ -0,0 +1,54 @@
+# Security Notes
+# Only selected Actions are allowed within this repository. Please refer to (https://github.com/nodejs/nodejs.org/settings/actions)
+# for the full list of available actions. If you want to add a new one, please reach out a maintainer with Admin permissions.
+# REVIEWERS, please always double-check security practices before merging a PR that contains Workflow changes!!
+# AUTHORS, please only use actions with explicit SHA references, and avoid using `@master` or `@main` references or `@version` tags.
+
+name: Create Release Blog Post
+
+on:
+  workflow_dispatch:
```
aduh95, line 10: Have we considered using [`release`](https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows#release) trigger on the nodejs/node repo instead? I guess the workflow on nodejs/node could call this one

RafaelGSS: That was my initial thought. Whenever a tag is pushed to nodejs/node it would trigger this repository. But then, we'll need to work around situations like the v24.10.0 or similar - or in cases where SHASUMS wasn't sent to the dist machine due to load conditions.

targos: Pushing the tag doesn't trigger the `release` event. It's triggered by the last manual step that we do (https://github.com/nodejs/node/blob/main/doc/contributing/releases.md#18-create-the-release-on-github).

aduh95: > But then, we'll need to work around situations like the v24.10.0 or similar

That should be fine, we only open the PR, the releaser can add a request for changes if it's not ready to land. My suggestion is use the "release" trigger, not the "tag" one, so another option for the releaser would be to not create the GH release if it's not ready for publication

avivkeller: > I guess the workflow on nodejs/node could call this one

That's the plan. We do something similar in the release worker

.github/workflows/create-release-post.yml
avivkeller: This won't run on PRs, only dispatches (which requires `write` access to the repo anyway)

.github/workflows/create-release-post.yml
avivkeller: Why? We have the ability to infer it from the latest version

.github/workflows/create-release-post.yml
avivkeller: No, those checks are optional, and are related to lighthouse and orama, which aren't really affected by the addition of a blog post

.github/workflows/create-release-post.yml
aduh95: If the goal is to have this workflow called by another workflow, it’s indeed a bit moot to keep this thing optional. Let’s make it mandatory and simplify the implementation

avivkeller: In https://github.com/nodejs/node/blob/main/.github/workflows/update-release-links.yml#L18, we don't pass the version to that workflow, I would assume it wouldn't be passed to this one.

aduh95: > I would assume it wouldn't be passed to this one.

I would not assume that, that would be a very suboptimal decision

.github/workflows/create-release-post.yml
mcollina: Exactly. But adding the secret would make it accessible to every CI job running, which is a significantly bigger surface area.

.github/workflows/create-release-post.yml
RafaelGSS: For release blog posts, they aren't optional; at least, we always need to add this flag before merging the PR. Not sure if things have changed recently. cc: @ovflowd

avivkeller: Things have changed (a month or two ago ish, I don't recall exactly)

Those checks are now optional, since we run the majority of checks on the `pull_request` trigger.

.github/workflows/create-release-post.yml
avivkeller: We've settled on _not_ using the secret. Pushing a commit shortly

avivkeller:
The passing of an input version is no longer optional, and a draft PR is used in place of requiring a custom token

avivkeller:
@mcollina could you please take another look?

mcollina (approved):
lgtm

MattIPv4 (approved):

aymen94 (approved):

marco-ippolito:
Just a random idea, maybe we could trigger automatically when a release is created on nodejs/node with https://docs.github.com/en/webhooks/webhook-events-and-payloads#repository_dispatch

richardlau:
> Just a random idea, maybe we could trigger automatically when a release is created on nodejs/node with https://docs.github.com/en/webhooks/webhook-events-and-payloads#repository_dispatch

I think that's worth exploring separately from this. Just having this (and being able to manually trigger) would be a good stepping point -- for me creating the blog posts is currently one of the more time consuming parts of the release process on release day (due to needing to switch to a dev environment with pnpm and Husky pre-commit hooks).

richardlau (approved):

MattIPv4:
> Just a random idea, maybe we could trigger automatically when a release is created on nodejs/node with https://docs.github.com/en/webhooks/webhook-events-and-payloads#repository_dispatch

Can this be triggered from https://github.com/nodejs/node/blob/main/.github/workflows/update-release-links.yml, which is the same automation that triggers the release worker to update the files available when a new release is cut?

avivkeller:
That's the plan I had in mind

avivkeller:
See https://github.com/nodejs/node/pull/60366

avivkeller:
@nodejs/web-infra 
```
The action gr2m/create-or-update-pull-request-action@b65137ca591da0b9f43bad7b24df13050ea45d1b is not allowed in nodejs/nodejs.org because all actions must be from a repository owned by your enterprise, created by GitHub, or match one of the patterns: LoveToKnow/slackify-markdown-action@*, MishaKav/jest-coverage-comment@*, actions-ecosystem/action-remove-labels@*, chromaui/action@*, codecov/codecov-action@*, codecov/test-results-action@*, crowdin/github-action@*, ossf/scorecard-action@*, patrickedqvist/wait-for-vercel-preview@*, pnpm/action-setup@*, rtCamp/action-slack-notify@*, stefanzweifel/git-auto-commit-action@*, step-security/harden-runner@*, thollander/actions-comment-pull-request@*, treosh/lighthouse-ci-action@*.
```

MattIPv4:
We cannot fix that, new actions need to be added to the repo settings by @ovflowd or @bmuenzenmeyer (I don't see the web-admin team yet)

bmuenzenmeyer:
> We cannot fix that, new actions need to be added to the repo settings by @ovflowd or @bmuenzenmeyer (I don't see the web-admin team yet)

added

PR #8245, opened 2025-10-19
feat(cicd): add bundle stats comparison

### LINKED ISSUE #8240: Build Size Difference
Can we have a CI step that comments the build size differences between runs?

_Originally posted by @ovflowd in https://github.com/nodejs/nodejs.org/pull/8239#issuecomment-3419670559_

avivkeller:
Adds a new `workflow_run` workflow which compares the bundle size of the current PR and the base branch.

See https://github.com/nodejs/nodejs.org/pull/8245#issuecomment-3419953311, which shows the diff between a previous commit of this PR and the current commit (which, since no Next.js code changed, is nothing).

> [!NOTE]
> If this workflow fails, it will _not_ display 🔴 on the PR. If it succeeds, it won't display 🟢 either.

> [!NOTE]
> This workflow _will_ fail for PRs open up unto this point, but should succeed for PRs opened after this is merged.

### DIFF
--- a/.github/workflows/bundle-compare.yml
+++ b/.github/workflows/bundle-compare.yml
@@ -1,0 +1,66 @@
+name: Compare Bundle Size
+
+on:
+  pull_request:
+  # workflow_run:
+  #   workflows: ['Build']
+  #   types: [completed]
+
+permissions:
+  contents: read
+  actions: read
+  # To create the comment
+  pull-requests: write
+
+jobs:
+  compare:
+    name: Compare Bundle Stats
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Harden Runner
+        uses: step-security/harden-runner@f4a75cfd619ee5ce8d5b864b0d183aff3c69b55a # v2.13.1
+        with:
+          egress-policy: audit
+
+      - name: Download Stats (HEAD)
+        uses: actions/download-artifact@018cc2cf5baa6db3ef3c5f8a56943fffe632ef53 # v6.0.0
+        with:
+          name: webpack-stats
+          path: head-stats
+          run-id: 18860058714
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Get Run ID from BASE
+        id: base-run
+        env:
+          WORKFLOW_ID: ${{ github.event.workflow_run.workflow_id || 64771488 }}
+          GH_TOKEN: ${{ github.token }}
+        run: |
+          ID=$(gh run list --repo $GITHUB_REPOSITORY -c 95f43b01b7790b2af237d3dd6f015d2212dd495f -w $WORKFLOW_ID -L 1 --json databaseId --jq ".[].databaseId")
+          echo "run_id=$ID" >> $GITHUB_OUTPUT
+
+      - name: Download Stats (BASE)
+        uses: actions/download-artifact@018cc2cf5baa6db3ef3c5f8a56943fffe632ef53 # v6.0.0
+        with:
+          name: webpack-stats
+          path: base-stats
+          run-id: ${{ steps.base-run.outputs.run_id }}
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Compare Bundle Size
+        id: compare-bundle-size
+        uses: actions/github-script@ed597411d8f924073f98dfc5c65a23a2325f34cd # v8.0.0
+        env:
+          HEAD_STATS_PATH: ./head-stats/webpack-stats.json
+          BASE_STATS_PATH: ./base-stats/webpack-stats.json
+        with:
+          script: |
+            const { compare } = await import('${{github.workspace}}/apps/site/scripts/compare-size/index.mjs')
+            await compare({core})
+
+      - name: Add Comment to PR
+        uses: thollander/actions-comment-pull-request@e2c37e53a7d2227b61585343765f73a9ca57eda9 # v3.0.0
+        with:
+          comment-tag: "compare_bundle_size"
+          message: ${{ steps.compare-bundle-size.outputs.comment }}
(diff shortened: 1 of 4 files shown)


AugustinMauroy:
WOW look soo cool ! I'm surprised there are any tool like that on Vercel

avivkeller:
You can also [download the stats file](https://github.com/nodejs/nodejs.org/actions/runs/18636044224?pr=8245), as it's an artifact.

image

bmuenzenmeyer:
This is cool. I am probably naive to some inner working here - but surprised it works with the chunks being generated hash filenames

.github/workflows/bundle-compare.yml
```diff
@@ -0,0 +1,103 @@
...
+            });
+
+            fs.mkdirSync('head-stats', { recursive: true });
+            fs.writeFileSync('head-stats/stats.zip', Buffer.from(download.data));
+
+      - name: Download main branch artifact (base)
+        uses: actions/github-script@ed597411d8f924073f98dfc5c65a23a2325f34cd # v8.0.0
+        with:
+          script: |
+            const fs = require('fs');
```
ovflowd: Not much a fan of inlining JavaScript here.

.github/workflows/bundle-compare.yml
```diff
@@ -0,0 +1,103 @@
...
+    steps:
+      - name: Harden Runner
+        uses: step-security/harden-runner@f4a75cfd619ee5ce8d5b864b0d183aff3c69b55a # v2.13.1
+        with:
+          egress-policy: audit
+
+      - name: Download PR artifact (head)
+        uses: actions/github-script@ed597411d8f924073f98dfc5c65a23a2325f34cd # v8.0.0
+        with:
+          script: |
```
ovflowd: Could you please move these scripts to our scripts folder and make them actual scripts and run them via node? Thanks!

.github/workflows/bundle-compare.yml
avivkeller: I know of an action that does this, but it's authored by an individual, which we are typically against.

.github/workflows/bundle-compare.yml
avivkeller: These are pretty simple scripts used in many many workflows, I don't think it's worth it to make new script files, and require this workflow perform a checkout, when we can just inline them

ovflowd: Sorry, I just don't feel happy approving a PR with JavaScript being inlined this way. Not even sure our security checks with CodeQL can detect these. Plus all our prettier and lint rules don't work here. Please eject this into proper files under a scripts folder (could even be under .github/workflows/scripts), that said, I think there are other workflows doing this, so I'd also ask to change that. Wdyt @bmuenzenmeyer

avivkeller: We use inline bash syntax in several workflows, why is inline JS such a big risk?

This practice of inlineing GitHub Actions Script is quite common. https://github.com/actions/github-script

avivkeller: Bump @ovflowd

ovflowd: Sorry, I'm not +1 with this solution and won't approve it :/

avivkeller: I'm still strongly against adding files to perform such a simple and common task. I understand that if this was more complex, maybe, but this is common practice

avivkeller:
Bump @nodejs/web-infra

.github/workflows/bundle-compare.yml
MattIPv4: Can we not use GitHub's actual action for downloading artifacts instead, so we avoid this completely?

avivkeller: No, because we are downloading from a different branch (main), and not the parent workflow

avivkeller: @ovflowd would you feel better if I used the `gh` CLI to get the workflow run number, and then use the action @MattIPv4 mentioned?

MattIPv4: 👍 I would be in favor of a CLI call + using a standard action, rather than rolling 50-100 lines of custom scripting

avivkeller: I need to do some last minute tests, but can you check that 44670eb11 resolved this concern?

MattIPv4: Much better, though I'm not quite sure I understand how that gh command is getting the base stats when it is using GITHUB_SHA?

avivkeller: `GITHUB_SHA` for `workflow_run` is the sha of the `main` branch commit it ran off of

MattIPv4: TIL

avivkeller:
So apparently the action doesn't support `workflow_run` internally (it's a needless check, as the action _would_ support the trigger).

So, I need to write a little custom action to do _all_ of our special logic

MattIPv4:
Can we not just open a PR upstream to add support, rather than rolling our own?

avivkeller:
> Can we not just open a PR upstream to add support, rather than rolling our own?

I'm planning to do both. I'll open a Pr upstream, and in the meantime, have a simple script here, unless we'd rather wait for the upstream PR to land?

MattIPv4:
I'm personally fine with waiting until upstream support is landed, unless others have a strong need for this urgently.

ovflowd:
I'm fine merging this without workflow_run for now. @avivkeller can you mark the PR as ready and open a follow-up issue?

avivkeller:
@ovflowd we need workflow_run for this, otherwise, it's useless to have as an extra file

ovflowd:
Ack, any issue created upstream?

avivkeller:
Yea, I opened a PR

avivkeller:
It seems nothing is happening upstream, so I'm implementing something here

avivkeller:
Bump @ovflowd @mattipv4

AugustinMauroy (approved):
that's cool and clever.

MattIPv4 (approved):

PR #8254, opened 2025-10-22
Clarify race condition in blocking vs non-blocking docs (fixes #8253)

### LINKED ISSUE #8253: A code example doesn't seem to work as described
### Affected URL

https://nodejs.org/en/learn/asynchronous-work/overview-of-blocking-vs-non-blocking

### Describe the issue in detail:

Hi everyone,

In the [Overview of Blocking vs Non-Blocking](https://nodejs.org/en/learn/asynchronous-work/overview-of-blocking-vs-non-blocking#dangers-of-mixing-blocking-and-non-blocking-code) article, section "Dangers of Mixing Blocking and Non-Blocking Code", a code example is given:
```js
const fs = require('node:fs');

fs.readFile('/file.md', (err, data) => {
  if (err) {
    throw err;
  }

  console.log(data);
});
fs.unlinkSync('/file.md');
```
It's followed by this text: 
"In the above example, `fs.unlinkSync()` is likely to be run before `fs.readFile()`, _which would delete `file.md` before it is actually read_".

I tried the example, but while `fs.unlinkSync()` is indeed run before `fs.readFile()`, `console.log(data)` still outputs (despite the italized part above) the starting byte values of `file.md`, so it looks if `file.md` was first deleted and then read.
Could anyone explain to me how that is possible (or correct the article if it's worded poorly) please?

OS: Windows 10 Home 22H2,
Node.js v22.20.0, also tried v25.0.0 with the same result.

Himanshu25Sahu:
# Clarify Race Condition in Blocking vs Non-Blocking Docs

This PR updates the **“Dangers of Mixing Blocking and Non-Blocking Code”** section with a revised example to reliably demonstrate the danger of mixing blocking and non-blocking calls.

Tested on **Windows 10 Home 22H2** with **Node.js v22.20.0** and **v25.0.0**, where the original example showed inconsistent behavior due to Windows file handle persistence.

---

## 🧩 Original Issue Description

In the Overview of Blocking vs Non-Blocking article, the section *“Dangers of Mixing Blocking and Non-Blocking Code”* included this example:

```javascript
const fs = require('node:fs');

fs.readFile('/file.md', (err, data) => {
  if (err) {
    throw err;
  }
  console.log(data);
});
fs.unlinkSync('/file.md');
```

The text stated: “In the above example, fs.unlinkSync() is likely to be run before fs.readFile(), which would delete file.md before it is actually read.” However, testing revealed that fs.readFile() often succeeded on Windows due to file handle behavior, contradicting the implied failure.

## Changes Made

- Replaced the original example with a new one using a temporary file (temp.md) and fs.unlinkSync before fs.readFile to ensure a consistent ENOENT failure, demonstrating the danger reliably.
Retained the original description’s intent, clarifying the race condition without altering its core message.
Added a try-finally block for cleanup and a note about asynchronous output.

✅ Updated Documentation Text
The revised example:
```js
const fs = require('node:fs');
const path = require('node:path');

const tempFile = path.join(__dirname, 'temp.md');
try {
  fs.writeFileSync(tempFile, 'Temporary content');
  fs.unlinkSync(tempFile);
  fs.readFile(tempFile, (err, data) => {
    if (err) {
      console.log('Error: Read failed with', err.code, 'as expected due to prior deletion.');
    } else {
      console.log('Unexpected success:', data.toString());
    }
  });
} finally {
  if (fs.existsSync(tempFile)) {
    fs.unlinkSync(tempFile);
  }
}
```

Description: “In the above example, fs.unlinkSync() is run before the asynchronous fs.readFile() operation, which deletes temp.md before the read can occur. This results in a failure (e.g., ENOENT error), demonstrating the danger of mixing blocking and non-blocking calls without proper sequencing. The try-finally block ensures file management.”

>**Note**: Due to the asynchronous nature of `fs.readFile`, the `console.log` may not execute immediately. You might need to add a `setTimeout` or keep the process alive (e.g., with `process.stdin.resume()`) to see the output.

The safe alternative remains:
```js
const fs = require('node:fs');
const path = require('node:path');

const tempFile = path.join(__dirname, 'temp.md');
fs.writeFileSync(tempFile, 'Temporary content');

fs.readFile(tempFile, (readFileErr, data) => {
  if (readFileErr) {
    throw readFileErr;
  }
  console.log(data);
  fs.unlink(tempFile, unlinkErr => {
    if (unlinkErr) {
      throw unlinkErr;
    }
  });
});
```

References

Original issue context from https://nodejs.org/en/learn/asynchronous-work/overview-of-blocking-vs-non-blocking
Repository: https://github.com/nodejs/nodejs.org
fixes #8253

### DIFF
--- a/apps/site/pages/en/learn/asynchronous-work/overview-of-blocking-vs-non-blocking.md
+++ b/apps/site/pages/en/learn/asynchronous-work/overview-of-blocking-vs-non-blocking.md
@@ -115,42 +115,63 @@ threads may be created to handle concurrent work.
 
 ## Dangers of Mixing Blocking and Non-Blocking Code
 
-There are some patterns that should be avoided when dealing with I/O. Let's look
-at an example:
+There are some patterns that should be avoided when dealing with I/O. Let's look at an example:
 
 ```js
 const fs = require('node:fs');
+const path = require('node:path');
 
-fs.readFile('/file.md', (err, data) => {
-  if (err) {
-    throw err;
+// Create and use a temporary file
+const tempFile = path.join(__dirname, 'temp.md');
+try {
+  fs.writeFileSync(tempFile, 'Temporary content');
+
+  // Perform blocking delete first
+  fs.unlinkSync(tempFile);
+
+  // Attempt asynchronous read afterward
+  fs.readFile(tempFile, (err, data) => {
+    if (err) {
+      console.log('Error: Read failed with', err.code, 'as expected due to prior deletion.');
+    } else {
+      console.log('Unexpected success:', data.toString());
+    }
+  });
+  // Note: The callback may not log immediately; the process exits before I/O completes.
+} finally {
+  // Cleanup (though file is already deleted)
+  if (fs.existsSync(tempFile)) {
+    fs.unlinkSync(tempFile);
   }
-
-  console.log(data);
-});
-fs.unlinkSync('/file.md');
+}
 ```
 
-In the above example, `fs.unlinkSync()` is likely to be run before the asynchronous `fs.readFile()` operation completes, introducing a race condition. Depending on system load and timing, this may delete `file.md` before it is opened, causing the read to fail (e.g., with an ENOENT error). However, if the file is opened before deletion, the read may still succeed due to platform-specific behaviors—such as files being marked for deletion on Windows while remaining accessible via open handles.
+In the above example, fs.unlinkSync() is run before the asynchronous fs.readFile() operation, which deletes temp.md before the read can occur. This results in a failure (e.g., ENOENT error), demonstrating the danger of mixing blocking and non-blocking calls without proper sequencing. The try-finally block ensures the file is managed, though the read failure is the key lesson here. To avoid such issues, ensure operations are ordered correctly...
+
+
+>**Note**: Due to the asynchronous nature of `fs.readFile`, the `console.log` may not execute immediately. You might need to add a `setTimeout` or keep the process alive (e.g., with `process.stdin.resume()`) to see the output.
 
 ```js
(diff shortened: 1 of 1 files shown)


avivkeller (changes requested):
The example's description was accurate prior to this change.

The issue referenced described an issue with the example code, so, if anything, the example itself should be changed, not its description.

Himanshu25Sahu:
Hi @avivkeller I've updated the PR with the revised example. Please let me know if any modifications are required.

apps/site/pages/en/learn/asynchronous-work/overview-of-blocking-vs-non-blocking.md
```diff
@@ -115,45 +115,63 @@ threads may be created to handle concurrent work.
 
 ## Dangers of Mixing Blocking and Non-Blocking Code
 
-There are some patterns that should be avoided when dealing with I/O. Let's look
-at an example:
+There are some patterns that should be avoided when dealing with I/O. Let's look at an example:
```
avivkeller, line 118: Unrelated change

apps/site/pages/en/learn/asynchronous-work/overview-of-blocking-vs-non-blocking.md
```diff
@@ -115,45 +115,63 @@ threads may be created to handle concurrent work.
...
 
 ```js
```
avivkeller, line 120: This example change isn't what we want.

We want:

```js
operationWithCallback(() => { ... })
operationWithoutCallback()
```

avivkeller:
As pointed out in the original issue, the example works as intended.

PR #8261, opened 2025-10-22
chore(gov): remove governance, rely on web-team instead

### LINKED ISSUE #9: Governance: Sync up and GHA

avivkeller:
Fixes https://github.com/nodejs/web-team/issues/9.

An alternative to this approach is a governance file that contains the link within it, rather than removing it entirely.


flakey5 (approved):

MattIPv4 (approved):

bmuenzenmeyer (approved):

bmuenzenmeyer:
i felt like you already did this 😅

bjohansebas (approved):

avivkeller:
@nodejs/web-admins can we merge this w/o TSC approval, since the file itself is unchanged, only moved?

bmuenzenmeyer:
> @nodejs/web-admins can we merge this w/o TSC approval, since the file itself is unchanged, only moved?

This would make sense to me if https://github.com/nodejs/web-team/blob/main/GOVERNANCE.md was owned by TSC, yes?

mcollina (approved):
lgtm

PR #8263, opened 2025-10-23
feat: add toggle to show/hide LTS releases

### LINKED ISSUE #8017: Enhance Vulnerability Table to Toggle LTS Versions
An ask for later (definitely not for this PR): It would be nice to have a toggle for the table that hides any rows for non-LTS versions.

_Originally posted by @jasnell in https://github.com/nodejs/nodejs.org/issues/7990#issuecomment-3113632788_

efekrskl:
## Description

Adds a toggle to the EOL table that filters non LTS releases

## Validation

Content being filtered:
image

Other:
image
image
image

## Related Issues

Closes #8017

### Check List


apps/site/components/EOL/EOLReleaseTable/TableClient.tsx
```diff
@@ -0,0 +1,96 @@
+'use client';
```
avivkeller: Why was this renamed?

apps/site/next-data/generators/releaseData.mjs
```diff
@@ -63,7 +63,7 @@ const generateReleaseData = async () => {
       version: latestVersion.semver.raw,
       versionWithPrefix: `v${latestVersion.semver.raw}`,
       codename: major.support.codename || '',
-      isLts: status.endsWith('LTS'),
+      isLts: support.ltsStart !== undefined,
```
avivkeller: What's the difference between these two? Is it related to this PR?

packages/ui-components/src/Common/Switch/index.tsx
```diff
@@ -0,0 +1,46 @@
...
+const Switch: FC<PropsWithChildren<SwitchProps>> = ({
+  label,
+  checked,
+  onCheckedChange,
+  className,
+  thumbClassname,
+  ...props
+}) => {
+  const id = useId();
+
```
avivkeller:
```suggestion
```

packages/ui-components/src/Common/Switch/index.tsx
```diff
@@ -0,0 +1,46 @@
...
+  onCheckedChange?: (checked: boolean) => void;
+  thumbClassname?: string;
+};
+
+const Switch: FC<PropsWithChildren<SwitchProps>> = ({
+  label,
+  checked,
+  onCheckedChange,
+  className,
+  thumbClassname,
```
avivkeller:
```suggestion
  thumbClassName,
```

avivkeller (commented):
Wow! Thanks for this!

apps/site/components/EOL/EOLReleaseTable/TableClient.tsx
efekrskl: Originally the file had only a table body, now it hosts the entire table so the new name felt like a better fit (because I also changed the component name)

apps/site/next-data/generators/releaseData.mjs
efekrskl: The table only holds EOL versions, so `getNodeReleaseStatus` would always return `End-of-life`. Therefore, `isLts` was always false (for eol releases)

packages/ui-components/src/Common/Switch/index.tsx
efekrskl: We need the `id` for wiring the label and the input itself, so that the label itself is also clickable. Do you have an alternative?

packages/ui-components/src/Common/Switch/index.tsx
efekrskl: Whoops 🫣 fixed

efekrskl:
> Wow! Thanks for this!

Thanks for the quick review!

flakey5 (approved):
lgtm for web infra bits, thank you!

AugustinMauroy (approved):
LGMT !

avivkeller (approved):

mikeesto (approved):

aymen94 (approved):

bmuenzenmeyer (approved):
love it - my only nit, as a fast follower or even as a never - is that it'd be nice if these occupied the same vertical spacing. that greatly complicates the MDX / header / component relationship though, so no worries

Image

bmuenzenmeyer:
this test error seems unrelated 🤷

bjohansebas (approved):
LGTM, it looks cool to be able to do that

bmuenzenmeyer:
rebased, the doc kit lockfile bump was causing errors - resolved via https://github.com/nodejs/nodejs.org/pull/8275

if my local testing is correct, [there is a unit test yet to resolve](https://github.com/nodejs/nodejs.org/pull/8263#issuecomment-3445592998)

bmuenzenmeyer:
the logic changes here 

image

make me increasingly sour on the strategy we have for determine node release states, and I think we should move toward the emerging plan within https://github.com/nodejs/nodejs.org/issues/8277 - even if this lands as a nice to have feature. looking for other thoughts on this

bmuenzenmeyer (changes requested):
blocking til we have alignment among @nodejs/nodejs-website  - not necessarily directly related to this PR (the work is great and an improvement on UX, thank you, but more about our data source)

ovflowd (changes requested):
This PR is doing something that it shouldn't (Out of Scope): https://github.com/efekrskl/nodejs.org/blob/f530878450c63aee83ab09bbaa615bfe45105601/apps/site/next-data/generators/releaseData.mjs

packages/ui-components/src/Common/Switch/index.tsx
ovflowd, line 1: Please add `'use client';`, Radix is explicitly client-side only.

efekrskl: Thanks, done!

efekrskl:
> This PR is doing something that it shouldn't (Out of Scope): [efekrskl/nodejs.org@`f530878`/apps/site/next-data/generators/releaseData.mjs](https://github.com/efekrskl/nodejs.org/blob/f530878450c63aee83ab09bbaa615bfe45105601/apps/site/next-data/generators/releaseData.mjs)

Yeah, in hindsight, maybe I should've created an issue or something. Because right now `isLts` just doesn't work at all for EOL releases, so at least some fix is needed.

efekrskl:
> blocking til we have alignment among @nodejs/nodejs-website - not necessarily directly related to this PR (the work is great and an improvement on UX, thank you, but more about our data source)

all good! If I can in any way would be happy to help.

packages/ui-components/src/Common/Switch/index.module.css
```diff
@@ -0,0 +1,59 @@
...
+      relative
+      inline-flex
+      h-6
+      cursor-pointer
+      items-center
+      rounded-full
+      bg-black
+      transition-colors
+      duration-100
+      ease-out
```
canerakdas:
```suggestion
      motion-safe:transition-colors
      motion-safe:duration-100
      motion-safe:ease-out
```

As a best practice for accessibility, it would be beneficial to apply animations using `motion-safe` https://www.w3.org/WAI/WCAG21/Understanding/animation-from-interactions.html

packages/ui-components/src/Common/Switch/index.module.css
```diff
@@ -0,0 +1,59 @@
...
+      focus:outline-none
+      focus-visible:ring-2
+      focus-visible:ring-green-500
+      focus-visible:ring-offset-2
+      focus-visible:ring-offset-neutral-900
```
canerakdas: I think the outline looks great! We can discuss this and later apply it to other components in a separate issue for design consistency

Just one little nitpick: it might be better to use the current outline for the dark theme and add a lighter background color for the light theme

image
image

packages/ui-components/src/Common/Switch/index.module.css
```diff
@@ -0,0 +1,59 @@
...
+
+  .root[data-state='checked'] {
+    @apply bg-green-600;
+  }
+
+  .thumb {
+    @apply pointer-events-none
+      block
+      h-5
+      w-5
```
canerakdas:
```suggestion
      size-5
```

packages/ui-components/src/Common/Switch/index.module.css
```diff
@@ -0,0 +1,59 @@
...
+      translate-x-0.5
+      rounded-full
+      bg-white
+      ring-0
+      transition-transform
+      duration-100
+      ease-out;
```
canerakdas:
```suggestion
      motion-safe:transition-transform
      motion-safe:duration-100
      motion-safe:ease-out;
```

packages/ui-components/src/Common/Switch/index.tsx
```diff
@@ -0,0 +1,48 @@
+'use client';
+
+import * as SwitchPrimitive from '@radix-ui/react-switch';
+import classNames from 'classnames';
+import { FC, PropsWithChildren, useId } from 'react';
```
canerakdas:
```suggestion
import { useId } from 'react';
import type { FC, PropsWithChildren } from 'react';
```

canerakdas (commented):
Sorry for the late review. I really like the feature, it makes reading the table easier! 🚀

packages/ui-components/src/Common/Switch/index.module.css
efekrskl: Do you mind elaborating on this? not sure if I'm following 😅

canerakdas: The ring offset color remains `neutral-900` even in light mode when focused. IMO to make it more consistent with the background and ensure the offset stands out more clearly, this approach would work better;

```css
dark:focus-visible:ring-offset-neutral-900
focus-visible:ring-offset-neutral-100
```

canerakdas (approved):

packages/ui-components/src/Common/Switch/index.module.css
efekrskl: Ah, yeah. Looks much cleaner now, thanks!

image
image

avivkeller:
@ovflowd can you re review

apps/site/components/EOL/EOLReleaseTable/TableClient.tsx
ovflowd, line 1: This is a very unusual component name. Why TableClient? Because it is a client Component? Unfortunately that's an awful name 😅

apps/site/components/EOL/EOLReleaseTable/TableClient.tsx
```diff
@@ -0,0 +1,96 @@
...
+    : eolReleases;
+
+  return (
+    <>
+      <Switch
+        label={t('components.eolTable.hideNonLts')}
+        checked={hideNonLts}
+        onCheckedChange={setHideNonLts}
+      />
+      <table id="tbVulnerabilities">
```
ovflowd: Why did you move this here and rename the file? Could you please revert this change? :)

apps/site/components/EOL/EOLReleaseTable/TableClient.tsx
```diff
@@ -0,0 +1,96 @@
...
+  const [currentModal, setCurrentModal] = useState<string | undefined>();
+  const [hideNonLts, setHideNonLts] = useState(false);
+
+  const filteredReleases = hideNonLts
+    ? eolReleases.filter(release => release.isLts)
+    : eolReleases;
+
+  return (
+    <>
+      <Switch
```
ovflowd: My gut feeling tells me you can do a pure HTML switch with CSS and have the table being hidden purely with CSS.

Can you please do that? On this website we want to avoid feature-gating behind JavaScript.

apps/site/next-data/generators/releaseData.mjs
ovflowd: This has unwanted side-effects. This flag isn't used to say if a version is a LTS version-line or not, but if it is currently LTS. Please revert this change and use the codename as a field to verify a release line is a LTS release line.

packages/ui-components/src/Common/Switch/index.tsx
```diff
@@ -0,0 +1,49 @@
+'use client';
```
ovflowd: As much as I love Radix! I feel we should make the switch purely with CSS and HTML and have the table entries be affected based on that.

ovflowd (commented):
Appreciate your contribution! Thanks for working on this :)

I've left a few comments.

apps/site/components/EOL/EOLReleaseTable/TableClient.tsx
efekrskl: Yep, that was my reasoning. Seeing that I got multiple comments on it, it was a bad call 😅 I will rename it after we agree on the other changes about the table

This is now reverted

apps/site/next-data/generators/releaseData.mjs
efekrskl: Whoops, reverted this back + using `codename` for checking for LTS releases now

apps/site/components/EOL/EOLReleaseTable/TableClient.tsx
efekrskl: Reverted

packages/ui-components/src/Common/Switch/index.tsx
efekrskl: Fair, I changed the implementation to use html/css only. However, the only way I could get it to work was through using `has`, which is newly available (https://caniuse.com/css-has). Does this work?

efekrskl:
Thanks for the review @ovflowd, I've changed the implementation to use html/css only.

The changes to the data sources is also reverted (as they were wrong in the first place). Could you unblock your review @bmuenzenmeyer ?

apps/site/components/EOL/EOLReleaseTable/TableBody.tsx
```diff
@@ -28,7 +28,7 @@ const EOLReleaseTableBody: FC<EOLReleaseTableBodyProps> = ({
     <tbody>
       {eolReleases.map(release => (
         <Fragment key={release.major}>
-          <tr>
+          <tr data-lts={!!release.codename}>
```
bmuenzenmeyer, line 31: smart!

bmuenzenmeyer (approved):

bmuenzenmeyer:
I cannot get prettier to overwrite the current contents, which makes me think the failure is a false positive or caching problem

ovflowd:
> Thanks for the review @ovflowd, I've changed the implementation to use html/css only.
> 
> The changes to the data sources is also reverted (as they were wrong in the first place). Could you unblock your review @bmuenzenmeyer ?

Howdy! I'm on work travel, I'll be reviewing this later today or tomorrow!

apps/site/components/EOL/EOLReleaseTable/index.tsx
```diff
@@ -1,10 +1,12 @@
+import Switch from '@node-core/ui-components/Common/Switch';
 import { getTranslations } from 'next-intl/server';
 import type { FC } from 'react';
 
 import provideReleaseData from '#site/next-data/providers/releaseData';
 import provideVulnerabilities from '#site/next-data/providers/vulnerabilities';
 import { EOL_VERSION_IDENTIFIER } from '#site/next.constants.mjs';
 
+import styles from './index.module.css';
```
ovflowd, line 9: nit: style imports must happen after all other imports

apps/site/components/EOL/EOLReleaseTable/TableBody.tsx
ovflowd, line 31: What is this for?

ovflowd (approved):
SGTM besides of one nit :)

apps/site/components/EOL/EOLReleaseTable/index.tsx
bmuenzenmeyer: i checked - our import order rules complain when they are swapped -so this is correct per the current guidelines

ovflowd: wdym? The guidelines should be style imports always on the bottom

ovflowd: That's literally how all other components are doing, style imports on the bottom... If it is not erroring, it should.

PR #8282, opened 2025-10-28
fix(cicd): add `--` before passing an argument

### LINKED ISSUE #8281: Create Release Blog Post workflow failed for v24.11.0
For Node.js v24.11.0 I tried to run the [workflow](https://github.com/nodejs/nodejs.org/actions/workflows/create-release-post.yml) but it failed: https://github.com/nodejs/nodejs.org/actions/runs/18882884840/job/53890578483

I've manually created https://github.com/nodejs/nodejs.org/pull/8280.

avivkeller:
Fixes #8281 
cc @richardlau

While the actions logs don't show it, the actual error that occured here was that the workflow did not correctly pass `v24.11.0` as an argument.

### DIFF
--- a/.github/workflows/create-release-post.yml
+++ b/.github/workflows/create-release-post.yml
@@ -33,7 +33,7 @@ jobs:
           pnpm: true
           use-version-file: true
 
-      - run: node --run scripts:release-post "$VERSION"
+      - run: node --run scripts:release-post -- "$VERSION"
         working-directory: apps/site
         env:
           VERSION: ${{ inputs.version }}


richardlau (approved):

MattIPv4 (approved):

avivkeller:
I'm merging this without a "formal" fast-track approval as a hotfix

richardlau:
ah right, I ran the script directly with `node` and not via the script in `package.json` which is why I didn't get the error locally.

Issue #8092, opened 2025-08-20
TODO list for v24 + Next.js v16


avivkeller:
This issue is the general tracker for things we need / should do when Node.js v24 becomes LTS in October, and Next.js 16 is released:

### Node.js v24 LTS

- Remove npm re-installation (Added in #8091)
- Update the `.nvmrc`
- Update `@types/node`

### Next.js v16

- #8207 (See https://nextjs.org/blog/next-16-beta#improved-caching-apis + https://react.dev/blog/2025/10/01/react-19-2#cache-signal)
- Possibly use [`turbopackFileSystemCacheForDev`](https://nextjs.org/blog/next-16-beta#turbopack-file-system-caching-beta)?
- Possibly use [The React Compiler](https://nextjs.org/blog/next-16-beta#react-compiler-support-stable)
- Possibly use [`Activity` component](https://react.dev/reference/react/Activity)

AugustinMauroy:
as my knowledge from userland-migrations we are good with this list

PR #8289, opened 2025-10-29
chore(versioning): bump node to v24, pnpm to v10.20

avivkeller:
This PR bumps our Node.js version to v24(.11.0), and our pnpm version to 10.20.

Ref: #8092


pnpm-workspace.yaml
```diff
@@ -3,7 +3,7 @@ packages:
   - apps/*
 
 catalog:
-  '@types/node': 22.18.6
+  '@types/node': 24.9.2
```
MattIPv4: Why is this pinned to a specific version?

pnpm-lock.yaml
MattIPv4, line 1: Why such a giant diff on this? A bunch of transient versions seem to have changed?

avivkeller: pnpm re-generated the lockfile during the version bump, I can have it not do that if you'd like

MattIPv4: Yes, please, I don't think we should be bumping a bunch of dependencies at the same time.

MattIPv4:
👀 so Vercel doesn't appear to support 24.x yet: https://vercel.com/docs/functions/runtimes/node-js/node-js-versions

I'm quite surprised the build didn't fail? Are we not setting `engines.node` for the website?

avivkeller:
> 👀 so Vercel doesn't appear to support 24.x yet: https://vercel.com/docs/functions/runtimes/node-js/node-js-versions
> 
> 
> 
> I'm quite surprised the build didn't fail? Are we not setting `engines.node` for the website?

We are on the root, but iirc we use `node --run`, which doesn't fail on unsupported Node.js versions.

MattIPv4:
I see, I think we want to set `engines.node` to `24.x` in the website package itself, which is what will tell Vercel to actually use v24 (which will then fail currently, I believe, as they don't support it).

.nvmrc
bmuenzenmeyer, line 1: I'd really prefer that we pin to a specific version.

bmuenzenmeyer (dismissed):

avivkeller:
Blocked until Vercel supports v24. They currently do not have a target date, but expect it sometime soon.

cc @styfle Can you update us once this is available?

.nvmrc
avivkeller: iiuc we only pinned to a specific version since we needed to be >= it. We support any v24.x, why do we need to pin to a specific one? (If we did, tho, I would assume it to be `.11.0`)

pnpm-workspace.yaml
avivkeller: `24.9.2` is the latest type available from `@types/node`

MattIPv4:
@avivkeller can you try setting `engines.node` to `24.x` in the website? I'd like to see us hit the same failure I hit in https://github.com/alveusgg/alveusgg/pull/1604#issuecomment-3465416559 so that we know Vercel is actually looking at the package.json for what version to use?

avivkeller:
Vercel now supports this version, rebasing now

.nvmrc
MattIPv4: I lean towards preferring this be just v24, as that is all Vercel guarantees us.

pnpm-workspace.yaml
MattIPv4: Right, but why do we need to pin, why can't this use `^`?

MattIPv4 (changes requested):
I'd still like to see `engines.node` set to `24.x` in the website `package.json`, please, so that Vercel actually builds this with Node.js 24

avivkeller:
Looks like the deployment overrides the Node.js version set (v22). @ovflowd or @bmuenzenmeyer can you change the builds to use v24.x once this lands?

image

https://vercel.com/docs/functions/runtimes/node-js/node-js-versions#setting-the-node.js-version-in-project-settings

apps/site/package.json
```diff
@@ -119,5 +119,8 @@
       "./*.mjs",
       "./*/index.mjs"
     ]
+  },
+  "engines": {
+    "node": ">=24"
```
MattIPv4:
```suggestion
    "node": "24.x"
```
I think we should make this `24.x` so that we're not inadvertently allowing `25.x` (or `26.x`) in the future? Using `24.x` is what Vercel recommends: https://vercel.com/docs/functions/runtimes/node-js/node-js-versions#version-overrides-in-package.json

avivkeller: Noted. Ideally the Vercel default should _also_ be set post this PR

MattIPv4 (approved):
Lovely, Vercel picked up the Node.js version override with that

Image

apps/site/package.json
MattIPv4: Yah, agree, once this is in we can bump the default, this just ensures this PR builds correctly and I think its a good habit to have encoded for future version bumps too.

avivkeller:
@nodejs/nodejs-website bump

araujogui (approved):

PR #8304, opened 2025-10-31
chore(open-next): apply various improvements

dario-piotrowicz:
## Description

This PR adds various open-next improvements to the site (mainly for better performance)

The changes are:
 - bumped `@opennextjs/cloudflare` to the latest version (`1.12.0`)
 - replaced the kv incremental cache with the more performance r2 + regional cache implementation
 - Added a DurableObject queue that was missing for time-based revalidation
 - Enabled the open-next cache interception
 - Added a custom image loader for the Cloudflare deployment
 - Build open-next version of the site using `--turbo` 🚀 
 - Remove the no longer needed `build:default` script

## Validation

I've validated these changes locally by running `pnpm cloudflare:preview`

The workflow running on this PR also validates the open-next build.

### Check List


dario-piotrowicz:
@vicb it would be great if you could have a look and validate the open-next changes if you get a chance 🙂

apps/site/cloudflare/image-loader.ts
avivkeller, line 1: Can this go in a `cloudflare/` subdirectory (`cloudflare/images.ts` or something)?

apps/site/cloudflare/image-loader.ts
```diff
@@ -0,0 +1,22 @@
...
+    // Serve the original image when using `next dev`
+    return src;
+  }
+  const params = [`width=${width}`];
+  if (quality) {
+    params.push(`quality=${quality}`);
+  }
+  const paramsString = params.join(',');
+  return `/cdn-cgi/image/${paramsString}/${normalizeSrc(src)}`;
+}
```
avivkeller, line 33: A few notes with this implementation:
1. is `normalizeSrc` needed? Won't the leading `/` be normalized by the browser? If not, shouldn't there not be a slash / always be a slash passed here? This should _only_ load images, inconsistencies should be resolved at their root, not here?

2. I think this can be simplified to:


```suggestion
import type { ImageLoaderProps } from 'next/image';

// This can also be a `.replace(...)`?
const normalizeSrc = (src: string) => src.startsWith('/') ? src : `/${src}`;

// For tree-shaking, do this outside
export default process.env.NODE_ENV === 'development' ?
  ({ src }) => src :
  ({
    src,
    width,
    quality,
  }: ImageLoaderProps) => `/cdn-cgi/image/width=${width}${quality ? `quality=${quality}` : ''}${normalizeSrc(src)}`;
```

apps/site/next.config.mjs
```diff
@@ -13,44 +13,61 @@ const nextConfig = {
...
-      },
-      {
-        protocol: 'https',
-        hostname: 'website-assets.oramasearch.com',
-        port: '',
-        pathname: '/**',
-      },
-    ],
-  },
+  images:
```
avivkeller: As a follow-up, I think we should make an `apps/site/vercel` folder and `apps/site/cloudflare` folder (well, it'll already exist, see my other comment) for the specific logic. Then, the config file itself will be simple, and call things like `getCloudflareImageConfiguration()` and `getVercelImageConfiguration()` or something?

wdyt @nodejs/nodejs-website

apps/site/wrangler.jsonc
```diff
@@ -33,10 +33,24 @@
     "fs": "./.wrangler/fs-assets-polyfilling/polyfills/node/fs.ts",
     "fs/promises": "./.wrangler/fs-assets-polyfilling/polyfills/node/fs/promises.ts",
   },
-  "kv_namespaces": [
```
avivkeller, line 36: I think we explain this somewhere in the docs, can we update that file if so?

apps/site/cloudflare/image-loader.ts
ovflowd: Part of me really does not like this function. something about how it's designed 😅 

I bert this can be simplified/cleaned up. + I'd rather not use the frigging process.env.NODE_ENV here and actually use the constants we have defined on next.constants. Please also document what this function is doing, why it works and Cloudflare external links. We need to know this exists and howit works.

apps/site/next.config.mjs
```diff
@@ -13,44 +13,61 @@ const nextConfig = {
...
+    // Important: The custom loader ignores `remotePatterns` as those are configured as allowed source origins
+    //            (https://developers.cloudflare.com/images/transform-images/sources/)
+    //            in the Cloudflare dashboard itself instead (to the exact same values present in `remotePatterns` below).
+    //
+    // TODO: The `OPEN_NEXT_CLOUDFLARE` environment variable is being
+    //       defined in the worker building script, ideally the open-next
+    //       adapter should set it itself when it invokes the Next.js build
+    //       process, once it does that remove the manual `OPEN_NEXT_CLOUDFLARE`
+    //       definition in the package.json script.
+    process.env.OPEN_NEXT_CLOUDFLARE
```
ovflowd: Can we not use direct process.env? Thanks! We use next.constants.mjs for all relevant constants.

ovflowd: Also both these could be extracted into their own constants instead of a big ternary, easier to read.

apps/site/open-next.config.ts
```diff
@@ -1,8 +1,16 @@
 import type { OpenNextConfig } from '@opennextjs/cloudflare';
 import { defineCloudflareConfig } from '@opennextjs/cloudflare';
-import incrementalCache from '@opennextjs/cloudflare/overrides/incremental-cache/kv-incremental-cache';
+import r2IncrementalCache from '@opennextjs/cloudflare/overrides/incremental-cache/r2-incremental-cache';
+import { withRegionalCache } from '@opennextjs/cloudflare/overrides/incremental-cache/regional-cache';
+import doQueue from '@opennextjs/cloudflare/overrides/queue/do-queue';
 
-const cloudflareConfig = defineCloudflareConfig({ incrementalCache });
+const cloudflareConfig = defineCloudflareConfig({
+  incrementalCache: withRegionalCache(r2IncrementalCache, {
```
ovflowd, line 14: Can you add inline comments reference `@see` to relevant docs? And also inline comment what these settings do? 🙇

apps/site/wrangler.jsonc
ovflowd: +1

apps/site/next.config.mjs
```diff
@@ -13,44 +13,61 @@ const nextConfig = {
...
+          loader: 'custom',
+          loaderFile: './cloudflare-image-loader.ts',
+        }
+      : {
+          // We disable image optimisation during static export builds
+          unoptimized: ENABLE_STATIC_EXPORT,
+          // We add it to the remote pattern for the static images we use from multiple sources
+          // to be marked as safe sources (these come from Markdown files)
+          remotePatterns: [
+            {
```
vicb: not directly related to this PR but I think readability could be improved here by using `URL`s:

```ts:
remotePatterns: [
   new URL('https://avatars.githubusercontent.com/**'),
   new URL(...),
   ...
]
```

see https://nextjs.org/docs/pages/api-reference/components/image#remotepatterns

apps/site/open-next.config.ts
vicb: The PR description mentions

> replaced the kv incremental cache with the more performance r2 + regional cache implementation

The actual issue with KV is that it is eventually consistent which might lead to unexpected behavior. This is why we recommend R2 instead - R2 is strongly consistent.

apps/site/package.json
```diff
@@ -83,7 +83,7 @@
     "@flarelabs-net/wrangler-build-time-fs-assets-polyfilling": "^0.0.1",
     "@next/eslint-plugin-next": "15.5.4",
     "@node-core/remark-lint": "workspace:*",
-    "@opennextjs/cloudflare": "^1.6.4",
+    "@opennextjs/cloudflare": "^1.11.0",
```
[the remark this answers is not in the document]
vicb: +1 wrangler should be updated to the latest available version

apps/site/cloudflare/image-loader.ts
```diff
@@ -0,0 +1,22 @@
+import type { ImageLoaderProps } from 'next/image';
+
+const normalizeSrc = (src: string) => {
+  return src.startsWith('/') ? src.slice(1) : src;
+};
+
+export default function cloudflareLoader({
```
vicb, line 7: Note that this requires images to be enabled and configured as described [in the docs](https://opennext.js.org/cloudflare/howtos/image#enable-cloudflare-images) before it can be used.

vicb (commented):
Changes look good to me when comments are addressed, thanks Dario!

A few notes:
- Cloudflare images need to be enabled/configured for the account
- R2 batch upload had been implemented recently to speed up cache asset upload. There are a few environment variables to set for this to work. I'll write the doc for it this week. Before that R2 upload could be slow.

apps/site/cloudflare/image-loader.ts
dario-piotrowicz: Thanks, yes I did that already 🙂

apps/site/open-next.config.ts
ovflowd: I think you misunderstood, I'm asking for comments to be added on the code, so that it stays there for reference :)

apps/site/package.json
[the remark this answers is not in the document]
dario-piotrowicz: I already did that 🙂

ovflowd:
Thanks Victor for the updates. Will keep an eye on your docs. Also, where can we enabled Cloudflare Images?

dario-piotrowicz:
> where can we enabled Cloudflare Images?

https://developers.cloudflare.com/images/get-started/#enable-transformations-on-your-zone

I already did that on the testing/staging account

apps/site/open-next.config.ts
vicb: I got that, I added the comment mostly for Dario to update the PR description and add comments here

vicb:
> Thanks Victor for the updates. Will keep an eye on your docs. Also, where can we enabled Cloudflare Images?

https://github.com/opennextjs/docs/pull/189 about R2 batching was just merged [to the docs](https://opennext.js.org/cloudflare/cli#populatecache-command).

The 3 environments variables should be added the [build time time env vars](https://opennext.js.org/cloudflare/howtos/dev-deploy#deploy-your-application-to-cloudflare-workers) (provided you use Cloudflare CD).

apps/site/open-next.config.ts
dario-piotrowicz: @ovflowd I've added an inline comment https://github.com/nodejs/nodejs.org/pull/8304/commits/65b4b4b39306e307764a4e1afb986bbbabe3d122

please let me know if this works for you 🙏

apps/site/next.config.mjs
dario-piotrowicz: > Can we not use direct process.env? Thanks! We use next.constants.mjs for all relevant constants.

Done 👍 

> Also both these could be extracted into their own constants instead of a big ternary, easier to read.

I hope this is what you had in mind: https://github.com/nodejs/nodejs.org/pull/8304/commits/302ba1b9b5d08591a18ad7fae4656e0d9f99349c

apps/site/wrangler.jsonc
dario-piotrowicz: Definitely! thanks a lot for the reminder! 🙏 

https://github.com/nodejs/nodejs.org/pull/8304/commits/57d3a4a1cdb6718c59577fa0f407425db066d968

apps/site/cloudflare/image-loader.ts
dario-piotrowicz: This function is taken straight from the open-next documentation: https://opennext.js.org/cloudflare/howtos/image#use-a-custom-loader

I think it might be beneficial (since there are no like custom requirements or anything like that) to keep it as close as possible as the version in the open-next docs? That way if anything on the open-next site changes an updated loader version can just be copy-pasted here instead or having a more bespoke version that we need to more directly maintain? 🤔

apps/site/cloudflare/image-loader.ts
dario-piotrowicz: I'm happy to go with your version, however as I mentioned there might be benefits in keeping this in sync with the open-next documentation? 🤔 

(PS: if you want you can also try to upstream these changes in the open-next docs (cc. @vicb))

apps/site/next.config.mjs
dario-piotrowicz: I'm resolving this since you've already opened a [PR proposing this change](https://github.com/nodejs/nodejs.org/pull/8325) 🙂

apps/site/cloudflare/image-loader.ts
dario-piotrowicz: > Can this go in a cloudflare/ subdirectory (cloudflare/images.ts or something)?

Moved 👍

.github/workflows/tmp-cloudflare-open-next-deploy.yml
```diff
@@ -64,3 +64,8 @@ jobs:
         env:
           CF_WORKERS_SCRIPTS_API_TOKEN: ${{ secrets.CLOUDFLARE_API_TOKEN }}
           CLOUDFLARE_API_TOKEN: ${{ secrets.CLOUDFLARE_API_TOKEN }}
+          CLOUDFLARE_ACCOUNT_ID: fb4a2d0f103c6ff38854ac69eb709272
+          # Note the R2_* env variables below are useful for speeding up the assets upload
+          # process, see: https://opennext.js.org/cloudflare/cli#populatecache-command
+          R2_ACCESS_KEY_ID: ${{ env.R2_ACCESS_KEY_ID }}
+          R2_SECRET_ACCESS_KEY: ${{ secrets.R2_SECRET_ACCESS_KEY }}
```
dario-piotrowicz: I've added these two to the repository, I've created a token specific for the account that only has access to the site's incremental cache r2 bucket, I hope this works for you, if it doesn't I'm happy to change things 🙂

dario-piotrowicz:
> The 3 environments variables should be added the [build time time env vars](https://opennext.js.org/cloudflare/howtos/dev-deploy#deploy-your-application-to-cloudflare-workers) (provided you use Cloudflare CD).

Added 🫡

apps/site/cloudflare/image-loader.ts
flakey5: Imo, the initial implementation is a lot more readable than this (though normalizeSrc may be better as a just a .replace but I'm either or on it)

flakey5 (approved):

avivkeller:
@dario-piotrowicz can you resolve conflicts? I'll give this another look today

dario-piotrowicz:
> @dario-piotrowicz can you resolve conflicts? I'll give this another look today

Hi @avivkeller, sorry I've only now seen your comment, I just rebased, please have another look whenever you get the chance 🙏

.github/workflows/tmp-cloudflare-open-next-deploy.yml
```diff
@@ -64,3 +64,8 @@ jobs:
         env:
           CF_WORKERS_SCRIPTS_API_TOKEN: ${{ secrets.CLOUDFLARE_API_TOKEN }}
           CLOUDFLARE_API_TOKEN: ${{ secrets.CLOUDFLARE_API_TOKEN }}
+          CLOUDFLARE_ACCOUNT_ID: fb4a2d0f103c6ff38854ac69eb709272
```
avivkeller, line 67:
```suggestion
          CLOUDFLARE_ACCOUNT_ID: ${{ vars.CLOUDFLARE_TEST_ACCOUNT_ID }}
```

avivkeller (approved):

.github/workflows/tmp-cloudflare-open-next-deploy.yml
```diff
@@ -64,3 +64,8 @@ jobs:
...
+          # Note the R2_* env variables below are useful for speeding up the assets upload
+          # process, see: https://opennext.js.org/cloudflare/cli#populatecache-command
+          R2_ACCESS_KEY_ID: ${{ env.R2_ACCESS_KEY_ID }}
```
avivkeller:
```suggestion
          R2_ACCESS_KEY_ID: ${{ vars.R2_ACCESS_KEY_ID }}
```

vicb:
Oh actually we are going to remove the need for specific R2 accounts/variables for batch upload. Creating S3 keys was required to use rclone but Wrangler will support a 'r2 bulk put' command in the next release and the adapter will leverage it. 
@dario-piotrowicz

avivkeller:
@dario-piotrowicz this is almost ready!

.github/workflows/tmp-cloudflare-open-next-deploy.yml
dario-piotrowicz: Could you explain why the suggestion?

This is what I see in the [GitHub docs](https://docs.github.com/en/actions/reference/workflows-and-actions/contexts?versionId=free-pro-team%40latest&productId=actions&restPage=reference%2Cworkflows-and-actions%2Cvariables#available-contexts):
Screenshot 2025-11-21 at 02 05 50

I do think that `env` is appropriate here, is it not?

(Also note that I can't see any other use of `vars` in the repos github workflow files)

dario-piotrowicz:
> Oh actually we are going to remove the need for specific R2 accounts/variables for batch upload. Creating S3 keys was required to use rclone but Wrangler will support a 'r2 bulk put' command in the next release and the adapter will leverage it. @dario-piotrowicz

I've bumped the open-next version and removed the R2 variables 🙂👍

.github/workflows/tmp-cloudflare-open-next-deploy.yml
dario-piotrowicz: 👍 

Although as I mentioned in [my other comment](https://github.com/nodejs/nodejs.org/pull/8304#discussion_r2548297238), why using `vars` and why just for this specific instance?

.github/workflows/tmp-cloudflare-open-next-deploy.yml
avivkeller: I don't think this variable is stored in the environment, but it is stored in our repository variables.

(Note that the values here are not sensitive, and the image is a few days old)

image

.github/workflows/tmp-cloudflare-open-next-deploy.yml
avivkeller: Well, if we change the account ID, it's tedious to open a PR to adjust the workflow, when, as the alternative, we can simply update the variable in the repository settings.

apps/site/cloudflare/image-loader.ts
```diff
@@ -0,0 +1,22 @@
...
+  src,
+  width,
+  quality,
+}: ImageLoaderProps) {
+  if (process.env.NODE_ENV === 'development') {
```
ovflowd, line 12: I still feel this method can be tidied up a little bit.

Can we do something a little bit cleaner?

```tsx
if (process.env.NODE_ENV === 'development') {
  // Serve the original image when using `next dev`
  return src;
}

// Sets the requested width by next/image
const params = new Map();

if (width > 0) {
  params.set('width', width);

if (quality > 0) {
  params.set('quality', quality);
}

const pathParams = // map with key=value, (could be .reduce, .map, a collector, etc)

return `/cdn-cgi/image/${pathParams}/${normalizeSrc(src)}`;
```

apps/site/next.config.mjs
```diff
@@ -1,7 +1,11 @@
 'use strict';
 import createNextIntlPlugin from 'next-intl/plugin';
 
-import { BASE_PATH, ENABLE_STATIC_EXPORT } from './next.constants.mjs';
+import {
+  BASE_PATH,
+  ENABLE_STATIC_EXPORT,
+  OPEN_NEXT_CLOUDFLARE,
```
ovflowd: nit: can we move all cloudflare-specific constants to `next.cloudflare.constants.mjs` (so we can do the same with Vercel-ones to `next.vercel.constants.mjs` in a follow-up or in this PR if you wish)

apps/site/next.config.mjs
```diff
@@ -81,17 +73,47 @@ const nextConfig = {
...
-  //       adapter should set it itself when it invokes the Next.js build
-  //       process, onces it does that remove the manual `OPEN_NEXT_CLOUDFLARE`
-  //       definition in the package.json script.
-  deploymentId: process.env.OPEN_NEXT_CLOUDFLARE
-    ? (await import('@opennextjs/cloudflare')).getDeploymentId()
-    : undefined,
+  deploymentId: await getDeploymentId(),
 };
 
+function getImagesConfig() {
```
ovflowd: Can we move this to a specific file? `next.image.config.mjs`?

ovflowd: Also, we use ES6 functions across the board :)

apps/site/next.config.mjs
```diff
@@ -81,17 +73,47 @@ const nextConfig = {
...
+    };
+  }
+
+  return {
+    // We disable image optimisation during static export builds
+    unoptimized: ENABLE_STATIC_EXPORT,
+    // We add it to the remote pattern for the static images we use from multiple sources
+    // to be marked as safe sources (these come from Markdown files)
+    remotePatterns: [
+      new URL('https://avatars.githubusercontent.com/**'),
```
ovflowd: Can we make this an array (constant) and iterate with .map(=> new URL)?

ovflowd (dismissed):
I havent' finished full review, but still believe we should have a specific document/documentation for the specific image loader and how it works and external ref.

ovflowd:
(apologies, accidentally made this a blocking review 🤦)

ovflowd (commented):
@dario-piotrowicz I've finished reviewing it, pretty much all my comments left are what needs to be addressed :)

Don't forget to update documentation!

Once all get resolved, I'll approve the PR!

apps/site/cloudflare/image-loader.ts
dario-piotrowicz: how does this look? 
26a76e86eaa0aaa0d790086da288bfe85e83fcbd

apps/site/next.config.mjs
dario-piotrowicz: Sure 🙂 

but instead of a `next.cloudflare.constants.mjs` file, what about a `next.constants.mjs` file inside the new `cloudflare` directory?

(PS: I am totally happy to do the same for Vercel, but I'd like to do it as a followup if ok not to scope-creep the PR 🙂)

apps/site/next.config.mjs
dario-piotrowicz: > Can we move this to a specific file? next.image.config.mjs?

Done 👍 

> Also, we use ES6 functions across the board :)

oh... sorry 😓, done 👍

apps/site/next.config.mjs
dario-piotrowicz: done 🙂 

https://github.com/nodejs/nodejs.org/pull/8304/commits/1170f1b5c4dbe8cb2a579c887d5924e81e5ea1e5

.github/workflows/tmp-cloudflare-open-next-deploy.yml
dario-piotrowicz: oh right! good point, my bad, thanks 👍

apps/site/next.config.mjs
ovflowd: I feel keeping it in the same directory is better for discoverability.

ovflowd:
@dario-piotrowicz PR needs rebasing and I left a final comment, after that we're good to go :D

apps/site/next.config.mjs
```diff
@@ -1,9 +1,21 @@
...
+import { OPEN_NEXT_CLOUDFLARE } from './next.constants.cloudflare.mjs';
 import { BASE_PATH, ENABLE_STATIC_EXPORT } from './next.constants.mjs';
+import { getImagesConfig } from './next.image.config.mjs';
 import { redirects, rewrites } from './next.rewrites.mjs';
 
+const getDeploymentId = async () => {
+  if (OPEN_NEXT_CLOUDFLARE) {
+    // If we're building for the Cloudflare deployment we want to set
+    // an appropriate deploymentId (needed for skew protection)
+    return (await import('@opennextjs/cloudflare')).getDeploymentId();
```
ovflowd:
```suggestion
    const openNextAdapter = await import('@opennextjs/cloudflare');

    return openNextAdapter.getDeploymentId();
```

ovflowd: I swear, last nit 🙈

.github/workflows/tmp-cloudflare-open-next-deploy.yml
dario-piotrowicz: I was about to update this, but I noticed that anyways the account id (`fb4a2d0f103c6ff38854ac69eb709272`) is already hardcoded in the wrangler.jsonc file ([L7](https://github.com/nodejs/nodejs.org/blob/563f9eb6f23f164486b7417dcfba5039bd2421d8/apps/site/wrangler.jsonc#L7) and [L21](https://github.com/nodejs/nodejs.org/blob/563f9eb6f23f164486b7417dcfba5039bd2421d8/apps/site/wrangler.jsonc#L21)). So not having it hardcoded here is not going to make things any more convenient (actually the contrary, since having the variable would mean that updating the id would need to be done both in the code and also in the repository's settings). 

So I think we should just keep this as is, what do you think? 🙂

apps/site/next.config.mjs
dario-piotrowicz: don't worry it's all good 😄

.github/workflows/tmp-cloudflare-open-next-deploy.yml
dario-piotrowicz: They are no longer necessary and I've removed them from the repository's setting 🙂

flakey5 (approved):

ovflowd (approved):
LGTM!

ovflowd:
:shipit: :shipit: :rocket:

dario-piotrowicz:
> we should have a specific document/documentation for the specific image loader and how it works and external ref.

@ovflowd regarding the image loader documentation, I've added a very fried section in the md file with a link to the cloudflare docs: https://github.com/nodejs/nodejs.org/pull/8304/commits/dbf3b69426db2e45dcc7cbaf43de5b77ea190096 as I didn't feel like more was needed, but please let me know if you think that this might not suffice 🙂

dario-piotrowicz:
> LGTM!

Thanks @ovflowd! Sorry this PR took a while! 😅

ovflowd:
> > we should have a specific document/documentation for the specific image loader and how it works and external ref.
> 
> @ovflowd regarding the image loader documentation, I've added a very fried section in the md file with a link to the cloudflare docs: [dbf3b69](https://github.com/nodejs/nodejs.org/commit/dbf3b69426db2e45dcc7cbaf43de5b77ea190096) as I didn't feel like more was needed, but please let me know if you think that this might not suffice 🙂

Yeah, if it has a link to other external resources (if any available) that'd be good, except that. Feel free to merge the PR

> > LGTM!
> 
> Thanks @ovflowd! Sorry this PR took a while! 😅

You good! Appreciate for the patience here :heart:

dario-piotrowicz:
I've enabled auto-merge 🙂 

@avivkeller could you go back to the unresolved conversations you started and make sure you're happy where we landed, and in that case resolve the conversations? 🙂 
(of course if you're not happy with any them let's discuss further 😄)

.github/workflows/tmp-cloudflare-open-next-deploy.yml
avivkeller: Yup, keeping it is fine

dario-piotrowicz:
thanks @avivkeller 🙂🙏

PR #8332, opened 2025-11-05
docs(learn): correct another example code for the consumption of response body

### LINKED ISSUE #8331: [Docs] incorrect example code in docs
### Affected URL

https://nodejs.org/en/learn/manipulating-files/reading-files-with-nodejs

### Describe the issue in detail:

Same as #8321, it is important to always either consume or cancel the response body anyway, but it isn't correctly handled in the example code, which may mislead new learners:

> ```mjs
>   const response = await fetch(url);
> 
>   if (!response.ok || !response.body) {
>     throw new Error(`Failed to fetch ${url}. Status: ${response.status}`);
>   }
> ```

tenkirin:
## Description

Docs, fix #8331.

Added `await response.body?.cancel();` to explicitly destroy the response body before throwing an error.

```diff
  if (!response.ok || !response.body) {
+    // consuming the response body is mandatory: https://undici.nodejs.org/#/?id=garbage-collection
+    await response.body?.cancel();
    throw new Error(`Failed to fetch ${url}. Status: ${response.status}`);
  }
```

### Check List

### DIFF
--- a/apps/site/pages/en/learn/manipulating-files/reading-files-with-nodejs.md
+++ b/apps/site/pages/en/learn/manipulating-files/reading-files-with-nodejs.md
@@ -101,6 +101,8 @@ async function downloadFile(url, outputPath) {
   const response = await fetch(url);
 
   if (!response.ok || !response.body) {
+    // consuming the response body is mandatory: https://undici.nodejs.org/#/?id=garbage-collection
+    await response.body?.cancel();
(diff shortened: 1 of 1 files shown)


avivkeller (approved):
Thanks! In the future, there's no need to open an issue for these little things. In fact, if you want, you can also group them all into a single PR.

mikeesto (approved):

PR #8334, opened 2025-11-05
fix: download page mobile layout spacing

canerakdas:
## Description

With the addition of sponsors on the download page, the order of the sections has changed, which caused the selects to appear too close to each other on mobile viewports. This PR aims to fix that visual issue.

Adding ids to the sections would probably be a better long-term solution, but instead of resetting all existing translations(mdx files), it would make more sense to introduce those ids after a major structural update

### Before

image

#### After
image

### Check List

### DIFF
--- a/apps/site/layouts/layouts.module.css
+++ b/apps/site/layouts/layouts.module.css
@@ -114,7 +114,7 @@
 }
 
 .downloadLayout {
-  section:nth-last-child(1) {
+  section:nth-last-child(3) {
     @apply flex
       flex-col
       gap-2;
@@ -124,7 +124,7 @@
     }
   }
 
-  section:nth-last-child(2) p {
+  section:nth-last-child(4) p {
     @apply flex
       flex-row
       flex-wrap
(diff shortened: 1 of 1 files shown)


mikeesto (commented):
This looks like a good change! Although prettier isn't happy

bmuenzenmeyer:
> Adding ids to the sections would probably be a better long-term solution, but instead of resetting all existing translations(mdx files), it would make more sense to introduce those ids after a major structural update

yeah this looks just as prone to brittleness 

more thorough chromatic visual regression testing could fix this - though i dont think we've used that to stop builds much

canerakdas:
> This looks like a good change! Although prettier isn't happy

It seems that the sorting algorithm has changed with version [`0.7.0`](https://github.com/tailwindlabs/prettier-plugin-tailwindcss/releases) of `prettier-plugin-tailwindcss`. I’ve updated it for this style file, but after this PR, I’ll run a general linter check to see if there are other places in the project that need updating as well 🙌

avivkeller (approved):

PR #8340, opened 2025-11-07
fix(pnpm): pin doc-kit

avivkeller:
Pin doc-kit to fix tests


bmuenzenmeyer (approved):
I should have been louder about this, but then again this long standing feature branch might not have pulled it in anyway

bmuenzenmeyer:
failures should be resolved within #8339

PR #8346, opened 2025-11-10
Add a learn article explaining HTTP status code

raghavexe:
## Description

A new article for HTTP explaining various HTTP status codes 

## Validation

article is located on the learn page under the http header. Article name: "Understanding HTTP Status Codes"
### Check List

- [X ] I have read the [Contributing Guidelines](https://github.com/nodejs/nodejs.org/blob/main/CONTRIBUTING.md) and made commit messages that follow the guideline.
- [X ] I have run `pnpm format` to ensure the code follows the style guide.
- [X ] I have run `pnpm test` to check if all tests are passing.
- [ X] I have run `pnpm build` to check if the website builds without errors.
- [ X] I've covered new added functionality with unit tests if necessary.

### DIFF
--- a/apps/site/pages/en/learn/http/understanding-http-status-codes.md
+++ b/apps/site/pages/en/learn/http/understanding-http-status-codes.md
@@ -1,0 +1,164 @@
+---
+title: Understanding HTTP Status Codes
+layout: learn
+authors: raghavexe
+---
+
+# Understanding HTTP status codes
+
+The purpose of this guide is to provide a clear understanding of HTTP status
+codes and how to use them effectively when building servers in Node.js.
+If you’ve followed along with [`Anatomy of an HTTP Transaction`](https://nodejs.org/en/learn/http/anatomy-of-an-http-transaction), you’ve already seen how to send responses to clients. This article builds on that foundation by exploring how to communicate meaningful outcomes through proper HTTP status codes.
+
+## What Are Status Codes?
+
+Every HTTP response begins with a three-digit status code. These codes tell the client whether the request succeeded, failed, or requires further action. Without them, clients wouldn’t know what to do with your response.
+
+In Node.js, a response’s status code is controlled through the
+`statusCode` property:
+
+```js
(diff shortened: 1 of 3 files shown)


apps/site/pages/en/learn/http/understanding-http-status-codes.md
```diff
@@ -0,0 +1,164 @@
...
+title: Understanding HTTP Status Codes
+layout: learn
+authors: raghavexe
+---
+
+# Understanding HTTP status codes
+
+The purpose of this guide is to provide a clear understanding of HTTP status
+codes and how to use them effectively when building servers in Node.js.
+If you’ve followed along with [`Anatomy of an HTTP Transaction`](https://nodejs.org/en/learn/http/anatomy-of-an-http-transaction), you’ve already seen how to send responses to clients. This article builds on that foundation by exploring how to communicate meaningful outcomes through proper HTTP status codes.
```
avivkeller, line 11: Instead, this should be bold, not code.

Also, the link should be `/learn/...`

avivkeller (changes requested):
My concern here is that we are re-inventing the wheel.

If the user wants to see the different codes they can use, they should visit https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Status.

raghavexe:
> My concern here is that we are re-inventing the wheel.
> 
> If the user wants to see the different codes they can use, they should visit https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Status.

I do agree with that, but i saw it in a way that if the user is utilizing the learn path on the website, it would be convenient for them to have that concept cleared without having to deviating from the page

efekrskl:
Nice write-up, but I also don't think this should belong nodejs.org. HTTP concepts are general web fundamentals, not Node specific behaviour. We already explain [how to send the status code](https://nodejs.org/en/learn/http/anatomy-of-an-http-transaction#http-status-code), which covers the only Node related part of this article.

PR #8356, opened 2025-11-19
chore(ui-components): add README

avivkeller:
This PR gives our @node-core/ui-components package a nice little README for npm

### DIFF
--- a/packages/ui-components/scripts/publish.mjs
+++ b/packages/ui-components/scripts/publish.mjs
@@ -1,5 +1,5 @@
 import { spawnSync } from 'node:child_process';
-import { writeFile } from 'node:fs/promises';
+import { writeFile, copyFile } from 'node:fs/promises';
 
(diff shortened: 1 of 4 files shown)


packages/ui-components/scripts/publish.mjs
```diff
@@ -17,6 +17,8 @@ await writeFile(
   'utf8'
 );
 
+await copyFile('README.md', 'dist/README.md');
```
bmuenzenmeyer, line 20: IIRC, READMEs were included by default in `npm publish` - is that not the case anymore, or for some other reason?

avivkeller: This is different. Only the built JavaScript files are copied to the `dist` directory, and thus, we need to explicitly copy the README as well.

bmuenzenmeyer (approved):

PR #8377, opened 2025-11-27
feat: standardize sidebars

bmuenzenmeyer:
relates to https://github.com/nodejs/web-team/issues/56

## Description

Within https://github.com/nodejs/web-team/blob/main/meetings/2025-10-27.md we discussed https://github.com/nodejs/learn/issues/38, and specific to this PR, how multiple attendees noted that Learn content seems to indicate there is a progression (I mean, it's labeled that way in our code even!), when the content is often not linear at all. In addition to re-ordering and presenting content differently, I want to explore doing away with the progression logic altogether.

This PR therefore removes progress the icon variant to unify all sidebar experiences, and no longer suggests linear flow.

I first thought about simply removing the vertical lines between hexagons, but that looks like oddly executed list items then. I think this proposed approach makes the most sense from a consistency standpoint, so the UX is similar throughout the whole page to indicate active, hover, current states. Of particular improvement is the hover state, which does not exist on learn items currently.

## Validation

[live preview](https://nodejs-org-git-no-path-openjs.vercel.app/en/learn/getting-started/introduction-to-nodejs)

| BEFORE | AFTER|
| - | - |
| image | image |

### Check List


avivkeller (approved):

packages/ui-components/src/Containers/Sidebar/SidebarGroup/index.tsx
```diff
@@ -13,35 +13,26 @@ type SidebarGroupProps = {
...
-  showProgressionIcons,
   className,
   ...props
 }) => (
   <section
     className={classNames(
       {
         [styles.group]: true,
-        [styles.progression]: showProgressionIcons,
       },
```
avivkeller:
```suggestion
styles.group
```

packages/ui-components/src/Containers/Sidebar/SidebarItem/index.tsx
```diff
@@ -6,35 +6,24 @@ import BaseActiveLink from '#ui/Common/BaseActiveLink';
...
-  link,
-  showProgressionIcons = false,
-  ...props
-}) => (
+const SidebarItem: FC<SidebarItemProps> = ({ label, link, ...props }) => (
   <BaseActiveLink
     className={classNames({
       [styles.item]: true,
-      [styles.progression]: showProgressionIcons,
     })}
```
avivkeller:
```suggestion
className={styles.item}
```

PR #8378, opened 2025-11-27
refactor: CrossLink `Prev` to `Previous` and cleanup of `Pagination` keys

bmuenzenmeyer:
## Description

While browsing the site, I got bothered by the abbreviation `Prev` showing to users. We have the screen real estate to use the full word here. I was initially concerned about internationalization too, which made me look at `en.json` key composition. I then noticed we were defining `prev` and `next` in two places... and only using one, or worse, intermingling them. This PR unifies the keys under the more structured `common.pagination` ...now the text and the label draw from the same object.

I chose not to refactor `PaginationProps` or non-user-facing references at this time which still use things like `prev` and `prevAria`. I could be convinced to do so, but wanted to start here. This might also thrash Crowdin translations.

## Validation

image

### Check List

### DIFF
--- a/packages/i18n/src/locales/en.json
+++ b/packages/i18n/src/locales/en.json
@@ -251,7 +251,7 @@
         "navigateToHome": "Navigate to Home"
       },
       "crossLink": {
-        "previous": "Prev",
+        "previous": "Previous",
         "next": "Next"
       },
       "codebox": {
--- a/packages/i18n/src/locales/en.json
+++ b/packages/i18n/src/locales/en.json
@@ -246,17 +246,13 @@
       "breadcrumbs": {
         "navigateToHome": "Navigate to Home"
       },
-      "crossLink": {
-        "previous": "Previous",
-        "next": "Next"
-      },
       "codebox": {
         "copy": "Copy to clipboard",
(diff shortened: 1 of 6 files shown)


packages/i18n/src/locales/en.json
```diff
@@ -251,7 +247,7 @@
         "navigateToHome": "Navigate to Home"
       },
       "crossLink": {
-        "previous": "Prev",
+        "previous": "Previous",
         "next": "Next"
       },
```
avivkeller: Can we use `components.common.pagination` instead?

bmuenzenmeyer: we certainly can - that would eliminate this completely

bmuenzenmeyer: addressed via https://github.com/nodejs/nodejs.org/pull/8378/commits/1577fbbb4d797be4cb64abdcf532bda97648a4de - as i mentioned in the OP, this got larger because of the dynamic reference to the key which was using `prev` not `previous`

avivkeller (approved):

mikeesto (approved):

PR #8404, opened 2025-12-08
fix: anticipate banner for security release

marco-ippolito:
### Check List

### DIFF
--- a/apps/site/site.json
+++ b/apps/site/site.json
@@ -28,7 +28,7 @@
   ],
   "websiteBanners": {
     "index": {
-      "startDate": "2025-12-08T23:00:00.000Z",
+      "startDate": "2025-12-08T17:00:00.000Z",
       "endDate": "2025-12-15T23:00:00.000Z",
       "text": "New security releases to be made available Monday, December 15, 2025",
       "link": "https://nodejs.org/en/blog/vulnerability/december-2025-security-releases",


richardlau:
This only updates the blog post and not the banner.

MattIPv4 (commented):
This is just changing the timestamp of the blog post, not the banner?

MattIPv4:
(I think changing the date on the blog post would also cause RSS readers to pick it up again, which would not be ideal)

apps/site/pages/en/blog/vulnerability/december-2025-security-releases.md
```diff
@@ -1,5 +1,5 @@
 ---
-date: 2025-12-08T23:00:00.000Z
+date: 2025-12-08T17:00:00.000Z
```
MattIPv4: Changing this date is going to cause RSS readers to pick it up as a new post, which I doubt we want?

richardlau (approved):

marco-ippolito:
> (I think changing the date on the blog post would also cause RSS readers to pick it up again, which would not be ideal)

so we should not update the blogpost?

MattIPv4:
> so we should not update the blogpost?

Yeh I think it'd be best to revert that

MattIPv4 (approved):

PR #8431, opened 2025-12-19
chore: server-side render partners

avivkeller:
This PR moves all the partner page logic to the server side, reducing the size of the client bundle


apps/site/components/Common/Partners/index.tsx
```diff
@@ -0,0 +1,66 @@
...
+};
+
+const PartnersList: FC<PartnersListProps> = async ({
+  size = 'small',
+  category,
+  length,
+}) => {
+  let partners = await providePartners();
+
+  if (category) {
```
ovflowd: I know it is an async function so settings things on root of component doesn't matter, bujt can we wrap it on a function maybe? (outside)

apps/site/components/Common/Partners/index.tsx
```diff
@@ -0,0 +1,66 @@
...
+  if (size === 'small' || length) {
+    partners = partners.slice(0, length ?? 6);
+  }
+
+  return (
+    <div className={style[size]}>
+      {partners.map(partner => {
+        const Logo = PartnerLogos[partner.id as keyof typeof PartnerLogos];
+
+        if (size === 'small') {
```
ovflowd: Also kinda feels weird an ife else like this here? No need for an else right? Just a return?

ovflowd: Also this is evaluating size every time although it only needs to evaluate once?

apps/site/next-data/generators/partners.mjs
```diff
@@ -0,0 +1,13 @@
...
+
+/**
+ * Builds the final partner list used by the UI.
+ *
+ * @returns {Promise<import('#site/types/partners').Partner[]>}
+ */
+export default async function getPartnersList() {
+  const seed = Math.floor(Date.now() / (60 * 1000));
+
+  return await shuffle(partners, seed);
```
ovflowd: No need of await on return statement? You can just do Promise.resolve() to ensure it is a Promise? Although I get what you're trying to accomplish.

avivkeller:
(Converting to Draft to implement the weight logic)

apps/site/package.json
ovflowd, line 1: I'd wish this was in another PR but all good.

ovflowd (approved):
SGTM!

apps/site/package.json
avivkeller: Whoops! I was testing the bundle analyzer, and I must've left some artifacts, I'll undo it

apps/site/components/Common/Partners/index.tsx
```diff
@@ -15,15 +16,45 @@ import style from './index.module.css';
...
+    const seed = Math.floor(Date.now() / (minutes * 1000));
+
+    // Create a copy of the array to avoid modifying the original
+    const shuffled = await shuffle(weightedPartners, seed);
+
+    // Remove duplicates while preserving order
+    const unique = Array.from(new Set(shuffled));
+
+    result = [...unique];
+  }
```
avivkeller: Can we make this part of a provider? Why do we need to re-compute this on every route, when we can cache the result?

`providePartners` returns the weighted array. If `sort == name`, we can call `toSorted()` on that output. I know it seems redundant to take an unsorted array, weight it, and then put it back to a sorted array, but I think, with caching, it'll be a smarter solution. 

(Ref: https://openjs-foundation.slack.com/archives/CVAMEJ4UV/p1766259930877069?thread_ts=1766181387.005589&cid=CVAMEJ4UV, let me know if you want me to implement it)

apps/site/components/Common/Partners/PartnerButton/index.module.css
```diff
@@ -17,3 +17,15 @@
       !w-auto;
   }
 }
+
+.small {
+  @apply h-9
+    w-auto
+    min-w-9
+    p-2;
```
avivkeller:
```suggestion
```

This padding isn't needed in this new layout. In fact, it actually makes the icons not padded enough

apps/site/components/Common/Partners/index.tsx
```diff
@@ -0,0 +1,92 @@
...
+    result = filtered.toSorted((a, b) => a.name.localeCompare(b.name));
+  } else {
+    // For weight sorting, use cached weighted partners
+    const minutes = 300; // Change every 5 minutes
+    const seed = Math.floor(Date.now() / (minutes * 1000));
+    result = await providePartners(seed, category);
+  }
+
+  return length ? result.slice(0, length) : result;
+};
```
avivkeller, line 33: IMO `providePartners` should take no arguments. Each time the cache expires (every few minutes), it'll re-run, calculating the seed each time.

```suggestion
const getPartners = async (
  length?: number,
  category?: PartnerCategory,
  sort?: 'name' | 'weight'
) => {
  let result = sort === 'name' ? partners : await providePartners();
  
  result = category
      ? result.filter(p => p.categories.includes(category))
      : result;

  return length ? result.slice(0, length) : result;
};
```

apps/site/components/Common/Partners/index.tsx
```diff
@@ -0,0 +1,92 @@
+'use server';
+
+import Tooltip from '@node-core/ui-components/Common/Tooltip';
+import * as PartnerLogos from '@node-core/ui-components/Icons/PartnerLogos';
+
+import { providePartners } from '#site/next-data/providers/partners';
+import { partners as allPartners } from '#site/next.json.mjs';
```
avivkeller:
```suggestion
import providePartners from '#site/next-data/providers/partners';
import { partners } from '#site/next.json.mjs';
```

apps/site/next-data/providers/partners.ts
```diff
@@ -0,0 +1,31 @@
...
+  });
+
+  // Shuffle and remove duplicates
+  const shuffled = await shuffle(weightedPartners, seed);
+  const unique = Array.from(new Set(shuffled));
+
+  return category
+    ? unique.filter(p => p.categories.includes(category))
+    : unique;
+};
```
avivkeller: All of this logic should go in a generator, which this file doing nothing but:
```mjs
'use cache';

import providePartners from '#site/next-data/generators/partners.mjs';

export default providePartners;
```

apps/site/next-env.d.ts
avivkeller, line 1: Can you revert this (it's my fault it's here).

Just run `git reset origin/main apps/site/next-env.d.ts apps/site/package.json pnpm-lock.yaml`

apps/site/package.json
avivkeller, line 1: Can you revert this (it's my fault it's here).

Just run `git reset origin/main apps/site/next-env.d.ts apps/site/package.json pnpm-lock.yaml`

pnpm-lock.yaml
avivkeller, line 1: Can you revert this (it's my fault it's here).

Just run `git reset origin/main apps/site/next-env.d.ts apps/site/package.json pnpm-lock.yaml`

apps/site/components/Common/Partners/index.tsx
```diff
@@ -0,0 +1,92 @@
...
+
+  if (sort === 'name') {
+    // For name sorting, just filter and sort directly - no need for weighted shuffling
+    const filtered = category
+      ? allPartners.filter(p => p.categories.includes(category))
+      : allPartners;
+    result = filtered.toSorted((a, b) => a.name.localeCompare(b.name));
+  } else {
+    // For weight sorting, use cached weighted partners
+    const minutes = 300; // Change every 5 minutes
```
efekrskl: Are these minutes or seconds?

apps/site/components/Common/Partners/index.tsx
```diff
@@ -0,0 +1,92 @@
...
+const PartnersList: FC<PartnersListProps> = async ({
+  size = 'small',
+  category,
+  sort = 'name',
+  length,
+}) => {
+  const isSmall = size === 'small';
+
+  const partners = await getPartners(
+    length ?? (isSmall ? 6 : undefined),
```
efekrskl: Maybe something like `SMALL_PARTNER_LIMIT` instead of a magic number?

apps/site/types/partners.ts
```diff
@@ -1,6 +1,6 @@
 import type { ReactElement, SVGProps } from 'react';
 
-export type Partners = {
+export type Partner = {
   id: string;
```
efekrskl: We could just set the id to `keyof typeof PartnerLogos` instead of asserting it each time we want to use it as a key. We have the values hardcoded anyway, so I don't think it would cause any issues + would provide some nice type completion

apps/site/components/Common/Partners/PartnerButton/index.module.css
```diff
@@ -1,6 +1,6 @@
 @reference "../../../../styles/index.css";
 
-.partnerIcon {
+.big {
```
canerakdas: As with the other components, it would be more appropriate to use `small` `medium` or `large` for sizing, since `big` is a somewhat subjective naming choice

apps/site/components/Common/Partners/PartnerButton/index.module.css
```diff
@@ -17,3 +17,15 @@
...
+
+.small {
+  @apply h-9
+    w-auto
+    min-w-9
+    p-2;
+
+  svg {
+    @apply !h-4
+      !w-auto;
```
canerakdas: It doesn't seem like we need to use important here

apps/site/components/Common/Partners/PartnerButton/index.tsx
```diff
@@ -0,0 +1,26 @@
+import Button from '#site/components/Common/Button';
+
+import type { AnchorHTMLAttributes, FC } from 'react';
+
+import style from './index.module.css';
+
+type PartnerProps = AnchorHTMLAttributes<HTMLAnchorElement> & {
+  size?: 'big' | 'small';
```
canerakdas:
```suggestion
  size?: 'small' | 'large';
```

I think small and large would be better naming choices, since we use similar naming conventions in our other components as well

apps/site/pages/fr/about/partners.mdx
```diff
@@ -18,7 +18,7 @@ Les partenaires infrastructurels apportent un soutien important au projet Node.j
 en fournissant le matériel et les machines nécessaires à nos processus d'intégration et de test continus,
 sans lesquels nous ne pourrions pas tester et publier de nouvelles versions de Node.js.
 
-<PartnersLogoList categories="infrastructure" maxLength={null} sort="name" />
+<PartnersList size="big" category="infrastructure" />
```
canerakdas: We usually make changes in en and then remove the other languages to let them sync again. Would manually editing these cause any issues on the Crowdin side?

avivkeller: That is usually the case, but we are changing the structure of this component, and I fear that it'll cause crashes on the other locales if we don't change it here (but I haven't tested that theory)

bjohansebas: Yep, it’s necessary here, because otherwise the i18n breaks.

apps/site/components/Common/Partners/index.tsx
bjohansebas: From what I understood from the documentation, the cache is invalidated based on the arguments, if they’re different, then it’s invalidated. 

How often does the cache expire here in Next.js?

apps/site/types/partners.ts
bjohansebas: I can’t quite understand what you mean, could you show it as a code suggestion, please?

efekrskl:
```suggestion
  id: keyof typeof PartnerLogos;
```

Then we could remove the assertions from `apps/site/components/Common/Partners/index.tsx` lines 47 and 62

apps/site/components/Common/Partners/index.tsx
avivkeller: > How often does the cache expire here in Next.js?

5 minutes - 15 minutes

bjohansebas:
going back to the draft, why can't we the current solution, Or no, wait, let's wait for it to be fixed in main, and then check it here

avivkeller:
@bjohansebas I'll rebase this PR once my fix is opened + lands

apps/site/components/Common/Partners/index.tsx
```diff
@@ -0,0 +1,85 @@
+'use server';
```
bjohansebas: now that I remember, the reason why it wasn't done on the server side, was because since we supported static export, we wanted the randomizer to keep working even if the page was static, and this was imposed in the issue, so we could make it work on the server side and still work statically, unfortunately the randomizer for the home page won't work anymore and will always show the partners in the same order 😔

Image

https://github.com/nodejs/nodejs.org/issues/7909#issuecomment-3080114652


so do we want to proceed with this pr?

avivkeller: I think we should _attempt_ to proceed as much as possible, and if it comes to a point where we can't, then we can't.

avivkeller:
@nodejs/nodejs-website PTAL

ovflowd:
> # 📦 Build Size Comparison
> 
> ## Summary
> 
> | Metric | Value |
> |--------|-------|
> | Old Total Size | 4.41 MB |
> | New Total Size | 3.98 MB |
> | Delta | -439.56 KB (-9.74%) |
> 
> ### Changes
> 
> <details>
> <summary>➕ Added Assets <strong>(9)</strong></summary>
> 
> | Name | Size |
> | --- | --- |
> | `.next/static/chunks/d7e20b6111f3e408.js` | 205.09 KB |
> | `.next/static/chunks/45b88b380febe3f1.js` | 101.82 KB |
> | `.next/static/chunks/7d19e315e828bf5f.js` | 101.82 KB |
> | `.next/static/chunks/8d389e1cd3e91770.js` | 27.57 KB |
> | `.next/static/chunks/a017a80ed43c7efc.js` | 101.82 KB |
> | `.next/static/chunks/340f0ef2ee3cd2f5.js` | 27.74 KB |
> | `.next/static/chunks/8288238ba8b20c26.js` | 561.22 KB |
> | `.next/static/chunks/74fbea75b8b661b2.js` | 561.22 KB |
> | `.next/static/chunks/84b4b6c5f0c7e49f.js` | 101.82 KB |
> 
> </details>
> 
> <details>
> <summary>➖ Removed Assets <strong>(10)</strong></summary>
> 
> | Name | Size |
> | --- | --- |
> | `.next/static/chunks/948dbad21b6ae055.js` | 204.26 KB |
> | `.next/static/chunks/472c79f905eeb9d5.js` | 101.79 KB |
> | `.next/static/chunks/39fc6610b9d853cf.js` | 101.79 KB |
> | `.next/static/chunks/83746613e033c5eb.js` | 27.28 KB |
> | `.next/static/chunks/c3e76e8ccf908f28.js` | 101.79 KB |
> | `.next/static/chunks/440c65068d6fbae2.js` | 27.74 KB |
> | `.next/static/chunks/c7363092071b3daf.js` | 767.74 KB |
> | `.next/static/chunks/11128c08875d4595.js` | 27.74 KB |
> | `.next/static/chunks/eb4fdb24df4a0411.js` | 767.74 KB |
> | `.next/static/chunks/d90f22899206d0e3.js` | 101.79 KB |
> 
> </details>
> 

Completely unrelated thing, but can we decrease the heading size of this comment? It is gigantic compared to the other ones 😭

ovflowd (approved):
LGTM! Good stuff!

PR #8441, opened 2025-12-22
chore(next): bump version of nextjs and related deps

avivkeller:
This PR bumps the dependencies related to Next.js, bringing in new features + smaller install bundles.

Notably, `next experimental-analyze` will allow us to better identify bundle improvements.


MattIPv4:
Is there a reason we're doing this by hand rather than allowing Dependabot to its job for us?

avivkeller:
> Is there a reason we're doing this by hand rather than allowing Dependabot to its job for us?

Yes, I want to dig more into bundle size improvements using features introduced in 16.1 (i.e. `next experimental-analyze`)

MattIPv4:
I see -- should this be a draft then if you're just using it to explore the new features? And then once Dependabot does the actual updates, we can use those features?

avivkeller:
> I see -- should this be a draft then if you're just using it to explore the new features? And then once Dependabot does the actual updates, we can use those features?

I'd prefer to land this, since it'll allow me to `experimental-analyze` compare main to proposed bundle improvements, rather than comparing a soon-to-be out of date feature branch. If you'd like, I can trigger Dependabot to run early?

MattIPv4:
I'm not sure I follow why this needs to land to allow you to do that? Why can't you bump the packages locally on main, and then compare that against other changes you wish to make locally?

But yes, if we do need to do dependency updates sooner, I'd prefer that Dependabot is still used for it, so that the team can review those bumps in isolation, rather than a bunch of bumps in a single PR that is much harder to review safely.

avivkeller:
> But yes, if we do need to do dependency updates sooner, I'd prefer that Dependabot is still used for it, so that the team can review those bumps in isolation, rather than a bunch of bumps in a single PR that is much harder to review safely.

Noted, no need to run early, I think it'll run soon anyway.

> I'm not sure I follow why this needs to land to allow you to do that? Why can't you bump the packages locally on main, and then compare that against other changes you wish to make locally?

I suppose that's better. We can wait the few days for Dependabot to run.

PR #8444, opened 2025-12-23
refactor: searchbox flash issue and minor improvements

canerakdas:
## Description
With this PR, we aim to fix the flashing issue in the search box for users using dark theme. In addition to this fix, I have increased the spacing between the icon and the text, and hide the keyboard shortcut on mobile resolutions since the keyboard is usually not present in most scenarios.

Even if it doesn’t address the root cause, it solves the problem for now. I’ll create a separate issue for the hydration problems, which I believe are the underlying cause 🖖 

## Validation

There should be no flash in the searchbox in the preview build

#### Before

Default;

image
image

On Hover;

image
image

On mobile;
image

#### After

Default;

image
image

On Hover;

image
image

On mobile;

image

### Check List

### DIFF
--- a/packages/ui-components/src/Common/Search/Modal/index.module.css
+++ b/packages/ui-components/src/Common/Search/Modal/index.module.css
@@ -22,13 +22,21 @@
     bg-white
     p-1.5
     text-neutral-900
-    duration-300
-    hover:bg-neutral-100
-    motion-safe:transition-colors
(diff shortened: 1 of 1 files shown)


avivkeller (approved):

ovflowd (approved):

ovflowd:
Let's fast-track to fix this issue.

PR #8447, opened 2025-12-24
chore(bundle): don't send translations to client

avivkeller:
Currently, per Next.js's bundle analyzer, all locales are sent to the client.

This PR removes the `importLocale` function, as I suspect it was the culprit.


packages/i18n/src/index.mjs
```diff
@@ -2,40 +2,22 @@
 
 import localeConfig from './config.json' with { type: 'json' };
 
-/**
```
ovflowd, line 5: You do realize by that removing this you're completely breaking HMR for next-intl, right?

ovflowd (changes requested):
I don't think OP understands the side effects of their changes.

Please read next-intl docs and triple check you're not breaking things unintentionally...

<sup>Edit: the phrasing above was poor, I meant to say that Im not confident of the changes and that OP has checked if every next-intl feature is still working as expected. I absolutely believe on OP's skill and just wanted to ensure that they verify that everything is still working as expected</sup>

packages/i18n/src/index.mjs
avivkeller: I don't think I am? We are still using dynamic import, just not this function.

Based on my analysis, this function was causing the locales to be bundled _everywhere_, but using `await import` directly (in `apps/site/i18n.tsx`) did not cause that

ovflowd: Can you triple confirm hmr still works both on english and translated languages?

ovflowd: If it is, then LGTM

avivkeller: > Can you triple confirm hmr still works both on english and translated languages?

Yes, it works.

ovflowd (approved):
SGTM! Appreciate for triple checking on my concerns.

PR #8448, opened 2025-12-24
fix(ci): use `##` instead of `#` when comparing

avivkeller:
cc @ovflowd 

> Completely unrelated thing, but can we decrease the heading size of this comment? It is gigantic compared to the other ones 😭

---

# This is a `#`

### This is a `###`


ovflowd (approved):
The PR description is legendary lol

canerakdas (approved):

avivkeller:
(I'm aware the comment above still used the large header. For safety reasons, we generate the comment from the main branch, and not the PR branch)

mikeesto (approved):

PR #8450, opened 2025-12-25
feat(index): add empty line to even code box size

### LINKED ISSUE #8425: [Suggestion] Expand the streams pipeline example to prevent the screen from flickering

An example area is flickering on example switching. All examples consists of 14 lines but only "streams pipeline" have 13 lines. Just add an empty line before a last comment to prevent flickering.

ATiwari03-03-2004:
## Description

This PR fixes the variable height issue of code block by introducing fixed height for the code block

## Validation

https://github.com/user-attachments/assets/36dfa588-1968-4606-a952-844912dce35d

## Related Issues

Fixes #8425

### Check List

### DIFF
--- a/apps/site/pages/en/index.mdx
+++ b/apps/site/pages/en/index.mdx
@@ -103,6 +103,7 @@ layout: home
     );
 
     // run with `node streams.mjs`
+
     ```
 
     ```js displayName="Work with Threads"
--- a/apps/site/pages/en/index.mdx
+++ b/apps/site/pages/en/index.mdx
@@ -102,8 +102,8 @@ layout: home
       createWriteStream('package.json.gz')
     );
 
-    // run with `node streams.mjs`
 
+    // run with `node streams.mjs`
     ```
 
     ```js displayName="Work with Threads"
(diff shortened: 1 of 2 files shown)


avivkeller (changes requested):
Don't we just need to make all the code boxes the same height, without CSS changes?

ATiwari03-03-2004:
@avivkeller  I used CSS because the height difference was coming from content length across tabs, and there wasn’t an existing prop or layout constraint to normalize it. When you say “without CSS changes,” do you mean setting a fixed height at the component level or something else?

avivkeller:
I mean just adding a line to the code block, surely there's a way we can get it to the 14? lines it needs to be

ATiwari03-03-2004:
Oh ok, then i will add an extra line to code block of Streams pipeline section, that ought to do it

apps/site/pages/en/index.mdx
```diff
@@ -103,6 +103,7 @@ layout: home
     );
 
     // run with `node streams.mjs`
+
```
avivkeller, line 105: Perhaps put the line before the comment?

ATiwari03-03-2004: I have shifted it just above comment

avivkeller (approved):

ovflowd (approved):
This fix is so funny. Anyhow, SGTM

ovflowd:
@nodejs/nodejs-website fast-tracking this PR as minor UX fix.

PR #8458, opened 2025-12-26
feat: inline copy feedback for code blocks

### LINKED ISSUE #8357: Replace Toast notifications with inline feedback for "Code Copied" action
### **Description**

**Problem**
Currently, when a user clicks the "Copy" button on a code block, the website triggers a global toast notification at the bottom of the screen to indicate success.

While visually common, this pattern introduces significant accessibility and usability issues. Recent discussions (referenced below) highlight that toast notifications are often problematic because:

1.  **Timing:** They often disappear too quickly for some users to read.
2.  **Focus/Screen Readers:** Assistive technology often misses them if focus isn't managed perfectly, leaving the user unsure if the action succeeded.
3.  **Cognitive Load:** They appear far away from the user's current locus of attention (the code block), making them easy to overlook.

**Reference:** [Primer Design System: Toast Accessibility](https://primer.style/accessibility/toasts/)

### **Proposed Solution**

We should remove the Toast dependency for the "Copy" action and instead implement **inline feedback**.

When the user clicks the copy button:

  * The feedback should appear **inside the code block header**, directly adjacent to or replacing the button itself.
  * **Example:** The copy icon could briefly change to a "Checkmark" icon, or the text "Copy" could change to "Copied\!".
  * **Benefit:** This keeps the feedback within the user's current field of vision and is more logically associated with the trigger action.

[Ref Slack message](https://openjs-foundation.slack.com/archives/CVAMEJ4UV/p1763518987820279?thread_ts=1763518987.820279&cid=CVAMEJ4UV)

cnaples79:
## Summary
- replace toast-based copy confirmation with inline feedback on the code block button
- reuse the existing copy hook to toggle button text/icon to "Copied" after clicking

## Rationale
- keeps feedback adjacent to the action and removes the global toast dependency (accessibility and clarity)

## Changes
- CodeBox now uses useCopyToClipboard state for inline status and no longer calls NotificationProvider
- BaseCodeBox accepts an optional copied flag to swap the copy icon for a checkmark

Fixes #8357


mikeesto (commented):
Looks good, thanks! Could you please run prettier on `apps/site/components/Common/CodeBox.tsx` to fix CI?

packages/ui-components/src/Common/BaseCodeBox/index.tsx
```diff
@@ -81,6 +83,7 @@ const BaseCodeBox: FC<PropsWithChildren<CodeBoxProps>> = ({
   buttonText,
   as = 'a',
   showCopyButton = true,
+  copied = false,
```
ovflowd: I don't think you should add this prop, you can change the `buttonText` prop to be of type ReactNode and pass the icon directly **with** the text.

ovflowd (commented):
I believe you can remove the whole source of NotificationProvider, including tests, stories and remove it from the Layout too. Which would also allow you to remove the "use client" from the layout. (I think?)

packages/ui-components/src/Common/BaseCodeBox/index.tsx
avivkeller: In that case, we should also rename it to be something more descriptive, i.e. `buttonContent`

avivkeller:
@cnaples79 Please run `pnpm version minor` in the `packages/ui-components` directory, so we can re-use this change on `doc-kit`. Let me know if you need help with @ovflowd's request above

avivkeller:
Hi @cnaples79! Thanks for the PR, let me know if you need help with @ovflowd's request above, i'd be happy to push a commit helping :-)

apps/site/components/Common/CodeBox.tsx
```diff
@@ -14,34 +15,30 @@ import type { FC, PropsWithChildren } from 'react';
...
+      onCopy={copyToClipboard}
+      buttonContent={
+        copied ? (
+          <>
+            <DocumentDuplicateIcon className="size-4" />
+            {t('components.common.codebox.copied')}
+          </>
+        ) : (
+          <>
+            <CodeBracketIcon className="size-4" />
```
ovflowd, line 36: What is this icon?

**Also nit:** (Cleared and less code)


```tsx

...

buttonContext={
  <>
    <ButtonIcon className="size-4" />
    {t(copied ? 
        'components.common.codebox.copied' :
        'components.common.codebox.copy'
    )}
  </>
}
```

apps/site/components/Common/CodeBox.tsx
```diff
@@ -14,34 +15,30 @@ import type { FC, PropsWithChildren } from 'react';
 type CodeBoxProps = {
   language: string;
   className?: string;
-  showCopyButton?: boolean;
```
ovflowd, line 17: Why have you removed this? There are situations we don't want to render the copy button.

ovflowd: Or was it not being used on this component?

packages/rehype-shiki/src/plugin.mjs
```diff
@@ -181,15 +181,6 @@ export default async function rehypeShikiji(options) {
         codeLanguage
       );
 
-      // Adds a Copy Button to the CodeBox if requested as an additional parameter
```
ovflowd, line 184: 🤔

ovflowd (changes requested):
These changes diverge so much from the original PR provided by OP that these should be a new PR. IMO, with how distinct HEAD is now from the previous commit, this completely removes the "co-author" factor from the original PR author.

apps/site/components/Common/CodeBox.tsx
avivkeller: > There are situations we don't want to render the copy button.

There are no situations where `showCopyButton` is falsey, so I removed it

avivkeller:
> These changes diverge so much from the original PR provided by OP that these should be a new PR. IMO, with how distinct HEAD is now from the previous commit, this completely removes the "co-author" factor from the original PR author.

Sorry! I meant to help OP expand their PR by removing the NotificationProvider + fixup some Code box changes, I definitely did not mean to overtake this PR 😓

apps/site/components/Common/CodeBox.tsx
avivkeller: > What is this icon?

What do you mean? This is the icon we used previously, is it not?

ovflowd: Yeah, I realized that afterwards.

ovflowd:
> > These changes diverge so much from the original PR provided by OP that these should be a new PR. IMO, with how distinct HEAD is now from the previous commit, this completely removes the "co-author" factor from the original PR author.
> 
> Sorry! I meant to help OP expand their PR by removing the NotificationProvider + fixup some Code box changes, I definitely did not mean to overtake this PR 😓

It's fine to address some changes, but the extra additions completely overtake the PR, it's "fine", I do believe OP abandoned the PR (possibly), but please make a new PR. (and if possible revert your commit in this PR), then we approve yours, merge and close this one.

ovflowd:
@avivkeller to clarify, as I'm unsure if I came across wrong, your changes are 100% ✅, just wanted if possible to have them done in another PR (the other PR can be the exact 1:1 diff of this PR) but authored by you :)

PR #8459, opened 2025-12-26
[automated]: crowdin sync

nodejs-crowdin:
New Crowdin translations from the [Node.js Crowdin project](https://crowdin.com/project/nodejs-web)


avivkeller (approved):

packages/i18n/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@node-core/website-i18n",
-  "version": "1.1.14",
+  "version": "1.1.15",
```
bjohansebas, line 3: Is this necessary?

bjohansebas (approved):

packages/i18n/package.json
avivkeller: Yes. If/when we use these translations in doc-kit, it's important they stay up-to-date.

PR #8493, opened 2025-12-30
feat(ui): inline code block feedback

### LINKED ISSUE #8357: Replace Toast notifications with inline feedback for "Code Copied" action

avivkeller:
Closes #8458
Fixes #8357

This PR removes the `NotificationProvider` in favor of inline feedback, and removes the un-used `showCopyButton` property from `CodeBox`


apps/site/components/Common/CodeBox.tsx
```diff
@@ -14,34 +15,30 @@ import type { FC, PropsWithChildren } from 'react';
...
-      ),
-    });
-  };
-
   return (
     <BaseCodeBox
       as={Link}
-      onCopy={onCopy}
+      onCopy={copyToClipboard}
+      buttonContent={
```
ovflowd, line 30: **nit:** (Cleared and less code)


```tsx

...

buttonContext={
  <>
    <ButtonIcon className="size-4" />
    {t(copied ? 
        'components.common.codebox.copied' :
        'components.common.codebox.copy'
    )}
  </>
}
```

packages/ui-components/package.json
```diff
@@ -44,7 +44,6 @@
     "@radix-ui/react-select": "~2.2.6",
     "@radix-ui/react-separator": "~1.1.8",
     "@radix-ui/react-tabs": "~1.1.13",
-    "@radix-ui/react-toast": "~1.2.15",
```
ovflowd, line 47: Remove also from next.config.mjs

ovflowd (approved):
SGTM with a few suggstions

ovflowd:
> 36.81 KB (+0.97%)

Interestingly the bundle became bigger?

apps/site/components/Common/CodeBox.tsx
avivkeller: I prefer the current implementation, do you insist on this?

ovflowd: I'd insist yes, if possible. I'm ok with not doing it, but genuinely believe it is cleaner. Less JSX snippets, the ternaries are smaller and occupy less space. Less code duplication. We can get more opinions if you wish 🙇

ovflowd:
Now we have a funny situation, technically speaking these changes (some of them) come from another PR that has been there for days (also the original commit comes from a commit you authored on that same PR). So technically speaking this isn't a fresh PR.

@nodejs/nodejs-website do y'all agree into fast-tracking this? This PR can also sit and wait.

PR #8500, opened 2025-12-31
chore(ci): bump scorecard

avivkeller:
Blocked by https://github.com/nodejs/web-team/pull/85


.github/workflows/scorecard.yml
MattIPv4, line 1: We should probably also switch to `permissions: {}` at the top-level here too?

MattIPv4 (approved):

PR #8516, opened 2026-01-03
chore: simplify hooks

avivkeller:
This PR simplfiies our usage of `useXYZ` hooks. In the process, it moves `useCopyToClipboard` to a similar `useCopy` hook in `@node-core/ui-components`, so that `CodeBox` performs it's copy logic internally. (This allows for less code to duplicate in `doc-kit`)


packages/ui-components/src/hooks/useCopy.ts
```diff
@@ -0,0 +1,34 @@
...
+  } catch {
+    return false;
+  }
+};
+
+export default function useCopy() {
+  const [copied, setCopied] = useState(false);
+
+  const copyText = useCallback(async (text?: string) => {
+    const success = await copyToClipboard(text);
```
araujogui, line 20: I’m not sure if ignoring the error is the best idea.

araujogui: Maybe we should show user that the text could not be copied

araujogui: Althought I think this is a very rare case

packages/ui-components/src/Common/Select/utils.ts
```diff
@@ -0,0 +1,24 @@
+export const isStringArray = (
+  values: Array<unknown>
+): values is Array<string> =>
+  Boolean(values[0] && typeof values[0] === 'string');
+
+export const isValuesArray = <T>(values: Array<unknown>): values is Array<T> =>
+  Boolean(values[0] && typeof values[0] === 'object' && 'value' in values[0]);
+
+export const mapValues = (values: Array<unknown>) => {
+  let mappedValues = values;
```
araujogui, line 10: nit: I usually prefer a more FP approach, I think it's easier to read.

araujogui (changes requested):

packages/ui-components/src/hooks/useCopy.ts
avivkeller: Can that be a follow-up? The ignore is based on the previous iteration:
```js
return navigator.clipboard
    .writeText(value)
    .then(() => true)
    .catch(() => false);
```

packages/ui-components/src/Common/Select/utils.ts
avivkeller: If this is a nit, why is this a requested changes?

packages/ui-components/src/hooks/useCopy.ts
araujogui: Sure! Np

packages/ui-components/src/Common/Select/utils.ts
araujogui: mistake, feel free to ignore this comment

araujogui (approved):
LGTM

PR #8517, opened 2026-01-06
feat(ui): implement sidebar scroll position preservation

### LINKED ISSUE #8336: Selected article in sidebar doesn’t scroll into view

In the Learn section sidebar, the currently selected article (highlighted in the list) should automatically scroll into view when navigating between pages

Right now, when you open or switch to a new article, the sidebar doesn't adjust (it actually resets). We have plenty of articles, and it is really hard to navigate the page like this.

I think [the Rust Book](https://doc.rust-lang.org/book/ch08-00-common-collections.html)'s sidebar behaviour could be a good example

Steps to reproduce:

- Navigate to an article that is deep down in the sidebar (e.g https://nodejs.org/en/learn/diagnostics/memory)
- Sidebar reset back to the top

See:

Image

malav2110:
## Description

This PR adds functionality to preserve sidebar scroll location on Learn Page

## Validation

Before:
https://github.com/user-attachments/assets/41ca31eb-72c2-447a-a6a7-2d451171775e

After:
https://github.com/user-attachments/assets/656713ee-5d03-430f-beda-8ae491f415a8

## Related Issues

Fixes https://github.com/nodejs/nodejs.org/issues/8336

### Check List

### DIFF
--- a/apps/site/components/withSidebar.tsx
+++ b/apps/site/components/withSidebar.tsx
@@ -1,15 +1,13 @@
 'use client';
 
 import Sidebar from '@node-core/ui-components/Containers/Sidebar';
-import { usePathname } from 'next/navigation';
-import { useLocale, useTranslations } from 'next-intl';
+import { useTranslations } from 'next-intl';
 import { useRef } from 'react';
 
 import Link from '#site/components/Link';
-import { useClientContext } from '#site/hooks/client';
-import { useNavigationState } from '#site/hooks/client';
+import { useClientContext, useNavigationState } from '#site/hooks/client';
 import { useSiteNavigation } from '#site/hooks/generic';
-import { useRouter } from '#site/navigation.mjs';
+import { useRouter, usePathname } from '#site/navigation.mjs';
 
 import type { NavigationKeys } from '#site/types';
 import type { RichTranslationValues } from 'next-intl';
@@ -23,14 +21,11 @@ type WithSidebarProps = {
 const WithSidebar: FC<WithSidebarProps> = ({ navKeys, context, ...props }) => {
   const { getSideNavigation } = useSiteNavigation();
   const pathname = usePathname()!;
-  const locale = useLocale();
   const t = useTranslations();
   const { push } = useRouter();
   const { frontmatter } = useClientContext();
   const sidebarRef = useRef<HTMLElement>(null);
   const sideNavigation = getSideNavigation(navKeys, context);
-
-  const localePathname = pathname.replace(`/${locale}`, '');
 
   // Preserve sidebar scroll position across navigations
   useNavigationState('sidebar', sidebarRef);
@@ -49,7 +44,7 @@ const WithSidebar: FC<WithSidebarProps> = ({ navKeys, context, ...props }) => {
     <Sidebar
       ref={sidebarRef}
       groups={mappedSidebarItems}
-      pathname={localePathname}
+      pathname={pathname}
       title={t('components.common.sidebar.title')}
       placeholder={frontmatter?.title}
       onSelect={push}
--- a/apps/site/components/withSidebar.tsx
+++ b/apps/site/components/withSidebar.tsx
@@ -1,7 +1,7 @@
 'use client';
 
 import Sidebar from '@node-core/ui-components/Containers/Sidebar';
-import { useTranslations } from 'next-intl';
+import { useLocale, useTranslations } from 'next-intl';
 import { useRef } from 'react';
 
 import Link from '#site/components/Link';
@@ -21,6 +21,7 @@ type WithSidebarProps = {
 const WithSidebar: FC<WithSidebarProps> = ({ navKeys, context, ...props }) => {
   const { getSideNavigation } = useSiteNavigation();
   const pathname = usePathname()!;
+  const locale = useLocale();
   const t = useTranslations();
   const { push } = useRouter();
   const { frontmatter } = useClientContext();
@@ -44,7 +45,7 @@ const WithSidebar: FC<WithSidebarProps> = ({ navKeys, context, ...props }) => {
     <Sidebar
       ref={sidebarRef}
       groups={mappedSidebarItems}
-      pathname={pathname}
+      pathname={pathname.replace(`/${locale}`, '')}
       title={t('components.common.sidebar.title')}
       placeholder={frontmatter?.title}
       onSelect={push}
--- a/apps/site/components/withSidebar.tsx
+++ b/apps/site/components/withSidebar.tsx
@@ -6,14 +6,14 @@ import { useLocale, useTranslations } from 'next-intl';
 import { useRef } from 'react';
 
 import Link from '#site/components/Link';
+import { useClientContext } from '#site/hooks/client';
+import { useNavigationState } from '#site/hooks/client';
+import { useSiteNavigation } from '#site/hooks/generic';
 import { useRouter } from '#site/navigation.mjs';
 
 import type { NavigationKeys } from '#site/types';
 import type { RichTranslationValues } from 'next-intl';
 import type { FC } from 'react';
-
-import { useClientContext, useNavigationState } from '../hooks/client';
-import { useSiteNavigation } from '../hooks/generic';
 
 type WithSidebarProps = {
   navKeys: Array<NavigationKeys>;
--- a/apps/site/components/withSidebar.tsx
+++ b/apps/site/components/withSidebar.tsx
@@ -5,7 +5,7 @@ import { useTranslations } from 'next-intl';
 import { useRef } from 'react';
 
 import Link from '#site/components/Link';
-import { useClientContext, useNavigationState } from '#site/hooks/client';
+import { useClientContext, useScrollToElement } from '#site/hooks/client';
 import { useSiteNavigation } from '#site/hooks/generic';
 import { useRouter, usePathname } from '#site/navigation.mjs';
 
@@ -28,7 +28,7 @@ const WithSidebar: FC<WithSidebarProps> = ({ navKeys, context, ...props }) => {
   const sideNavigation = getSideNavigation(navKeys, context);
 
   // Preserve sidebar scroll position across navigations
-  useNavigationState('sidebar', sidebarRef);
+  useScrollToElement('sidebar', sidebarRef);
 
   const mappedSidebarItems =
     // If there's only a single navigation key, use its sub-items
--- a/apps/site/components/withSidebar.tsx
+++ b/apps/site/components/withSidebar.tsx
@@ -3,15 +3,17 @@
 import Sidebar from '@node-core/ui-components/Containers/Sidebar';
 import { usePathname } from 'next/navigation';
 import { useLocale, useTranslations } from 'next-intl';
+import { useRef } from 'react';
 
 import Link from '#site/components/Link';
-import { useClientContext } from '#site/hooks/client';
-import { useSiteNavigation } from '#site/hooks/generic';
 import { useRouter } from '#site/navigation.mjs';
 
 import type { NavigationKeys } from '#site/types';
 import type { RichTranslationValues } from 'next-intl';
 import type { FC } from 'react';
+
+import { useClientContext, useNavigationState } from '../hooks/client';
+import { useSiteNavigation } from '../hooks/generic';
 
 type WithSidebarProps = {
   navKeys: Array<NavigationKeys>;
@@ -25,7 +27,13 @@ const WithSidebar: FC<WithSidebarProps> = ({ navKeys, context, ...props }) => {
   const t = useTranslations();
   const { push } = useRouter();
   const { frontmatter } = useClientContext();
+  const sidebarRef = useRef<HTMLElement>(null);
   const sideNavigation = getSideNavigation(navKeys, context);
+
+  const localePathname = pathname.replace(`/${locale}`, '');
+
+  // Preserve sidebar scroll position across navigations
+  useNavigationState('sidebar', sidebarRef);
 
   const mappedSidebarItems =
     // If there's only a single navigation key, use it's sub-items
@@ -39,8 +47,9 @@ const WithSidebar: FC<WithSidebarProps> = ({ navKeys, context, ...props }) => {
 
   return (
     <Sidebar
+      ref={sidebarRef}
       groups={mappedSidebarItems}
-      pathname={pathname.replace(`/${locale}`, '')}
+      pathname={localePathname}
       title={t('components.common.sidebar.title')}
       placeholder={frontmatter?.title}
       onSelect={push}
--- a/apps/site/components/withSidebar.tsx
+++ b/apps/site/components/withSidebar.tsx
@@ -1,7 +1,7 @@
 'use client';
 
 import Sidebar from '@node-core/ui-components/Containers/Sidebar';
-import { useLocale, useTranslations } from 'next-intl';
+import { useTranslations } from 'next-intl';
 import { useRef } from 'react';
 
 import Link from '#site/components/Link';
@@ -21,7 +21,6 @@ type WithSidebarProps = {
 const WithSidebar: FC<WithSidebarProps> = ({ navKeys, context, ...props }) => {
   const { getSideNavigation } = useSiteNavigation();
   const pathname = usePathname()!;
-  const locale = useLocale();
   const t = useTranslations();
   const { push } = useRouter();
   const { frontmatter } = useClientContext();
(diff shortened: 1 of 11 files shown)


apps/site/hooks/client/useScroll.ts
malav2110, line 1: Reintroducing the file, removed in https://github.com/nodejs/nodejs.org/pull/8516

packages/ui-components/src/Containers/Sidebar/index.tsx
malav2110, line 1: Major changes in the file are adding `forwardRef` and `ref`

apps/site/components/withSidebar.tsx
```diff
@@ -3,16 +3,18 @@
...
-import { useClientContext } from '#site/hooks/client';
-import { useSiteNavigation } from '#site/hooks/generic';
 import { useRouter } from '#site/navigation.mjs';
 
 import type { NavigationKeys } from '#site/types';
 import type { RichTranslationValues } from 'next-intl';
 import type { FC } from 'react';
 
+import { useClientContext, useNavigationState } from '../hooks/client';
+import { useSiteNavigation } from '../hooks/generic';
```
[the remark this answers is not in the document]
malav2110: addressed

apps/site/hooks/client/useNavigationState.ts
```diff
@@ -0,0 +1,42 @@
...
+        y: ref.current.scrollTop,
+      };
+    }
+  }, debounceTime);
+
+  useEffect(() => {
+    const element = ref.current;
+    if (element) {
+      if (navigationState[id] && navigationState[id].y !== element.scrollTop) {
+        element.scroll({ top: navigationState[id].y, behavior: 'instant' });
```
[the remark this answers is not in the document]
malav2110: addressed

apps/site/components/withSidebar.tsx
```diff
@@ -25,8 +27,14 @@ const WithSidebar: FC<WithSidebarProps> = ({ navKeys, context, ...props }) => {
...
+  const sidebarRef = useRef<HTMLElement>(null);
   const sideNavigation = getSideNavigation(navKeys, context);
 
+  const localePathname = pathname.replace(`/${locale}`, '');
+
+  // Preserve sidebar scroll position across navigations
+  useNavigationState('sidebar', sidebarRef);
+
   const mappedSidebarItems =
     // If there's only a single navigation key, use it's sub-items
```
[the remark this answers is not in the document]
malav2110: addressed

apps/site/hooks/client/useNavigationState.ts
```diff
@@ -0,0 +1,42 @@
...
+  const navigationState = useContext(NavigationStateContext);
+
+  const handleScroll = debounce(() => {
+    if (ref.current) {
+      navigationState[id] = {
+        x: ref.current.scrollLeft,
+        y: ref.current.scrollTop,
+      };
+    }
+  }, debounceTime);
```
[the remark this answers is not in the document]
malav2110: addressed

packages/ui-components/src/Containers/Sidebar/index.tsx
```diff
@@ -17,52 +19,46 @@ type SidebarProps = {
...
+            items={items}
+            pathname={pathname}
+            as={as}
+            className={styles.navigation}
+          />
+        ))}
+      </aside>
+    );
+  }
+);
```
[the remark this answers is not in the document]
malav2110: addressed

apps/site/components/withSidebar.tsx
```diff
@@ -3,9 +3,11 @@
 import Sidebar from '@node-core/ui-components/Containers/Sidebar';
 import { usePathname } from 'next/navigation';
 import { useLocale, useTranslations } from 'next-intl';
+import { useRef } from 'react';
 
 import Link from '#site/components/Link';
 import { useClientContext } from '#site/hooks/client';
+import { useNavigationState } from '#site/hooks/client';
```
avivkeller: This can be a single import

apps/site/components/withSidebar.tsx
```diff
@@ -25,10 +27,16 @@ const WithSidebar: FC<WithSidebarProps> = ({ navKeys, context, ...props }) => {
   const t = useTranslations();
   const { push } = useRouter();
   const { frontmatter } = useClientContext();
+  const sidebarRef = useRef<HTMLElement>(null);
   const sideNavigation = getSideNavigation(navKeys, context);
 
+  const localePathname = pathname.replace(`/${locale}`, '');
```
avivkeller: Don’t we have a `usePathname` from `#site/navigation.mjs` that strips this?

If not, I know this is how it was previously

packages/ui-components/src/Containers/Sidebar/index.tsx
```diff
@@ -17,52 +19,48 @@ type SidebarProps = {
...
+
+SideBar.displayName = 'SideBar';
```
avivkeller: Contrary to what copilot said, I’d rather not do this at the top level, and just keep the display name as Anonymous

apps/site/hooks/client/useScroll.ts
avivkeller, line 1: Can you add some code comments?

ovflowd, line 1: Scroll is one of the ways of doing this, the other one are element IDs (ish) and anchoring to the element ID of the navigation entry.

Anyhow, could you make another hook called `useScroll` that hanldes just scrolling/scroll listeners. And then this hook uses that hook, ensuring that each hook does only one thing.

apps/site/hooks/client/useNavigationState.ts
```diff
@@ -0,0 +1,55 @@
+'use client';
+
+import { useCallback, useContext, useEffect, useRef } from 'react';
+
+import { NavigationStateContext } from '#site/providers/navigationStateProvider';
+
+import type { RefObject } from 'react';
+
+const useNavigationState = <T extends HTMLElement>(
```
ovflowd: Name is a bit misleading, I'd probably call `useScrollToElement`

apps/site/hooks/client/useNavigationState.ts
```diff
@@ -0,0 +1,55 @@
...
+  id: string,
```
ovflowd: Do you need the ID if you have the ref or is the id used for something else?

apps/site/hooks/client/useNavigationState.ts
```diff
@@ -0,0 +1,55 @@
...
+  ref: RefObject<T | null>,
+  debounceTime = 300
+) => {
+  const navigationState = useContext(NavigationStateContext);
+  const timeoutRef = useRef<NodeJS.Timeout | null>(null);
+
+  const handleScroll = useCallback(() => {
```
ovflowd: Do you really need to make it memoized? This memoization is going to reocurr on every change of ref... I don't think you need a useCallback here.

apps/site/hooks/client/useNavigationState.ts
```diff
@@ -0,0 +1,55 @@
...
+        navigationState[id] = {
+          x: ref.current.scrollLeft,
+          y: ref.current.scrollTop,
+        };
+      }
+    }, debounceTime);
+  }, [id, ref, navigationState, debounceTime]);
+
+  useEffect(() => {
+    const element = ref.current;
```
ovflowd: Let's remove the redundant variable assignment. Let's use direct reference. I don't see why doing the assignment is necessary

apps/site/hooks/client/useNavigationState.ts
```diff
@@ -0,0 +1,55 @@
...
+      }
+
+      element.addEventListener('scroll', handleScroll, { passive: true });
+
+      return () => {
+        element.removeEventListener('scroll', handleScroll);
+        // Clear any pending debounced calls
+        if (timeoutRef.current) {
+          clearTimeout(timeoutRef.current);
+          timeoutRef.current = null;
```
ovflowd: You don't need to assign it to null, this is being executed on component unmount. Next time this hook executes the timeoutRef will have its initial value.

apps/site/hooks/client/useNavigationState.ts
```diff
@@ -0,0 +1,55 @@
...
+
+import type { RefObject } from 'react';
+
+const useNavigationState = <T extends HTMLElement>(
+  id: string,
+  ref: RefObject<T | null>,
+  debounceTime = 300
+) => {
+  const navigationState = useContext(NavigationStateContext);
+  const timeoutRef = useRef<NodeJS.Timeout | null>(null);
```
ovflowd:
```suggestion
  const timeoutRef = useRef<NodeJS.Timeout>();
```

No need to have duality of null. By default a `useRef` .current will be undefined, so no need for explicit null value.

apps/site/hooks/client/useScroll.ts
ovflowd: Agreed

packages/ui-components/src/Containers/Sidebar/index.tsx
```diff
@@ -1,8 +1,10 @@
+import { forwardRef } from 'react';
```
ovflowd, line 1: I'm not sure forwardRef is compartible with RSC. Wouldn't this make this component inplicitly a client component?

ovflowd: If so, you gotta ensure the `'use client';` at the top to keep things explicit.

ovflowd (commented):
I feel we're missing tests that verify the hook works as expected and scrolls to an element that was outside of the view.

apps/site/components/withSidebar.tsx
malav2110: addressed.

apps/site/components/withSidebar.tsx
malav2110: addressed.

packages/ui-components/src/Containers/Sidebar/index.tsx
malav2110: removed

packages/ui-components/src/Containers/Sidebar/index.tsx
malav2110: `forwardRef` itself is compatible with RSC. However, the component is being used by withSidebar.tsx which is a client component (has `'use client'`). So adding `use client`.

apps/site/hooks/client/useNavigationState.ts
malav2110: TS overloads for `useRef` get a bit noisy without an explicit initial value, so I’m switching this to `NodeJS.Timeout | undefined` and initializing with `undefined` instead of `null`.

apps/site/hooks/client/useNavigationState.ts
malav2110: We do need `id` as multiple components might use this hook with different elements and the scroll positions are stored in the shared context object keyed by `id`.

apps/site/hooks/client/useScrollToElement.ts
```diff
@@ -0,0 +1,49 @@
...
+
+    // Scroll only if the saved position differs from current
+    if (savedState && savedState.y !== element.scrollTop) {
+      element.scroll({ top: savedState.y, behavior: 'auto' });
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [id, ref]);
+
+  // Save scroll position on scroll
+  const handleScroll = useCallback(
```
ovflowd: nit: not sure if this needs to be a callback

apps/site/hooks/client/useScrollToElement.ts
```diff
@@ -0,0 +1,49 @@
...
+    }
+
+    // Restore scroll position if saved state exists
+    const savedState = navigationState[id];
+
+    // Scroll only if the saved position differs from current
+    if (savedState && savedState.y !== element.scrollTop) {
+      element.scroll({ top: savedState.y, behavior: 'auto' });
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
ovflowd, line 33: nit: add a comment explaining why only id and ref changes are enough.

apps/site/hooks/client/index.ts
```diff
@@ -1,3 +1,5 @@
 export { default as useDetectOS } from './useDetectOS';
 export { default as useMediaQuery } from './useMediaQuery';
 export { default as useClientContext } from './useClientContext';
+export { default as useScrollToElement } from './useScrollToElement';
```
ovflowd, line 4: You also gotta make stub exports on hooks/server that when called throw exceptions. (see examples on hooks/server folder)

ovflowd (approved):
Left a few nits, but SGTM! And thanks for the work and effort here 🙇

avivkeller (approved):
Can you `pnpm version patch` in `packages/ui-components`?

apps/site/hooks/client/useScrollToElement.ts
malav2110: addressed.

apps/site/hooks/client/useScrollToElement.ts
malav2110: addressed.

apps/site/hooks/client/index.ts
malav2110: addressed.

malav2110:
Addressed nitpicks and ran `pnpm version patch` command in `ui-packages` directory.  :)

packages/ui-components/src/Containers/Sidebar/index.tsx
avivkeller: No, let's remove `use client`, can't this component itself be RSC if `withSidebar` changes? IMO we should only `'use client'` if absolutely needed.

malav2110: addressed.

ovflowd: @avivkeller are you sure we should remove the use client directive? Id assume that Next.js is implicitly making this a client or at least mixed component, no?

apps/site/components/withSidebar.tsx
```diff
@@ -39,8 +42,9 @@ const WithSidebar: FC<WithSidebarProps> = ({ navKeys, context, ...props }) => {
 
   return (
     <Sidebar
+      ref={sidebarRef}
       groups={mappedSidebarItems}
-      pathname={pathname.replace(`/${locale}`, '')}
```
ovflowd, line 43: Wondering why this got changed? 👀

apps/site/hooks/client/useScrollToElement.ts
```diff
@@ -0,0 +1,48 @@
...
+const useScrollToElement = <T extends HTMLElement>(
+  id: string,
+  ref: RefObject<T | null>,
+  debounceTime = 300
+) => {
+  const navigationState = useContext(NavigationStateContext);
+
+  // Restore scroll position on mount
+  useEffect(() => {
+    const element = ref.current;
```
ovflowd: nit: no const assignment when using references

ovflowd (approved):
SGTM, but left final two nits!

apps/site/components/withSidebar.tsx
malav2110: ~It was put in while I was exploring the solution. Reverted back.~

Actually, Changed `next/navigation`'s `usePathname` to `next-intl` version from `navigation.mjs`. The `next-intl` version already returns pathnames without the locale prefix, so `.replace()` call is not needed.

apps/site/hooks/client/useScrollToElement.ts
malav2110: addressed.

packages/ui-components/src/Containers/Sidebar/index.tsx
malav2110: @avivkeller thoughts / suggestions ?

avivkeller: > @avivkeller are you sure we should remove the use client directive? Id assume that Next.js is implicitly making this a client or at least mixed component, no?

Yes, I'm sure.

If Next.js wants to make this a client component, it can, however, I don't want it to be _forced_ into a client component in a case where it doesn't have to be.

I'd argue that if the context can be inferred without the directive, the directive isn't needed.

ovflowd: > I'd argue that if the context can be inferred without the directive, the directive isn't needed.

I wonder if it actually gets treated as a client component.. Could you check that? Might be that forwardRef treats as RSC.

PR #8520, opened 2026-01-06
fix: semantic version in release blog posts

aduh95:
## Description

SImilar to https://github.com/nodejs/nodejs.org/pull/7998, our guidelines indicates we should be using Node.js x.y.z instead of Node.js vx.y.z.

Refs: https://github.com/nodejs/node/blob/668e52339261ec21c7388884620987914c833395/doc/README.md#L220-L222
Fixes: https://github.com/nodejs/node/issues/61231

### Check List


avivkeller (commented):
Can we leave the previous blog posts the same, and only start applying this to future posts?

avivkeller:
I'll update @node-core/remark-lint with this new requirement, as well

aduh95:
> Can we leave the previous blog posts the same, and only start applying this to future posts?

Why? We did update old posts in https://github.com/nodejs/nodejs.org/pull/7998, I don't think it broke anything

avivkeller:
That change was a bit different, IMO. This is referring to the actual version, and I think it's important to maintain consistency, however, I'm not going to force it, if you think this is best, go ahead

aduh95:
> This is referring to the actual version

I think there's a misunderstanding, the actual version does not change, e.g. v0.10.0 has always been Node.js 0.10.0, before or after this change. This PR is simply updating the blog posts to match our style guide, it is not a renaming nor a new requirement (very similar to https://github.com/nodejs/nodejs.org/pull/7998).

MattIPv4:
Has the tooling that generates these posts been updated?

aduh95:
> Has the tooling that generates these posts been updated?

Yes, as you can see in the diff (https://github.com/nodejs/nodejs.org/pull/8520/files#diff-873e2ee117570f0836a600113925cbeb0929ece48acb9df0076d06b54f17d0e4)

avivkeller (approved):

MikeMcC399:
This looks like it fixes part of the problem described in https://github.com/nodejs/node/issues/61231, so although I don't have privileges to approve PRs, I can at least write that I think this would be a good thing to merge!

RafaelGSS (approved):

PR #8522, opened 2026-01-07
fix: ensure last sidebar item is visible when scrolling

### LINKED ISSUE #8521: Cannot see last item in sidebar after scrolling down, need to scroll down the whole page
### URL:

### Browser Name:

Firefox

### Browser Version:

146.0.1

### Operating System:

NixOS 26.05.20251229.346dd96

Originally reported at https://github.com/nodejs/node/issues/52343#issuecomment-3716945992

**Steps to reproduce**
Scroll down in the sidebar

**Expected Results**
See the last item of the sidebar ("Collecting code coverage in Node.js")

**Actual Result**
Last item is not visible. 
Need to scroll down the whole page content to make it visible

**Video**

https://github.com/user-attachments/assets/38f12d29-23aa-4f1b-b57e-8b6add98a1eb

(in the video the last 2 items are not visible due to the additional security release banner)

SimoHypers:
## Description

Fixed the sidebar scrolling issue where the last item ("Collecting code coverage in Node.js") was not visible unless the entire page was scrolled down. 

The sidebar now uses sticky positioning with a proper max-height calculation that accounts for the navbar height, allowing users to scroll within the sidebar independently of the page content and see all items.

**Changes made:**
- Added `sm:sticky` positioning to keep sidebar in viewport
- Added `sm:top-16` to position below navbar
- Set `sm:max-h-[calc(100vh-6.5rem)]` to constrain height accounting for navbar
- Changed to `sm:overflow-y-auto` for vertical scrolling

## Validation

**Before:** Last sidebar items were cut off when scrolling within the sidebar. Users had to scroll the entire page content to reveal them.

**After:** All sidebar items are now visible when scrolling within the sidebar itself, independent of page scroll position.

**Testing performed:**
- ✅ Verified last item "Collecting code coverage in Node.js" is now visible when scrolling sidebar to bottom
- ✅ Tested on Firefox 146.0.1 (where issue was originally reported)
- ✅ Tested on Chrome
- ✅ Verified with security banner present (as mentioned in original issue)

**File changed:** `packages/ui-components/src/Containers/Sidebar/index.module.css`

## Related Issues

Fixes #8521

### Check List

### DIFF
--- a/packages/ui-components/src/Containers/Sidebar/index.module.css
+++ b/packages/ui-components/src/Containers/Sidebar/index.module.css
@@ -10,7 +10,10 @@
     bg-white
     px-4
     py-6
-    sm:overflow-auto
+    sm:sticky
+    sm:top-16
+    sm:max-h-[calc(100vh-6.5rem)]
+    sm:overflow-y-auto
     sm:border-r
     md:max-w-xs
     lg:px-6


ovflowd:
image

There's an interesting artifact where the scrollbar ends before the end of the container? Probaby the fix should slightly different?

packages/ui-components/src/Containers/Sidebar/index.module.css
```diff
@@ -10,7 +10,10 @@
     bg-white
     px-4
     py-6
-    sm:overflow-auto
+    sm:sticky
+    sm:top-16
+    sm:max-h-[calc(100vh-6.5rem)]
```
danielpza, line 15: nitpick: you might want to leave a comment explaining what the 6.5rem value means (eg header height)

danielpza:
> The sidebar now uses sticky positioning with a proper max-height calculation that accounts for the navbar height, allowing users to scroll within the sidebar independently of the page content and see all items.

Might fall outside of this PR, but notice in the original issue there's also the occasional security release banner which makes this extra top bar height variable.

ovflowd:
> > The sidebar now uses sticky positioning with a proper max-height calculation that accounts for the navbar height, allowing users to scroll within the sidebar independently of the page content and see all items.
> 
> Might fall outside of this PR, but notice in the original issue there's also the occasional security release banner which makes this extra top bar height variable.

Yeah, I don't think adding a fixed 6.5rem is a valid solution; that is hacky, and I'm pretty sure this can be fixed in a different manner.

avivkeller:
Bump!

malav2110:
Thought: Instead of hardcoding 6.5rem, could we derive the offset from an existing CSS variable - `var(--header-height)`, so it adapts when the security banner or header height changes? Happy to create a small follow-up PR if this makes sense.

ovflowd:
> Thought: Instead of hardcoding 6.5rem, could we derive the offset from an existing CSS variable - `var(--header-height)`, so it adapts when the security banner or header height changes? Happy to create a small follow-up PR if this makes sense.

@malav2110 I believe the author has abandoned the PR, feel free to work on this if you have capacity :)

malav2110:
Got it, thanks @ovflowd :)

PR #8532, opened 2026-01-09
feat(ci): pnpm updater

avivkeller:
Add a workflow to update pnpm once a month via the `pnpm self-update` command.

### DIFF
--- a/.github/workflows/pnpm-updater.yml
+++ b/.github/workflows/pnpm-updater.yml
@@ -1,0 +1,60 @@
+# Security Notes
+# Only selected Actions are allowed within this repository. Please refer to (https://github.com/nodejs/nodejs.org/settings/actions)
+# for the full list of available actions. If you want to add a new one, please reach out a maintainer with Admin permissions.
+# REVIEWERS, please always double-check security practices before merging a PR that contains Workflow changes!!
+# AUTHORS, please only use actions with explicit SHA references, and avoid using `@master` or `@main` references or `@version` tags.
+
+name: PNPM updater
+
+on:
+  schedule:
+    - cron: "0 0 1 * *"
+
+permissions:
+  contents: write
+  pull-requests: write
+
+jobs:
+  update-pnpm:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Harden Runner
+        uses: step-security/harden-runner@20cf305ff2072d973412fa9b1e3a4f227bda3c76 # v2.14.0
+        with:
+          egress-policy: audit
+
+      - name: Git Checkout
+        uses: actions/checkout@8e8c483db84b4bee98b60c0593521ed34d9990e8 # v6.0.1
+
+      - name: Setup pnpm
+        uses: pnpm/action-setup@a7487c7e89a18df4991f7f222e4898a00d66ddda # v4.1.0
+
+      - name: Update pnpm and package.json
+        id: update-pnpm
+        run: |
+          echo "old=$(pnpm --version)" >> $GITHUB_OUTPUT
+          pnpm self-update
+          echo "new=$(pnpm --version)" >> $GITHUB_OUTPUT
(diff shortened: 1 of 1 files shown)


.github/workflows/pnpm-updater.yml
```diff
@@ -0,0 +1,60 @@
...
+      - name: Setup pnpm
+        uses: pnpm/action-setup@a7487c7e89a18df4991f7f222e4898a00d66ddda # v4.1.0
+
+      - name: Update pnpm and package.json
+        id: update-pnpm
+        run: |
+          echo "old=$(pnpm --version)" >> $GITHUB_OUTPUT
+          pnpm self-update
+          echo "new=$(pnpm --version)" >> $GITHUB_OUTPUT
+          cat <<< $(jq '.devEngines.packageManager.version = (.packageManager | split("@")[1])' package.json) > package.json
```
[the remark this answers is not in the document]
avivkeller: `pnpm self-update` updates the `packageManager` field, and the `jq` command copies that to the `devEngines` field

.github/workflows/pnpm-updater.yml
```diff
@@ -0,0 +1,61 @@
...
+          update-pull-request-title-and-body: true
+          branch: chore/update-pnpm
+          body: |
+            Updates pnpm from ${{ steps.update-pnpm.outputs.old }} to version ${{ steps.update-pnpm.outputs.new }}
+
+            cc @nodejs/web-infra
+
+            Check this workflow's logs at ${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}.
+          commit-message: 'chore: update pnpm to ${{ steps.update-pnpm.outputs.new }}'
+          labels: fast-track
```
MattIPv4: Why are we marking the PRs as fast-track? I imagine this wants to go through the same review process as an other dependency update PR?

.github/workflows/pnpm-updater.yml
```diff
@@ -0,0 +1,61 @@
...
+          title: 'chore: update pnpm to ${{ steps.update-pnpm.outputs.new }}'
```
MattIPv4: Can we match the format Dependabot uses? `meta: bump <thing> from <version> to <version>`?

.github/workflows/pnpm-updater.yml
avivkeller: I copied the step from `create-release-post`, I'll undo that

MattIPv4 (approved):

flakey5 (approved):

.github/workflows/pnpm-updater.yml
```diff
@@ -0,0 +1,60 @@
...
+          pnpm self-update
+          echo "new=$(pnpm --version)" >> $GITHUB_OUTPUT
+          cat <<< $(jq '.devEngines.packageManager.version = (.packageManager | split("@")[1])' package.json) > package.json
+
+      - name: Open pull request
+        if: steps.update-pnpm.outputs.old != steps.update-pnpm.outputs.new
+        uses: gr2m/create-or-update-pull-request-action@b65137ca591da0b9f43bad7b24df13050ea45d1b # v1.10.1
+        # Creates a PR or update the Action's existing PR, or
+        # no-op if the base branch is already up-to-date.
+        env:
```
ovflowd, line 47: IMO, this workflow should instead of be cron be on workflow_dispatch and not use a secret but the permissions from the user that triggered it, but also fine as it is.

ovflowd (approved):

.github/workflows/pnpm-updater.yml
avivkeller: > but the permissions from the user that triggered it, but also fine as it is.

GitHub Actions doesn't work that way. If it did, it would pose a huge security risk.

There's no way for an action to act on behalf of a user without having their API key.

Issue #8549, opened 2026-01-14
Enhance CodeTabs tab naming


canerakdas:
The tab naming in our current `CodeTabs` component isn’t very descriptive and doesn’t clearly indicate that the displayed code is an example. This can be confusing for screen reader users and for people who are new to the ecosystem

Renaming them to something like `Example CJS`, `Example MJS`, and displaying the type separately could be more intuitive.

* Since this improvement is implemented in the `ui-components` package and indirectly affects [`doc-kit`](https://github.com/nodejs/doc-kit) as well, it should also cover scenarios like `ESM (2)`, `CJS (2)`, `JS` etc.

The design can be reviewed in [Figma](https://www.figma.com/design/a10cjjw3MzvRQMPT9FP3xz/Node.js?node-id=6289-1939&t=7tWL9fjIKRNuT6im-0)  (apologies in advance for my terrible Figma skills 😅).

<img width="744" height="274" alt="Image" src="https://github.com/user-attachments/assets/08475b5f-3aae-4f5d-b055-2ddaf28c1006" />

naveenkarmegam:
This makes a lot of sense — the current tab labels aren’t very descriptive, especially for screen readers and for developers who are new to the Node.js ecosystem. Shall i proceed ?

avivkeller:
I'm afraid I don't understand this, why should we give every code box the same name (`Example`), aren't they assumed to be examples already?

ovflowd:
> I'm afraid I don't understand this, why should we give every code box the same name (`Example`), aren't they assumed to be examples already?

Not every codebox. It allows us to add actual titles to the CodeBoxes and keep the "language tag". Right now in most we're just saying ESM/CJS but that isn't much clear. Imagine if you have a CodeBox with 4 tabs CJS, MJS, CJS, MJS... It becomes confusing what it is. So naming it as Example 1 (CJS), Example 1 (MJS), Example 2 (CJS), Example 2 (MJS) makes it easier to understand.

avivkeller:
Ahhh, I see. So, to clarify, we are putting the language in a secondary label _if_ a custom `displayTitle` is specified

ovflowd:
> Ahhh, I see. So, to clarify, we are putting the language in a secondary label _if_ a custom `displayTitle` is specified

I guess? It really depends on implementation, hence why I'm unsure if #8552 is actually a proper solution implementation-wise.

PR #8552, opened 2026-01-15
feat: Improve CodeTabs tab naming 

naveenkarmegam:
## Description

This PR updates the secondary label styling of the tab components to improve visual clarity and consistency. It also updates the Storybook examples to reflect the new design for easier reference and validation.
Relevant screenshots have been attached for comparison and review

#### Light mode:
image

#### Dark mode :
image

## Related Issues
Related to #8549 

### Check List

### DIFF
--- a/packages/ui-components/src/Common/CodeTabs/index.stories.tsx
+++ b/packages/ui-components/src/Common/CodeTabs/index.stories.tsx
@@ -67,8 +67,8 @@ export default {
     children: <TabsContent />,
     defaultValue: 'mjs',
     tabs: [
-      { key: 'mjs', label: 'Example', secondaryLabel: 'MJS' },
-      { key: 'cjs', label: 'Example', secondaryLabel: 'CJS' },
+      { key: 'mjs', label: 'Example', extension: 'MJS' },
+      { key: 'cjs', label: 'Example', extension: 'CJS' },
     ],
   },
 } as Meta;
--- a/packages/ui-components/src/Common/CodeTabs/index.stories.tsx
+++ b/packages/ui-components/src/Common/CodeTabs/index.stories.tsx
@@ -61,14 +61,23 @@ const TabsContent: FC = () => (
 
 export const Default: Story = {};
 
+export const WithExtension: Story = {
+  args: {
+    tabs: [
+      { key: 'mjs', label: 'Example', extension: 'MJS' },
+      { key: 'cjs', label: 'Example', extension: 'CJS' },
+    ],
+  },
+};
+
 export default {
   component: CodeTabs,
   args: {
     children: <TabsContent />,
     defaultValue: 'mjs',
     tabs: [
-      { key: 'mjs', label: 'Example', extension: 'MJS' },
-      { key: 'cjs', label: 'Example', extension: 'CJS' },
+      { key: 'mjs', label: 'Example' },
+      { key: 'cjs', label: 'Example' },
     ],
   },
 } as Meta;
--- a/packages/ui-components/src/Common/CodeTabs/index.stories.tsx
+++ b/packages/ui-components/src/Common/CodeTabs/index.stories.tsx
@@ -67,8 +67,8 @@ export default {
     children: <TabsContent />,
     defaultValue: 'mjs',
     tabs: [
-      { key: 'mjs', label: 'MJS' },
-      { key: 'cjs', label: 'CJS' },
+      { key: 'mjs', label: 'Example', secondaryLabel: 'MJS' },
+      { key: 'cjs', label: 'Example', secondaryLabel: 'CJS' },
     ],
   },
 } as Meta;
--- a/packages/ui-components/src/Common/CodeTabs/index.stories.tsx
+++ b/packages/ui-components/src/Common/CodeTabs/index.stories.tsx
@@ -76,8 +76,8 @@ export default {
     children: <TabsContent />,
     defaultValue: 'mjs',
     tabs: [
-      { key: 'mjs', label: 'Example' },
-      { key: 'cjs', label: 'Example' },
+      { key: 'mjs', label: 'MJS' },
+      { key: 'cjs', label: 'CJS' },
     ],
   },
 } as Meta;
--- a/packages/ui-components/src/Common/Tabs/index.module.css
+++ b/packages/ui-components/src/Common/Tabs/index.module.css
@@ -23,8 +23,15 @@
         dark:text-neutral-200;
 
       .tabSecondaryLabel {
-        @apply pl-1
+        @apply ml-1
+          rounded
+          border
+          border-neutral-500
+          px-1
+          py-0.5
+          text-xs
           text-neutral-500
+          dark:border-neutral-800
           dark:text-neutral-800;
       }
 
@@ -35,7 +42,9 @@
           dark:text-green-400;
 
         .tabSecondaryLabel {
-          @apply text-green-800
+          @apply border-green-800
+            text-green-800
+            dark:border-green-600
             dark:text-green-600;
         }
       }
--- a/packages/ui-components/src/Common/Tabs/index.module.css
+++ b/packages/ui-components/src/Common/Tabs/index.module.css
@@ -27,11 +27,11 @@
           rounded-xs
           border
           border-neutral-200
-          p-0.5
-          text-[10px]
+          px-1
+          py-0.5
+          text-xs
           font-normal
-          text-neutral-800
-          dark:text-neutral-200;
+          text-neutral-200;
       }
 
       .tabSecondaryLabel {
@@ -47,10 +47,8 @@
           dark:text-green-400;
 
         .tabExtension {
-          @apply border-green-600
-            text-green-600
-            dark:border-green-400
-            dark:text-green-400;
+          @apply border-green-400
+            text-green-400;
         }
 
         .tabSecondaryLabel {
--- a/packages/ui-components/src/Common/Tabs/index.module.css
+++ b/packages/ui-components/src/Common/Tabs/index.module.css
@@ -22,16 +22,21 @@
         text-neutral-800
         dark:text-neutral-200;
 
-      .tabSecondaryLabel {
+      .tabExtension {
         @apply ml-1
-          rounded
+          rounded-xs
           border
-          border-neutral-500
-          px-1
-          py-0.5
-          text-xs
+          border-neutral-200
+          p-0.5
+          text-[10px]
+          font-normal
+          text-neutral-800
+          dark:text-neutral-200;
+      }
+
+      .tabSecondaryLabel {
+        @apply pl-1
           text-neutral-500
-          dark:border-neutral-800
           dark:text-neutral-800;
       }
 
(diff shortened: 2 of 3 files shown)


ovflowd:
**nit:** the secondary label border style (padding) differs from Figma.

ovflowd:
**nit:** my understanding is that this change won't actually affect existing code snippet boxes, for the website, (cc @nodejs/nodejs-website) do we want to update existing code snippet labels?

For doc-kit, cc @avivkeller I believe we can simply call most of the in-box code-snippet as "Example", but I worry it would _require_ us to actually change in-source code-snippets (the raw md files)... I wonder if this is the right way of achieving what @canerakdas proposed?

packages/ui-components/src/Common/CodeTabs/index.stories.tsx
```diff
@@ -67,8 +67,8 @@ export default {
     children: <TabsContent />,
     defaultValue: 'mjs',
     tabs: [
-      { key: 'mjs', label: 'MJS' },
-      { key: 'cjs', label: 'CJS' },
+      { key: 'mjs', label: 'Example', secondaryLabel: 'MJS' },
+      { key: 'cjs', label: 'Example', secondaryLabel: 'CJS' },
```
avivkeller: You should have tabs with and without secondary labels

avivkeller:
> For doc-kit, cc @avivkeller I believe we can simply call most of the in-box code-snippet as "Example", but I worry it would require us to actually change in-source code-snippets (the raw md files)... I wonder if this is the right way of achieving what @canerakdas proposed?

We'd probably have update the Shiki plugin to change the way it works, rather than changing the raw markdown files. This would also apply to the website.

canerakdas:
> For doc-kit, cc @avivkeller I believe we can simply call most of the in-box code-snippet as "Example", but I worry it would _require_ us to actually change in-source code-snippets (the raw md files)... I wonder if this is the right way of achieving what @canerakdas proposed?

I’m wondering whether we could handle this with only CSS by adding content: "Example" via `::before` when there are more than two tabs, without making many changes in `doc-kit`. Applying it unconditionally to all tabs is also an option, but I’m not sure how it would look in general usage 🤔

naveenkarmegam:
@ovflowd @avivkeller 

Thanks for calling this out — this change is mainly meant to showcase the visual output and behavior as a reference and to gather feedback before we lock anything in.
That’s why, for now, I’ve only updated the CSS and Storybook to demonstrate the idea.
Once the direction is finalized, we can decide the best approach and proceed with broader updates if needed.
Do you have any idea ?

ovflowd:
> > For doc-kit, cc @avivkeller I believe we can simply call most of the in-box code-snippet as "Example", but I worry it would _require_ us to actually change in-source code-snippets (the raw md files)... I wonder if this is the right way of achieving what @canerakdas proposed?
> 
> I’m wondering whether we could handle this with only CSS by adding content: "Example" via `::before` when there are more than two tabs, without making many changes in `doc-kit`. Applying it unconditionally to all tabs is also an option, but I’m not sure how it would look in general usage 🤔

Not sure I'm a fan of CSS-only solution... That'd force the "Example" tag on all codeboxes

naveenkarmegam:
image

In cases where we only have an MJS example today, do we plan to introduce a CJS example later?

avivkeller:
> In cases where we only have an MJS example today, do we plan to introduce a CJS example later?

No. Some code snippets will be only MJS/only CJS.

packages/ui-components/src/Common/Tabs/index.module.css
```diff
@@ -23,8 +23,15 @@
         dark:text-neutral-200;
 
       .tabSecondaryLabel {
-        @apply pl-1
+        @apply ml-1
+          rounded
+          border
+          border-neutral-500
```
canerakdas: The border and text colors don't match Figma and aren't sufficient for the minimum contrast ratio;

image

packages/ui-components/src/Common/Tabs/index.module.css
```diff
@@ -23,8 +23,15 @@
         dark:text-neutral-200;
 
       .tabSecondaryLabel {
-        @apply pl-1
```
canerakdas, line 26: The secondary label was added for the search box module we used [previously](https://github.com/nodejs/nodejs.org/pull/6537) and is no longer a feature in use. Since we’ll only be using it in the code box, adding just the dark mode variant will be better

packages/ui-components/src/Common/CodeTabs/index.stories.tsx
canerakdas: `secondaryLabel` doesn’t feel like the right naming. Since the `Tab` component is used in many places, it could lead to misuse. IMO using the name `extension` directly would be clearer and more self-explanatory

ovflowd:
cc @naveenkarmegam bump?

packages/ui-components/src/Common/CodeTabs/index.stories.tsx
naveenkarmegam: Agreed. ```secondaryLabel``` is ambiguous for a shared component. ```extension``` would make the intent clearer. I’ll update it.

packages/ui-components/src/Common/Tabs/index.module.css
naveenkarmegam: Let me check and update it.

naveenkarmegam: can you please mention the light mode color for ```extension``` ?  i think , i can see only the dark mode color

canerakdas: The light version of the `CodeTabs` component isn’t used / doesn’t exist, so you can use the colors defined in Figma as the default

naveenkarmegam:
@canerakdas i have updated few things please review and let me know.

packages/ui-components/src/Common/Tabs/index.module.css
```diff
@@ -22,6 +22,18 @@
...
+      .tabExtension {
+        @apply ml-1
+          rounded-xs
+          border
+          border-neutral-200
+          p-0.5
+          text-[10px]
+          font-normal
+          text-neutral-800
+          dark:text-neutral-200;
```
canerakdas: The `CodeTabs` component only has a dark theme;

```suggestion
          text-neutral-200;
```

packages/ui-components/src/Common/Tabs/index.module.css
```diff
@@ -22,6 +22,18 @@
         text-neutral-800
         dark:text-neutral-200;
 
+      .tabExtension {
+        @apply ml-1
+          rounded-xs
+          border
+          border-neutral-200
+          p-0.5
+          text-[10px]
```
canerakdas: This one my bad, lets update with the 12px one 😅 

```suggestion
          text-xs
```

canerakdas: Also font should be `font-ibm-plex-mono`

packages/ui-components/src/Common/Tabs/index.module.css
```diff
@@ -34,6 +46,13 @@
           dark:border-b-green-400
           dark:text-green-400;
 
+        .tabExtension {
+          @apply border-green-600
+            text-green-600
+            dark:border-green-400
+            dark:text-green-400;
```
canerakdas: The `CodeTabs` component only has a dark theme;

```suggestion
          @apply border-green-400
            text-green-400;
```

packages/ui-components/src/Common/Tabs/index.tsx
```diff
@@ -37,6 +38,9 @@ const Tabs: FC<PropsWithChildren<TabsProps>> = ({
           className={classNames(styles.tabsTrigger, triggerClassName)}
         >
           {tab.label}
+          {tab.extension ? (
+            <span className={styles.tabExtension}>{tab.extension}</span>
+          ) : null}
```
canerakdas:
```suggestion
          {tab.extension && (
            <span className={styles.tabExtension}>{tab.extension}</span>
          )}
```

packages/ui-components/src/Common/Tabs/index.module.css
```diff
@@ -22,6 +22,18 @@
         text-neutral-800
         dark:text-neutral-200;
 
+      .tabExtension {
+        @apply ml-1
+          rounded-xs
+          border
+          border-neutral-200
+          p-0.5
```
canerakdas: Adjustment for the following font size changes;

```suggestion
          px-1
          py-0.5
```

canerakdas (commented):
I’ve added my final comments, thanks for the effort you’ve put into this 🚀 🐢

packages/ui-components/src/Common/Tabs/index.module.css
naveenkarmegam: Got it, Thanks

naveenkarmegam:
@canerakdas Thanks for suggesting me. I have done

packages/ui-components/src/Common/Tabs/index.module.css
```diff
@@ -22,6 +22,18 @@
...
+          px-1
+          py-0.5
+          text-xs
+          font-normal
+          text-neutral-200;
+      }
+
       .tabSecondaryLabel {
         @apply pl-1
           text-neutral-500
```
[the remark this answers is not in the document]
ovflowd: @naveenkarmegam is this relevant? Justt rying to see how useful the output of this AI tool is?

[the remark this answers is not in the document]
naveenkarmegam: @ovflowd 

but in this case it’s not applicable.

```.tabExtension``` is currently only used by ```CodeTabs```, which is rendered exclusively in dark mode. Because of that, ```text-neutral-200``` is intentional and matches the existing design tokens for dark backgrounds.

The ```Tabs``` component isn’t used on light backgrounds with extensions today, so adding a light-mode override here would be premature. If we later introduce a light-mode variant or reuse this style elsewhere, we can revisit and adjust the colors accordingly.

packages/ui-components/src/Common/CodeTabs/index.stories.tsx
canerakdas: This issue doesn’t seem to have been addressed yet @naveenkarmegam  As Aviv mentioned, it would be great to see both the extension and non-extension versions in Storybook 👀

naveenkarmegam: Added a Storybook variant for ```CodeTabs``` with and without extension.

naveenkarmegam: Without extension (default)
image



With extension
image

canerakdas (approved):
LGTM

naveenkarmegam:
@ovflowd @avivkeller have any suggestion ?

araujogui (approved):
SGTM

naveenkarmegam:
anyone please give me chromatic access ?

canerakdas:
> anyone please give me chromatic access ?

Why do you need this access permission? 🤔

AugustinMauroy (approved):

naveenkarmegam:
@canerakdas I just want to see the deployed storybook component.

ovflowd:
> @canerakdas I just want to see the deployed storybook component.

Let me change the permissions so anyone can see the deployed Storybook. BTW, you can also run Storybook locally.

naveenkarmegam:
@ovflowd Thanks!. I’ll take a look once the permissions are updated.

aymen94 (approved):

PR #8561, opened 2026-01-25
chore: UX improvements

canerakdas:
## Description

With this PR, aiming to make some small UX improvements that I noticed while using the new `doc-kit` (some of them might be nitpicks 🙈 )

### AlerBox
image

Links added inside the `AlertBox` can cause the content to overflow outside of the `AlertBox`

### Badge
image

The Badge component we use in the Sidebar isn’t circular, and its width changes depending on the text inside. I added a prop for single-character badges that makes the text monospace (also did on `DataTag`) and fixes the width and height, ensuring all badges are consistent with each other

### TableOfContents

image

image

Because there’s no line break, some texts can overflow. In addition, when a line break is applied to lines with indents, ignores the indent.

And replaced the marker icon since it was causing issues on iOS and didn’t align with the design system

## Validation

The changes made in Storybook/Chromatic should appear without any issues

Example screenshots;
image

image

image

### Check List

### DIFF
--- a/packages/ui-components/src/Common/AlertBox/index.module.css
+++ b/packages/ui-components/src/Common/AlertBox/index.module.css
@@ -50,6 +50,10 @@
       size-5;
   }
 
+  .content {
+    @apply wrap-anywhere;
+  }
+
   &.info {
     @apply bg-info-600;
 
--- a/packages/ui-components/src/Common/Badge/index.tsx
+++ b/packages/ui-components/src/Common/Badge/index.tsx
@@ -10,17 +10,25 @@ type BadgeSize = 'small' | 'medium';
 type BadgeProps = HTMLAttributes<HTMLSpanElement> & {
   size?: BadgeSize;
   kind?: BadgeKind;
+  circular?: boolean;
 };
 
 const Badge: FC<PropsWithChildren<BadgeProps>> = ({
   kind = 'default',
   size = 'medium',
+  circular = false,
   className,
   children,
   ...props
 }) => (
   <span
-    className={classNames(styles.badge, styles[kind], styles[size], className)}
+    className={classNames(
+      styles.badge,
+      styles[kind],
+      styles[size],
+      { [styles.circular]: circular },
+      className
+    )}
     {...props}
   >
     {children}
--- a/packages/ui-components/src/Common/TableOfContents/index.module.css
+++ b/packages/ui-components/src/Common/TableOfContents/index.module.css
@@ -8,19 +8,39 @@
     lg:hidden
     dark:bg-neutral-900;
 
+  &[open] {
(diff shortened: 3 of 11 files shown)


packages/ui-components/src/Common/TableOfContents/index.module.css
```diff
@@ -8,19 +8,39 @@
...
+      @apply size-5;
+    }
   }
 
   .list {
     @apply space-y-1
       px-4
-      pb-2;
+      pb-2
+      wrap-anywhere;
```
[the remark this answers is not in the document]
canerakdas: https://tailwindcss.com/docs/overflow-wrap#wrapping-anywhere

packages/ui-components/src/Common/AlertBox/index.module.css
```diff
@@ -50,6 +50,10 @@
       size-5;
   }
 
+  .content {
+    @apply wrap-anywhere;
```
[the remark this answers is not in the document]
canerakdas: https://tailwindcss.com/docs/overflow-wrap#wrapping-anywhere

avivkeller (approved):

packages/ui-components/src/Common/Badge/index.tsx
```diff
@@ -10,17 +10,25 @@ type BadgeSize = 'small' | 'medium';
 type BadgeProps = HTMLAttributes<HTMLSpanElement> & {
   size?: BadgeSize;
   kind?: BadgeKind;
+  circular?: boolean;
 };
 
 const Badge: FC<PropsWithChildren<BadgeProps>> = ({
   kind = 'default',
   size = 'medium',
+  circular = false,
```
ovflowd, line 19: instead of this, could we enforce "circular" if children.length === 1?

ovflowd (approved):
SGTM

ovflowd:
I'm fast-tracking this as bug fixes, cc @nodejs/nodejs-website

packages/ui-components/src/Common/Badge/index.tsx
ovflowd: cc @canerakdas feel free to address this on follow-up PR, I believe the extra prop is unnecessary

canerakdas: I thought the same, but since `children` is a `ReactNode`, converting it to a string just to check its length doesn’t feel quite right, IDK 🤔

ovflowd: I mean... ReactNode has string as one of its subtypes. YHou could just check:

```
typeof children === 'string' && children.length === 1
```

PR #8563, opened 2026-01-26
fix(ui): Sidebar scroll not showing last element

### LINKED ISSUE #8521: Cannot see last item in sidebar after scrolling down, need to scroll down the whole page

malav2110:
## Description

Fixed the Sidebar scrolling issue where last item was not visible

## Validation

Before:

https://github.com/user-attachments/assets/7b4f11f9-b065-4e40-8ea7-6e420c2bc315

After:

https://github.com/user-attachments/assets/87ad40d1-e3fd-47a8-b143-ffec9609617b

## Related Issues

Fixes: https://github.com/nodejs/nodejs.org/issues/8521 
Closes: https://github.com/nodejs/nodejs.org/pull/8522

### Check List


ovflowd:
@malav2110 there's an odd effect here:

image

[Screencast_20260126_144928.webm](https://github.com/user-attachments/assets/7be93138-380e-4c5d-b672-4cf02b0267eb)

There's a weird effect where the sidebar is not full width.

This is how's supposed to look:

image

This is how it looks now in certain scroll positions:

image

malav2110:
@ovflowd Thanks for the video. I want to make sure I’m looking at the right things.

1. In the clip, are you referring to the padding at the bottom of the sidebar, or the horizontal bar that appears near the bottom? If it’s the latter, that section is the LTS / Current / social links block, which I understand to be separate from the sidebar items — but happy to confirm if I’m mistaken.

2. For the sidebar not staying full width, I’m not entirely sure which visual change you’re pointing to in the screenshots. A quick pointer to the specific area or box highlight would really help me focus on the right fix.

ovflowd:
> In the clip, are you referring to the padding at the bottom of the sidebar, or the horizontal bar that appears near the bottom? If it’s the latter, that section is the LTS / Current / social links block, which I understand to be separate from the sidebar items — but happy to confirm if I’m mistaken.

Not really, what I'm referring is the Sidebar is not full width, the Sidebar should go all the way to the bottom bar (not the footer), but the bar that has the breadcrumbs.

ovflowd:
image

This is what Im referring to, the Sidebar should be occupying this area

malav2110:
Ah, got it. I see the gap you’re referring to now. I’ll fix this.

packages/ui-components/src/Containers/Article/index.module.css
```diff
@@ -18,8 +18,8 @@
   > *:nth-child(1) {
     @apply [grid-area:sidebar]
       lg:sticky
-      lg:top-0
-      lg:h-[100vh]
+      lg:top-[var(--header-height,0)]
+      lg:h-[calc(100vh-var(--header-height,0))]
```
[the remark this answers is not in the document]
malav2110: False positive - As per project's existing pattern and https://github.com/tailwindlabs/tailwindcss/discussions/12147 .

ovflowd:
@malav2110 with the current code now the issue is at the beginning of the sidebar ;)

image

malav2110:
Sharp eyes  :) . I’ll address the top offset issue and make sure all scroll states are covered before the next push.

packages/ui-components/src/Containers/Sidebar/index.module.css
```diff
@@ -12,6 +12,7 @@
     py-6
     sm:overflow-auto
     sm:border-r
+    sm:pb-16
```
ovflowd: Padding could probably be a bit smaller.

malav2110: With `pb-14`, the last item on the list is barely visible thus added `pb-16`.

Screenshot 2026-01-28 at 3 50 34 PM

ovflowd: Is that really the case? This is how much padding I see 

image

ovflowd: You're probably referring to the padding when the footer isn't visible.

image

So this is probably not the place you want to change the padding.

ovflowd: For example, if you see the beginning of the sidebar has the same margins/paddings no matter if the header is visible or not

image

image

But the same isn't truthy for the end of the sidebar:

image

image

So this means this place is probably not the right one to add padding. Or it must be a mix of two different places.

malav2110: Fair point, thanks for the context. I’ll investigate and see if a different approach makes more sense. :)

malav2110: Hey @ovflowd , I did some digging and I think I understand what’s causing the inconsistency now.

The sidebar is `position: sticky` with `height: 100vh` and `top: 0`, but the navbar (and the security banner when present) scroll away. That makes the effective offset dynamic, which is why static values behave differently depending on scroll state.

I tried a few CSS-only approaches, but they all come with trade-offs. They either fix the “navbar visible” case or introduce extra space and create a gap at the top or bottom once it scrolls away, as you noticed.

From what I can see, the options are roughly:
- keep a static CSS-based solution and accept the trade-off
- make the navbar sticky so `calc(100vh - header-height)` becomes reliable
- or introduce JS to react to navbar visibility

Before I push anything else, wanted to check what direction you’d prefer here. Happy to explore other ideas if you think there’s a better approach 🙂

ovflowd: That's weird. the Metabar is also stick and doesn't have said issue, so I'm pretty sure the same can be accomplished to the Sidebar :thinking:

malav2110: I did review the MetaBar to compare.

It is also `position: sticky` with `top: 0`, but the key difference is that it does not have a height restriction or internal scrolling. Its content is short enough to fit naturally in the viewport.

The SideBar is different. On Learn page, it has many more items, so it needs `max-height` + `overflow-y-auto` to scroll internally.

That's where the issue comes from:
`100vh` uses the full viewport height, but when the NavBar (+ security banner) is visible, it takes up space at top. The Sidebar still assumes the full height, so the last item/s ends up getting clipped.

I tested both cases:
- Without the height constraint -> everything is visible, but user do not see the internal scrollbar and have to scroll the whole page.
- With the constraint -> internal scrolling works, last item gets cutoff when the NavBar is present.

Because the MetaBar rarely overflows, it avoids this edge case entirely.

ovflowd: > It is also position: sticky with top: 0, but the key difference is that it does not have a height restriction or internal scrolling. Its content is short enough to fit naturally in the viewport.

Unfortunately, you're wrong. Example: https://nodejs-api-docs-tooling.vercel.app/child_process.html / https://nodejs-api-docs-tooling.vercel.app/globals.html

ovflowd: (I would recommend to avoid making assumptions, if I tell you it is possible or something the contrary, then I'm pretty sure it is possible :stuck_out_tongue:)

ovflowd: (I just want to avoid giving you a solution plated, so you can try it yourself :eyes:)

malav2110: Got it. Thanks for the examples and for calling that out. That helps reset my assumptions.

Appreciate you pushing me to think it through rather than handing me the solution 🙂

ovflowd: Absolutely! Although, I know I used a bit of a strong tone, apologies for that. Let me know if you get stuck and need support tho :bow:

malav2110: Thanks, appreciate that 🙂
I pushed a fix with reduced padding to address the issue that started the thread, and also resolved the merge conflict. Curious to hear your thoughts.

ovflowd:
Hey @malav2110 I ended up fixing this in a similar way on https://github.com/nodejs/nodejs.org/pull/8618, but please feel free to rebase this PR and improve on top of that with suggestions/improvements on my initial solution :)

malav2110:
Sounds good. I'll rebase and iterate on top of it. :)

malav2110:
Quick update -  I’ve been a bit tied up with work the past couple of weeks, but I haven’t abandoned this. I’ll rebase and push an update in the next few days.

ovflowd:
> Quick update - I’ve been a bit tied up with work the past couple of weeks, but I haven’t abandoned this. I’ll rebase and push an update in the next few days.

Appreciate you!

malav2110:
Seems like I have messed up with the rebase process. Closing this and starting fresh.

PR #8596, opened 2026-02-05
feat(ui): remove meta bar

avivkeller:
This PR removes the `MetaBar` in order to give the main content **more space** on the website.

While collecting feedback for `doc-kit`, many people noted that the main content is simply too narrow.

This, obviously, isn't the only solution to that problem, and it's worth noting that many sites use a "three column system", so this may go against the norm.

Alternatively, we could update the paddings to give the next column much more space.

---

After:
image

image


ovflowd (changes requested):
This is huge change on the overall design of the webiste... That I don't think should simply drop in like a PR.

I'm unfortunately also not a fan at all of these design choices, sorry. Not sure what you're trying to solve with removing the MetaBar.

PR #8597, opened 2026-02-05
feat(ui): lessen padding of side bars

avivkeller:
Alternate #8596.

This PR greatly reduces the padding of the side bars, giving the content more visibility on the page.

---
After:
image


ovflowd:
Can you add a before/after, I unfortunately am not able to see what changed here :thinking: 

(Just asking because the before/after isn't obvious for me :sweat_smile:)

PR #8618, opened 2026-02-11
feat: conditional exports, numerous bug fixes and ux improvements

ovflowd:
## Description

This PR implements several design improvements, infrastructure changes, and bug fixes for the website and the rehype-shiki plugin — primarily addressing feedback tracked in https://github.com/nodejs/doc-kit/issues/596.

> [!NOTE]
> The ui-components publishing changes have been split out to https://github.com/nodejs/nodejs.org/pull/8620.

> [!NOTE]
> This PR fixes critical timeouts, theme flashing issues and hydration issues with the Partners List 

### 1. Global CSS-only tooltips via `data-tooltip`

Replaced all per-component tooltip implementations (DataTag, Badge) and the Radix `Tooltip` wrapper on the Partners page with a single global `[data-tooltip]` CSS rule in `base.css`. Any element with `data-tooltip="…"` now shows a styled tooltip on hover/focus using `::after`/`::before` pseudo-elements — no JS required.

- **Global styles in `base.css`**: White background, border, shadow, arrow with border; dark mode handled via `[data-theme='dark'] [data-tooltip]` ancestor selector (avoids Tailwind v4's `dark:` variant limitation where `@apply dark:…` generates a same-element `:where([data-theme=dark])` check)
- **DataTag**: Renders a human-readable label (e.g. "Method", "Constructor", "Deprecated") on hover/focus via `data-tooltip`, improving discoverability for the E/S/D/M/P/C indicator letters
- **Badge**: Per-component `[data-tooltip]` styles removed — global styles apply automatically
- **Partners page**: Replaced native `title` attribute with `data-tooltip` for consistent styled tooltips

Screenshot

image

image

### 2. Monospace font for code-like entries in Table of Contents and MetaBar

The `TableOfContents` and `MetaBar` components now apply IBM Plex Mono (`.codeLink`) to headings whose `type` is `method`, `classMethod`, `function`, or `ctor`. This makes function/method names visually distinct from prose headings.

Screenshot

image

### 3. Heading anchor `#` indicator

Added an anchor link indicator on headings that appears on hover. Headings with an `id` now have a `.anchor` child link with a `::after` pseudo-element (`content: '#'`) that fades in on hover/focus. Also added `scroll-margin-top` to headings for proper scroll-to behavior with the sticky header.

> [!NOTE]
> This only renders on HOVER 

Screenshot

image

### 4. Slimmer code box footer and hidden line numbers for short snippets

- Reduced `BaseCodeBox` footer paddings (`px-3 py-2`) and copy button size (`px-2.5 py-1`, smaller icon)
- `CodeTabs` header padding reduced (`pt-2`)
- `Tabs` trigger alignment adjusted (`pt-0 pb-2`, smaller extension badge `py-0`)
- The `rehype-shiki` plugin now counts lines and:
  - Adds `no-line-numbers` class when a code block has fewer than 5 lines
  - Adds `no-footer` class for single-line snippets (hides footer entirely)
- Plain-text code blocks now use a CSS class (`plain-text`) instead of inline `style`

Screenshot — short snippet (no line numbers)

image

image

### 5. Increased MetaBar width and responsive layout adjustments

- `Article` grid columns widened: sidebar `52` → `56`, metabar `52` → `64` at `xl`, `72` at `2xl`, `80` at `3xl`
- Sidebar now sticky at `ml:` breakpoint instead of `xl:`, uses `h-svh` instead of `h-screen`, with `ml:pb-[var(--header-height)]` for correct scroll height
- `MetaBar` max-width constraint removed (`xl:max-w-xs` → `xl:max-w-none`)
- `TableOfContents` details element hides at `xl:` instead of `lg:` (since MetaBar is visible on xl+)
- `baseLayout` in `layouts.module.css` now applies grid only at `ml:` breakpoint
- Content area padding adjusted: `xl:p-10 2xl:p-12` instead of flat `xl:p-12`

### 6. Better list spacing and responsive paddings

- Lists (`ul`, `ol`) now have `mt-2` between consecutive `li` elements
- List padding reduced and made responsive: `max-ml:pr-3 max-ml:pl-5 pr-4 pl-6`
- Table cells responsive padding: `max-ml:px-3 max-ml:py-1.5`
- Content gap reduced on mobile: `max-ml:gap-4` for `main`
- Replaced deprecated `break-words` with `wrap-break-word`

### 7. Simplified avatar count logic in MetaBar

Replaced two separate media queries (`isMobileResolution`, `isTabletResolution`) with a single `isSmallerThanDesktop` check. Avatar limit simplified from `isMobileResolution ? 7 : isTabletResolution ? 5 : 9` to `isSmallerThanDesktop ? 5 : 8`.

### 8. Removed `motion-safe:transition-colors` from interactive elements

Stripped `motion-safe:transition-colors` from `BaseButton`, `SearchButton`, `ThemeToggle`, and `NavItem` — these transitions added no meaningful UX value and introduced unnecessary style overhead. `motion-safe:duration-300` is retained where hover effects still use it.

### 9. Tailwind v4 class name fixes

- Fixed deprecated `bg-gradient-to-r` → `bg-linear-to-r` in `BaseButton`'s special variant (Tailwind v4 renamed gradient utilities)

### 10. Fetch retry with timeout

Added `AbortSignal.timeout(30000)` to `fetchWithRetry` to prevent hanging requests during data fetching.

### 11. DigitalOcean SVG path cleanup

Collapsed the multi-line SVG `d` attribute in the DigitalOcean favicon into a single line (no visual change, just formatting).

---

Refs: https://github.com/nodejs/doc-kit/issues/596
Refs: https://github.com/nodejs/node/issues/52343


avivkeller (dismissed):
Can we split this up, please

.github/workflows/publish-packages.yml
```diff
@@ -116,10 +116,11 @@ jobs:
           # Check if a custom publish script exists in package.json
           if jq -e '.scripts.release' package.json > /dev/null; then
             pnpm run release
-          else
-            pnpm publish --access public --no-git-checks
           fi
 
+          # Then publish the package to npm
+          pnpm publish --access public --no-git-checks
+
```
avivkeller: This is important

ovflowd:
> Can we split this up, please

What exactly you want to have split up, could you be more precise? I'd agree splitting up the conditional exports, but everything else IMO is fair game as it is just visual bug fixes and ux improvements (no new feature or refactor or major breaking change)

avivkeller:
I'd like if all the publishing changes are one PR, and everything else separate

packages/ui-components/package.json
```diff
@@ -20,7 +27,8 @@
     "compile:ts": "tsc",
     "compile:css": "postcss --dir dist --base src \"src/**/*.module.css\" src/styles/index.css",
     "compile": "node --run compile:ts && node --run compile:css",
-    "release": "node --run compile && node scripts/publish.mjs",
+    "compile:watch": "node --run compile:ts -- --watch & node --run compile:css -- --watch",
+    "release": "node --run compile",
```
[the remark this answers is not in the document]
ovflowd: I ain't adding a 3rd party package due to that.

ovflowd:
> I'd like if all the publishing changes are one PR, and everything else separate

Fair, can you do that if you have bandwidth? I unfortunately went overboard open source for the week, and gotta focus on my job for the rest of the week :(

ovflowd:
Let me ask Copilot to do this lol

apps/site/components/Common/Partners/index.tsx
```diff
@@ -36,15 +35,15 @@ const renderSmallPartner = (partner: Partner) => {
   const Logo = PartnerLogos[partner.id];
 
   return (
-    <Tooltip
+    <PartnerButton
+      aria-label={partner.name}
       key={partner.id}
-      asChild
-      content={<div className={style.tooltip}>{partner.name}</div>}
```
avivkeller, line 42: Shouldn't you also remove the `tooltip` style, in this case

apps/site/components/Common/Partners/index.tsx
```diff
@@ -36,15 +35,15 @@ const renderSmallPartner = (partner: Partner) => {
...
+      size="small"
+      href={partner.href}
+      data-tooltip={partner.name}
     >
-      <PartnerButton aria-label={partner.name} size="small" href={partner.href}>
-        <Logo.Favicon />
-      </PartnerButton>
-    </Tooltip>
+      <Logo.Favicon />
+    </PartnerButton>
```
avivkeller, line 46: Looks like this method of tooltip shows a help cursor:
Image

apps/site/next-env.d.ts
avivkeller, line 1: In a follow-up, we should gitignore this file, imo

apps/site/next.fetch.mjs
```diff
@@ -26,7 +26,7 @@ export const fetchWithRetry = async (
   const backoff = Math.max(0, Number(delay) || 0);
 
   const attemptFetch = attempt =>
-    fetch(url, options).catch(e => {
+    fetch(url, { ...options, signal: AbortSignal.timeout(30000) }).catch(e => {
```
avivkeller, line 29:
```suggestion
    fetch(url, { ...options, signal: AbortSignal.timeout(timeout) }).catch(e => {
```

packages/ui-components/src/Common/BaseCodeBox/index.module.css
```diff
@@ -53,6 +53,19 @@
             [counter-increment:line];
         }
       }
+
+      &[class*='plain-text'] {
```
avivkeller, line 57: -1 on this kind of CSS, let's use `.[class]` selectors, they are better for minification

packages/ui-components/src/Common/BaseCodeBox/index.module.css
```diff
@@ -53,6 +53,19 @@
...
+
+    &[class*='no-line-numbers'] > code > [class='line'] {
+      @apply pl-0;
+
+      &:not(:empty:last-child)::after {
+        @apply content-none
+          [counter-reset:none];
+      }
     }
   }
```
avivkeller, line 70: Why do we need to override the CSS here, can't we just exclude `no-line-numbers` from the line number styles above, rather than set them and then override them here?

packages/rehype-shiki/src/plugin.mjs
```diff
@@ -13,6 +13,11 @@ const languagePrefix = 'language-';
 // The regex to match metadata
 const rMeta = /(\w+)(?:=(?:"([^"]+)"|(\S+)))?/g;
 
+const getLineCount = value =>
+  value
+    .split('\n')
+    .filter((_, i, arr) => !(i === arr.length - 1 && arr[i] === '')).length;
+
```
avivkeller: This seems overly complex


```suggestion
const getLineCount = value => value.split('\n').length
```

Plus, any/all trailing lines that made it this far are there intentionally

packages/ui-components/src/Common/BaseCodeBox/index.tsx
```diff
@@ -89,8 +85,11 @@ const BaseCodeBox: FC<PropsWithChildren<CodeBoxProps>> = ({
   const containerRef = useRef<HTMLPreElement>(null);
 
   const handleCopy = () => copy(containerRef.current?.textContent);
+
   const ButtonIcon = copied ? DocumentDuplicateIcon : CodeBracketIcon;
 
+  const hideFooter = className?.includes('no-footer');
+
```
avivkeller, line 92: My major concern is that these classes will not be minified, since they are used like this. I'd rather have Shiki pass additional props (i.e. `footer={false}`)

packages/ui-components/src/Common/TableOfContents/index.tsx
```diff
@@ -44,7 +46,12 @@ const TableOfContents: FC<TableOfContentsProps> = ({
           <li key={head.data?.id ?? index}>
             <Component
               href={head.data?.id && `#${head.data.id}`}
-              className={classNames(styles.link, depthClasses[head.depth])}
+              className={classNames(
+                styles.link,
+                depthClasses[head.depth],
+                CODE_LIKE_TYPES.has((head.data?.type as string) ?? '') &&
+                  styles.codeLink
+              )}
```
avivkeller: What's this?

packages/ui-components/src/Containers/MetaBar/index.tsx
```diff
@@ -55,7 +57,14 @@ const MetaBar: FC<MetaBarProps> = ({
...
                   >
-                    <Component href={`#${head.data?.id}`}>
+                    <Component
+                      href={`#${head.data?.id}`}
+                      className={
+                        CODE_LIKE_TYPES.has((head.data?.type as string) ?? '')
+                          ? styles.codeLink
+                          : undefined
+                      }
+                    >
```
avivkeller, line 68: You use this style twice, IMO it should be shared somewhere, no need to duplicate code.

packages/ui-components/src/Icons/PartnerLogos/DigitalOcean/Favicon.tsx
```diff
@@ -18,9 +18,7 @@ const DigitalOcean: FC<SVGProps<SVGSVGElement>> = props => (
                 <path
                   id="XMLID_49_"
                   fill="currentColor"
-                  d="M155.2,351.7v-34.2c36.2,0,64.3-35.9,50.4-74c-5.1-14.1-16.4-25.4-30.5-30.5
-                  c-38.1-13.8-74,14.2-74,50.4l0,0H67c0-57.7,55.8-102.7,116.3-83.8c26.4,8.3,47.5,29.3,55.7,55.7
-                  C257.9,295.9,213,351.7,155.2,351.7z"
+                  d="M155.2,351.7v-34.2c36.2,0,64.3-35.9,50.4-74c-5.1-14.1-16.4-25.4-30.5-30.5c-38.1-13.8-74,14.2-74,50.4l0,0H67c0-57.7,55.8-102.7,116.3-83.8c26.4,8.3,47.5,29.3,55.7,55.7C257.9,295.9,213,351.7,155.2,351.7z"
```
avivkeller, line 21: Unrelated

packages/ui-components/README.md
avivkeller, line 1: This should go with the other PR

apps/site/components/Common/Partners/index.tsx
ovflowd: Correct, is that not OK?

apps/site/next.fetch.mjs
ovflowd: wdym by this?

packages/ui-components/src/Common/BaseCodeBox/index.module.css
ovflowd: In this situation we can't really use as they're invoked from outside a component, but from rehype.

packages/ui-components/src/Common/BaseCodeBox/index.module.css
ovflowd: Not sure I follow.

packages/rehype-shiki/src/plugin.mjs
ovflowd: I disagree. You can run your example locally and see that it produces incorrect line number results.

packages/ui-components/src/Common/BaseCodeBox/index.tsx
ovflowd: Wdym will not be minified? The styles will absolutely be minified. rehype can't pass react props to the component. It can pass HTML props. So you can't from there pass a boolean... Just a string with the value false, but that's the same as using classNames imo.

packages/ui-components/src/Common/TableOfContents/index.tsx
ovflowd: Monofonot on ToC

packages/ui-components/src/Icons/PartnerLogos/DigitalOcean/Favicon.tsx
ovflowd: wdym? This actually fixes a hydration issue.

apps/site/next.fetch.mjs
avivkeller: Let's specify the `timeout` in the options

apps/site/components/Common/Partners/index.tsx
avivkeller: It's fine, just making sure it's an intentional change

packages/rehype-shiki/src/plugin.mjs
avivkeller: > You can run your example locally and see that it produces incorrect line number results.

No, I think it produces the correct number of lines:

```js
const original = value =>
    value
        .split('\n')
        .filter((_, i, arr) => !(i === arr.length - 1 && arr[i] === '')).length;

const suggested = value => value.split('\n').length;


const cases = [
    "Hello\nWorld", // Two lines
    "Hello\nWorld\n", // Three lines
    "Hello\n\nWorld", // Three lines
]

cases.map(c => ([original(c), suggested(c)])).forEach(([o, s]) => console.log(`Original: ${o} | Suggested: ${s}`));
```

packages/ui-components/src/Common/BaseCodeBox/index.module.css
avivkeller: On line 32, we have an `& > [class='line'] { ...` CSS opener.

Why don't we change that to `&:not([class='no-line-number']) > [class='line']`? I think it's additional work for the browser to set the line styles, then apply `content:none`, when we can just not set the styles in the first place

packages/rehype-shiki/src/plugin.mjs
ovflowd: Yeah the original is correct. Suggested isn't.

apps/site/next.fetch.mjs
ovflowd: I don't think that's needed. timeout isn't even a valid fetch option. Let's not break the spec.

packages/ui-components/src/Common/BaseCodeBox/index.module.css
ovflowd: Could you give a full example? Im failing to understand what the issue is.

packages/rehype-shiki/src/plugin.mjs
avivkeller: I disagree. Any trailing lines should be counted as part of the line count. We trim the one trailing line when we do `const preElementContents = toString(preElement).replace(/\n$/, '');`, so why are we trimming a _second_ one here?

ovflowd: We're not triming twice, technically speaking since there's `const preElementContents = toString(preElement)`, but your code and mine will give the same result. So, actually, I believe it can be changed to your code. The `.replace(/\n$/, '');` was added by me.

apps/site/components/Common/Partners/index.tsx
ovflowd: Hmm, I think it does look odd indeed, for buttons. Wdyt, I think I should revert the cursor style...

ovflowd:
cc @nodejs/nodejs-website for reviews?

packages/ui-components/src/Common/BaseCodeBox/index.module.css
```diff
@@ -53,6 +53,19 @@
             [counter-increment:line];
         }
       }
+
+      &[class*='plain-text'] {
+        @apply font-mono;
```
canerakdas: To use IBM Plex Mono font, we need to apply the following utility class;

```suggestion
        @apply font-ibm-plex-mono;
```

packages/ui-components/src/Common/BaseCodeBox/index.module.css
```diff
@@ -72,9 +85,15 @@
...
-      @apply px-3
-        py-1.5
+      @apply px-2.5
+        py-1
+        text-xs
         font-medium;
+
+      > svg {
+        @apply h-4
+          w-4;
```
canerakdas: nit:
```suggestion
        @apply size-4;
```

packages/ui-components/src/styles/base.css
```diff
@@ -15,3 +15,75 @@ body {
...
+      px-2.5
+      py-1.5
+      text-sm
+      font-medium
+      text-neutral-900
+      opacity-0
+      shadow-lg
+      transition-opacity;
+
+    content: attr(data-tooltip);
```
canerakdas: To improve readability, it might be better to add it like this;

```css
@apply [content:attr(data-tooltip)]
```

maybe;
```css
@apply content-[attr(data-tooltip)]
```

packages/ui-components/src/styles/base.css
```diff
@@ -15,3 +15,75 @@ body {
...
+
+  &::before {
+    @apply pointer-events-none
+      absolute
+      top-full
+      left-1/2
+      z-10
+      mt-1
+      h-2
+      w-2
```
canerakdas:
```suggestion
      size-2
```

packages/ui-components/src/styles/base.css
```diff
@@ -15,3 +15,75 @@ body {
...
+      rotate-45
+      border
+      border-r-0
+      border-b-0
+      border-neutral-200
+      bg-white
+      opacity-0
+      transition-opacity;
+
+    content: '';
```
canerakdas: Similar to the suggestion above, keeping this inside `@apply` as well would make it more readable IMO

canerakdas:
Not having line numbers for short code snippets is a nice touch ✨ However, to be honest, I’m not a big fan of the codebox footer paddings the button feels a bit too small and cramped 🥹 

If showing multiple code boxes within a single scroll doesn’t provide a clear benefit, I’d rather not optimize for that. I think the current spacing already makes the page more readable

image

Also the alignment of the “JavaScript” label with the tab names and line numbers doesn’t look quite right to me

ovflowd:
> Not having line numbers for short code snippets is a nice touch ✨ However, to be honest, I’m not a big fan of the codebox footer paddings the button feels a bit too small and cramped 🥹
> 
> If showing multiple code boxes within a single scroll doesn’t provide a clear benefit, I’d rather not optimize for that. I think the current spacing already makes the page more readable
> 
> image
> Also the alignment of the “JavaScript” label with the tab names and line numbers doesn’t look quite right to me

Feel free to make a commit to the PR with your suggestion :)

canerakdas:
> Feel free to make a commit to the PR with your suggestion :)

This PR does exactly what’s described in the description and resolves many issues and added great features. We can move it forward without waiting for me. I can address my comments in a separate PR 😄

canerakdas (approved):
LGTM

PR #8623, opened 2026-02-13
docs: update explanation of setImmediate vs setTimeout

sebastiendan:
## Description

Clarifies the execution order of setTimeout and setImmediate when called from the main module and inside an I/O callback.

The current `they will be both run in the next iteration of the event loop` phrasing is confusing without the explicit context of where the code is run (main module vs I/O callback).

### Check List


AugustinMauroy:
cc @nodejs/timers

avivkeller (approved):

PR #8626, opened 2026-02-13
chore(ui): use compiled exports by default

avivkeller:
It makes little sense to me why the version _not_ published to npm is the default. It should be the other way around.

### DIFF
--- a/.github/workflows/publish-packages.yml
+++ b/.github/workflows/publish-packages.yml
@@ -114,8 +114,8 @@ jobs:
         working-directory: packages/${{ matrix.package }}
         run: |
           # Check if a custom publish script exists in package.json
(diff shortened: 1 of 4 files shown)


ovflowd (changes requested):
Might not make sense to you, but I've done this after several testing, the other order, things break, hard.

ovflowd:
(as you noticed yourself)

avivkeller:
I see. Had just run `compile` before testing, so my terminal didn't report any errors :-/

ovflowd:
> I see. Had just run `compile` before testing, so my terminal didn't report any errors :-/

I can explain via Slack, why I had to make this way if it helps :melting_face:

PR #8631, opened 2026-02-15
Blog: Evolving the Node.js Release Schedule

UlisesGascon:
Preview url: https://nodejs-org-git-fork-ulisesgascon-release-announcement-openjs.vercel.app/en/blog/announcements/evolving-the-nodejs-release-schedule

**PUBLICATION DATE: April 2nd**

**Objective**

~~This is an initial draft of the blog post, intended to collect feedback and iterate until we have a clear plan for communicating the release schedule change.~~

Explain to the users that we are planning a change in the Release Schedule starting on Node@27 and provide context on how the schedule works in general lines and the motivation for this change.

**Context**

We have been discussing this topic for a while in https://github.com/nodejs/Release/issues/1113, https://github.com/nodejs/Release/issues/953 and at the [Collaboration Summit Chesapeake 2025](https://youtu.be/ppi87YjU9x0?si=NFF5WKIGDJE_U-_V&t=6524).

The goal of this post is to communicate the change clearly to users as part of the messaging around Node 26.x, since the new plan will take effect starting with Node 27.x. We decided to write this blog post during our last [Release WG meeting](https://www.youtube.com/live/ulMhCu8Bz30?t=322).


apps/site/pages/en/blog/announcements/evolving-the-nodejs-release-schedule.md
```diff
@@ -0,0 +1,114 @@
+---
+date: '2026-04-01T00:00:00.000Z'
```
UlisesGascon: This is just a placeholder, no idea when makes sense to publish it.

avivkeller (approved):

apps/site/pages/en/blog/announcements/evolving-the-nodejs-release-schedule.md
```diff
@@ -0,0 +1,114 @@
...
+
+## What's Changing
+
+Starting with `27.x` in 2027:
+
+- **One major release per year** (April), with LTS promotion in October
+- **Every release becomes LTS**. No more odd/even distinction.
+- **Alpha channel replaces odd-numbered releases** for early testing
+- **Version numbers align with years**: `27.x` in 2027, `28.x` in 2028
+- **Maximum 3 active release lines** (down from 5)
```
ChALkeR: > down from 5

When did that last happen?

ChALkeR: also s/active/supported/ as "active lts" was a defined term in release phases

aduh95: It happens every year, e.g. next time will be between 2026-04-22 and 2026-04-30 there will be:

- 20.x
- 22.x
- 24.x
- 25.x
- 26.x

And there's always `main` as well

gurgunday (approved):
lgtm

apps/site/pages/en/blog/announcements/evolving-the-nodejs-release-schedule.md
ChALkeR: @aduh95 I don't think _every_? Last time the overlap was in 2023 for under 2 weeks.

My point is: this change doesn't remove two supported branches, it removes at most one.
There were never two supported odd branches at the same moment.

It can't go from 5 to 3.

Current schedule gives 4 with sometimes small overlaps raising that to 5 for a few days.
And those overlaps always involve even branches, not the removed odd ones.

This new schedule gives 3 _but only assuming overlaps never happen anymore_:
image

> And there's always `main` as well

And if we count `main` it definitely is not down to 3.

apps/site/pages/en/blog/announcements/evolving-the-nodejs-release-schedule.md
```diff
@@ -0,0 +1,114 @@
...
+| Current | 6 months  | Apr to Oct. Stabilization                       |
+| LTS     | 29 months | Long-term support with security fixes           |
+
+Total support: 35 months from release to End of Life.
+
+This model follows a pattern similar to Ubuntu's release cycle: predictable April/October anchors, with interim releases for testing and LTS releases for production.
+
+### About the Alpha Channel
+
+The Alpha channel replaces odd-numbered releases. Alpha releases are signed, tagged, and tested through CITGM. This is different from [nightly builds](https://nodejs.org/download/nightly/), which remain available as automated untested builds from main.
```
pimterry: Nit: imo CITGM could use a link or explanation - I expect very very few of the people reading this will know what it is.

pimterry (approved):

aymen94 (approved):

apps/site/pages/en/blog/announcements/evolving-the-nodejs-release-schedule.md
aduh95: > I don't think _every_? Last time the overlap was in 2023 for under 2 weeks.

The only exception has been 2024 because Node.js 16 was cut early due to the OpenSSL EOL. It looks like two weeks (and on 2025, it even looks like negative time, but let me tell it was not!) but that's not telling the whole story: we have to prepare the new release line in advance and maintain it on top of the other lines ahead of its release date, so there's at least 4 weeks of overlap.

With the new schedule, we can avoid the overlap altogether because:

- cutting the alpha in October, we don't have to worry too much about making sure every deps is up-to-date etc, those can come later, meaning less preparation work
- cutting the semver in April, much of the preparation work would have happened in advance, and delays would be much lower stakes (last year 24.0.0 was delayed by the V8 update, which was a quite stressful situation for everyone involved because we ended up in a situation where the project had _no_ current release lines ready, which can't happen with the proposed schedule).

Whether alpha releases should or should not count is debatable, one can argue that it will still be 4 release lines.

That being said, I don't mind this being re-written to focus less on the number of release lines and more on the difference it makes for releasers.

apps/site/pages/en/blog/announcements/evolving-the-nodejs-release-schedule.md
MikeMcC399: Perhaps just remove "through CITGM" and leave it to say "and extensively tested"? That still gives a difference between Alpha and nightly, which says "automated untested".

But then CITGM is referred to later on, so perhaps it does need to be explained?

apps/site/pages/en/blog/announcements/evolving-the-nodejs-release-schedule.md
```diff
@@ -0,0 +1,114 @@
...
+date: '2026-04-01T00:00:00.000Z'
+category: announcements
+title: Evolving the Node.js Release Schedule
+layout: blog-post
+author: The Node.js Release Team
+---
+
+Starting with `27.x`, Node.js will move from two major releases per year to one. This post explains what's changing, why, and what it means for users.
+
+## Why This Change
```
RafaelGSS, line 21: Let's add the issue where we discussed it extensively: https://github.com/nodejs/Release/issues/1113

apps/site/pages/en/blog/announcements/evolving-the-nodejs-release-schedule.md
```diff
@@ -0,0 +1,114 @@
...
+
+We now have a decade of data showing how people actually use Node.js:
+
+- Odd-numbered releases see minimal adoption. Users wait for LTS.
+- The odd/even distinction confuses newcomers.
+- Many organizations skip odd releases entirely, upgrading only between LTS versions.
+
+### Volunteer Sustainability
+
+Node.js is maintained primarily by volunteers. While some contributors receive sponsorship, most of the work (reviewing PRs, handling security issues, cutting releases, backporting fixes) is done by people in their spare time.
```
RafaelGSS: We should also mention that we care about predictability for enterprises, so any new schedule will be well-defined so companies can prepare their teams to handle it.

apps/site/pages/en/blog/announcements/evolving-the-nodejs-release-schedule.md
```diff
@@ -0,0 +1,114 @@
...
+
+Managing security releases across four or five active release lines has become difficult to sustain. Each additional line increases backporting complexity. By reducing concurrent release lines, we can focus on better supporting the releases people actually use.
+
+## What's Changing
+
+Starting with `27.x` in 2027:
+
+- **One major release per year** (April), with LTS promotion in October
+- **Every release becomes LTS**. No more odd/even distinction.
```
RafaelGSS:
```suggestion
- **Every release becomes LTS**. No more odd/even distinction - Node.js 27 will be LTS.
```

It always good to clarify.

apps/site/pages/en/blog/announcements/evolving-the-nodejs-release-schedule.md
```diff
@@ -0,0 +1,114 @@
...
+This model follows a pattern similar to Ubuntu's release cycle: predictable April/October anchors, with interim releases for testing and LTS releases for production.
+
+### About the Alpha Channel
+
+The Alpha channel replaces odd-numbered releases. Alpha releases are signed, tagged, and tested through CITGM. This is different from [nightly builds](https://nodejs.org/download/nightly/), which remain available as automated untested builds from main.
+
+**Who it's for:** Library authors and CI pipelines testing compatibility with upcoming breaking changes. Not intended for production use.
+
+**What to expect:**
+- Semver-major changes land during this phase
```
RafaelGSS: We should clarify that although we might have two releases per month for Alpha, it's not written in stone, and it's up to the release team to define when to ship updates.

apps/site/pages/en/blog/announcements/evolving-the-nodejs-release-schedule.md
```diff
@@ -0,0 +1,114 @@
...
+| 31.x    | Oct 2030 | Apr 2031 | Oct 2031 | Mar 2034    |
+| 32.x    | Oct 2031 | Apr 2032 | Oct 2032 | Mar 2035    |
+| 33.x    | Oct 2032 | Apr 2033 | Oct 2033 | Mar 2036    |
+| 34.x    | Oct 2033 | Apr 2034 | Oct 2034 | Mar 2037    |
+| 35.x    | Oct 2034 | Apr 2035 | Oct 2035 | Mar 2038    |
+| 36.x    | Oct 2035 | Apr 2036 | Oct 2036 | Mar 2039    |
+
+## Thank You
+
+This change is the result of discussions across GitHub issues, Release Working Group meetings, and [the Collaboration Summit Chesapeake 2025](https://youtu.be/ppi87YjU9x0?si=NFF5WKIGDJE_U-_V&t=6524). We thank everyone who contributed feedback.
```
RafaelGSS: I think it's worth mentioning that we are going to discuss and present it again at the upcoming collaborator summit in London.

RafaelGSS (changes requested):
Despite those changes, I think we need to mention that it's up to the release team to define what the rules will be to ship semver-major commits in alpha versions.

apps/site/pages/en/blog/announcements/evolving-the-nodejs-release-schedule.md
aduh95: Node.js 27.0.0 (released circa April 2027) will not be LTS, LTS comes later, in October (likely 27.10.0 or something like that)

apps/site/pages/en/blog/announcements/evolving-the-nodejs-release-schedule.md
```diff
@@ -0,0 +1,149 @@
...
+### About the Alpha Channel
+
+The Alpha channel replaces odd-numbered releases. Alpha releases are signed,
+tagged, and tested through CITGM. CITGM (Canary in the Goldmine) is a tool we
+maintain that runs the test suite of major open-source packages on the upcoming
+version of Node.js, which can let us detect ecosystem breakage and notify the
+package authors ahead of the release.
+
+This is different from [Nightly builds](https://nodejs.org/download/nightly/), which remain
+available as automated untested builds from `main` – Alpha release may not contain all changes from
```
MikeMcC399:
```suggestion
available as automated untested builds from `main` – Alpha releases may not contain all changes from
```

apps/site/pages/en/blog/announcements/evolving-the-nodejs-release-schedule.md
```diff
@@ -0,0 +1,149 @@
...
+`main`, a change may be not included in an Alpha release if:
+
+- during Pull Request review, reviewers add a label requesting the change to not be backported
+  (e.g. if an API is getting runtime deprecated in an Alpha release, the change actually removing
+  that API should not land until the next release line).
+- during the Alpha release preparation, the releaser ultimately which commits actually makes the
```
MikeMcC399:
```suggestion
- during the Alpha release preparation, the releaser ultimately decides which commits actually make the
```

apps/site/pages/en/blog/announcements/evolving-the-nodejs-release-schedule.md
RafaelGSS: updated

BridgeAR (approved):

Waldenesque:
Folks this looks really nice, and it's a pleasant touch that the version numbers will match the year of release.

I would suggest adding one short section to the beginning of the blog post, something along these lines...


> Starting with 27.x, Node.js will move from two major releases per year to one. This post explains what's changing, why, and what it means for users.

**TL;DR**

*For most developers who occasionally switch to newer even-numbered LTS releases*

Almost nothing will change for you in practical terms, except starting in 2027 LTS releases of Node.js will begin including odd version numbers: 27, 28, 29 and so on will all be LTS releases. This will have little to no impact on new features and upgrades being included in LTS releases of Node.js. The primary impact of this change will be a significant reduction of unproductive busywork for the Node.js project, allowing everyone contributing to Node.js to spend a greater portion of their time and energy on useful activity and constructive effort.

*For developers more familiar with the details of the Node.js release process*

Effectively this change will eliminate short-lived odd-numbered releases, and replace them with an "Alpha" period preceding "Current" release for LTS versions of Node.js, allowing similar early compatibility testing and library & addon development with less ceremony.

> **Why This Change**
> 
> The current release schedule is 10 years old. Etc., etc...


Here is the goal. Someone on Hacker News or Reddit posts some low-effort fearmongering. You want someone else to be able to simply copy-and-paste a short summary from the blog post that will defuse any further drama within forty-five seconds. That is how long you have to frame the discussion before they all head off on some tangent. Then a routine post about Node.js inevitably transforms into dozens of comments debating the borrow checker, or higher-kinded types, or how Ken Thompson and Rob Pike miraculously reincarnated as freshman CS students at a community college in West Virginia after pioneering Unix. (Not that I've seen this movie before or anything.)

I bet you can avoid much of that nonsense if you lead with a succinct sound bite, targeting the majority of your audience with a quick recap of the bare basics they need to understand. Make up some brief message similar to what I just suggested. Or I don't care, copy some of my verbiage, or copy the whole thing, or anything you prefer. I just think you'll have less drama if you offer a sound bite of your own, rather than letting someone else (who might have an axe to grind) hype up whatever sound bite they prefer.

apps/site/pages/en/blog/announcements/evolving-the-nodejs-release-schedule.md
```diff
@@ -0,0 +1,149 @@
...
+reducing the number of concurrent release lines, we can focus on better
+supporting the releases people actually use.
+
+## What's Changing
+
+As of October 2026:
+
+- **One major release per year** (April), with LTS promotion in October
+- **Every release becomes LTS**. No more odd/even distinction.
+- **Alpha channel replaces odd-numbered releases** for early testing (for librairies)
```
panva:
```suggestion
- **Alpha channel replaces odd-numbered releases** for early testing
```

tchetwin: If you _don't_ choose panva's suggestion, there is still a typo in "librairies" -> libraries

apps/site/pages/en/blog/announcements/evolving-the-nodejs-release-schedule.md
```diff
@@ -0,0 +1,149 @@
...
+- **Version numbers align with years**: 27.0.0 in 2027, 28.0.0 in 2028
```
Renegade334: It's a nice coincidence, but not sure this is worth stating here; it's not a stated objective or an explicit commitment of the new release plan.

apps/site/pages/en/blog/announcements/evolving-the-nodejs-release-schedule.md
```diff
@@ -0,0 +1,149 @@
...
+Nightly builds lack. Also allows landing V8 updates earlier in the cycle.
+
+## What's NOT Changing
+
+- **Long-Term Support duration** remains similar (29 months)
+- **Migration windows preserved**. Overlap between LTS versions remains.
+- **Quality standards unchanged**. Same testing, same CITGM, same security process.
+- **Predictable schedule**. April releases, October LTS promotion.
+- **V8 adoption cycle**. Node.js latest releases will still include a version of
+  V8 that's at most about 6-month old.
```
Renegade334:
```suggestion
  V8 that's at most about 6 months old.
```

apps/site/pages/en/blog/announcements/evolving-the-nodejs-release-schedule.md
```diff
@@ -0,0 +1,149 @@
...
+
+## Timeline
+
+![New Node.js Release Schedule](/static/images/blog/announcements/2026-new-release-schedule.svg)
+
+### Node.js 26 Schedule (exiting one)
```
Renegade334:
```suggestion
### Node.js 26 Schedule (existing model)
```

apps/site/pages/en/blog/announcements/evolving-the-nodejs-release-schedule.md
```diff
@@ -0,0 +1,149 @@
...
+| Milestone   | Date         |
+| ----------- | ------------ |
+| 26.0.0      | April 2026   |
+| Enters LTS  | October 2026 |
+| Maintenance | October 2027 |
+| End of Life | April 2029   |
+
+Node.js 26 follows the existing schedule. This is the last release line under the current model.
+
+### Node.js 27 Schedule (new one)
```
Renegade334:
```suggestion
### Node.js 27 Schedule (new model)
```

apps/site/pages/en/blog/announcements/evolving-the-nodejs-release-schedule.md
```diff
@@ -0,0 +1,149 @@
...
+date: '2026-04-01T00:00:00.000Z'
+category: announcements
+title: Evolving the Node.js Release Schedule
+layout: blog-post
+author: The Node.js Release Team
+---
+
+Starting with 27.x, Node.js will move from two major releases per year to one.
+This post explains what's changing, why, and what it means for users.
+
```
Waldenesque, line 16: I think it would be helpful to add a quick summary statement here before going into further details:

**TL;DR: We expect most developers, who upgrade even-numbered LTS releases yearly or biennially, will notice only a cosmetic change in version numbering, without experiencing any significant differences or disruptions.**

I don't know if the Node project has specific guidelines for wording or formatting, so that's a loose suggestion. Some kind of short and sweet summary along those lines sets a comforting tone from the start, and can be easily copied and pasted from the blog to other discussion forums.

I deleted a more elaborate comment in favor of this focused suggestion. Should have thought more carefully, sorry my bad. This new release process looks like it will be really nice.

apps/site/pages/en/blog/announcements/evolving-the-nodejs-release-schedule.md
UlisesGascon: I mentioned in the last part, but I can bring it here too 🤔

apps/site/pages/en/blog/announcements/evolving-the-nodejs-release-schedule.md
UlisesGascon: Addressed in https://github.com/nodejs/nodejs.org/pull/8631/changes/edf1129287b2c614ce4266942fb5cfad3ed7b878

apps/site/pages/en/blog/announcements/evolving-the-nodejs-release-schedule.md
UlisesGascon: Thanks @Waldenesque! Addressed in https://github.com/nodejs/nodejs.org/pull/8631/commits/79d6b8778a5a0d174d7cdb9e706563504b8eb05a

apps/site/pages/en/blog/announcements/evolving-the-nodejs-release-schedule.md
UlisesGascon: Addressed in https://github.com/nodejs/nodejs.org/commit/79d6b8778a5a0d174d7cdb9e706563504b8eb05a

apps/site/pages/en/blog/announcements/evolving-the-nodejs-release-schedule.md
UlisesGascon: Addressed in https://github.com/nodejs/nodejs.org/commit/79d6b8778a5a0d174d7cdb9e706563504b8eb05a

apps/site/pages/en/blog/announcements/evolving-the-nodejs-release-schedule.md
UlisesGascon: Addressed in https://github.com/nodejs/nodejs.org/commit/79d6b8778a5a0d174d7cdb9e706563504b8eb05a

apps/site/pages/en/blog/announcements/evolving-the-nodejs-release-schedule.md
UlisesGascon: Addressed in https://github.com/nodejs/nodejs.org/commit/79d6b8778a5a0d174d7cdb9e706563504b8eb05a

UlisesGascon:
I think the PR is ready for review (even CI is green now :partying_face: ), but I am not sure what is the publication date... so leaving as draft until we can clarify that detail :thinking:

apps/site/pages/en/blog/announcements/evolving-the-nodejs-release-schedule.md
```diff
@@ -0,0 +1,165 @@
+---
+date: '2026-04-01T00:00:00.000Z'
+category: announcements
+title: Evolving the Node.js Release Schedule
+layout: blog-post
+author: The Node.js Release Team
```
bjohansebas: Could you add this author here so it shows up properly with the logo?

https://github.com/nodejs/nodejs.org/blob/main/apps/site/authors.json

apps/site/pages/en/blog/announcements/evolving-the-nodejs-release-schedule.md
```diff
@@ -0,0 +1,165 @@
...
+- **Reduced Releasers burden**
+
+### New Schedule
+
+| Phase   | Duration  | Description                                     |
+| ------- | --------- | ----------------------------------------------- |
+| Alpha   | 5 months  | Oct to Mar. Early testing, semver-major allowed |
+| Interim | 6 months  | Apr to Oct. Stabilization                       |
+| LTS     | 29 months | Long-term support with security fixes           |
+| EOL     | Infinity  | The project no longer provides any support      |
```
bjohansebas, line 66: Maybe a link to the EOL page could be helpful.
```suggestion
| [EOL](https://nodejs.org/en/about/eol) | Infinity  | The project no longer provides any support      |
```

apps/site/pages/en/blog/announcements/evolving-the-nodejs-release-schedule.md
```diff
@@ -0,0 +1,165 @@
...
+Total support: 35 months from release to End of Life.
+
+We took some inspiration from Ubuntu's release cycle: predictable
+April/October anchors, with Interim releases for testing and LTS releases for
+production.
+
+### About the Alpha Channel
+
+The Alpha channel replaces odd-numbered releases. Alpha releases are signed,
+tagged, and tested through CITGM. CITGM (Canary in the Goldmine) is a tool we
```
bjohansebas:
```suggestion
tagged, and tested through [CITGM](https://github.com/nodejs/citgm). CITGM (Canary in the Goldmine) is a tool we
```

bjohansebas (approved):

apps/site/pages/en/blog/announcements/evolving-the-nodejs-release-schedule.md
```diff
@@ -0,0 +1,165 @@
...
+supporting the releases people actually use.
+
+## What's Changing
+
+As of October 2026:
+
+- **One major release per year** (April), with LTS promotion in October
+- **Every release becomes LTS**. No more odd/even distinction - Node.js 27 will be LTS.
+- **Alpha channel replaces odd-numbered releases** for early testing
+- **Version numbers align with years**: 27.0.0 in 2027, 28.0.0 in 2028
```
MikeMcC399:
```suggestion
- **Version numbers align with the year of the first Interim release and transition to LTS**: 27.0.0 in 2027, 28.0.0 in 2028
```

I'm not sure the best way to write this but I found it confusing when I read the table later on

| Version | Alpha    | Interim  | LTS      | End of Life |
| ------- | -------- | -------- | -------- | ----------- |
| 27.x    | Oct 2026 | Apr 2027 | Oct 2027 | Mar 2030    |

apps/site/pages/en/blog/announcements/evolving-the-nodejs-release-schedule.md
```diff
@@ -0,0 +1,165 @@
...
+- **Reduced Releasers burden**
```
MikeMcC399:
```suggestion
- **Reduced Releasers' burden**
```

grammar

apps/site/pages/en/blog/announcements/evolving-the-nodejs-release-schedule.md
```diff
@@ -0,0 +1,165 @@
...
+- **Long-Term Support duration** remains similar (29 months)
+- **Migration windows preserved**. Overlap between LTS versions remains.
+- **Quality standards unchanged**. Same testing, same CITGM, same security process.
+- **Predictable schedule**. April releases, October LTS promotion.
+- **V8 adoption cycle**. Node.js latest releases will still include a version of
+  V8 that's at most about 6 months old.
+
+## Timeline
+
+![New Node.js Release Schedule](/static/images/blog/announcements/2026-new-release-schedule.svg)
```
MikeMcC399, line 116: It's not picking up this image in preview mode on https://github.com/UlisesGascon/nodejs.org/blob/release-announcement/apps/site/pages/en/blog/announcements/evolving-the-nodejs-release-schedule.md#timeline

Does this get fixed during publication or does this need to be a relative reference?

```text
../../../../public/static/images/blog/announcements/2026-new-release-schedule.svg
```

apps/site/pages/en/blog/announcements/evolving-the-nodejs-release-schedule.md
UlisesGascon: Solved in 7dd8066

apps/site/pages/en/blog/announcements/evolving-the-nodejs-release-schedule.md
UlisesGascon: Solved in 7dd8066

apps/site/pages/en/blog/announcements/evolving-the-nodejs-release-schedule.md
UlisesGascon: Solved in 7dd8066

apps/site/pages/en/blog/announcements/evolving-the-nodejs-release-schedule.md
UlisesGascon: Yes, in the GH preview mode is not  working as it is designed for Next.js, but you can see it working at https://nodejs-org-git-fork-ulisesgascon-release-announcement-openjs.vercel.app/en/blog/announcements/evolving-the-nodejs-release-schedule#timeline so I assume that the reference is correct for publication.

MikeMcC399: Thanks and apologies for missing the Vercel preview!

apps/site/pages/en/blog/announcements/evolving-the-nodejs-release-schedule.md
RafaelGSS: Let's align it with Kylie and Robin. We should create an official tweet using Node.js account along with it.

apps/site/pages/en/blog/announcements/evolving-the-nodejs-release-schedule.md
```diff
@@ -0,0 +1,165 @@
...
+layout: blog-post
+author: Node.js Releasers
+---
+
+Starting with 27.x, Node.js will move from two major releases per year to one.
+This post explains what's changing, why, and what it means for users. For the
+full discussion and background, see [nodejs/Release#1113](https://github.com/nodejs/Release/issues/1113).
+
+**TL;DR:** If you already upgrade between LTS versions, little changes beyond
+version numbering. LTS support windows remain similar, and the upgrade path
```
RafaelGSS: That doesn't "stay the same" since they can now use odd versions as LTS Prod

RafaelGSS (commented):
Final nits, let's coordinate with social team to publish it!

apps/site/pages/en/blog/announcements/evolving-the-nodejs-release-schedule.md
Waldenesque: Maybe slight rewording from...

**"LTS support windows remain similar, and the upgrade path stays the same."**

to...

**"LTS support windows and upgrade path remain similar."**

or just...

**"LTS support windows and upgrading remain similar."**

The reason I suggested this is so the reader is not overwhelmed by details, when most developers won't be affected beyond just the numbering of versions. If you're not already intimately familiar with all of this history, it's a lot of information to absorb.

UlisesGascon: I believe that is solved now in 7503c54 :)

apps/site/pages/en/blog/announcements/evolving-the-nodejs-release-schedule.md
```diff
@@ -0,0 +1,165 @@
...
+- **Alpha channel replaces odd-numbered releases** for early testing
+- **Version numbers align with the year of the first Interim release and transition to LTS**: 27.0.0 in 2027, 28.0.0 in 2028
+- **Reduced Releasers' burden**
+
+### New Schedule
+
+| Phase   | Duration  | Description                                     |
+| ------- | --------- | ----------------------------------------------- |
+| Alpha   | 5 months  | Oct to Mar. Early testing, semver-major allowed |
+| Interim | 6 months  | Apr to Oct. Stabilization                       |
```
aduh95: From what I understand from https://ubuntu.com/about/release-cycle, using `Interim` might not be a wise choice since Ubuntu seems to be using it only for releases that do not get long-term support (i.e. it would fit the current schedule's odd releases, but the new one no longer have those). So if we release 27.0.0 as Interim, it's actually sending the wrong signal.

Wdwt about "Latest"? 
```suggestion
| Latest  | 6 months  | Apr to Oct. Stabilization                       |
```

apps/site/pages/en/blog/announcements/evolving-the-nodejs-release-schedule.md
```diff
@@ -0,0 +1,165 @@
...
+| LTS     | 29 months | Long-term support with security fixes           |
+| EOL     | Infinity  | The project no longer provides any support      |
+
+Total support: 35 months from release to [End of Life (EOL)](https://nodejs.org/en/about/eol).
+
+We took some inspiration from Ubuntu's release cycle: predictable
+April/October anchors, with Interim releases for testing and LTS releases for
+production.
```
aduh95: If we don't keep the `Interim` nomenclature, we should probably remove that paragraph since we have very little in common left.

apps/site/pages/en/blog/announcements/evolving-the-nodejs-release-schedule.md
```diff
@@ -0,0 +1,165 @@
...
+title: Evolving the Node.js Release Schedule
+layout: blog-post
+author: Node.js Releasers
+---
+
+Starting with 27.x, Node.js will move from two major releases per year to one.
+This post explains what's changing, why, and what it means for users. For the
+full discussion and background, see [nodejs/Release#1113](https://github.com/nodejs/Release/issues/1113).
+
+**TL;DR:** If you already upgrade between LTS versions, little changes beyond
```
MattIPv4: `between` feels a bit ambiguous here, I think it could be read as to imply upgrading from a LTS release to current, and then from current to the next LTS release -- upgrading between the LTS releases?

 
```suggestion
**TL;DR:** If you already only upgrade to LTS versions, little changes beyond
```

apps/site/pages/en/blog/announcements/evolving-the-nodejs-release-schedule.md
```diff
@@ -0,0 +1,165 @@
...
+
+The current release schedule is 10 years old. It was created during the io.js
+merger to balance the needs of a growing ecosystem. As one contributor put it at
+the time, it was "a guess of what enterprises would need."
+
+We now have a decade of data showing how people actually use Node.js:
+
+- Odd-numbered releases see minimal adoption. Most users wait for Long-Term Support.
+- The odd/even distinction confuses newcomers.
+- Many organizations skip odd releases entirely, upgrading only between LTS versions.
```
MattIPv4: Ditto on the `between` ambiguity

```suggestion
- Many organizations skip odd releases entirely, upgrading only to LTS versions.
```

apps/site/pages/en/blog/announcements/evolving-the-nodejs-release-schedule.md
MikeMcC399: The exact comparison with Ubuntu is tricky, because their Interim releases are killed and renumbered as LTS releases. Ubuntu LTS releases are LTS from the first day of release. https://ubuntu.com/about shows that pictorially.

Beta would not be a good term either.

Maybe just call it what it is: "Stabilization" instead of "Interim", if the term "Interim" may be misinterpreted? (Except that's a long word!)

apps/site/pages/en/blog/announcements/evolving-the-nodejs-release-schedule.md
```diff
@@ -0,0 +1,165 @@
...
+difficult to sustain. Each additional line increases backporting complexity. By
+reducing the number of concurrent release lines, we can focus on better
+supporting the releases people actually use.
+
+## What's Changing
+
+As of October 2026:
+
+- **One major release per year** (April), with LTS promotion in October
+- **Every release becomes LTS**. No more odd/even distinction - Node.js 27 will be LTS.
```
MattIPv4: Nit, while the line above does clarify, I feel `will be LTS` could be read as it being LTS at the point it is first released, rather than being promoted/graduating to LTS later in the cycle?

```suggestion
- **Every release becomes LTS**. No more odd/even distinction - Node.js 27 will become LTS.
```

apps/site/pages/en/blog/announcements/evolving-the-nodejs-release-schedule.md
```diff
@@ -0,0 +1,165 @@
...
+author: Node.js Releasers
+---
+
+Starting with 27.x, Node.js will move from two major releases per year to one.
+This post explains what's changing, why, and what it means for users. For the
+full discussion and background, see [nodejs/Release#1113](https://github.com/nodejs/Release/issues/1113).
+
+**TL;DR:** If you already upgrade between LTS versions, little changes beyond
+version numbering. LTS support windows remain similar, and now every release
+becomes LTS.
```
aduh95, line 15: IMO it would be worth adding a note specific to library authors for Alpha releases

```suggestion
becomes LTS.
**Library authors:** Please integrate Alpha releases to your CI as early as possible;
if you only test on LTS releases, you will not be able to report bugs before they
affect your users.
```

apps/site/pages/en/blog/announcements/evolving-the-nodejs-release-schedule.md
```diff
@@ -0,0 +1,165 @@
...
+
+## What's Changing
+
+As of October 2026:
+
+- **One major release per year** (April), with LTS promotion in October
+- **Every release becomes LTS**. No more odd/even distinction - Node.js 27 will be LTS.
+- **Alpha channel replaces odd-numbered releases** for early testing
+- **Version numbers align with the year of the first Interim release and transition to LTS**: 27.0.0 in 2027, 28.0.0 in 2028
+- **Reduced Releasers' burden**
```
MattIPv4:
```suggestion
- **One major release per year** (April), with LTS promotion in October.
- **Every release becomes LTS**. No more odd/even distinction - Node.js 27 will be LTS.
- **Alpha channel replaces odd-numbered releases** for early testing.
- **Version numbers align with the year of the first Interim release and transition to LTS**: 27.0.0 in 2027, 28.0.0 in 2028.
- **Reduced Releasers' burden**.
```
Grammar nit, consistent periods.

apps/site/pages/en/blog/announcements/evolving-the-nodejs-release-schedule.md
```diff
@@ -0,0 +1,165 @@
...
+**Who it's for:** Library authors and CI pipelines testing compatibility with
+upcoming breaking changes. Not intended for production use.
+
+**What to expect:**
+
+- Semver-major changes may land during this phase
+- Releases are signed and tagged (unlike nightly)
+- API may change between releases
+- The release cadence is flexible; the Release Team will determine the timing
+  and frequency of Alpha releases based on the volume of changes and project needs
```
MattIPv4:
```suggestion
- Semver-major changes may land during this phase.
- Releases are signed and tagged (unlike nightly).
- API may change between releases.
- The release cadence is flexible; the Release Team will determine the timing
  and frequency of Alpha releases based on the volume of changes and project needs.
```
Grammar nit, consistent periods.

apps/site/pages/en/blog/announcements/evolving-the-nodejs-release-schedule.md
```diff
@@ -0,0 +1,165 @@
...
+
+**Why:** Provides early feedback on breaking changes with quality gates that
+Nightly builds lack. Also allows landing V8 updates earlier in the cycle.
+
+The rules for shipping semver-major commits in Alpha versions will be defined by
+the Release Team and documented in the [Release repository](https://github.com/nodejs/Release).
+
+## What's NOT Changing
+
+- **Long-Term Support duration** remains similar (29 months)
```
MattIPv4:
```suggestion
- **Long-Term Support duration** remains similar (29 months).
```
Grammar nit, consistent periods.

apps/site/pages/en/blog/announcements/evolving-the-nodejs-release-schedule.md
MikeMcC399: I'd suggest removing that paragraph anyway because it starts to get confusing when you compare Node.js old model / new model and Ubuntu's model in any detail.

Ubuntu's model is more like the old Node.js model, where the Interim releases are never promoted to LTS.
In the Node.js new model, an Interim release is promoted to an LTS release by releasing a minor version update.

In Ubuntu's model each new LTS release is a new major version with a one year gap in between, 20.04, 22.04, 24.04 and coming up would be 26.04.

Also the April and October dates have different usage. Ubuntu releases LTS in April, with Interims in both April and October. Node.js releases Interims in April and promotes to LTS in October.

apps/site/pages/en/blog/announcements/evolving-the-nodejs-release-schedule.md
UlisesGascon: Solved in 13f3189

apps/site/pages/en/blog/announcements/evolving-the-nodejs-release-schedule.md
UlisesGascon: Solved in 13f3189

apps/site/pages/en/blog/announcements/evolving-the-nodejs-release-schedule.md
UlisesGascon: Solved in 13f3189

apps/site/pages/en/blog/announcements/evolving-the-nodejs-release-schedule.md
UlisesGascon: Solved in 13f3189

apps/site/pages/en/blog/announcements/evolving-the-nodejs-release-schedule.md
UlisesGascon: Solved in 13f3189

apps/site/pages/en/blog/announcements/evolving-the-nodejs-release-schedule.md
UlisesGascon: Solved in 13f3189

apps/site/pages/en/blog/announcements/evolving-the-nodejs-release-schedule.md
UlisesGascon: Solved in 13f3189

apps/site/pages/en/blog/announcements/evolving-the-nodejs-release-schedule.md
UlisesGascon: Removed in 13f3189

apps/site/pages/en/blog/announcements/evolving-the-nodejs-release-schedule.md
UlisesGascon: what about using Current (23ce3d1)? Maybe is less confusing as we use a term that is already known by our users? I you don't agree I can drop that commit :+1: . 


Super open to suggestions anyway, but I agree that maybe Interm is not clear enough. While I like  "Stabilization" maybe is translating the message that the release is not yet ready? but at the same time we finished Alfa but there is no Beta :face_with_diagonal_mouth:

Waldenesque: I suggest "Preview."

People heavily involved in open source speak so much English they're like native speakers, but that's not true for all developers. I expect a term like "pilot" would risk confusion for non-native English speakers, and can be ambiguous even if you are a native speaker. Maybe "promotion" since a new release is being promoted to LTS status. That's not too exotic, but it's not too familiar either, and feels subtle even for native speakers.

Perhaps "preview" is the least bad term. It borrows from everybody on earth watching Hollywood movies so they know about previews. The movie itself is largely finished, but it's being prepared for distribution to a wide audience. If you watch a preview, you get a valid taste of what's coming, but you also know that it's still possible something might change before final release. Even for native speakers this offers more instant recognition than other terms, and for people with limited English skills it's likely the easiest term.

apps/site/pages/en/blog/announcements/evolving-the-nodejs-release-schedule.md
```diff
@@ -0,0 +1,165 @@
...
+| ------- | --------- | ----------------------------------------------- |
+| Alpha   | 5 months  | Oct to Mar. Early testing, semver-major allowed |
+| Current | 6 months  | Apr to Oct. Stabilization                       |
+| LTS     | 29 months | Long-term support with security fixes           |
+| EOL     | Infinity  | The project no longer provides any support      |
+
+Total support: 35 months from release to [End of Life (EOL)](https://nodejs.org/en/about/eol).
+
+### About the Alpha Channel
+
```
Waldenesque, line 71: Sorry to stir up trouble here but...

If the goal is to invite library authors to integrate Alpha releases into their CI, the term "Alpha" is... bad. Nobody outside the Node inner circle has any idea how much or how little the details of release channels will change, so they cannot know how they should interpret terminology. If the proposed "Alpha" channel is intended to replace odd-numbered Current releases, it's a misleading name. The word "alpha" has strong connotations with decades of precedent that contradict such usage. I doubt many people would expect "alpha" releases to be tested through Canary in the Goldmine.

For a very mature project like Node, I suspect Nightly builds from main are closer to what most think of as "alpha" software. Today's odd-numbered Current releases are more equivalent to release candidates or at least beta versions. If someone from a Java background just starting serious work with Node sees the term "alpha," without knowing this history or context they might think "buggy prototype" and ignore it completely. Yet the purpose of this channel is to attract attention from library authors, right?

The term "beta" tells library authors it's time for testing without having to read a blog post. Was that term avoided deliberately, perhaps to allow more wiggle room for occasional buggy releases in this channel? I suppose "testing" would be misleading, since although this is meant for testing of libraries, the subsequent channel is also meant for testing, by end-user developers before LTS for production. Maybe the best option is already right under everyone's noses from CITGM: call this channel "Canary."

I think the non-native English speaker concern is less relevant here, since library authors already have to deal with more troublesome terminology anyway (and probably get more practice speaking English). The biggest precedent most people will have for the term "canary" is the React channel used by Next.js. That's probably an acceptable association, since it roughly matches what the experience will be for Node: generally stable, but expect occasional breakage or bugs. The name "Canary" avoids any "flaky prototype" connotations of "alpha," yet appropriately warns people not to expect reliable stability either.

After "Canary" to attract library authors, "Preview" would attract attention from early adopters (as opposed to those who stay on the older LTS). Imagine the viewpoint of Java, C#, Go, Rust, C++, Python and PHP developers, and hope for a match between their natural expectations and the future release channels being proposed for Node. I suspect "Canary" to "Preview" to "LTS" is less confusing or misleading than many other possibilities.

apps/site/pages/en/blog/announcements/evolving-the-nodejs-release-schedule.md
Renegade334: "Preview" suggests pre-release imo.

apps/site/pages/en/blog/announcements/evolving-the-nodejs-release-schedule.md
RafaelGSS: >  If the proposed "Alpha" channel is intended to replace odd-numbered Current releases

That's not the intention, Alpha releases can be very different from Current releases as we *can* ship semver-majors on it. 

I wouldn't be opposed to a different term, but Beta or Canary doesn't seem different from Alpha in my experience - all of them tell the same: It's an experimental version, subject to break.

Waldenesque: This is a direct quote from the blog post draft:

> The Alpha channel replaces odd-numbered releases.

So presumably that wording should be adjusted?

I think this perception of terminology is a difference between experts and everybody else. When you reach a certain level of expertise, you realize there isn't really a fundamental difference between categories like "alpha" or "beta" or whatever. But outside that circle of expertise, those words have a lot more subjective impact. I think "alpha" has stronger implications than experts realize.

MattIPv4: I personally think Alpha is also fine, it clearly communicates that the releases may still contain breaking changes, and that folks shouldn't rely on them for day-to-day use. We're explicitly stating that library authors are encouraged to test against these releases, whatever their perception of Alpha may be.

If there is really that much concern about the specific word being used, I could see Canary being a fine alternative, as it even more directly has connotations to being a release that we want library authors and such to test.

apps/site/pages/en/blog/announcements/evolving-the-nodejs-release-schedule.md
Waldenesque: I also thought about "Early," though I don't think that's been used by other software before, and it's vague. LibreOffice used to use the term "Fresh," though my own memory of confusion around that term partly triggered me when I saw this blog post. The problem in that case was they didn't explain anything clearly, so it was hard to understand how their releases worked (even though it wasn't actually complicated at all).

It's hard to find a word that strikes a good balance. Essentially the two channels (Alpha and Interim) are meant for library testing and application testing, right? Theoretically they could be called "Libtest" and "Apptest" which isn't pretty but certainly avoids confusion.

EDIT:

I tried a thesaurus for ideas, but nothing else seems to be clearly better. The way @UlisesGascon has it right now with "Current" may be best. If there's no obvious improvement available, the least bad option is probably just to avoid introducing new terminology at all.

aduh95 (approved):

MattIPv4 (approved):

apps/site/pages/en/blog/announcements/evolving-the-nodejs-release-schedule.md
RafaelGSS: > So presumably that wording should be adjusted?

Yes. Can you adjust? @UlisesGascon

UlisesGascon: solved in https://github.com/nodejs/nodejs.org/pull/8631/commits/01ecbebb99ab2e7b1237636850f24ee6e514b6fb

RafaelGSS (approved):

UlisesGascon:
Updated publication date to April 2nd

apps/site/pages/en/blog/announcements/evolving-the-nodejs-release-schedule.md
```diff
@@ -0,0 +1,165 @@
...
+- **Version numbers align with the year of the first Current release and transition to LTS**: 27.0.0 in 2027, 28.0.0 in 2028.
+- **Reduced Releasers' burden**.
+
+### New Schedule
+
+| Phase   | Duration  | Description                                     |
+| ------- | --------- | ----------------------------------------------- |
+| Alpha   | 5 months  | Oct to Mar. Early testing, semver-major allowed |
+| Current | 6 months  | Apr to Oct. Stabilization                       |
+| LTS     | 29 months | Long-term support with security fixes           |
```
richardlau: Why have we shortened the supported lifetime (of LTS) by a month? Did I miss a discussion?

apps/site/pages/en/blog/announcements/evolving-the-nodejs-release-schedule.md
```diff
@@ -0,0 +1,165 @@
...
+reducing the number of concurrent release lines, we can focus on better
+supporting the releases people actually use.
+
+## What's Changing
+
+As of October 2026:
+
+- **One major release per year** (April), with LTS promotion in October.
+- **Every release becomes LTS**. No more odd/even distinction - Node.js 27 will become LTS.
+- **Alpha channel for early testing** with semver-major changes allowed.
```
nschonni, line 54: Would be helpful to describe how the version numbering for these will work. EX: will it follow https://semver.org/ with a pattern like `27.0.0-alpha.1` or some variation on that

UlisesGascon: Solved in https://github.com/nodejs/nodejs.org/pull/8631/commits/2661a1df71cff60231343e405b4e76a59bdd82cb

apps/site/pages/en/blog/announcements/evolving-the-nodejs-release-schedule.md
UlisesGascon: Solved in https://github.com/nodejs/nodejs.org/pull/8631/commits/2661a1df71cff60231343e405b4e76a59bdd82cb

apps/site/authors.json
```diff
@@ -158,6 +158,11 @@
     "name": "Julián Duque",
     "website": "https://github.com/julianduque"
   },
+  "Node.js Releasers": {
+    "id": "nodejs",
```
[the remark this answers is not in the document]
UlisesGascon: Ignored

RafaelGSS:
@UlisesGascon any reason to not land that this week or early next week?

avivkeller:
(FYI Without a rebase Vercel will continue to fail, however, since building locally works, you can force merge this, though I'd recommend rebasing and using the Queue)

apps/site/pages/en/blog/announcements/evolving-the-nodejs-release-schedule.md
```diff
@@ -0,0 +1,166 @@
+---
+date: '2026-04-02T00:00:00.000Z'
```
aduh95: I suggest targeting next Tuesday at 1400UTC (3PM CET, 6AM PST, 7:30PM IST)
```suggestion
date: '2026-03-10T16:00:00.000Z'
```

apps/site/pages/en/blog/announcements/evolving-the-nodejs-release-schedule.md
```diff
@@ -0,0 +1,166 @@
...
+
+**Library authors:** Please integrate Alpha releases to your CI as early as possible;
+if you only test on LTS releases, you will not be able to report bugs before they
+affect your users.
+
+## Why This Change
+
+The current release schedule is 10 years old. It was created during the io.js
+merger to balance the needs of a growing ecosystem. As one contributor put it at
+the time, it was "a guess of what enterprises would need."
```
jasnell: Being the person that made that guess, I would qualify this as a well-informed guess, but still a guess ;-)

apps/site/pages/en/blog/announcements/evolving-the-nodejs-release-schedule.md
```diff
@@ -0,0 +1,166 @@
...
+reducing the number of concurrent release lines, we can focus on better
+supporting the releases people actually use.
+
+## What's Changing
+
+As of October 2026:
+
+- **One major release per year** (April), with LTS promotion in October.
+- **Every release becomes LTS**. No more odd/even distinction - Node.js 27 will become LTS.
+- **Alpha channel for early testing** with semver-major changes allowed.
```
trivikr, line 54: This might have been discussed before, but were other names considered instead of Alpha?
In semver, the `alpha` prerelease tag is used for the earliest and most unstable prerelease versions.
They're optionally followed by more stable `beta` releases, and `rc` tags close to final release.

Some options I can think of:
* `prerelease`: aligns with [semver](https://semver.org/). It's a little verbose though.
* `preview`: a shorter version of prerelease. This is often used by the enterprise, including Amazon, Microsoft, Google, Adobe, etc.
* `beta`: Chrome, whose V8 engine Node.js includes, uses [Beta](https://www.chromium.org/getting-involved/chrome-release-channels/)
* `dev`/`canary`/`nightly`: these terms are used for a more unstable versions, and may not be appropriate.
* `insiders`: some Microsoft products use this label.

Since Node.js stops taking semver major breaking changes one month before the `X.0.0` release, I believe the prerelease tag should change from `alpha` to `rc`. Some users might get confused with `rc` label and `Alpha` channel.

I think `preview` would be a better name for the channel. We can continue to use `alpha` and `rc` prerelease tags. Since the blog post mentions enterprise multiple times, we can consider using `preview` tag instead of `alpha` too. WDYT?

MikeMcC399: Given the statement:

> The Alpha channel fills the early-testing role that odd-numbered releases once served, but with a key difference: semver-major changes are allowed during Alpha.

it seems to be a more unstable version, with the possibility of semver-major changes, than `beta` or `rc` would imply

trivikr: I think the release team will continue publishing `rc` versions as they’ve done previously. These are currently released before the `current` phase (which doesn’t yet have an official name) and do not include `semver-major` changes.

Example for v25:

| Version                                                      | Release Date |
|--------------------------------------------------------------|--------------|
| [v25.0.0-rc.0](https://nodejs.org/download/rc/v25.0.0-rc.0/) | Sep 18, 2025 |
| [v25.0.0-rc.1](https://nodejs.org/download/rc/v25.0.0-rc.1/) | Oct 9, 2025  |
| [v25.0.0-rc.2](https://nodejs.org/download/rc/v25.0.0-rc.2/) | Oct 13, 2025 |
| [v25.0.0](https://nodejs.org/download/release/v25.0.0/)      | Oct 15, 2025 |

I believe that after this announcement, `rc` releases will continue to occur about a month before the actual releases and will not include any `semver-major` changes. However, they will take place during the `Alpha` phase, which could be confusing.

Example for v27 (dates are approx)

| Version                                 | Release Date | Phase | semver-major allowed? |
|-----------------------------------------|--------------|-------|-----------------------|
| v27.0.0-alpha.0                         | Oct 15, 2026 | Alpha | Yes                   |
| v27.0.0-alpha.1                         | Oct 30, 2026 | Alpha | Yes                   |
| // More alpha releases twice a month |              |       |                       |
| v27.0.0-alpha.9                         | Feb 28, 2027 | Alpha | Yes                   |
| v27.0.0-rc.0                            | Mar 15, 2027 | Alpha | No                    |
| v27.0.0-rc.1                            | Mar 28, 2027 | Alpha | No                    |
| v27.0.0-rc.2                            | Apr 9, 2027  | Alpha | No                    |
| v27.0.0                                 | Apr 15, 2027 | Current | No                    |

My concern is March 2027, when `rc` releases start appearing in the `Alpha` phase, which could cause confusion. Users might ask questions such as:
* Why are the prerelease tag `rc` instead of `alpha`?
* Will these releases still include `semver-major` changes? Should I avoid using them and wait for a stable version?

Renaming the phase would address both of these issues. We could call it the `preview` phase and use prerelease labels permitted by semver:
* `alpha`: `semver-major` changes allowed.
* `beta`: Not required at the moment, and the release team may never introduce it. However, the naming scheme leaves room for adding beta prereleases in the future.
* `rc`: `semver-major` changes not allowed. Users can adopt `rc` releases with more confidence compared to `alpha`.

The `preview` name is not new, and is also used by the multiple companies, as noted in the previous post.

aduh95: Bikeshedding of the release "phases" need to happen before we cut the 27.x release line, but I don't think it's a blocker for landing this blog post as is. Whether the releases in the alpha channel end up being called `preview`, `canary`, or something else, I don't expect that it's going to affect the ecosystem much. We will probably need to make adjustments anyway based on the feedback we'll receive from this blog post.

apps/site/pages/en/blog/announcements/evolving-the-nodejs-release-schedule.md
indexzero: +1 to it being based on the best information available at the time 🧠 of course I was a lowly Board collaborator and not a TSC member at the time (even more lowly today I suppose)

A few things are clear to me when comparing then to now when I reflect on this long strange trip:

1. The rocks to move up the mountain were less well defined then than they are now
2. The reverberation of the shockwaves stemming from the _"everything broke in every release"_ period of early on in the project's history could still be felt and seen in a way that isn't true today
3. The project's adoption – while significant then – had not yet reached the scale collaborators knew it could achieve

I recall fondly the Node.js Foudnation board meeting where Mikeal Rogers shared that _**"Node is bigger than Java"**_ which was ~18 months after this process was established

I imagine if he were here today he'd have an even more elegant words to say, but tragically, the lossy compression algorithm will have to suffice 🫶 

All of that of course surmounts to this being a good idea. Speaking in more recent terms I've been eager to use `node@25` features and would use `node@25` today if this process were active 👍

apps/site/pages/en/blog/announcements/evolving-the-nodejs-release-schedule.md
```diff
@@ -0,0 +1,166 @@
...
+- **Alpha versioning** follows semver prerelease format (e.g., `27.0.0-alpha.1`).
+- **Version numbers align with the calendar year of their initial Current release**: 27.0.0 in 2027, 28.0.0 in 2028.
+- **Reduced Releasers' burden**.
+
+### New Schedule
+
+| Phase   | Duration  | Description                                     |
+| ------- | --------- | ----------------------------------------------- |
+| Alpha   | 6 months  | Oct to Mar. Early testing, semver-major allowed |
+| Current | 6 months  | Apr to Oct. Stabilization                       |
```
trivikr, line 64: Since we're aligning the version number with the year, would it make sense to align the release months as well?

Instead of the prerelease phase running from October to March, shifting it to January through July would offer a cleaner calendar alignment:

```suggestion
| Alpha   | 6 months  | Jan to June. Early testing, semver-major allowed |
| Current | 6 months  | July to Dec. Stabilization                       |
```

With this schedule, a new Node.js version would release and the previous version would move to LTS both in January. The current proposal has both events landing in October, which is close to holiday season in the US and Europe. As Matteo's [downloads graph](https://nodedownloads.nodeland.dev/) shows, LTS download numbers tend to plateau or even dip in December, making October a less-than-ideal release window.

richardlau: A January release is unrealistic. It would involve preparation work happening in December.

apps/site/pages/en/blog/announcements/evolving-the-nodejs-release-schedule.md
aduh95:
```suggestion
the time, it was "an educated guess of what enterprises would need."
```

trivikr (approved):

apps/site/pages/en/blog/announcements/evolving-the-nodejs-release-schedule.md
trivikr: Resolving for discussion after blog post goes live.

Commented at https://github.com/nodejs/Release/issues/1113#issuecomment-4033172738

MikeMcC399 (commented):
It's probably too late now, but I just noticed on the apps/site/public/static/images/blog/announcements/2026-new-release-schedule.svg

there is no key that says green is "CURRENT"

Alpha isn't specifically labeled, although at least "UNSTABLE" is the same colour.

apps/site/pages/en/blog/announcements/evolving-the-nodejs-release-schedule.md
trivikr: I suggested extending the "Current" window to 12 months as an alternative https://github.com/nodejs/Release/issues/1143

PR #8632, opened 2026-02-15
meta: export types for rehype-shiki

ovflowd:
This PR introduces type exports for @node-core/rehype-shiki, which IMO is important for downstream consumers/importers of the package, such as doc-kit. Even if the runtime doesn't require the types.


ovflowd:
cc @nodejs/nodejs-website

avivkeller:
I don't think this is necessary, but I could be wrong

ovflowd:
> I don't think this is necessary, but I could be wrong

Kinda. Technically speaking VS Code will automatically resolve with its own IntelliSense, yes and use the JSDocs. But, adding this actually allows us to check for type errors. For example, this allowed me to see that HighlighterCoreSync type didn't exist anymore. So this allows us to catch errors. Even if the source is pure .mjs and TS is not needed tho.

We could even simply not ship the d.ts, but I want to also be a TypeScript citizen without really needing TypeScript lol

ovflowd:
cc @nodejs/nodejs-website

packages/rehype-shiki/package.json
```diff
@@ -1,22 +1,33 @@
...
   },
   "repository": {
     "type": "git",
     "url": "https://github.com/nodejs/nodejs.org",
     "directory": "packages/rehype-shiki"
   },
   "scripts": {
+    "compile:ts": "tsc",
+    "compile": "node --run compile:ts",
+    "release": "node --run compile",
```
avivkeller, line 25: (In a follow up) we should change the action to just run `compile` directly

packages/ui-components/package.json
avivkeller, line 1: Unrelated

avivkeller (commented):
Minus the ui-components changes, LGTM

packages/ui-components/package.json
ovflowd: Not really, it is needed otherwise the build breaks. The pinning change is unrelated tho, but the order and placement of dependencies, is. You could argue I could do that first in another PR but I don't think it really matters (we don't need to be that strict in this repo)

avivkeller: The order shouldn't matter, nor should the version. Sure, moving dependencies from devDeps may be rveelant, but I can't imagine anything else is

ovflowd: > Sure, moving dependencies from devDeps may be rveelant

It is relevant. CI installs pnpm i --prod, those dependencies should have been prod from the beginning, the fact the CI failed because I added `compile` to turbo.json shows the careless and lack of contributors following our dependency guidelines.

ovflowd: So in order for the compile to work, I had to fix the dependencies for ui-components. That could be another PR that acts as a prequel for this one? Sure, but I don't think it hurts adding them here and is related enough as ie: ensuring the pipeline for the rehype-shiki works...

avivkeller (approved):

packages/ui-components/package.json
ovflowd: I just want to ensure @avivkeller we're on the same page and you understand why I believe these changes are needed :sweat_smile:

AugustinMauroy (approved):
LGMT ! that's good idea

PR #8641, opened 2026-02-19
Blog: update new HackerOne signal requirements

RafaelGSS:
PTAL: @nodejs/security


BethGriggs (approved):

aduh95 (approved):

MattIPv4:
👀 Just an FYI for the future here, unsure if it was intentional or not, but bumping the updated date has caused this to be picked up as a new article by RSS readers.

RafaelGSS:
> 👀 Just an FYI for the future here, unsure if it was intentional or not, but bumping the updated date has caused this to be picked up as a new article by RSS readers.

That's was intentional.

Issue #8628, opened 2026-02-14
switch the nvm version of 0.40.3 to 0.40.4


Heitor-hub-del:
### What is the problem this feature will solve?

node 25.6.2 (or like that)

add the newest nvm version

### What is the feature you are proposing to solve the problem?

nvm (new version (0.40.4)

Heitor-hub-del:
basically a simple feature request, i know that it doesnt switch a version to another one but.... its always good to be up to date :)

Renegade334:
Node.js does not bundle nvm. If you have it installed, then updating it is your responsibility.

ljharb:
I think this is for the website.

Heitor-hub-del:
> Node.js does not bundle nvm. If you have it installed, then updating it is your responsibility.

<img width="1115" height="628" alt="Image" src="https://github.com/user-attachments/assets/3d230322-a67e-448e-a608-d9984704755c" />

Heitor-hub-del:
> Node.js does not bundle nvm. If you have it installed, then updating it is your responsibility.

wdym it "doesnt bundle"

<img width="1920" height="1032" alt="Image" src="https://github.com/user-attachments/assets/730de0d6-5431-47f4-9b76-c30ba4126326" />

ljharb:
`nvm install` is for installing node versions. To install nvm itself, you need to use nvm's install command, listed in its readme: https://github.com/nvm-sh/nvm

avivkeller:
@ljharb @Heitor-hub-del Frankly, it's hard for us to keep updating nvm every time a patch version is released. Would one the following solutions be possible

- Use the `HEAD` tag.
- Have nvm PR here when they want the website to change?

fyi @nodejs/nodejs-website

ljharb:
No to the first one; and the second one is adding the burden of me remembering to do it. We could of course reopen #7296 and it’d never be out of date, with zero effort.

ovflowd:
> reopen [#7296](https://github.com/nodejs/nodejs.org/pull/7296) and it’d never be out of date

I don't think we want to do that. Simply because it doesn't make sense for us to adhere to the whims of a version manager.

> with zero effort.

Or you could do 1) as all other version managers... but we won't force you into something, so please don't do the same.

ovflowd:
> Or you could do 1) as all other version managers... but we won't force you into something, so please don't do the same.

[...]

PR #8643, opened 2026-02-20
meta: bump nvm to v0.40.4

ljharb:
Updates the English nvm install snippet to [`v0.40.4`](https://github.com/nvm-sh/nvm/releases/tag/v0.40.4). The translation system handles other locales.

Ref: https://github.com/nodejs/nodejs.org/issues/8628

### DIFF
--- a/apps/site/snippets/en/download/nvm.bash
+++ b/apps/site/snippets/en/download/nvm.bash
@@ -1,5 +1,5 @@
 # Download and install nvm:
-curl -o- https://raw.githubusercontent.com/nvm-sh/nvm/v0.40.3/install.sh | bash
+curl -o- https://raw.githubusercontent.com/nvm-sh/nvm/v0.40.4/install.sh | bash
 
 # in lieu of restarting the shell
 \. "$HOME/.nvm/nvm.sh"


ljharb:
note: this is an automated PR, created by a GHA workflow on nvm. i'm hoping to sidestep arguments about what should be automated or not, and what node's responsibilities to its users are. The reality is that anything that removes burden from open source maintainers should be HIGHLY prioritized, and so this is my attempt to do just that.

okuryu:
Would it be a problem if we automatically update the nvm version for non-English pages as well?

Whenever the English page is updated, we get a notification from Crowdin and the number of untranslated strings increases. However, this change doesn't actually require any translation work. From a translator's perspective, it's not really a relevant change to process manually.

AugustinMauroy (approved):

ovflowd:
> note: this is an automated PR, created by a GHA workflow on nvm. i'm hoping to sidestep arguments about what should be automated or not, and what node's responsibilities to its users are. The reality is that anything that removes burden from open source maintainers should be HIGHLY prioritized, and so this is my attempt to do just that.

No no, you're good. I personally even think we could do the automation here if really needed, but it does make sense too to have it on nvm-side.

ovflowd:
> Would it be a problem if we automatically update the nvm version for non-English pages as well?
> 
> Whenever the English page is updated, we get a notification from Crowdin and the number of untranslated strings increases. However, this change doesn't actually require any translation work. From a translator's perspective, it's not really a relevant change to process manually.

Agree, the snippets also need to be translated, if the versions on the translated snippets could be updated, that'd be great.

avivkeller (approved):

avivkeller:
> if the versions on the translated snippets could be updated, that'd be great.

Won't Crowdin handle that?

ovflowd:
> > if the versions on the translated snippets could be updated, that'd be great.
> 
> Won't Crowdin handle that?

It might consider them as new strings, IDK.

ljharb:
In the past I updated them all at once, and was blocked until I only changed the english ones. Should I change my automation to change every language's version?

ovflowd:
> In the past I updated them all at once, and was blocked until I only changed the english ones. Should I change my automation to change every language's version?

I'm unsure at the moment, in general policy indeed is that... But really depends if these would get auto translated 🤔 or not even marked for translation (ie seen as variables), let's merge and see what happens.

PR #8672, opened 2026-02-28
chore(linting): update ESLint to v10, bump lint group deps

avivkeller:
This PR updates ESLint to v10, and replaces `eslint-plugin-react` with `@eslint-react/eslint-plugin` for the following reasons:
- `eslint-plugin-react` does not support ESLint v10
- `@eslint-react/eslint-plugin` is **much** smaller and **much** faster (in CI, it's 2m43s vs 3m12s


apps/site/hooks/client/index.ts
avivkeller, line 1: This is related, as the new lint rules include ensuring that `use`* hooks **only appear when they are actually used.

MattIPv4 (changes requested):
Not sure I'm overly a fan of this at first glance, it seems like we're losing linting on things like making sure hook calls have the correct dependencies?

Is there reason not to just wait until the existing React plugins are updating for eslint v10?

avivkeller:
> Not sure I'm overly a fan of this at first glance, it seems like we're losing linting on things like making sure hook calls have the correct dependencies?

That's [coming soon](https://github.com/Rel1cx/eslint-react/blob/main/packages/plugins/eslint-plugin-react-x/src/rules/exhaustive-deps/exhaustive-deps.mdx).

> Is there reason not to just wait until the existing React plugins are updating for eslint v10?

I don't think [they are planning on doing that anytime soon](https://github.com/jsx-eslint/eslint-plugin-react/issues/3977#issuecomment-3863644745).


---

Let's draft this until either one of the following:
1. This plugin releases v3
2. @ljharb's plugin supports v10

Whichever happens happens, how does that sound?

MattIPv4:
👍 Sounds reasonable

avivkeller:
3.0.0 is out, and the previous plugin still has not updated support for v10. I'll rebase this PR today

avivkeller:
Just need to fix the egregious lockfile diff...

apps/site/components/Common/FormattedTime.tsx
```diff
@@ -13,6 +13,8 @@ type FormattedTimeProps = {
 const FormattedTime: FC<FormattedTimeProps> = ({ date, format }) => {
   const formatter = useFormatter();
 
+  // "date" is deterministic
```
ovflowd, line 16: Since the Date won't change, doesn't make sense to create a new instance on every render. I agree either: useMemo() can be used here, but also the footprint here is so small and I doubt re-renders will happen often here to the point this is even remotely an issue. So also agree with dismissing it.

apps/site/components/EOL/EOLModal.tsx
```diff
@@ -26,6 +26,7 @@ const EOLModal: FC<EOLModalProps> = ({
     ? t('components.eolModal.title', { version, codename })
     : t('components.eolModal.titleWithoutCodename', { version });
 
+  // eslint-disable-next-line @eslint-react/use-memo
```
ovflowd: Why it is complaining about useMemo? 🤔

apps/site/hooks/server/useClientContext.ts
ovflowd, line 1: Why are you removing this? Server and Client contexts are different. One uses the getClientContext from server-side the other uses the Client-side Context Provider.

Could you explain the why on this change? Or has something changed on React allowing use() to be used both client/server and guarantee same context on client/server?

apps/site/eslint.config.js
```diff
@@ -10,11 +9,18 @@ export default baseConfig.concat([
...
-  next.configs['core-web-vitals'],
+  eslintReact.configs['recommended-typescript'],
+  next.configs.recommended,
   mdx.flatCodeBlocks,
 
+  // React
+  {
+    rules: {
+      '@eslint-react/no-array-index-key': 'off',
+      '@eslint-react/web-api/no-leaked-event-listener': 'off',
```
ovflowd: Could you add inline comments explaining (in the code) why these rules should be turned off?

packages/ui-components/src/Common/AvatarGroup/Avatar/index.tsx
```diff
@@ -17,54 +16,43 @@ export type AvatarProps = {
...
+const Avatar = ({
+  image,
+  nickname,
+  name,
+  fallback,
+  url,
+  size = 'small',
+  as: Component = 'a',
+  ...props
+}: HTMLAttributes<HTMLSpanElement> & AvatarProps) => {
```
ovflowd: nit: can you extract these props to our standard `FC<HTMLAttributes<HTMLSpanElement> & AvatarProps>` pattern?

packages/ui-components/src/Common/AvatarGroup/Avatar/index.tsx
```diff
@@ -17,54 +16,43 @@ export type AvatarProps = {
...
+  if (!url) {
+    Component = 'div';
+  }
 
-    return (
-      <RadixAvatar.Root
-        {...props}
-        ref={ref}
```
ovflowd, line 44: don't we need ref anymore? why this removed? 🤔

packages/ui-components/src/Containers/Sidebar/index.tsx
```diff
@@ -17,48 +15,56 @@ type SidebarProps = {
...
+const SideBar = ({
+  groups,
+  pathname,
+  title,
+  onSelect,
+  as,
+  children,
+  placeholder,
+  ...props
+}: PropsWithChildren<SidebarProps>) => {
```
ovflowd: nit: use our FC pattern

ovflowd (commented):
Left a bunch of review comments.

apps/site/hooks/server/useClientContext.ts
avivkeller: One of the new ESLint rules is that `use*` functions must contain an actual hook (e.g. `useState`, `useMemo`). This hook is a wrapper of a `get*` function, so I removed the wrapper, and it's consumers just call `getClientContext` directly

apps/site/components/EOL/EOLModal.tsx
avivkeller: > The return value of 'useMemo' must be assigned to a variable. Calling 'useMemo' without capturing its return value is likely a mistake — use 'useEffect' for side effects instead.

Fixed in a commit I'll push shortly

packages/ui-components/src/Common/AvatarGroup/Avatar/index.tsx
avivkeller: `ref` is passed as a part of `{...props}`

ljharb:
fwiw v10 support in eslint-plugin-react will happen soon, but obv if there's an urgency to upgrade to eslint 10, feel free to use an alternative.

avivkeller:
I can't speak for the rest of the team, but I think it's also important to note that the alternative plugin is much smaller, and also claims to be much faster.

ljharb:
Package size (including dep count) doesn't meaningfully affect local dev or CI times, and performance claims absent benchmarks aren't particularly useful :-)

ovflowd:
It's not that we have urgency, but it does indeed seem to have a smaller footprint. And it seems to introduce rules the other didn't? To be fair I'm neutral. Fine first merging this PR and then moving back if needed.

ovflowd:
cc @MattIPv4 re-review, @avivkeller can you rebase?

MattIPv4:
👍 I don't have strong opinions either way here, it looks like the new package is now doing what it was missing before and seems to be catching everything the old one was and more.

MattIPv4 (approved):
ESLint changes seem fine, can't speak to the changes relating to the client/server hook stuff though.

MattIPv4:
> ReferenceError: window is not defined

This build failure may suggest the server/client hook changes aren't quite right though?

avivkeller:
> > ReferenceError: window is not defined
> 
> This build failure may suggest the server/client hook changes aren't quite right though?

It's missing a:
```
    if (typeof window === 'undefined') {
      return false;
    }
```

from the previous implementation (which probably shouldn't be needed in the first place, tbh)

packages/ui-components/package.json
```diff
@@ -69,21 +69,22 @@
     "@storybook/addon-themes": "~10.3.3",
     "@storybook/addon-webpack5-compiler-swc": "~4.0.3",
     "@storybook/react-webpack5": "~10.3.3",
+    "@eslint-react/eslint-plugin": "~3.0.0",
```
bmuenzenmeyer, line 72: nit: our other eslint dependencies are directly pinned

bmuenzenmeyer (commented):
some food for thought - is this PR only making changes per the application of the new eslint rules? some of it to my quick glance suggests no.

relates to https://github.com/nodejs/web-team/pull/112

avivkeller:
> is this PR only making changes per the application of the new eslint rules?

As oddly as it seems, yes, however, these changes can definitely be broken up into a pre requisite PR

ovflowd (approved):

PR #8714, opened 2026-03-11
chore(ci): make cost-saving changes

aduh95:
## Description

IIUC we don't need full blown runners for most jobs, let's see if the slim ones do the job

Refs: https://docs.github.com/en/billing/reference/actions-runner-pricing

### Check List

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -47,7 +47,7 @@ jobs:
           - os: windows-latest
             # Skipping the Windows build on the private repo to save on costs
             repo: nodejs-private/nodejs.org-private
-        
+
     steps:
       - uses: nodejs/web-team/actions/setup-environment@9f3c83af227d721768d9dbb63009a47ed4f4282f
         with:
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -31,7 +31,7 @@ env:
 
 jobs:
   build-linux:
-    name: Build on Linux
+    name: Build on ubuntu-latest # Name must match the repo setting
     runs-on: ubuntu-slim
 
     steps:
@@ -77,7 +77,7 @@ jobs:
           NEXT_PUBLIC_STATIC_EXPORT_LOCALE: false # ${{ github.event_name == 'push' }}
 
   build-windows:
-    name: Build on windows
+    name: Build on Windows
     runs-on: windows-latest
     
     if: github.event_name != 'merge_group' || github.repository == 'nodejs/nodejs.org'
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -37,8 +37,17 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ubuntu-latest, windows-latest]
-
+        os: [ubuntu-slim, windows-latest]
+        event-name: ${{ github.event_name }}
+        repo: ${{ github.repository }}
+        exclude:
+          - os: windows-latest
+            # Skipping the Windows build on the merge queue to speed it up
+            event-name: merge_group
+          - os: windows-latest
+            # Skipping the Windows build on the private repo to save on costs
+            repo: nodejs-private/nodejs.org-private
+        
     steps:
       - uses: nodejs/web-team/actions/setup-environment@9f3c83af227d721768d9dbb63009a47ed4f4282f
         with:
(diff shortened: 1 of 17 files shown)


.github/workflows/build.yml
```diff
@@ -37,7 +37,16 @@ jobs:
...
+        os: [ubuntu-slim, windows-latest]
+        event-name: ${{ github.event_name }}
+        repo: ${{ github.repository }}
+        exclude:
+          - os: windows-latest
+            # Skipping the Windows build on the merge queue to speed it up
+            event-name: merge_group
+          - os: windows-latest
+            # Skipping the Windows build on the private repo to save on costs
+            repo: nodejs-private/nodejs.org-private
```
avivkeller: We (the website team) don't have access to this reppsitory

.github/workflows/build.yml
MattIPv4, line 1: Why has this been split into two separate jobs, rather than a matrix?

MattIPv4, line 1: I'm a bit concerned about using the lightweight runners here, given we've struggled with long build times, and recently needed to switch to more powerful Vercel runners as well?

aduh95: If you look at the commit history, you'll see I've started with that exact approach, and gave up as I wasn't able to make it work. Open to suggestions

MattIPv4: Yes, it looks like the runs failed with `Error when evaluating 'strategy' for job 'build'. .github/workflows/build.yml (Line: 41, Col: 21): Unexpected value 'pull_request', .github/workflows/build.yml (Line: 42, Col: 15): Unexpected value 'nodejs/nodejs.org'`, which seems like it was just a syntax error?

aduh95: Like I said, open to suggestions?

MattIPv4: I'd probably start with keeping it as the matrix it was before, and switching to the slim runner? Then if that works, look at adding the other changes you made that exclude certain runs etc.

aduh95: Maybe that's just a me-thing, but your use of question mark on sentences that are clearly not questions is... frustrating. It reads to me as the equivalent of writing `if(console.log('Hello'));` instead of `console.log('hello');`.
We can discuss decoupling (and if someone wants to open an alternative PR, that's great), but the more pressing problems is that it looks like lots of jobs are not compatible with the slim runners.

MattIPv4: They are questions. I am asking if that change can be done, as I would not like to land this as-is with the jobs being split out of what was a working matrix before, creating duplication.

ovflowd: Any update here?

aduh95: I still need help to unblock the `Error: EACCES: permission denied, open '/home/agent/post_event.json'` failure

bmuenzenmeyer:
this looks interesting to me, more from a speed perspective than costs savings. pulling a smaller image should be faster in most circumstances.

@aduh95 i am naive to this - but Node.js or OpenJS Foundation doesn't actually pay GitHub for our CI costs, right?

aduh95:
> Node.js or OpenJS Foundation doesn't actually pay GitHub for our CI costs, right?

We do on the private fork used to prepare security releases – GH only lets public repo use GHA for free.

bmuenzenmeyer:
> We do on the private fork used to prepare security releases – GH only lets public repo use GHA for free.

Yikes. That seems like a tax on security. I did some digging and found this explicitly mentioned on a Discussion Ticket: [Investing in the security advisory experience on GitHub](https://github.com/orgs/community/discussions/189802)

> Enabling CI on advisory workspace private forks

> One of the longest-standing requests we hear is the inability to run GitHub Actions on the temporary private forks created for security advisories. Right now, maintainers can't run their CI pipeline against a security fix before merging, which means you're either merging untested patches or maintaining a separate private fork workflow outside of our advisory tooling.

> We're actively working through the security model required to enable Actions on these workspaces safely. The core challenge is ensuring that embargoed vulnerability details can't leak through webhook payloads, third-party app integrations, or untrusted workflow execution. This is complex work with real security implications, and we're being deliberate about getting it right - the integrity of the embargo model is foundational to the trust that makes coordinated disclosure work.

PR #8726, opened 2026-03-16
docs(learn): expand type stripping explanation and constraints

ibiraza1077-pixel:
## Description

Expands the `run-natively` page to address the request for a more in-depth conceptual explanation of Node.js type stripping, as requested in nodejs/learn#34.

Changes:
- Adds a "What is Type Stripping?" section explaining the concept and how it differs from transpilation
- Adds a version support table covering v22.6.0 through v23.6.0
- Completes the Constraints section, which previously contained an incomplete sentence with no list

No existing content was removed or restructured.

## Validation

This is a documentation-only change. Reviewers can verify by previewing the markdown renders correctly and checking the version numbers against the [Node.js changelog](https://nodejs.org/en/blog/release).

## Related Issues

Addresses nodejs/learn#34

### Check List

### DIFF
--- a/apps/site/pages/en/learn/typescript/run-natively.md
+++ b/apps/site/pages/en/learn/typescript/run-natively.md
@@ -9,13 +9,11 @@ authors: AugustinMauroy, khaosdoctor, jakebailey, robpalme
 You can write code that's valid TypeScript directly in Node.js without the need to transpile it first.
 
 If you are using v22.18.0 or later and your source code contains only [erasable TypeScript syntax](https://devblogs.microsoft.com/typescript/announcing-typescript-5-8-beta/#the---erasablesyntaxonly-option), you can execute TypeScript code without any flags.
-
 ```bash
 node example.ts
 ```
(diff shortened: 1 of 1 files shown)


avivkeller (changes requested):
This has several unrelated changes, and fails to provide any additional information that isn't mentioned previously.

Simply re-stating information doesn't resolve any open issues.

ibiraza1077-pixel:
Thank you for the feedback. I understand — the additions were 
accurate but largely restated information available elsewhere. 

Could you point me toward what genuinely new content would be 
most valuable to add? For example, would an explanation of why 
Node.js chose type stripping over full compilation, or guidance 
on native running vs transpilation tradeoffs, be the right 
direction?

JakobJingleheimer:
I think the info is scattered around. It would likely take a lot of digging to find (and would be faster for someone who's already familiar to make the updates—probably me or Marco).

I think this is not a "good first issue" 😅

PR #8728, opened 2026-03-16
feat(ci): add zizmor

avivkeller:
Ref: https://github.com/nodejs/web-team/issues/113
Ref: https://openjs-foundation.slack.com/archives/C0ALRN98G8K/p1773689960844809 (🔒)

### DIFF
--- a/.github/workflows/zizmor.yml
+++ b/.github/workflows/zizmor.yml
@@ -1,4 +1,4 @@
-name: GitHub Actions Security Analysis with zizmor 🌈
+name: Zizmor
 
 on:
   push:
@@ -10,10 +10,17 @@ on:
 
 permissions: {}
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: ${{ github.event_name == 'pull_request' }}
+
 jobs:
   zizmor:
+    name: Zizmor Security Analysis
     runs-on: ubuntu-latest
     permissions:
+      contents: read
+      actions: read
       security-events: write
     steps:
       - name: Checkout repository
--- a/.github/workflows/zizmor.yml
+++ b/.github/workflows/zizmor.yml
@@ -1,0 +1,25 @@
+name: GitHub Actions Security Analysis with zizmor 🌈
+
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+    branches:
+      - main
+
+permissions: {}
+
+jobs:
+  zizmor:
+    runs-on: ubuntu-latest
+    permissions:
+      security-events: write
+    steps:
+      - name: Checkout repository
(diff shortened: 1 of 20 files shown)


flakey5 (approved):
lgtm other than https://github.com/nodejs/nodejs.org/pull/8728/changes#r2943202991

MattIPv4:
🤔 I don't see this having run on this PR? I would imagine there will be a lot of changes to other workflows required before Zizmor can pass?

.github/workflows/zizmor.yml
```diff
@@ -0,0 +1,25 @@
+name: GitHub Actions Security Analysis with zizmor 🌈
```
MattIPv4: This seems like an excessively long title to show up in our PR checks?

avivkeller:
> 🤔 I don't see this having run on this PR? I would imagine there will be a lot of changes to other workflows required before Zizmor can pass?

I copied this workflow from Zizmor's documentation, I can, of course, remove that trigger.

Im not sure if it will fail PRs, or just report them as having security concerns.

Perhaps I should push another commit with the fixes to the issues it would find?

MattIPv4:
We're also going to want a branch ruleset change I imagine to block PRs where we have Zizmor failings if we're using the advanced security mode?

MattIPv4:
> Im not sure if it will fail PRs, or just report them as having security concerns.

It will just report them unless we add a new rule. _But_, I don't see it running at all? It would show up as a passing Actions check still as it is being run on the PR trigger?

avivkeller:
> > Im not sure if it will fail PRs, or just report them as having security concerns.
> 
> It will just report them unless we add a new rule. _But_, I don't see it running at all? It would show up as a passing Actions check still as it is being run on the PR trigger?

It won't run on this PR, since it's the one that added it (GitHub is weird that way)

ovflowd (approved):
SGTM

avivkeller:
@nodejs/web-admins remember to add the action to the allowlist

MattIPv4:
> It won't run on this PR, since it's the one that added it (GitHub is weird that way)

This is false... When you add a new workflow with a `pull_request` trigger, it _will_ run on the PR that introduces it (assuming the PR isn't a fork).

> @nodejs/web-admins remember to add the action to the allowlist

This is why the workflow hasn't run, if you'd actually checked...

image

https://github.com/nodejs/nodejs.org/actions/runs/23171995051

Let's actually check the workflow runs and works before we blindly merge the change, please.

MattIPv4:
I've added the action to the list. If you can push a change to cause the CI to run again, that'd be appreciated so we can actually validate the workflow is going to work.

avivkeller:
> This is why the workflow hasn't run, if you'd actually checked...

I'm sorry I was mistaken previously

aymen94 (approved):

ovflowd (approved):

PR #8730, opened 2026-03-17
fix(ui-components): guard ThemeToggle against invalid stored theme

### LINKED ISSUE #8729: ThemeToggle crashes with "Element type is invalid" when localStorage theme value is not a valid Theme
### URL:

https://nodejs.org/en

### Browser Name:

Brave

### Browser Version:

1.88.132

### Operating System:

Windows 11, 25H2, 26200.8037

### Environment
- Next.js: 16.1.6 (Turbopack)
- OS: Windows
- Node: v24.14.0

Image

Image

### Steps to reproduce
1. `pnpm install --frozen-lockfile && pnpm dev`
2. Open `http://localhost:3000`
3. In DevTools console run:
   localStorage.setItem('theme', 'invalid-theme'); location.reload();

### Actual
Runtime crash: "Element type is invalid: expected a string (for built-in 
components) or a class/function (for composite components) but got: undefined.
Check the render method of `ThemeToggle`."

### Expected
Graceful fallback to `'system'` theme without crashing.

### Root cause
`themeIcons[currentTheme]` in ThemeToggle/index.tsx returns `undefined` 
for any value outside `'system' | 'light' | 'dark'`. This undefined is 
then used as a JSX element type, causing the React error.

The fix would be to normalize `currentTheme` before the icon lookup:
  const normalizedTheme = currentTheme in themeIcons ? currentTheme : 'system';

shivxmsharma:
## Description

This PR fixes a runtime crash in `ThemeToggle` when `currentTheme` contains an unexpected value.

Previously, the component assumed `currentTheme` was always one of `system | light | dark`. If a stale or invalid value was persisted in `localStorage` (for example, `invalid-theme`), `themeIcons[currentTheme]` resolved to `undefined`, causing React to throw:

`Element type is invalid: expected a string (for built-in components) or a class/function (for composite components) but got: undefined.`

The fix normalizes `currentTheme` to `system` when the value is not a known key, and uses the normalized value for both icon lookup and active item highlighting.

## Validation

I validated the fix with the following steps:

1. Run the site locally.
2. In browser console, execute:

```js
localStorage.setItem('theme', 'invalid-theme'); location.reload();
```

3. Before fix: app crashes in `ThemeToggle` with "Element type is invalid".
4. After fix: no crash; theme toggle renders and safely falls back to `system` behavior.

## Related Issues

Fixes #8729

### Check List

### DIFF
--- a/packages/ui-components/src/Common/ThemeToggle/index.tsx
+++ b/packages/ui-components/src/Common/ThemeToggle/index.tsx
@@ -33,7 +33,9 @@ const ThemeToggle: FC<ThemeToggleProps> = ({
   ariaLabel,
   themeLabels = { system: 'System', light: 'Light', dark: 'Dark' },
 }) => {
-  const TriggerIcon = themeIcons[currentTheme];
+  const normalizedTheme: Theme =
+    currentTheme in themeIcons ? currentTheme : 'system';
+  const TriggerIcon = themeIcons[normalizedTheme];
 
   return (
     <DropdownMenu.Root>
@@ -60,7 +62,7 @@ const ThemeToggle: FC<ThemeToggleProps> = ({
                 key={theme}
                 onClick={() => onChange(theme)}
                 className={classNames(styles.dropDownItem, {
-                  [styles.activeItem]: theme === currentTheme,
+                  [styles.activeItem]: theme === normalizedTheme,
                 })}
               >
                 <Icon height="16" />


avivkeller:
I don't think we need a type guard here, we _know_ that it's set by us

shivxmsharma:
Thanks for the review, that makes sense. If invalid localStorage theme values are considered out of scope and the project prefers relying on the controlled write path, I’m happy to close this out.

PR #8746, opened 2026-03-18
feat(remark-lint): improve validation

avivkeller:
Ref: https://github.com/nodejs/node/pull/62090

---

Adds deprecation validation based on https://github.com/nodejs/node/blob/main/tools/doc/deprecationCodes.mjs, and allows type validation to work in links (i.e. `[<Type>]()`)

### DIFF
--- a/packages/remark-lint/src/rules/invalid-deprecations.mjs
+++ b/packages/remark-lint/src/rules/invalid-deprecations.mjs
@@ -1,0 +1,69 @@
+import { lintRule } from 'unified-lint-rule';
+import { visit } from 'unist-util-visit';
+
+const DEPRECATION_COMMENT = /^<!-- md-lint skip-deprecation (DEP\d{4}) -->$/;
+const DEPRECATION_HEADING = /^(DEP\d{4}):/;
+const DEPRECATION_YAML = /^<!-- YAML\r?\nchanges:\r?\n/;
+
+const generateDeprecation = code => `DEP${code.toString().padStart(4, '0')}`;
+
+/**
+ * Ensures all needed metadata exists
+ * @type {import('unified-lint-rule').Rule}
+ */
+const invalidDeprecations = (tree, vfile) => {
+  if (vfile.stem !== 'deprecations') {
+    // Skip non-deprecation files
+    return;
+  }
+
+  let expectedDeprecationCode = 1;
+
+  visit(tree, ['heading', 'html'], (node, idx, parent) => {
+    // Get the current deprecation
+    const deprecationCode =
+      node.type === 'html'
+        ? node.value.match(DEPRECATION_COMMENT)?.[1]
+        : node.children[0].value.match(DEPRECATION_HEADING)?.[1];
+
+    if (node.type === 'heading') {
+      const typeNode = parent.children[idx + 1]?.children[0];
+
+      if (!typeNode?.value?.startsWith('Type:')) {
+        vfile.message(
+          `Deprecation "${deprecationCode}" is missing a "Type"`,
+          node
+        );
+      }
+
+      const changesNode = parent.children[idx + 2];
+
+      if (!DEPRECATION_YAML.test(changesNode?.value)) {
+        vfile.message(
+          `Deprecation "${deprecationCode}" is missing changes`,
+          node
+        );
+      }
+    }
+
+    // If not found, skip
+    if (!deprecationCode) {
+      return;
+    }
+
+    const generatedDeprecationCode = generateDeprecation(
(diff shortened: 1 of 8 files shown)


packages/remark-lint/src/rules/invalid-deprecations.mjs
```diff
@@ -0,0 +1,65 @@
...
+const DEPRECATION_YAML = /^<!-- YAML\r?\nchanges:\r?\n/;
+
+const generateDeprecation = code => `DEP${code.toString().padStart(4, '0')}`;
+
+/**
+ * Ensures all needed metadata exists
+ * @type {import('unified-lint-rule').Rule}
+ */
+const invalidDeprecations = (tree, vfile) => {
+  if (vfile.basename !== 'deprecations') {
```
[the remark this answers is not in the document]
avivkeller:
```suggestion
  if (vfile.stem !== 'deprecations') {
```

packages/remark-lint/src/rules/invalid-deprecations.mjs
```diff
@@ -0,0 +1,65 @@
...
+    return;
+  }
+
+  let expectedDeprecationCode = 1;
+
+  visit(tree, ['heading', 'html'], (node, idx, parent) => {
+    let currentDeprecationCode;
+
+    // Get the current deprecation
+    if (node.type === 'html') {
```
ovflowd: Can you explain what it really does with? `currentDeprecationCode = node.value.match(DEPRECATION_COMMENT)?.[1];` it seems like that's supposed to be when we skip a deprecation note?

avivkeller: Yes. This matches when we skip a deprecation code, and ensures that the code we skip is the next code in line (1, 2, etc).

ovflowd: Could you then better name this variable/constant or add some inline comments?

packages/remark-lint/src/rules/invalid-deprecations.mjs
```diff
@@ -0,0 +1,65 @@
...
+      currentDeprecationCode = node.value.match(DEPRECATION_COMMENT)?.[1];
+    } else {
```
ovflowd: Since the majority of the content is on the else, could maybe the if be completely avoided and be transformed into a ternary? Or have the else content on the body of if? (switching)

ovflowd (approved):
Ultimately, fine with the PR. Just wondering if it could be documented a bit more, like here `type[0] !== '{' || type[type.length - 1] !== '}'` you documented it of sorts but on the lines below not much.

packages/remark-lint/src/rules/invalid-deprecations.mjs
```diff
@@ -0,0 +1,65 @@
...
+          node
+        );
+      }
+
+      if (!DEPRECATION_YAML.test(parent.children[idx + 2]?.value)) {
+        vfile.message(
+          `Deprecation "${currentDeprecationCode}" is missing changes`,
+          node
+        );
+      }
```
canerakdas: We can improve readability by giving these nodes more meaningful names.

For example;

```mjs
const typeNode = parent.children[idx + 1]?.children[0];
const changesNode = parent.children[idx + 2];

if (!typeNode?.value?.startsWith('Type:')) {
  vfile.message(
    `Deprecation "${currentDeprecationCode}" is missing a "Type"`,
    node
  );
}

if (!DEPRECATION_YAML.test(changesNode?.value)) {
  vfile.message(
    `Deprecation "${currentDeprecationCode}" is missing changes`,
    node
  );
}
```

packages/remark-lint/src/rules/invalid-deprecations.mjs
```diff
@@ -0,0 +1,65 @@
...
+    }
+
+    if (
+      currentDeprecationCode !== generateDeprecation(expectedDeprecationCode)
+    ) {
+      vfile.message(
+        `Deprecation codes are out of order. Expected ${generateDeprecation(expectedDeprecationCode)}, saw "${currentDeprecationCode}"`,
+        node
+      );
+    }
```
canerakdas:
```suggestion
    const generatedDeprecationCode = generateDeprecation(expectedDeprecationCode);

    if (currentDeprecationCode !== generatedDeprecationCode) {
      vfile.message(
        `Deprecation codes are out of order. Expected ${generatedDeprecationCode}, saw "${currentDeprecationCode}"`,
        node
      );
    }
```

canerakdas (approved):

PR #8801, opened 2026-04-05
fix(ui): make mobile nav toggle keyboard accessible

### LINKED ISSUE #7895: The toggle navigation menu button is not keyboard accessible

abhijeetnardele24-hash:
## Description

Replaces the mobile navigation toggle's hidden-checkbox + label pattern with a real button so the control is focusable and exposed with the correct button semantics to keyboard and assistive technology users.

## Changes

- replace the `Label`-driven mobile toggle with a native `button`
- add `aria-controls` and `aria-expanded` to the toggle
- remove the now-unused hidden checkbox styling
- add a focused NavBar test covering the accessible toggle state

## Validation

- `node --enable-source-maps --import=global-jsdom/register --import=tsx --import=../../tests/setup.mjs --test src/Containers/NavBar/__tests__/index.test.jsx`
- `corepack pnpm exec prettier --check packages/ui-components/src/Containers/NavBar/index.tsx packages/ui-components/src/Containers/NavBar/index.module.css packages/ui-components/src/Containers/NavBar/__tests__/index.test.jsx`

Fixes #7895

### DIFF
--- a/packages/ui-components/src/Containers/NavBar/__tests__/index.test.jsx
+++ b/packages/ui-components/src/Containers/NavBar/__tests__/index.test.jsx
@@ -1,0 +1,54 @@
+import { describe, it } from 'node:test';
+import assert from 'node:assert/strict';
+
+import { render, screen } from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
+
+import NavBar from '..';
+
+const Logo = () => <span>Node.js</span>;
+
+describe('NavBar', () => {
+  it('uses a keyboard-accessible button to toggle the mobile navigation', async () => {
+    const user = userEvent.setup();
+
+    render(
+      <NavBar
+        as="a"
+        Logo={Logo}
+        pathname="/"
+        sidebarItemTogglerAriaLabel="Toggle navigation menu"
+        navItems={[
+          { text: 'Learn', link: '/learn' },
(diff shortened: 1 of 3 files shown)


packages/ui-components/src/Containers/NavBar/__tests__/index.test.jsx
avivkeller, line 1: This test seems unnecessary

bmuenzenmeyer (changes requested):
Functionally this represents an improvement - tested in Mac Chrome and Safari. Especially considering our mobile screen shows on rather large viewports (split screen mac book air, easily) this would be good.

I agree, to land this we should refactor the classname as copilot suggests, and...yes, as @avivkeller mentions in most cases we want robust test coverage - but if we are going to bother here I'd rather we use playwright... which actually we have already. 

https://github.com/nodejs/nodejs.org/blob/main/apps/site/tests/e2e/general-behavior.spec.ts#L142-L165

We should be able to pretty easily extend this not for a mobile viewport but a small viewport keyboard use case.

Issue #8699, opened 2026-02-08
Make NavBar Re-appear on Upward Scroll


constb:
### Enter your suggestions in details:

While reading a docs page, I decided to switch to the dark mode. But the mode button is not in the right section, it's in the navigation and **navigation is hidden** unless I scroll to the very top of the page, loosing the position I was reading at.

One possible solution would be to move mode switch to the right section (where ToC is). Another is to make navigation reappear whenever user starts scrolling upwards.

Other options to improve mode switching that come to mind are:
1. Add the "system" mode option that follows the light/dark OS setting, like the one is MacOS.
2. Add a hotkey to switch mode.

avivkeller:
> One possible solution would be to move mode switch to the right section (where ToC is). Another is to make navigation reappear whenever user starts scrolling upwards.

+1

codebyanas:
Hi @avivkeller,

I noticed that PR #8808 was closed. I would like to work on issue #8699 and implement the scroll-up-to-reveal navigation bar functionality.

Could you please assign this issue to me?

PR #8808, opened 2026-04-07
feat: make navbar re-appear on upward scroll

shivxmsharma:
## Description

This PR implements a "scroll-up-to-reveal" behavior for the navigation bar on the website, giving desktop users more screen real estate while scrolling through content.

When scrolling **down**, the navbar slides out of view. When scrolling **up** (or returning to the top of the page), it smoothly slides back into view.

**Key implementation details:**
- Added a new `useScrollDirection` client hook that uses `requestAnimationFrame` for high-performance, flicker-free scroll tracking
- Added a `SCROLL_THRESHOLD` of 10px to avoid direction flips on micro-scrolls
- Integrated the hook into `WithNavBar` to apply a `hidden` CSS class when scrolling down
- The sticky + hide behavior is scoped to the **`xl` breakpoint (≥1280px) only** to avoid side effects on the mobile navigation layout
- CSS transitions use `transform: translateY(-100%)` for smooth, hardware-accelerated animations

## Validation

- On a **desktop viewport**, scrolling down causes the navbar to slide up and out of view
- Scrolling back up reveals the navbar with a smooth transition
- At `scrollY === 0`, the navbar is always visible regardless of previous scroll direction
- On **mobile/tablet viewports**, the navbar behaves exactly as before — no regression

## Related Issues

Addresses #8699

### Check List

### DIFF
--- a/apps/site/components/withNavBar.tsx
+++ b/apps/site/components/withNavBar.tsx
@@ -8,6 +8,7 @@ import NavBar from '@node-core/ui-components/Containers/NavBar';
 import styles from '@node-core/ui-components/Containers/NavBar/index.module.css';
 import GitHubIcon from '@node-core/ui-components/Icons/Social/GitHub';
 import { availableLocales } from '@node-core/website-i18n';
+import classNames from 'classnames';
 import dynamic from 'next/dynamic';
 import { useLocale, useTranslations } from 'next-intl';
 import { useTheme } from 'next-themes';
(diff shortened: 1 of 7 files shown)


bmuenzenmeyer:
this is not working for me, as far as i can tell. i am assuming my laptop is xl, but it may not be i suppose.

edit: oh i see what's happening, the nav learn content is now going to nodejs.org/learn not the forked copy.

there is a small bug in that if you:

- scroll down to hide the nav bar
- scroll back up to show it
- the nav bar may clip the tops of left and right secondary navs, and the only way to restore them is to scroll all the way back up

PR #8817, opened 2026-04-14
chore: remove Active/Maintenance LTS distinction

### LINKED ISSUE #8277: Node 24.x not yet marked as LTS ("Krypton") in index.json (website lists v24.10.0 as Latest LTS)
### Version

Not applicable – issue concerns Node.js distribution metadata, not runtime behavior.

### Platform

```text
All platforms (observed on Windows, but affects all because index.json is global).
```

### Subsystem

Release / Distribution metadata

### What steps will reproduce the bug?

1. Visit https://nodejs.org/dist/index.json
2. Search for any entries for version v24.x.x
   → all of them have `"lts": false`.
3. Visit https://nodejs.org – it lists v24.10.0 as “Latest LTS”.
4. Tools that rely on index.json (Volta, nvm, fnm, etc.) still treat Node 22 ("Jod") as the current LTS.

### How often does it reproduce? Is there a required condition?

Always, until the index.json is updated to include `"lts": "Krypton"` for Node 24 entries.

### What is the expected behavior? Why is that the expected behavior?

When Node 24 is promoted to LTS (codename "Krypton"), each v24.x entry in index.json
should have an `"lts"` field set to `"Krypton"` rather than `false`.

Tools and automation use this field to determine the latest LTS line,
so keeping it accurate ensures consistency between nodejs.org and the dist metadata.

### What do you see instead?

All v24.x entries in index.json currently show `"lts": false`,
while the homepage lists v24.10.0 as “Latest LTS”.
The most recent version marked as LTS in index.json is still 22.21.0 ("Jod").

### Additional information

Date checked: 2025-10-28

Example entry from https://nodejs.org/dist/index.json:
{
  "version": "v24.10.0",
  "date": "2025-10-14",
  "lts": false
}

Expected:
{
  "version": "v24.10.0",
  "date": "2025-10-14",
  "lts": "Krypton"
}

This discrepancy causes tools such as Volta, nvm, and fnm — which rely on index.json
to determine the current LTS — to continue installing Node 22 instead of 24.

This seems like a metadata lag during the LTS promotion.
Just filing it here in case the release metadata hasn’t yet been updated.

Thanks for maintaining Node.js and for your work on the release process 🙏

MattIPv4:
## Description

In preparation for the change to how Node.js will be shipping releases, this removes the Active/Maintenance distinction for LTS release lines in the website's logic. This distinction was never shown to users; it was only used in internal logic and was always just reported as LTS to the user viewing the site.

Also, I've updated the version dropdown on the downloads page to make use of badges to make it clearer what status a given release has, as during the conversation around this change, it was flagged that it is odd we show EoL releases there and it isn't super obvious that they're EoL.

| Before | After |
|-|-|
| image | image |

## Validation

All release lines continue to report their correct statuses and dates on the site.

## Related Issues

I think we can say that this finally closes #8277.

### Check List

### DIFF
--- a/apps/site/components/Downloads/Release/VersionDropdown.tsx
+++ b/apps/site/components/Downloads/Release/VersionDropdown.tsx
@@ -5,36 +5,13 @@ import { useLocale, useTranslations } from 'next-intl';
 import { use } from 'react';
 
 import { redirect, usePathname } from '#site/navigation';
+import { STATUS_KIND_MAP } from '#site/next.constants.mjs';
 import {
   ReleaseContext,
   ReleasesContext,
 } from '#site/providers/releaseProvider';
 
-import type { NodeReleaseStatus } from '#site/types/releases.js';
 import type { FC } from 'react';
-
-const getDropDownStatus = (status: NodeReleaseStatus) => {
-  if (status === 'LTS') {
-    return {
-      label: 'LTS',
-      kind: 'info' as const,
-    };
-  }
-
-  if (status === 'Current') {
-    return {
-      label: 'Current',
-      kind: 'default' as const,
-    };
-  }
-
-  if (status === 'End-of-life') {
-    return {
-      label: 'EoL',
-      kind: 'warning' as const,
-    };
-  }
-};
 
 const VersionDropdown: FC = () => {
   const { releases } = use(ReleasesContext);
@@ -69,7 +46,7 @@ const VersionDropdown: FC = () => {
       values={releases.map(({ status, versionWithPrefix }) => ({
         value: versionWithPrefix,
         label: versionWithPrefix,
-        badge: getDropDownStatus(status),
+        badge: { label: status, kind: STATUS_KIND_MAP[status] },
       }))}
       defaultValue={release.versionWithPrefix}
       onChange={setVersionOrNavigate}
--- a/apps/site/components/Downloads/Release/VersionDropdown.tsx
+++ b/apps/site/components/Downloads/Release/VersionDropdown.tsx
@@ -13,20 +13,27 @@ import {
 import type { NodeReleaseStatus } from '#site/types/releases.js';
 import type { FC } from 'react';
 
-const getDropDownStatus = (version: string, status: NodeReleaseStatus) => {
+const getDropDownStatus = (status: NodeReleaseStatus) => {
   if (status === 'LTS') {
-    return `${version} (LTS)`;
+    return {
+      label: 'LTS',
+      kind: 'info' as const,
+    };
   }
 
   if (status === 'Current') {
-    return `${version} (Current)`;
+    return {
+      label: 'Current',
+      kind: 'default' as const,
+    };
   }
 
   if (status === 'End-of-life') {
-    return `${version} (EoL)`;
+    return {
+      label: 'EoL',
+      kind: 'warning' as const,
+    };
   }
-
-  return version;
 };
 
 const VersionDropdown: FC = () => {
@@ -61,7 +68,8 @@ const VersionDropdown: FC = () => {
       ariaLabel={t('layouts.download.dropdown.version')}
       values={releases.map(({ status, versionWithPrefix }) => ({
         value: versionWithPrefix,
-        label: getDropDownStatus(versionWithPrefix, status),
+        label: versionWithPrefix,
+        badge: getDropDownStatus(status),
       }))}
       defaultValue={release.versionWithPrefix}
       onChange={setVersionOrNavigate}
--- a/apps/site/components/Downloads/Release/VersionDropdown.tsx
+++ b/apps/site/components/Downloads/Release/VersionDropdown.tsx
@@ -10,15 +10,20 @@ import {
   ReleasesContext,
 } from '#site/providers/releaseProvider';
 
+import type { NodeReleaseStatus } from '#site/types/releases.js';
 import type { FC } from 'react';
 
-const getDropDownStatus = (version: string, status: string) => {
-  if (status.endsWith('LTS')) {
+const getDropDownStatus = (version: string, status: NodeReleaseStatus) => {
+  if (status === 'LTS') {
     return `${version} (LTS)`;
   }
 
   if (status === 'Current') {
     return `${version} (Current)`;
+  }
+
+  if (status === 'End-of-life') {
+    return `${version} (EoL)`;
   }
 
   return version;
@@ -38,7 +43,7 @@ const VersionDropdown: FC = () => {
       ({ versionWithPrefix }) => versionWithPrefix === version
     );
 
-    if (release?.isLts && pathname.includes('current')) {
+    if (release?.status === 'LTS' && pathname.includes('current')) {
       redirect({ href: '/download', locale });
       return;
     }
--- a/apps/site/next-data/generators/releaseData.mjs
+++ b/apps/site/next-data/generators/releaseData.mjs
@@ -3,31 +3,18 @@
 import getMajorNodeReleases from './majorNodeReleases.mjs';
 
 // Gets the appropriate release status for each major release
-const getNodeReleaseStatus = (latest, support) => {
+const getNodeReleaseStatus = (latest, eol) => {
   const now = new Date();
-  const { endOfLife, maintenanceStart, ltsStart, currentStart } = support;
 
-  if (endOfLife && now >= new Date(endOfLife)) {
+  if (eol && now >= new Date(eol)) {
     return 'End-of-life';
   }
 
-  if (
-    latest.lts.isLts &&
-    maintenanceStart &&
-    now >= new Date(maintenanceStart)
-  ) {
-    return 'Maintenance LTS';
+  if (latest.lts.isLts) {
+    return 'LTS';
   }
 
-  if (latest.lts.isLts && ltsStart && now >= new Date(ltsStart)) {
-    return 'Active LTS';
-  }
-
-  if (currentStart && now >= new Date(currentStart)) {
-    return 'Current';
-  }
-
-  return 'Pending';
+  return 'Current';
 };
 
 /**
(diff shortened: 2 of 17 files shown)


apps/site/components/Downloads/Release/VersionDropdown.tsx
```diff
@@ -10,17 +10,22 @@ import {
...
+const getDropDownStatus = (version: string, status: NodeReleaseStatus) => {
+  if (status === 'LTS') {
     return `${version} (LTS)`;
   }
 
   if (status === 'Current') {
     return `${version} (Current)`;
   }
 
+  if (status === 'End-of-life') {
```
ovflowd: Why we need a new (EoL) label and where would that be shown?

apps/site/types/releases.ts
```diff
@@ -1,21 +1,13 @@
...
   version: string;
   codename?: string;
-  currentStart: string;
-  ltsStart?: string;
-  maintenanceStart?: string;
-  endOfLife: string;
   npm?: string;
   v8: string;
   releaseDate: string;
+  initialDate: string;
```
aduh95: Is this relevant?

MattIPv4: https://nodejs-org-git-mattipv4-remove-active-maintenance-lts-openjs.vercel.app/en/download/archive/v25.9.0 shows the initial release date for each major, just keeping that page rendering as it was before.

aduh95: This might need some further change with the introduction of Alpha releases, or at least some clarification in the UI – but I guess it's not related to this PR

MattIPv4: Yeh, totally agree, I don't think it is super clear today either -- it can easily be read as the "first release" + "last updated" for the specific version you're looking at, not for the major as a whole.

Would you mind creating an issue to track this, @aduh95?

apps/site/scripts/orama-search/get-documents.mjs
```diff
@@ -13,17 +13,15 @@ const fetchOptions = process.env.GITHUB_TOKEN
...
- * for the current Active LTS version.
+ * for the current LTS version.
  */
 export const getAPIDocs = async () => {
   // Find the current Active LTS version
   const releaseData = await generateReleaseData();
-  const ltsRelease =
-    releaseData.find(r => r.status === 'Active LTS') ||
-    releaseData.find(r => r.status === 'Maintenance LTS');
+  const ltsRelease = releaseData.find(r => r.status === 'LTS');
```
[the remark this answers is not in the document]
MattIPv4: The default ordering will be correct.

avivkeller (approved):

apps/site/next-data/generators/releaseData.mjs
```diff
@@ -3,31 +3,18 @@
...
+    return 'EOL';
   }
 
-  if (
-    latest.lts.isLts &&
-    maintenanceStart &&
-    now >= new Date(maintenanceStart)
-  ) {
-    return 'Maintenance LTS';
+  if (latest.lts.isLts) {
```
ovflowd, line 13: Didn't you remove the lts check?

apps/site/components/Downloads/Release/VersionDropdown.tsx
```diff
@@ -56,7 +45,11 @@ const VersionDropdown: FC = () => {
       ariaLabel={t('layouts.download.dropdown.version')}
       values={releases.map(({ status, versionWithPrefix }) => ({
         value: versionWithPrefix,
-        label: getDropDownStatus(versionWithPrefix, status),
+        label: versionWithPrefix,
+        badge: {
```
ovflowd: **nit:** can be inlined

```suggestion
        badge: { label: status, kind: STATUS_KIND_MAP[status] }
```

apps/site/next-data/generators/releaseData.mjs
```diff
@@ -62,16 +47,15 @@ const generateReleaseData = async () => {
...
       status,
       major: latestVersion.semver.major,
       version: latestVersion.semver.raw,
       versionWithPrefix: `v${latestVersion.semver.raw}`,
       codename: major.support.codename || '',
-      isLts: status.endsWith('LTS'),
       npm: latestVersion.dependencies.npm || '',
       v8: latestVersion.dependencies.v8,
       releaseDate: latestVersion.releaseDate,
+      initialDate: initialVersion.releaseDate,
```
ovflowd: Wthat's the diff of releaseDate and initialDate? Is it the first release of said major or? Can we call it firstReleaseDate or initialReleaseDate?

apps/site/types/releases.ts
```diff
@@ -1,21 +1,13 @@
-export type NodeReleaseStatus =
-  | 'Active LTS'
-  | 'Maintenance LTS'
-  | 'Current'
-  | 'End-of-life'
-  | 'Pending';
+export type NodeReleaseStatus = 'LTS' | 'Current' | 'EOL';
```
ovflowd, line 1: Aren't they adding a new "Alpha" version/status?

packages/ui-components/src/Common/Select/StatelessSelect/index.tsx
```diff
@@ -66,6 +68,15 @@ const StatelessSelect = <T extends string>({
             <span className={styles.selectedValue}>
               {currentItem.iconImage}
               <span>{currentItem.label}</span>
+              {currentItem.badge && (
```
ovflowd, line 71: Screenshot of how it looks? Or a Story?

packages/ui-components/src/Common/Select/index.tsx
ovflowd, line 1: These Badge changes seem kinda unrelated to the LTS work, should this be on its own PR? I'm fine keeping here, but would be good updating the PR title/description to include such mentions.

ovflowd (approved):
SGTM! Left a few tiny comments and I just want @MattIPv4 to triple check the PR diff to see if he missed anything.

packages/ui-components/src/Common/Select/index.tsx
MattIPv4: > Also, I've updated the version dropdown on the downloads page to make use of badges to make it clearer what status a given release has, as during the conversation around this change, it was flagged that it is odd we show EoL releases there and it isn't super obvious that they're EoL.

Already in the description -- didn't include it in the title as it seemed like a minor additional change and would make the title rather long, it still relates to cleaning up our version statuses.

apps/site/components/Downloads/Release/VersionDropdown.tsx
MattIPv4: In the version dropdown on the downloads page.

apps/site/next-data/generators/releaseData.mjs
MattIPv4: We still need to check for whether a release is LTS, from the actual release data. This PR just removes the Active/Maintenance distinction which was being derived from the release schedule dates.

apps/site/components/Downloads/Release/VersionDropdown.tsx
MattIPv4: Interesting that Prettier didn't enforce this 🤔

apps/site/next-data/generators/releaseData.mjs
ovflowd: Right, Im a bit confused by the logic, maybe because `lts: { isLts: false },` is redundant. Can't we simplify this object?

apps/site/next-data/generators/releaseData.mjs
```diff
@@ -3,31 +3,18 @@
...
 // Gets the appropriate release status for each major release
-const getNodeReleaseStatus = (latest, support) => {
+const getNodeReleaseStatus = (latest, eol) => {
   const now = new Date();
-  const { endOfLife, maintenanceStart, ltsStart, currentStart } = support;
 
-  if (endOfLife && now >= new Date(endOfLife)) {
-    return 'End-of-life';
+  if (eol && now >= new Date(eol)) {
+    return 'EOL';
```
ovflowd, line 10: Can't we keep it End-of-Life? At the very least:


```suggestion
    return 'EoL';
```

apps/site/next-data/generators/releaseData.mjs
MattIPv4: Sure.

apps/site/types/releases.ts
MattIPv4: I think so yes, but I didn't worry about that for this PR, it's not clear yet how we'd even detect that unless a new property is being introduced to the release data.

packages/ui-components/src/Common/Select/StatelessSelect/index.tsx
MattIPv4: Added to the description.

apps/site/next-data/generators/releaseData.mjs
MattIPv4: We already used `EOL` capitalized in most places, including the canonical page (https://nodejs.org/en/about/eol), but I can switch to `EoL` if you'd prefer that?

apps/site/next-data/generators/releaseData.mjs
MattIPv4: We don't control that object, it comes from nodevu

apps/site/next-data/generators/releaseData.mjs
ovflowd: Weird, I feel we should use EoL as it is the proper capitalization of End-of-Life.

MattIPv4: I think maybe we should do that as a separate change? It'll involve updating the page, and in theory, a lot of existing blog posts etc. if we want to be consistent the other way (there are 297 ocurrences of `EOL` in this branch).

ovflowd: We don't need to change existing blog posts, just our core pages, and those should be small results IMO.

MattIPv4: After talking on Slack, even just core pages results in ~15 files touched to change this, plus another ~40 if translations are included. Going to do that separately.

PR #8819, opened 2026-04-14
meta: add next-env.d.ts to .gitignore

MattIPv4:
## Description

Folks frequently run into diffs being generated on `next-env.d.ts` locally when they run the site, and we've flip-flopped on the checked-in values a few times. I think it is easier to just ensure this file is never checked in, so it can contain whatever Next.js needs to generate for a person's local setup.

Next.js itself also recommends that this value is not checked in: https://nextjs.org/docs/app/api-reference/config/typescript#next-envdts + https://nextjs.org/docs/app/api-reference/cli/next#next-typegen-options

If you run `next typegen` or `next build`, you get this:
```ts
/// <reference types="next" />
/// <reference types="next/image-types/global" />
/// <reference types="next/navigation-types/compat/navigation" />
import "./.next/types/routes.d.ts";

// NOTE: This file should not be edited
// see https://nextjs.org/docs/app/api-reference/config/typescript for more information.
```

But if you run `next dev`, you get this:
```ts
/// <reference types="next" />
/// <reference types="next/image-types/global" />
/// <reference types="next/navigation-types/compat/navigation" />
import "./.next/dev/types/routes.d.ts";

// NOTE: This file should not be edited
// see https://nextjs.org/docs/app/api-reference/config/typescript for more information.
```

The `import` line changes...

## Validation

File no longer checked in. CI still passes for type-checking.

## Related Issues

N/A

### Check List

### DIFF
--- a/apps/site/package.json
+++ b/apps/site/package.json
@@ -21,7 +21,7 @@
     "lint:js:fix": "node --run lint:js -- --fix",
     "lint:md": "eslint \"**/*.md?(x)\" --cache --cache-strategy=content --cache-location=.eslintmdcache",
     "lint:md:fix": "node --run lint:md -- --fix",
-    "lint:types": "tsc --noEmit",
+    "lint:types": "next typegen && tsc --noEmit",
     "playwright": "playwright test",
     "scripts:release-post": "cross-env NODE_NO_WARNINGS=1 node scripts/release-post/index.mjs",
     "serve": "node --run dev",
--- a/.gitignore
+++ b/.gitignore
@@ -12,6 +12,7 @@ package-lock.json
 apps/site/.next
 apps/site/build
 apps/site/public/blog-data.json
+apps/site/next-env.d.ts
 
 # Test Runner
 junit.xml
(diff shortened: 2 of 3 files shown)


apps/site/package.json
```diff
@@ -21,7 +21,7 @@
     "lint:js:fix": "node --run lint:js -- --fix",
     "lint:md": "eslint \"**/*.md?(x)\" --cache --cache-strategy=content --cache-location=.eslintmdcache",
     "lint:md:fix": "node --run lint:md -- --fix",
-    "lint:types": "tsc --noEmit",
+    "lint:types": "next typegen && tsc --noEmit",
```
ovflowd, line 24: Can we somehow hook this command unto tsconfig itself? Like as a hook?

ovflowd (approved):
SGTM!

apps/site/package.json
MattIPv4: I don't believe so, no, I don't think tsconfig has any such hook functionality

ovflowd:
@nodejs/web-infra can we fast-track this?

ovflowd:
@MattIPv4 did you ask for fast-tracking?

MattIPv4:
I hadn't no, but no qualms with you asking for it.

PR #8833, opened 2026-04-22
chore: explicitly skip the Vercel OTEL when running open-next

dario-piotrowicz:
## Description

The Vercel OTEL implementation doesn't work on the open-next build (I haven't investigated why), but it doesn't seem to cause issues either. However it produces some error in the terminal on previews.

In any case for the open-next build Vercel OTEL couldn't be used anyways (since the app wouldn't be hosted on Vercel).

In order to remove the terminal error and also to make the code more clear/explicit this PR is making the Vercel OTEL setup conditional based on whether the application is being run by Cloudflare or not.

## Validation

I've validated that the open-next preview now does not present the telemetry error anymore.

### Check List

### DIFF
--- a/apps/site/instrumentation.ts
+++ b/apps/site/instrumentation.ts
@@ -1,5 +1,7 @@
-import { registerOTel } from '@vercel/otel';
-
-export function register() {
-  registerOTel({ serviceName: 'nodejs-org' });
+export async function register() {
+  if (!('Cloudflare' in globalThis)) {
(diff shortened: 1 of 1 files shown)


apps/site/instrumentation.ts
ovflowd, line 1: Does this work? Does Vercel support this function being async? Cool.

ovflowd (approved):
SGTM! Fast-tracking as a hot-fix for OpenNext.

PR #8838, opened 2026-04-23
refactor: split platform-specific code into platform-vercel and platform-cloudflare packages

ovflowd:
Extracts all Vercel- and Cloudflare-specific integrations out of `apps/site` and into dedicated workspace packages (`@node-core/platform-vercel`, `@node-core/platform-cloudflare`), selected at build time via `NEXT_PUBLIC_DEPLOY_TARGET` (set by `vercel.json` and the OpenNext `buildCommand`).

Each adapter exports a `next.platform.config.mjs` with a `{ nextConfig, aliases, images, mdx }` shape that `apps/site` merges into its Next.js config, MDX plugins, and Playwright config via dynamic import. No-op `apps/site/next.platform.config.mjs` and `apps/site/playwright.platform.config.mjs` keep standalone `pnpm dev` / `pnpm build` / `pnpm deploy` working when no target is set.

Removes the last platform forks from `apps/site` — `VERCEL_ENV`, `OPEN_NEXT_CLOUDFLARE`, `'Cloudflare' in global`, and `PLAYWRIGHT_RUN_CLOUDFLARE_PREVIEW` are all gone.


ovflowd:
FYI deployment will only work once this gets merged as we need to change the root directory within vercel.

.github/workflows/build.yml
```diff
@@ -55,21 +55,37 @@ jobs:
...
+      - name: Build apps/site (standalone)
         # We want a ISR build on CI to ensure that regular Next.js builds work as expected.
-        run: node_modules/.bin/turbo build --cache-dir=.turbo/cache
+        run: node_modules/.bin/turbo build --filter=@node-core/website --cache-dir=.turbo/cache
         env:
-          # We want to ensure we have enough RAM allocated to the Node.js process
-          # this should be a last resort in case by any chances the build memory gets too high
-          # but in general this should never happen
           NODE_OPTIONS: '--max_old_space_size=4096'
-          # See https://github.com/vercel/next.js/pull/81318
```
avivkeller, line 70: Why were theee comments removed?

.github/workflows/build.yml
```diff
@@ -55,21 +55,37 @@ jobs:
...
           TURBOPACK_STATS: ${{ matrix.os == 'ubuntu-latest' }}
```
avivkeller, line 71: Is this needed if we are building standalone?

.github/workflows/build.yml
```diff
@@ -78,11 +94,8 @@ jobs:
 
       - name: Build Next.js (Static Export)
         # We want to generate a static build, as it is a requirement of our website.
-        run: node_modules/.bin/turbo deploy --cache-dir=.turbo/cache
+        run: node_modules/.bin/turbo deploy --filter=@node-core/website --cache-dir=.turbo/cache
         env:
-          # We want to ensure we have enough RAM allocated to the Node.js process
-          # this should be a last resort in case by any chances the build memory gets too high
-          # but in general this should never happen
```
avivkeller, line 85: Why were these comments removed?

.github/workflows/build.yml
```diff
@@ -55,21 +55,37 @@ jobs:
...
+          TURBOPACK_STATS: ${{ matrix.os == 'ubuntu-latest' }}
+
+      - name: Build apps/cloudflare (Cloudflare / OpenNext deployment target)
+        # Ensures the Cloudflare deployment target builds with its Cloudflare
+        # Images loader and MDX_WASM=false Shiki path.
+        # Windows skip: OpenNext/wrangler asset polyfilling is POSIX-only.
+        if: matrix.os == 'ubuntu-latest'
+        run: node_modules/.bin/turbo cloudflare:build:worker --filter=@node-core/website-cloudflare --cache-dir=.turbo/cache
+        env:
+          NODE_OPTIONS: '--max_old_space_size=4096'
```
avivkeller: Can you duplicate the previous code comments to these?

apps/cloudflare/app/[locale]/[...path]
Comment (avivkeller) [inline apps/cloudflare/app/[locale]/[...path]:1]: I don't like all these symlinks, is there a better way to do this?

apps/cloudflare/cloudflare/deployment-id.mjs
avivkeller, line 1: No need to put this in a subdirectory if the entire directory is Cf related

apps/site/mdx/plugins.mjs
```diff
@@ -9,25 +9,16 @@ import readingTime from 'remark-reading-time';
...
-  // Currently we fall back to the JavaScript RegEx engine
-  // on Cloudflare workers because `shiki/wasm` requires loading via
-  // `WebAssembly.instantiate` with custom imports, which Cloudflare doesn't support
-  // for security reasons.
-  wasm: !OPEN_NEXT_CLOUDFLARE,
-
-  // TODO(@avivkeller): Find a way to enable Twoslash w/ a VFS on Cloudflare
-  twoslash: !OPEN_NEXT_CLOUDFLARE,
-});
+const singletonShiki = await rehypeShikiji({ wasm, twoslash });
```
avivkeller: Can you keep this inline?

apps/site/next.config.base.mjs
```diff
@@ -0,0 +1,93 @@
...
+ *
+ * @param {Partial<import('next').NextConfig>} [overrides]
+ * @returns {import('next').NextConfig}
+ */
+export const createNextConfig = (overrides = {}) => {
+  const {
+    images: imagesOverride,
+    experimental: experimentalOverride,
+    ...rest
+  } = overrides;
```
avivkeller: Use a deep merge utility, so we can easily expand this

apps/site/next.image.config.mjs
avivkeller: This seems odd, why would be even call this function if `overrides` is truthy

apps/site/package.json
```diff
@@ -122,6 +110,34 @@
...
+    "./styles/*": "./styles/*",
+    "./public/*": "./public/*",
+    "./pages/*": "./pages/*",
+    "./next.fonts": "./next.fonts.ts",
+    "./next.constants.mjs": "./next.constants.mjs",
+    "./next.config.base.mjs": "./next.config.base.mjs",
+    "./next.image.config.mjs": "./next.image.config.mjs",
+    "./next.rewrites.mjs": "./next.rewrites.mjs",
+    "./i18n.tsx": "./i18n.tsx",
+    "./middleware.ts": "./middleware.ts"
```
avivkeller: This shouldn't be needed

apps/site/tsconfig.json
```diff
@@ -1,25 +1,9 @@
 {
+  "extends": "../../tsconfig.base.json",
```
avivkeller: Why do we need a base TSconfig, can't this file itself serve as the base?

avivkeller (changes requested):
Really really against the symlinks, so blocking on that regard

ovflowd:
> Really really against the symlinks, so blocking on that regard

Why so? And what are your alternatives? Because Im unsure if there are any... Without duplicating every file 😅

ovflowd:
Aviv, this PR is a PoC no need to go super nitpicky on it. I haven't done myself a round of reviews yet as I just published this PR so we can start analyzing approaches, I might as well convert to a draft.

avivkeller:
> Aviv, this PR is a PoC no need to go super nitpicky on it. I haven't done myself a round of reviews yet as I just published this PR so we can start analyzing approaches, I might as well convert to a draft.

Noted, I've converted it to a draft so it's clearer for reviewers

avivkeller:
> Why so? And what are your alternatives?

I'm not sure what the alternatives are, but Symlinks are a, in my opinion, slippery slope of inconsistent behavior across operating systems and build tools. Is there a way to tell Next.js / OpenNext to look at `apps/site/` for the actual site, or something?

ovflowd:
> > Why so? And what are your alternatives?
> 
> I'm not sure what the alternatives are, but Symlinks are a, in my opinion, slippery slope of inconsistent behavior across operating systems and build tools. Is there a way to tell Next.js / OpenNext to look at `apps/site/` for the actual site, or something?

I'm investigating 👀

.github/CODEOWNERS
```diff
@@ -27,8 +27,10 @@ turbo.json @nodejs/nodejs-website @nodejs/web-infra
 crowdin.yml @nodejs/web-infra
 apps/site/redirects.json @nodejs/web-infra
 apps/site/site.json @nodejs/web-infra
-apps/site/wrangler.jsonc @nodejs/web-infra
-apps/site/open-next.config.ts @nodejs/web-infra
+packages/platform-cloudflare/wrangler.jsonc @nodejs/web-infra
+packages/platform-cloudflare/open-next.config.ts @nodejs/web-infra
+packages/platform-cloudflare/next.platform.config.mjs @nodejs/web-infra
+packages/platform-vercel/next.platform.config.mjs @nodejs/web-infra
```
avivkeller: Everything in `packages` is meant to be published, let's put these in `apps/vercel` and `apps/cloudflare`?

apps/site/mdx/plugins.mjs
```diff
@@ -7,27 +7,19 @@ import rehypeSlug from 'rehype-slug';
...
-// tree-shaking.
-// Reference: https://github.com/nodejs/nodejs.org/pull/7896#issuecomment-3009480615
-const OPEN_NEXT_CLOUDFLARE = 'Cloudflare' in global;
+// Load MDX overrides contributed by the active deployment target. Keeps
+// this module free of platform-specific branches — each platform owns
+// its own `{ wasm, twoslash }` defaults via `next.platform.config.mjs`,
+// with the in-repo default config serving as the standalone fallback.
+const { default: platform } = DEPLOY_TARGET
+  ? await import(`@node-core/platform-${DEPLOY_TARGET}/next.platform.config`)
+  : await import('../next.platform.config.mjs');
```
avivkeller: What if we use import maps? `#platform/` maps to `apps/[platform]` depending on the platform?

ovflowd: Interesting, let me check that 👀

apps/cloudflare/package.json
```diff
@@ -0,0 +1,45 @@
...
+  },
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/nodejs/nodejs.org",
+    "directory": "apps/cloudflare"
+  },
+  "scripts": {
+    "cloudflare:build:worker": "cd ../site && opennextjs-cloudflare build --openNextConfigPath ../cloudflare/open-next.config.ts --config ../cloudflare/wrangler.jsonc",
+    "cloudflare:deploy": "cd ../site && opennextjs-cloudflare deploy --openNextConfigPath ../cloudflare/open-next.config.ts --config ../cloudflare/wrangler.jsonc",
+    "cloudflare:preview": "cd ../site && wrangler dev --config ../cloudflare/wrangler.jsonc",
```
dario-piotrowicz: I feel that having build/deploy/preview scripts changing your working directory to be very weird/inconvenient 😕 

(also, is it necessary? is open-next not smart enough to work from a different directory? if it's not then we should amend that there)

dario-piotrowicz (commented):
I appreciate where the PR is coming from, however I don't think I am a huge fan of this, having 3 apps being `cloudflare`, `vercel` and `site` seems quite unclear and unintuitive to me (since there is only a single app, `site` which can have 2 different variations). There are also magic build time resolutions that I feel are very likely to confuse and alienate people not too familiar with the codebase (and me 😅).

Sorry but personally I feel like, instead of making things clearer the changes here introduce clear/strong boundaries between the different versions of the site at the cost of making the project less clear and navigatable (and approachable by newcommers).

If it were me and we wanted to isolate the vercel and cloudflare logic using different packages in the monorepo I think I'd try to keep `site` being the only app and instead add `cloudflare` and `vercel` as actual packages instead and having `site` depend on those (with some minor exceptions such as wrangler and open-next config files).

Sorry this is my general take on the approach 😓

(PS: if people are happy with this I'll definitely won't try to spot this direction though 🙂)

ovflowd:
> I'd try to keep site being the only app and instead add cloudflare and vercel as actual packages instead and having site depend on those (with some minor exceptions such as wrangler and open-next config files).

This is the exact opposite others asked. To be clear, apps/cloudflare and apps/vercel aren't "apps" or variations of the website, they are "packages", they are under "apps" because @avivkeller asked, as everything under "packages/" is supposed to be published. I'm fine making a new root folder named "platforms"

To also clarify here, @dario-piotrowicz, apps/site is the only package, all that happens is that when building for Vercel or for Cloudflare we tell Node.js to load the platform-specific config files and things that each one of these environments used.

apps/cloudflare/package.json
ovflowd: Let me 👀

dario-piotrowicz:
> This is the exact opposite others asked. 

Oh ok... different people have different preferences 🙈 

> To be clear, apps/cloudflare and apps/vercel aren't "apps" or variations of the website, they are "packages", they are under "apps" because @avivkeller asked, as everything under "packages/" is supposed to be published. I'm fine making a new root folder named "platforms"

I see... yeah probably platforms would be clearer in my opinion 🤔 , although they are both tied to the `site` project (they don't make sense as standalone projects)... mh... I can't remember, in turborepo I think you can have packages inside packages... what about having Vercel and Cloudflare inside `site/platforms/`? (as standalone/isolated packages but inside the `site` directory)

> To also clarify here, @dario-piotrowicz, apps/site is the only package, all that happens is that when building for Vercel or for Cloudflare we tell Node.js to load the platform-specific config files and things that each one of these environments used.

Yes I got that, but the dir structure really doesn't make it clear in my opinion, since `site`, `cloudflare` and `vercel` are sibling directories

ovflowd:
> I see... yeah probably platforms would be clearer in my opinion 🤔 , although they are both tied to the site project (they don't make sense as standalone projects)... mh... I can't remember, in turborepo I think you can have packages inside packages... what about having Vercel and Cloudflare inside site/platforms/? (as standalone/isolated packages but inside the site directory)

Im unsure how putting them as subfolders of apps/site makes things better. One of the goals here is that apps/site is the Next.js App, it is agnostic to any platform. But it is wired to support "platform" overrides, that's where the env flags come in, pretty much we're using Node.js to say "for cloudflare, please load the overrides that come from this apps/cloudflare" package. We aren't really building things within apps/cloudflare, nor apps/cloudflare is an actual app, I guess the name apps/ is the confusing factor. Think of apps/cloudflare or apps/vercel as bindings. Im fine moving them under another folder, but I'd like to avoid keep moving them around as @avivkeller said apps/ you said something else, so if we can reach an agreement of preference, that'd be great.

> Yes I got that, but the dir structure really doesn't make it clear in my opinion, since site, cloudflare and vercel are sibling directories

I get that. Anyhow, the goal of this PR is to:

1. installing apps/site doesn't require any dep of apps/cloudflare or apps/vercel, so reduced deps, better control of blast radius, if smth breaks, it must be Next.js-related things or of our sandboxed app.
2. removes specific code and checks and paths and settings and logical statements regarding platforms. This makes hosting plain Next.js or static exports easier, also allows the codebase to grow to a number of different platforms.
3. Easier to group all code retaining a given platform. No more search across all packages;apps/site where we have Vercel or Clopudflare or Platform X-specific code, it is all in one place, easy to find and maintain.

dario-piotrowicz:
> Im unsure how putting them as subfolders of apps/site makes things better. One of the goals here is that apps/site is the Next.js App, it is agnostic to any platform. But it is wired to support "platform" overrides, that's where the env flags come in, pretty much we're using Node.js to say "for cloudflare, please load the overrides that come from this apps/cloudflare" package. We aren't really building things within apps/cloudflare, nor apps/cloudflare is an actual app, I guess the name apps/ is the confusing factor. Think of apps/cloudflare or apps/vercel as bindings. Im fine moving them under another folder, but I'd like to avoid keep moving them around as @avivkeller said apps/ you said something else, so if we can reach an agreement of preference, that'd be great.

Agreed 👍 , I do think that `apps/` is a bit confusing here, but if everyone else is ok with it I'm ok with it as well (also the dir structure can be easy to change in the future anyways) 

> I get that. Anyhow, the goal of this PR is to:
> 
> 1. installing apps/site doesn't require any dep of apps/cloudflare or apps/vercel, so reduced deps, better control of blast radius, if smth breaks, it must be Next.js-related things or of our sandboxed app.
> 2. removes specific code and checks and paths and settings and logical statements regarding platforms. This makes hosting plain Next.js or static exports easier, also allows the codebase to grow to a number of different platforms.
> 3. Easier to group all code retaining a given platform. No more search across all packages;apps/site where we have Vercel or Clopudflare or Platform X-specific code, it is all in one place, easy to find and maintain.

Yes and I don't disagree that it's doing that well. My concern here is that in order to satisfy those 3 goals the structure of the monorepo is becoming (at least in my opinion) more complex, less intuitive (and more unconventional?). So I think there is a tradeoff to be aware of between isolating the platforms bits and the complexity of the repo's setup.

This might also be subjective and people might not find the updated structure more complex 🤔

bmuenzenmeyer:
In general, I'm in support of isolating the deployment targets from the rest of the code as best we can, and want to keep thinking about how best to do this. I suspect we need to put the branching logic somewhere, and I am thinking through the best options:

- in config/scripts via /apps as the PR does
- in new modules like /packages or /platforms

Naming things is hard, of course, and we are not constrained by the two monorepo directories we have now. Perhaps moving them to platforms/ helps ever so slightly. At the end of the day, the READMEs are doing the work.

>  So I think there is a tradeoff to be aware of between isolating the platforms bits and the complexity of the repo's setup.

I would agree with this in principle but also think this approach has made it simpler in a way to casual contribution. 
I think the goal of being able to run the site locally and see mostly content, without any divergent platform code is admirable. The repo is already complicated. Made more so by learn, the api docs, the data we consume, i18n, I've said before I think that boat sailed a long time ago :D This IMO might be a net improvement, and I can imagine a step simpler to maintain if, say, a third deployment target were ever to present itself.

> people might not find the updated structure more complex 

Maybe yes, maybe no. Is it slightly unconventional? I think it's pushing the boundaries of how apps can be constructed. Most apps don't have the luxury? the design goal? of deploying multiple ways. But this feels like a good use of a monorepo and perhaps a _slightly clever_ way to use import maps...but one that I see as innovative 😄 

I do think we should always try to optimize for casual contribution, which I think this still does via a simpler /apps/site

.github/CODEOWNERS
bmuenzenmeyer, line 1: Splitting by platform also opens  the door to more nuanced CODEOWNERS - if we have experts or partners in one platform or the other

platforms/cloudflare/image-loader.ts
bmuenzenmeyer, line 1: nit: this change feels arbitrary unless i am misunderstanding an eslint rule. i usually prefer braces because they are durable to more code being added later

platforms/vercel/vercel.json
bmuenzenmeyer, line 1: this file's presence in /site goes against the sprit of the separation
can it move?

bmuenzenmeyer (approved):
I am a 👍 on this proposal. 
Due to the nature of the change, I wonder if we should bring this up in a web meeting?

platforms/vercel/vercel.json
ovflowd: I want to move in a follow-up PR as it requires changes on Vercel.com to manually change the appRoot, and move the vercel.json file to the other folder, but I can also do it on this PR.

.github/CODEOWNERS
ovflowd: Any example?

ovflowd:
@bmuenzenmeyer I made the changes for vercel.json, Vercel build will fail till I move change the appRoot within the UI.

apps/site/mdx/plugins.mjs
```diff
@@ -7,27 +7,16 @@ import rehypeSlug from 'rehype-slug';
 import remarkGfm from 'remark-gfm';
 import readingTime from 'remark-reading-time';
 
-import remarkTableTitles from '../util/table';
+// MDX overrides contributed by the active deployment target. Resolved via
+// the `#platform/next.platform.config` import map in `package.json`; each
+// platform owns its own `{ wasm, twoslash }` defaults and the in-repo
+// default file acts as the standalone fallback.
```
avivkeller: nit: I don't think this comment is needed

apps/site/instrumentation.ts
```diff
@@ -1,7 +1 @@
-export async function register() {
-  if (!('Cloudflare' in globalThis)) {
-    // Note: we don't need to set up the Vercel OTEL if the application is running on Cloudflare
-    const { registerOTel } = await import('@vercel/otel');
-    registerOTel({ serviceName: 'nodejs-org' });
-  }
-}
+export { register } from '@platform/instrumentation';
```
avivkeller:
```suggestion
export { register } from '#platform/instrumentation';
```

Is the `@` here intentional?

apps/site/next.config.mjs
```diff
@@ -27,9 +21,14 @@ const nextConfig = {
   // We allow the BASE_PATH to be overridden in case that the Website
   // is being built on a subdirectory (e.g. /nodejs-website)
   basePath: BASE_PATH,
-  // Vercel/Next.js Image Optimization Settings
-  images: getImagesConfig(),
+  images: getImagesConfig(await platform.images?.()),
```
avivkeller:
```suggestion
```

Since we have two platforms, and they both have different configs, this can be defined in the platform

apps/site/platform/next.platform.config.mjs
avivkeller, line 1: This is the default, right? Maybe `platforms/default`?

apps/site/playwright.config.ts
```diff
@@ -12,9 +12,12 @@ export default defineConfig({
   retries: isCI ? 2 : 0,
   workers: isCI ? 1 : undefined,
   reporter: isCI ? [['html'], ['github']] : [['html']],
-  ...getWebServerConfig(),
+  ...(platform.webServer ? { webServer: platform.webServer } : {}),
```
avivkeller:
```suggestion
  ...platform
```

apps/site/tsconfig.json
```diff
@@ -20,7 +20,11 @@
         "name": "next"
       }
     ],
-    "baseUrl": "."
+    "baseUrl": ".",
+    "paths": {
+      "@platform/analytics": ["./platform/analytics.tsx"],
+      "@platform/instrumentation": ["./platform/instrumentation.ts"]
```
avivkeller: These can be Node.js aliases, right?

platforms/cloudflare/worker-entrypoint.ts
```diff
@@ -11,7 +11,7 @@ import type {
   Request,
 } from '@cloudflare/workers-types';
 
-import { default as handler } from '../.open-next/worker.js';
+import { default as handler } from '../../../apps/site/.open-next/worker.js';
```
avivkeller: This should go in this folder, can it?

platforms/cloudflare/next.config.mjs
avivkeller, line 1: 1. Let's just call this `next.config.mjs`, no need to `.platform.` it, we are already in a platform folder
2. This should only have the next config stuff. It should "extend" the default config in it's default export. The MDX config can go in a `config.ts` for _general_ configurations

platforms/cloudflare/open-next.config.ts
```diff
@@ -20,7 +20,8 @@ const cloudflareConfig = defineCloudflareConfig({
 
 const openNextConfig: OpenNextConfig = {
   ...cloudflareConfig,
-  buildCommand: 'pnpm build --webpack',
+  buildCommand:
```
avivkeller: @dario-piotrowicz Suggestion for OpenNext: `env` for setting environment variables like this?

platforms/cloudflare/package.json
```diff
@@ -0,0 +1,45 @@
...
+  "version": "1.0.0",
+  "private": true,
+  "type": "module",
+  "exports": {
+    "./analytics": "./src/analytics.tsx",
+    "./image-loader": "./src/image-loader.ts",
+    "./instrumentation": "./src/instrumentation.ts",
+    "./next.platform.config": "./next.platform.config.mjs",
+    "./playwright.platform.config": "./playwright.platform.config.mjs",
+    "./worker-entrypoint": "./src/worker-entrypoint.ts"
```
avivkeller:
```suggestion
    "./src/*": [
        "./src/*",
        "./src/*.ts",
        "./src/*.tsx",
        "./src/*.mjs"
      ]
```

platforms/cloudflare/image-loader.ts
ovflowd: It is arbitrary, but follows repository practices 👀

platforms/cloudflare/wrangler.jsonc
```diff
@@ -31,13 +31,14 @@
...
   },
   "alias": {
-    "node:fs": "./.wrangler/fs-assets-polyfilling/polyfills/node/fs.ts",
-    "node:fs/promises": "./.wrangler/fs-assets-polyfilling/polyfills/node/fs/promises.ts",
-    "fs": "./.wrangler/fs-assets-polyfilling/polyfills/node/fs.ts",
-    "fs/promises": "./.wrangler/fs-assets-polyfilling/polyfills/node/fs/promises.ts",
+    "node:fs": "../../apps/site/.wrangler/fs-assets-polyfilling/polyfills/node/fs.ts",
+    "node:fs/promises": "../../apps/site/.wrangler/fs-assets-polyfilling/polyfills/node/fs/promises.ts",
+    "fs": "../../apps/site/.wrangler/fs-assets-polyfilling/polyfills/node/fs.ts",
+    "fs/promises": "../../apps/site/.wrangler/fs-assets-polyfilling/polyfills/node/fs/promises.ts",
```
avivkeller, line 41: Can these go in the platform directory or no?

platforms/vercel/__tests__/next.platform.config.test.mjs
avivkeller, line 1: This test seems un-needed

platforms/vercel/package.json
```diff
@@ -0,0 +1,41 @@
+{
+  "name": "@node-core/platform-vercel",
+  "version": "1.0.0",
+  "private": true,
+  "type": "module",
+  "exports": {
+    "./analytics": "./src/analytics.tsx",
+    "./instrumentation": "./src/instrumentation.ts",
+    "./next.platform.config": "./next.platform.config.mjs",
+    "./playwright.platform.config": "./playwright.platform.config.mjs"
```
avivkeller: Use `*`

apps/site/instrumentation.ts
ovflowd: No, fixing, good catch!

apps/site/platform/next.platform.config.mjs
ovflowd: Like in a new package? Is this really needed? 🤔 -- I wonder if it is implied default platform / apps/site, but I guess it makes sense?

platforms/cloudflare/worker-entrypoint.ts
ovflowd: It can't, because .open-next generates its output on apps/site, tried but no way to change that....

platforms/cloudflare/package.json
ovflowd: Yeah, I was going to do that, also noticed this on self-review xD

platforms/cloudflare/wrangler.jsonc
ovflowd: Nope, they are auto-generated.

platforms/cloudflare/package.json
```diff
@@ -0,0 +1,45 @@
...
+  },
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/nodejs/nodejs.org",
+    "directory": "platforms/cloudflare"
+  },
+  "scripts": {
+    "cloudflare:build:worker": "pnpm --filter=@node-core/website exec opennextjs-cloudflare build --openNextConfigPath ../../platforms/cloudflare/open-next.config.ts --config ../../platforms/cloudflare/wrangler.jsonc",
+    "cloudflare:deploy": "pnpm --filter=@node-core/website exec opennextjs-cloudflare deploy --openNextConfigPath ../../platforms/cloudflare/open-next.config.ts --config ../../platforms/cloudflare/wrangler.jsonc",
+    "cloudflare:preview": "pnpm --filter=@node-core/website exec wrangler dev --config ../../platforms/cloudflare/wrangler.jsonc",
```
dario-piotrowicz: Nice scripts 🤩 

What about getting rid of the `cloudflare:` prefix as well? (and also build's `:worker` suffix)?

platforms/cloudflare/instrumentation.ts
```diff
@@ -0,0 +1 @@
+export const register = () => {};
```
dario-piotrowicz, line 1: Is this file actually needed?

docs/technologies.md
```diff
@@ -120,31 +122,35 @@ We chose Next.js because it is:
...
+│   │   ├── providers/           # React context providers
+│   │   ├── types/               # TypeScript definitions
+│   │   ├── next-data/           # Build-time data fetching
+│   │   ├── scripts/             # Utility scripts
+│   │   ├── snippets/            # Code snippets for download page
+│   │   └── tests/               # Test files
+│   │       └── e2e/             # End-to-end tests
+│   ├── vercel/                  # Vercel deployment adapter
+│   │                            #   (analytics, instrumentation, vercel.json)
+│   └── cloudflare/              # Cloudflare deployment adapter
```
dario-piotrowicz: These haven't been updated, they are no longer under `apps`

platforms/vercel/package.json
```diff
@@ -0,0 +1,41 @@
...
+  },
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/nodejs/nodejs.org",
+    "directory": "platforms/vercel"
+  },
+  "scripts": {
```
dario-piotrowicz, line 11: Should the vercel platform also have some `build`/`deploy` scripts?

dario-piotrowicz (approved):
This feels much better than the version I reviewed last, thanks for all the improvements @ovflowd! 🫶 

I also completely agree with @bmuenzenmeyer this simplifying the base app and allowing contributors to just work on it without caring about target platform details is a very strong plus of this approach 🚀 

Looks good to me, awesome work 🫶

platforms/cloudflare/open-next.config.ts
dario-piotrowicz: Sorry, can you clarify what you mean? 🙏

platforms/cloudflare/instrumentation.ts
ovflowd: Yes! Even if no-op.

platforms/vercel/next.config.mjs
```diff
@@ -0,0 +1,30 @@
...
+    return {
+      // Expose Vercel's auto-assigned deployment URL as a platform-agnostic
+      // `NEXT_PUBLIC_BASE_URL` so `apps/site` consumers can read a single
+      // canonical env var. A manually-set `NEXT_PUBLIC_BASE_URL` wins. Only
+      // contribute the entry when a value is actually present so downstream
+      // truthiness vs. existence checks behave the same as if the var were
+      // never set.
+      env: {
+        NEXT_PUBLIC_BASE_URL: process.env.NEXT_PUBLIC_BASE_URL || VERCEL_URL,
+      },
```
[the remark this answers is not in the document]
ovflowd: I think this is completely fine, wdyt @avivkeller?

apps/site/next.config.mjs
```diff
@@ -1,22 +1,30 @@
...
+ * inside Turbopack/webpack).
+ *
+ * @type {{ default: import('./next.platform.config.d.ts').PlatformConfig }}
+ */
+const { default: platform } = await import(`${PLATFORM_ALIAS}/next.config.mjs`);
 
-    return openNextAdapter.getDeploymentId();
-  }
+const platformImages = await platform.images?.();
+const platformNextConfig = await platform.nextConfig?.();
```
avivkeller: These don't need to be functions set this this way.

If we just pass `...nextConfig`, if the `nextConfig` defines an `images`, it'll work like that, right?

```
const { default: nextConfig } = await import(`${PLATFORM_ALIAS}/next.config.mjs`);
```

apps/site/next.config.mjs
```diff
@@ -1,22 +1,30 @@
...
 
-  return undefined;
-};
+// Turbopack's `resolveAlias` requires explicit `*` wildcards; webpack's
+// `resolve.alias` matches the bare prefix and the `/*` form is invalid,
+// so the two bundlers need different shapes for the same mapping.
+const turbopackPlatformAliases = { '@platform/*': `${PLATFORM_ALIAS}/*` };
+const webpackPlatformAliases = { '@platform': PLATFORM_ALIAS };
```
avivkeller: Can you add a source for why we can't use node aliases, so it's good to know for the future?

avivkeller:
I think the bug i was waiting for has been patched, so ill continue this work

avivkeller:
I'm going to open a separate PR with more improvements

PR #8839, opened 2026-04-23
fix: fix the incorrect architecture detection in PlatformDropdown for Cloudflare builds

dario-piotrowicz:
## Description

OS detection and architecture detection can arrive in the same React render cycle. When that happens, both `useEffect` hooks in `PlatformDropdown` fire at the same time, the first one correctly dispatches `setPlatform('arm64')`, but the second one reads `release.platform` from its stale closure (still `'x64'`) and immediately calls `setPlatform('x64')` overriding the correct value. On Vercel I believe this worked by accident of timing, as the two detections land in separate render cycles, for whatever reason that doesn't seem to be the case in Cloudflare.

So the fix is to compute the platform during render and use it in both `useEffect`s, so that they share the same up-to-date value when they fire together.

## Validation

I've run the app locally via `pnpm dev` and `pnpm cloudflare:preview` and verified that now both correctly detect the correct architecture

### Check List

### DIFF
--- a/apps/site/components/Downloads/Release/PlatformDropdown.tsx
+++ b/apps/site/components/Downloads/Release/PlatformDropdown.tsx
@@ -16,14 +16,18 @@ const PlatformDropdown: FC = () => {
   const { architecture, bitness } = useClientContext();
 
   const release = use(ReleaseContext);
+
   const t = useTranslations();
+
+  // Compute the platform during render so it's available to both `useEffect` calls below in the same cycle
+  // (avoiding race conditions when architecture detection and OS detection arrive in the same batch)
+  const currentPlatform =
+    architecture && bitness ? getUserPlatform(architecture, bitness) : null;
 
   useEffect(
     () => {
-      if (architecture && bitness) {
-        const autoDetectedPlatform = getUserPlatform(architecture, bitness);
-
-        release.setPlatform(autoDetectedPlatform);
+      if (currentPlatform) {
+        release.setPlatform(currentPlatform);
       }
     },
     // Only react on the change of the Client Context Architecture and Bitness
@@ -49,12 +53,14 @@ const PlatformDropdown: FC = () => {
   useEffect(
     () => {
       if (release.os !== 'LOADING' && release.platform !== '') {
-        release.setPlatform(nextItem(release.platform, parsedPlatforms));
+        // Use the current platform if available, otherwise fall back to the current release platform
+        const currentTargetPlatform = currentPlatform ?? release.platform;
+        release.setPlatform(nextItem(currentTargetPlatform, parsedPlatforms));
       }
     },
     // We only want to react on the change of the OS and Version
     // eslint-disable-next-line @eslint-react/exhaustive-deps
-    [release.os, release.version, release.platform]
+    [release.os, release.version]
   );
 
   return (
--- a/apps/site/components/Downloads/Release/PlatformDropdown.tsx
+++ b/apps/site/components/Downloads/Release/PlatformDropdown.tsx
@@ -2,7 +2,7 @@
 
 import Select from '@node-core/ui-components/Common/Select';
 import { useTranslations } from 'next-intl';
-import { useEffect, use, useMemo } from 'react';
+import { useEffect, use, useMemo, useRef } from 'react';
 
 import useClientContext from '#site/hooks/useClientContext';
 import { ReleaseContext } from '#site/providers/releaseProvider';
@@ -24,9 +24,15 @@ const PlatformDropdown: FC = () => {
   const currentPlatform =
     architecture && bitness ? getUserPlatform(architecture, bitness) : null;
 
+  // Track whether the user has manually selected a platform via the dropdown.
+  // When true the OS/version effect will respect their choice instead of
+  // resetting to the auto-detected value.
+  const userSelectedPlatformRef = useRef(false);
+
   useEffect(
     () => {
       if (currentPlatform) {
+        userSelectedPlatformRef.current = false;
         release.setPlatform(currentPlatform);
       }
     },
@@ -53,9 +59,14 @@ const PlatformDropdown: FC = () => {
   useEffect(
     () => {
       if (release.os !== 'LOADING' && release.platform !== '') {
-        // Use the current platform if available, otherwise fall back to the current release platform
-        const currentTargetPlatform = currentPlatform ?? release.platform;
-        release.setPlatform(nextItem(currentTargetPlatform, parsedPlatforms));
+        // If the user has not manually selected a platform and there is a currently
+        // auto-detected one then use it otherwise fallback to the current release platform
+        const basePlatform =
+          !userSelectedPlatformRef.current && currentPlatform
+            ? currentPlatform
+            : release.platform;
+
+        release.setPlatform(nextItem(basePlatform, parsedPlatforms));
       }
     },
     // We only want to react on the change of the OS and Version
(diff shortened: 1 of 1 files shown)


avivkeller (approved):

apps/site/components/Downloads/Release/PlatformDropdown.tsx
```diff
@@ -49,12 +53,14 @@ const PlatformDropdown: FC = () => {
   useEffect(
     () => {
       if (release.os !== 'LOADING' && release.platform !== '') {
-        release.setPlatform(nextItem(release.platform, parsedPlatforms));
+        // Use the current platform if available, otherwise fall back to the current release platform
+        const currentTargetPlatform = currentPlatform ?? release.platform;
+        release.setPlatform(nextItem(currentTargetPlatform, parsedPlatforms));
       }
```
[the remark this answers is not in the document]
dario-piotrowicz: also commented by cursor

apps/site/components/Downloads/Release/PlatformDropdown.tsx
```diff
@@ -49,12 +53,14 @@ const PlatformDropdown: FC = () => {
   useEffect(
     () => {
       if (release.os !== 'LOADING' && release.platform !== '') {
-        release.setPlatform(nextItem(release.platform, parsedPlatforms));
+        // Use the current platform if available, otherwise fall back to the current release platform
+        const currentTargetPlatform = currentPlatform ?? release.platform;
+        release.setPlatform(nextItem(currentTargetPlatform, parsedPlatforms));
```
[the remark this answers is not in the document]
dario-piotrowicz: Addressed in https://github.com/nodejs/nodejs.org/pull/8839/commits/e7a30c9e3b14f5d0984463f2823eb4a77c4a2960

ovflowd (commented):
I am trying to wrap my head around on why this would be needed 🤔

Why it works in Vercel but it breaks in Cloudflare 😅

apps/site/components/Downloads/Release/PlatformDropdown.tsx
```diff
@@ -16,14 +16,24 @@ const PlatformDropdown: FC = () => {
   const { architecture, bitness } = useClientContext();
 
   const release = use(ReleaseContext);
+
   const t = useTranslations();
 
+  // Compute the platform during render so it's available to both `useEffect` calls below in the same cycle
+  // (avoiding race conditions when architecture detection and OS detection arrive in the same batch)
+  const currentPlatform =
```
ovflowd, line 24: Is the getUserPlatform an expensive operation? This will be called on every render.

apps/site/components/Downloads/Release/PlatformDropdown.tsx
```diff
@@ -16,14 +16,24 @@ const PlatformDropdown: FC = () => {
...
+    architecture && bitness ? getUserPlatform(architecture, bitness) : null;
+
+  // Track whether the user has manually selected a platform via the dropdown.
+  // When true the OS/version effect will respect their choice instead of
+  // resetting to the auto-detected value.
+  const userSelectedPlatformRef = useRef(false);
```
ovflowd: Is using a ref the only alternative here? Why a ref and not an useState? Shouldn't this cause a re-render.

apps/site/components/Downloads/Release/PlatformDropdown.tsx
```diff
@@ -64,7 +81,12 @@ const PlatformDropdown: FC = () => {
       loading={release.os === 'LOADING' || release.platform === ''}
       placeholder={t('layouts.download.dropdown.unknown')}
       ariaLabel={t('layouts.download.dropdown.platform')}
-      onChange={platform => platform && release.setPlatform(platform)}
+      onChange={platform => {
+        if (platform) {
```
ovflowd: Won't platform always be a value? In which situation it would be null/empty, I think this check can be removed.

apps/site/components/Downloads/Release/PlatformDropdown.tsx
```diff
@@ -49,12 +59,19 @@ const PlatformDropdown: FC = () => {
   useEffect(
     () => {
       if (release.os !== 'LOADING' && release.platform !== '') {
-        release.setPlatform(nextItem(release.platform, parsedPlatforms));
+        // If the user has not manually selected a platform and there is a currently
+        // auto-detected one then use it otherwise fallback to the current release platform
+        const basePlatform =
```
ovflowd, line 65: I'm skeptical about this logic, and still trying to wrap my head around on *why* this has a different behavior on Cf workers.

dario-piotrowicz:
> I am trying to wrap my head around on why this would be needed 🤔
> 
> Why it works in Vercel but it breaks in Cloudflare 😅

yeah... I don't know... it is client side logic so this is pretty baffling... maybe on Cloudflare some assets gets served in parallel? 🤷

apps/site/components/Downloads/Release/PlatformDropdown.tsx
dario-piotrowicz: No, it's a very trivial function 🙂  (that's why I though it'd be ok to be re-run on every render)

https://github.com/nodejs/nodejs.org/blob/524e64b59c20916363d063b177efb3137d836910/apps/site/util/userAgent.ts#L39-L48

apps/site/components/Downloads/Release/PlatformDropdown.tsx
dario-piotrowicz: `userSelectedPlatformRef` is being set when the user selects a platform, so there will already be a re-render associated to that operation (the ref is just a flag to record that that event happened)

So I do think that ref is valid here 🤔 , but I can switch this to a `useState` if you prefer 🙂

ovflowd: If the both states are being set at the same time, React will only trigger one re-render, React lifecycle already optimizes for it. I'd say using a state here is probably better as this is associated to an eventual re-render.

apps/site/components/Downloads/Release/PlatformDropdown.tsx
dario-piotrowicz: > still trying to wrap my head around on why this has a different behavior on Cf workers.

I can try to investigate that more... but I am not really sure it might just be a very subtle difference that open-next introduces or maybe it's related to the assets serving 🤷

The way I see this is that the existing logic is flowed, since it does make the assumption that OS detection and architecture detection don't arrive in the same React render cycle. Removing that assumption makes the logic here more correct/robust while having the side-effect of making the Cloudflare deployment work

ovflowd: These are good arguments.

dario-piotrowicz: (what I was trying to say is that fixing code prone to race conditions seems a valid fix/improvement in any case even when taking Cloudflare out of the equation)

apps/site/components/Downloads/Release/PlatformDropdown.tsx
dario-piotrowicz: Good point 🙂 

https://github.com/nodejs/nodejs.org/pull/8839/commits/a095200e7470f58cdef4b9b9612de5f6cb6d2909

Too bad I had to use an eslint-disable comment 😓

ovflowd (approved):
SGTM!

PR #8842, opened 2026-04-25
feat(open-next): create cloudflare-sentry-tail package

flakey5:
Creates a `cloudflare-sentry-tail` package that enables us to add a tail worker to the open next deployment of the site.

This package should be publishable as well so that we can reuse it in the release worker.

### Check List


packages/cloudflare-sentry-tail/src/index.ts
```diff
@@ -0,0 +1,300 @@
...
+      value: exception.message,
+    };
+  }
+}
+
+function shouldSampleTraceItem(sampleRate: number) {
+  const buffer = new Uint32Array(1);
+  crypto.getRandomValues(buffer);
+
+  const random = buffer[0] / 4294967295;
```
flakey5: idk where this number comes from but it works so :shrug:

packages/cloudflare-sentry-tail/package.json
```diff
@@ -0,0 +1,31 @@
...
+    "type": "git",
+    "url": "https://github.com/nodejs/nodejs.org",
+    "directory": "packages/cloudflare-sentry-tail"
+  },
+  "scripts": {
+    "lint": "node --run lint:js",
+    "lint:fix": "node --run lint:js:fix",
+    "lint:js": "eslint \"**/*.ts\"",
+    "lint:js:fix": "node --run lint:js -- --fix"
+  },
```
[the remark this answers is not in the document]
dario-piotrowicz: this issue seems legit 🤔

packages/cloudflare-sentry-tail/src/index.ts
dario-piotrowicz: `4294967295` is the maximum value that an element in an Unit32Array could have:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/TypedArray#typedarray_objects

maybe we can add a constant here instead of this magic number?

packages/cloudflare-sentry-tail/src/index.ts
```diff
@@ -0,0 +1,304 @@
...
+    return;
+  }
+
+  let breadcrumbsIdx = sentryEvent.breadcrumbs.length;
+
+  // Allocate space for the elements we're gonna add
+  sentryEvent.breadcrumbs.length +=
+    item.logs.length +
+    item.diagnosticsChannelEvents.length +
+    item.exceptions.length;
```
dario-piotrowicz: Arrays in JavaScript automatically expand as you add elements to them, see:
example of adding elements to an incrementing index in javascript

So I don't think we need to allocate memory for the new elements here 🤔

packages/cloudflare-sentry-tail/src/index.ts
```diff
@@ -0,0 +1,304 @@
...
+      level: 'debug',
+      timestamp: payload.timestamp,
+    };
+  }
+
+  let fingerprintIdx = sentryEvent.fingerprint!.length;
+  let exceptionValueIdx = sentryEvent.exception!.values!.length;
+
+  sentryEvent.fingerprint!.length += item.exceptions.length;
+  sentryEvent.exception!.values!.length += item.exceptions.length;
```
dario-piotrowicz: Like my other comment, I don't think that any "allocation" is needed for the arrays

packages/cloudflare-sentry-tail/package.json
```diff
@@ -0,0 +1,31 @@
+{
+  "name": "@node-core/cloudflare-sentry-tail",
```
avivkeller, line 2: Can you open an issue in admin for this package?

packages/cloudflare-sentry-tail/src/index.ts
avivkeller: Also, let's be a bit more apparent in where it comes from:

`const UINT_32_LIMIT = 2 ** 32`

avivkeller (approved):

packages/cloudflare-sentry-tail/src/index.ts
flakey5: Unnecessary yes, however, preallocating makes it so it's only one memory allocation instead of multiple so it's faster and more efficient: 
image

(noteworthy that this is in Node, but ofc both runtimes are v8 based, also [benchmark src](https://gist.github.com/flakey5/17006b572060fb255aa29d2d43bafc97))

dario-piotrowicz: I see, thanks that's quite interesting 😄 

However for a tail worker I don't imagine that having a better performance for the tail worker would be in any way helpful/useful, would it? 🤔 

(also I am curious about the number of such events that you'd expect, obviously the less the number the less the gain here)

___

PS: I'm just asking out of curiosity since these lines seem to introduce unnecessary (even if very minimal of course) complexity for no real gain. But of course they don't cause any harm if left in.

flakey5: > I don't imagine that having a better performance for the tail worker would be in any way helpful/useful, would it

On a small scale, there really wouldn't be that much difference. But within past 24hrs alone there have been 79 million reqs to the website and given the performance difference of ~600ns between preallocating and not, that equates to about

$$600 \text{ ns } * 79,000,000 \text { requests } * \frac{1 \text{ ms}}{1000 \text{ns}} * \frac{1 \text{ seconds}}{1000 \text{ ms}} * \frac{1 \text{minute}}{60 \text{ seconds}} * \frac{1 \text{ hour }}{60 \text{ minutes}} = 13.2 \text{ hours}$$

of cpu time saved (if all of those requests made it to this point that is). I definitely don't think we'd be seeing that many of requests at this point in the tail worker (hopefully at least 😅 ), but regardless point stands that it still adds up over time especially since cpu time is how tail workers are billed

> (also I am curious about the number of such events that you'd expect, obviously the less the number the less the gain here)

Of exceptions, the max I'd expect would be 1-2, but I have no idea how many logs or other pieces we'd be expecting

packages/cloudflare-sentry-tail/package.json
flakey5: For what exactly?

dario-piotrowicz (approved):

avivkeller:
@flakey5 can you rebase?

aymen94 (approved):

PR #8878, opened 2026-05-05
chore(open-next): avoid the use of DurableObjects by using `queueCache` with `memoryQueue` instead

dario-piotrowicz:
## Description

The use of DurableObjects breaks the open-next skew protection, so in order to make sure skew protection works this PR updates the queue to use `queueCache` with `memoryQueue` instead of durable objects 

## Validation

Site built and deployed to my personal account: https://8bf0874a-nodejs-website-1.dario-test.workers.dev

### Check List

### DIFF
--- a/apps/site/open-next.config.ts
+++ b/apps/site/open-next.config.ts
@@ -1,7 +1,8 @@
 import { defineCloudflareConfig } from '@opennextjs/cloudflare';
 import r2IncrementalCache from '@opennextjs/cloudflare/overrides/incremental-cache/r2-incremental-cache';
 import { withRegionalCache } from '@opennextjs/cloudflare/overrides/incremental-cache/regional-cache';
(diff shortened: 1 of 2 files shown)


vicb (commented):
LGTM (I can not approve here it seems).

Do we need to add a delete migration for former DOs?

dario-piotrowicz:
> LGTM (I can not approve here it seems).

Thanks 🙂 

> Do we need to add a delete migration for former DOs?

I think a [delete migration](https://developers.cloudflare.com/durable-objects/reference/durable-objects-migrations/#delete-migration) is used when a class is deleted but the durable object binding is still used.
Since here we are instead completely removing the durable object I think that the simplest / most appropriate way to delete it is doing so though the dashboard. Don't you think?

vicb:
> the simplest / most appropriate way to delete it is doing so though the dashboard

+1

avivkeller (approved):

PR #8903, opened 2026-05-18
warm search indexes on search intent

CodingRule:
## Description

- warm the Orama search indexes before the first query runs
- trigger warm-up when the user shows search intent through hover, focus, pointer down, or `Ctrl/Cmd+K`
- reuse one cached warm-up promise so the search data is not loaded more than once
- add unit tests for the cached warm-up path and the modal warm-up triggers

## Validation

- the first search interaction should feel faster because index loading now starts before the first typed query

### Check List


avivkeller (changes requested):
Why is this needed? What real difference does this make?

CodingRule:
I believe search should feel more responsive, so this change starts loading the Orama indexes as soon as the user shows search intent. It makes the first search feel faster on slower networks.

PR #8917, opened 2026-05-27
chore(monorepo): break out platform-specific code

avivkeller:
Ref: #8838

PR #8920, opened 2026-05-30
fix: disable twoslash for cloudflare build

jhcpeixoto:
Targets #8917.

The Cloudflare OpenNext build fails while collecting page data because the Cloudflare MDX config imports a VFS Twoslash setup, which pulls Node.js filesystem modules into the Worker bundle. This keeps syntax highlighting enabled on Cloudflare but disables Twoslash for that deployment target, and removes the now-unused VFS generator/dependencies.


MattIPv4:
_Why?_ I'm not sure I follow removing functionality instead of fixing it...

And, was this PR generated with AI assistance that you haven't declared? Your PR body originally had escaped newline characters which is usually a tell-tale sign of that.

jhcpeixoto:
> _Why_?

Is fix of other PR

MattIPv4:
I don't think removing functionality is particularly a fix?

avivkeller:
That generation was specifically added because it works on CF deployments. Removing it would not make sense.

PR #8955, opened 2026-06-10
chore(ui): reduce prod deps

avivkeller:
This PR makes `@node-core/ui-components` ship a precompiled `dist/` and consume it from the site, instead of the site transpiling the package's raw source. As a result, the package's build-only tooling (`typescript`, `tailwindcss`, `postcss-*`, `@tailwindcss/postcss`, `@types/react`) moves from `dependencies` to `devDependencies`, where it belongs.


packages/ui-components/package.json
```diff
@@ -1,41 +1,32 @@
...
-        "./dist/*/index.js"
-      ],
-      "default": [
-        "./src/*",
-        "./src/*.tsx",
-        "./src/*/index.tsx",
-        "./src/*.ts",
-        "./src/*/index.ts"
-      ]
-    }
```
avivkeller, line 19: This is an added benefit, since for other consumers, needing to pass a `rolldown` option becomes quite annoying.

MattIPv4:
> cc @nodejs/web-infra **this changes our `vercel.json` to install development dependencies**

Can we make sure that post-build, we prune back to just production runtime dependencies?

avivkeller:
I believe vercel does that post-build anyway, it only ships the built `.next` folder, right?

MattIPv4:
Does it bundle everything for server-side logic too?

avivkeller:
> Does it bundle everything for server-side logic too?

This change pre-bundles `@node-core/ui-components`, which is used by both the client and the server

MattIPv4:
Right, but I don't really want any dev deps hanging around in node_modules after we've done the initial build

avivkeller:
> Right, but I don't really want any dev deps hanging around in node_modules after we've done the initial build

It shouldn't matter what's in `node_modules`, right? Doesn't Vercel/Open-Next not ship `node_modules`? Don't they only ship the output of `build`?

MattIPv4:
That's what I was asking, to check that it is just that output that's used, and that node_modules doesn't persist for server-side dependencies 🤔

avivkeller:
Ahh sorry if I misunderstood. No, `node_modules` is not preserved, even on the server. In fact, we actually explicitly need to mark the modules which we want to preserve on the server side (which we do for `@types/node` for typings). I've double-checked this to confirm.

avivkeller:
FYI @nodejs/web-infra

MattIPv4 (approved):
This seems fine, but I would appreciate a second review from @nodejs/web-infra on this

AugustinMauroy (approved):

aymen94 (approved):

shanpriyan (approved):

avivkeller:
Hmmm, maybe a rebase will fix this Chromatic issue?

PR #8992, opened 2026-07-01
chore(ci): no more pull_request_target

avivkeller:


.github/workflows/leave-comment.yml
MattIPv4, line 1: We should probably determine the PR ourselves from the workflow_run data, otherwise this would allow an attacker to post a comment on _any_ PR.

avivkeller: While `workflow_run` _does_ send PR data, I found that it's not 100% reliable, as often (for some reason) is omitted. That being said, we can also,

Rely on it anyway,
Compare the SHA of the HEAD with open PRs

avivkeller: WDYT?

MattIPv4: This is what we do for js.org: https://github.com/js-org/js.org/blob/30b6a762eff9c2bf8a2fde8c05f0f394c0f44ae0/.github/workflows/rename_comment.yml#L30-L39

I suppose including a PR number in the payload, and then doing that same head check, is also safe.

avivkeller:
Not sure why Next.js didn't generate the stats file... maybe something changed?

avivkeller:
Looks like `TURBOPACK_STATS` is broken, and has been for some time. @MattIPv4 It's not really idiomatic, but I'm going to update this PR to _also_ fix this.

Edgarruiz8585:
Verificado por Edgarruiz8585 `gh pr checkout 8992`

.github/workflows/leave-comment.yml
```diff
@@ -0,0 +1,89 @@
+name: Leave Comment
+
+on:
+  workflow_run:
+    # Any Workflow that uploads a `pr-comment` artifact should be listed here
+    workflows: ['Build', 'Lighthouse']
+    types: [completed]
```
MattIPv4: This will need a zizmor ignore comment with a justification, please

.github/workflows/leave-comment.yml
```diff
@@ -0,0 +1,96 @@
...
+            // 2. For forks that list is empty, so find the open Pull Request whose HEAD SHA matches
+            // the commit that triggered the run.
+            const pulls = await github.paginate(github.rest.pulls.list, {
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              state: 'open',
+              per_page: 100,
+            });
+
+            const match = pulls.find(pull => pull.head.sha === run.head_sha);
```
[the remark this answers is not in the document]
MattIPv4: @avivkeller I think this is valid? Probably want to switch to doing the head branch comparsion like JS.org does?

MattIPv4 (approved):

avivkeller:
@nodejs/nodejs-website for approval

.github/workflows/chromatic.yml
```diff
@@ -70,6 +67,6 @@ jobs:
         with:
           workingDir: packages/ui-components
           buildScriptName: storybook:build
-          projectToken: ${{ secrets.CHROMATIC_PROJECT_TOKEN }}
+          projectToken: ${{ vars.CHROMATIC_PROJECT_TOKEN }}
```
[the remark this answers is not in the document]
bmuenzenmeyer: can we make sure we talk about this in writing?

(edit so it's near the review: https://www.chromatic.com/docs/github-actions/#project-token-secret)

bmuenzenmeyer (commented):
one comment i want to better understand. making another pass now

bmuenzenmeyer (changes requested):
want to make sure we all look at this slower.
https://www.chromatic.com/docs/github-actions/#forked-repositories does suggest this pattern (plaintext vs vars), with it as a tradeoff for just this use case. are we accepting of the possible misuse of quota from forks?

.github/workflows/chromatic.yml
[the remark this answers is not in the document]
bmuenzenmeyer: yeah, per https://www.chromatic.com/docs/github-actions/#project-token-secret isnt this a secet?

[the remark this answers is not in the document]
avivkeller: A few sections later in that same file, https://www.chromatic.com/docs/github-actions/#run-chromatic-on-external-forks-of-open-source-projects:

> Sharing project tokens allows contributors and others to run Chromatic builds on your project, which can use snapshots. They cannot access your account, settings, or accept baselines. This can be an acceptable tradeoff for open source projects that value community contributions.

avivkeller:
(Copying my review comment for visibility)

A few sections later in that same file, https://www.chromatic.com/docs/github-actions/#run-chromatic-on-external-forks-of-open-source-projects:

> Sharing project tokens allows contributors and others to run Chromatic builds on your project, which can use snapshots. They cannot access your account, settings, or accept baselines. This can be an acceptable tradeoff for open source projects that value community contributions.

avivkeller:
IMHO the risk of a misused quota is substantially lower than the risk of a misconfigured/vulnerable `pull_request_target`. 

The worst case scenario is that we get some spam Storybook builds that affect our (very high) quota of which we use very little.

The worst case scenario of a compromised workflow, on the other hand, is far more dangerous

bmuenzenmeyer (approved):

bmuenzenmeyer:
agreed - i am just glad we are talking about these tradeoffs

PR #8993, opened 2026-07-02
fix(ui): show up to 12 partner logos and fix width

styfle:
## Description

The home page hero renders the small partner list from a weighted shuffle whose
seed changes every 5 minutes, then truncates it to the first 6 entries. With 9
partners in `constants.json`, each render dropped 3 of them — so depending on
when (or from which cached render) you loaded the page, a given partner's logo
was simply missing. Some wide logos (≈2.2:1 aspect ratio) also rendered about
twice as wide as the roughly square marks next to them. The defined `weight`
in `constants.json` did not impact the width of the logo as still doesn't.

### Changes
- Raised `SMALL_PARTNER_LIMIT` from 6 to 12 so all current partners always
  render; the weighted shuffle still impacts the sort order.
- Added `justify-center` to the small partner list so rows stay centered when
  the icons wrap on different viewports.
- Added `max-w-4` to the small partner button SVG. Since the logos have a
  `viewBox`, wide marks scale down proportionally to fit the cap instead of
  stretching their button, so all partner buttons render at a uniform width.

## Validation

1. Visit the homepage
2. Notice the partner logos just below the main Get Node and Get Security buttons
3. Up to 12 are visible (currently 9 visible)
4. Try again on different viewports/devices

### Before

nodejs-org-before

### After

nodejs-org-after

## Related Issues

None

### Check List

### DIFF
--- a/apps/site/components/Common/Partners/PartnerButton/index.module.css
+++ b/apps/site/components/Common/Partners/PartnerButton/index.module.css
@@ -25,6 +25,7 @@
 
   svg {
     @apply h-4
-      w-auto;
+      w-auto
+      max-w-4;
   }
 }
--- a/apps/site/components/Common/Partners/index.module.css
+++ b/apps/site/components/Common/Partners/index.module.css
@@ -5,6 +5,7 @@
     flex-row
     flex-wrap
     items-center
+    justify-center
     gap-2;
 }
 
--- a/apps/site/components/Common/Partners/index.tsx
+++ b/apps/site/components/Common/Partners/index.tsx
@@ -70,7 +70,7 @@ const PartnersList: FC<PartnersListProps> = async ({
 }) => {
   const isSmall = size === 'small';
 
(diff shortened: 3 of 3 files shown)


avivkeller (approved):

avivkeller:
requesting fast track to avoid confusion

bmuenzenmeyer (approved):
Glad this is simple and cosmetic. Good call.

https://github.com/nodejs/nodejs.org/pull/7991#issuecomment-3122156208 and https://github.com/nodejs/nodejs.org/blob/main/docs/partners.md#adding-or-removing-partners make clear that content inclusion-exclusion decisions are not the scope of the website team, and usually come from events like https://github.com/nodejs/nodejs.org/pull/8973

bjohansebas:
That was actually done intentionally when that part was developed. I don't remember the exact details, but I think it was one of the requirements.

PR #8998, opened 2026-07-05
Adds link to beta docs

bmuenzenmeyer:
## Description

Good feedback within https://github.com/nodejs/node/pull/62045 has mentioned that despite social media posts and github issues, most users may have not yet seen the [beta docs](https://beta.docs.nodejs.org/). To expand an ever-increasing feedback gathering campaign, I think we should publicize the beta docs on par with the official docs.

Unsure if this has been proposed before - apologies if I missed it.

Large Light/Dark
image
image

Mobile
image

## Validation

- View docs next to current.
- Should open in new tab

## Related Issues

https://github.com/nodejs/node/pull/62045

### Check List

### DIFF
--- a/packages/ui-components/src/Containers/NavBar/NavItem/index.module.css
+++ b/packages/ui-components/src/Containers/NavBar/NavItem/index.module.css
@@ -45,6 +45,30 @@
     }
   }
 
+  &.accent {
+    @apply bg-brand-100
+      ring-brand-600/20
+      dark:bg-brand-600/15
+      dark:ring-brand-400/20
+      ring-1
+      ring-inset;
+
+    .label {
+      @apply text-brand-700
+        dark:text-brand-400;
+    }
+
+    .icon {
+      @apply text-brand-600
+        dark:text-brand-400;
+    }
+
+    &:hover {
+      @apply bg-brand-200
+        dark:bg-brand-600/25;
+    }
+  }
+
   &.footer {
     .label {
       @apply text-neutral-800
--- a/packages/ui-components/src/Containers/NavBar/NavItem/index.module.css
+++ b/packages/ui-components/src/Containers/NavBar/NavItem/index.module.css
@@ -6,7 +6,8 @@
     gap-2
     rounded-sm
     px-3
-    py-2;
+    py-2
+    motion-safe:transition-colors;
 
   .label {
     @apply text-base
@@ -46,26 +47,15 @@
   }
 
   &.accent {
-    @apply bg-brand-100
-      ring-brand-600/20
-      dark:bg-brand-600/15
-      dark:ring-brand-400/20
+    @apply bg-transparent
       ring-1
-      ring-inset;
-
-    .label {
-      @apply text-brand-700
-        dark:text-brand-400;
-    }
-
-    .icon {
-      @apply text-brand-600
-        dark:text-brand-400;
-    }
+      ring-neutral-200
+      ring-inset
+      dark:ring-neutral-900;
 
     &:hover {
-      @apply bg-brand-200
-        dark:bg-brand-600/25;
+      @apply bg-neutral-200
+        dark:bg-neutral-900;
     }
   }
 
--- a/apps/site/components/withNavBar.tsx
+++ b/apps/site/components/withNavBar.tsx
@@ -54,11 +54,14 @@ const WithNavBar: FC = () => {
       <WithBanner section="index" />
 
       <NavBar
-        navItems={navigationItems.map(([, { label, link, target }]) => ({
-          link,
-          text: label,
-          target,
-        }))}
+        navItems={navigationItems.map(
+          ([, { label, link, target, accent }]) => ({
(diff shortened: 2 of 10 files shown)


bmuenzenmeyer (commented):
if we dont like the `accent` approach and how much it drills through the schema and makes file changes - we could try to key the styling entirely off of the URL value - which would be less intrusive long-term, but sorta brittle

bmuenzenmeyer:
both AI bots are too ill-informed to know how we translate based on en.json only - i'd be game to turn off cursor entirely

bmuenzenmeyer:
lol those functional tests are really banking on this being up to date in crowdin. brittle

AugustinMauroy:
okay 2l point:
* we should add small blog post to say "hey we are working on this ... it's beta so link will disapear"
* on the beta doc we didn't say where to report issue 

So in overall I think most user will not get what happened

AugustinMauroy:
about accent I found that cool but with current style it's confusing with active 



so In my opinion we play with an hover effect

packages/ui-components/src/Containers/NavBar/NavItem/index.module.css
```diff
@@ -45,6 +45,30 @@
     }
   }
 
+  &.accent {
+    @apply bg-brand-100
+      ring-brand-600/20
+      dark:bg-brand-600/15
+      dark:ring-brand-400/20
+      ring-1
+      ring-inset;
```
AugustinMauroy: we should add small transition because now on hover it's pass from something transparent to solid instantly

avivkeller:
I'm not sure how i feel about this. The beta docs aren't 1:1 with the new docs, and only serve as a visual preview of the latest `main` changes from node, and this gives the impression that they are 1:1, no?

araujogui (commented):
maybe should we start with a announcement banner only first?

bjohansebas:
> I'm not sure how i feel about this. The beta docs aren't 1:1 with the new docs, and only serve as a visual preview of the latest main changes from node, and this gives the impression that they are 1:1, no?

At the very least, it should be updated to the latest version, which is Node.js 26, right? The preview is still on Node.js 24.

> maybe should we start with a announcement banner only first?

I think that's the best approach for now

avivkeller:
> At the very least, it should be updated to the latest version, which is Node.js 26, right? The preview is still on Node.js 24.

The preview is purely visual (at the moment), and isn't populated with the correct data, for example, it's _generated_ with the doc files of the main branch of `nodejs/node`, but _claims_ it was generated for Node.js v24.x, since we set the `version` value as the latest LTS.

At the very least, the beta docs should be updated to generate with the latest tag on the `nodejs/node` repo, and correctly use that tag name (e.g. `v26.0.0`) as the version, rather than using `process.version`.

That being said, none of this is an 'issue' with doc-kit, it's just that our preview was set up in a way to showcase the visuals, and not in a way to make them accurate to readers.

bmuenzenmeyer:
all good points - i am trying to nudge us toward more exposure here. it helps us get more feedback from those not following along. i think its the right pathway, after 2+years

- [x] resolve beta.docs.nodejs.org deployment https://github.com/nodejs/doc-kit/pull/888
- [x] iterate on nav design vs banner via https://github.com/nodejs/nodejs.org/pull/8998/commits/e890574d7a9ea4e776934742b08340df3c12d20f now looks like secondary button

image


- [x] banner? it cannot happen before render (that one is active). i want something in place prior to that (resolved, use the blog and nav, no banner)
- [x] blog post https://github.com/nodejs/nodejs.org/pull/9002
   - [x] publicize feedback mechanisms

> The preview is purely visual

I'd disagree - the design is the user experience

bjohansebas:
I think we should add a call to action on the beta page itself—maybe in the sidebar or somewhere visible, letting users know that if they have feedback, they should submit it to a dedicated issue o something

bmuenzenmeyer:
> I think we should add a call to action on the beta page itself—maybe in the sidebar or somewhere visible, letting users know that if they have feedback, they should submit it to a dedicated issue o something


This means we need to impact the actual Node.js source code, or devise a way within doc-kit (or config) to inject extra content.

avivkeller:
We can pass a custom remote config that shows a banner?

bjohansebas:
> This means we need to impact the actual Node.js source code, or devise a way within doc-kit (or config) to inject extra content.

Maybe something similar to what the webpack redesign did to show sponsors in the sidebar? We wouldn't need to change anything in Node itself—just the configuration we use when deploying to Vercel.

bjohansebas (approved):
LGTM!

bmuenzenmeyer:
This has sat with one approval for over a week - will be releasing end-of-day

araujogui (approved):
LGTM

araujogui:
> This has sat with one approval for over a week - will be releasing end-of-day

We should merge https://github.com/nodejs/nodejs.org/pull/9002 at the same time

PR #8999, opened 2026-07-05
fix: navbar active link should match the most specific path

ryzrr:
## Description

fix navbar active-link matching that only compared the first path segment, so two items sharing a segment could highlight together. now the most specific link (longest matching prefix) wins.

## Validation

run `pnpm test` in packages/ui-components for the new getActiveNavLink and active-override tests. node's own top nav is unchanged since its items are single-segment, this fixes the nested case for package consumers whose nav has items like /about and /about/sponsors.

## Related Issues

none, raised with the maintainers beforehand.

### Check List


avivkeller (approved):

AugustinMauroy (approved):
LGTM!

aymen94 (approved):

PR #9005, opened 2026-07-14
chore(security.txt): make compliant by adding an `Expires` field

kobihikri:
## Description

The published security.txt at `https://nodejs.org/.well-known/security.txt` currently contains only:

```
Contact: https://hackerone.com/nodejs
Policy: https://github.com/nodejs/node/security/policy#security
```

[RFC 9116 §2.5.5](https://www.rfc-editor.org/rfc/rfc9116#section-2.5.5) states the `Expires` field **MUST** be present. Without it, a scanner or security researcher following the RFC has no signed expiry to validate the file against (some tooling flags a `security.txt` with no `Expires` as non-conformant).

This PR adds the single missing field:

```diff
 Contact: https://hackerone.com/nodejs
+Expires: 2027-06-30T23:59:59.000Z
 Policy: https://github.com/nodejs/node/security/policy#security
```

I set the value roughly a year out. Since a static date will itself lapse eventually, you may want to automate the refresh — glad to wire that up in a follow-up, but I kept this change scoped to the one missing mandatory field.

## Check List

- [x] I have read the [Contributing Guidelines](https://github.com/nodejs/nodejs.org/blob/main/CONTRIBUTING.md) and made my commit messages conform to the DCO (`Signed-off-by`).
- [x] Change is limited to a single concern.

---

I used AI assistance to help identify and draft this; I verified the RFC reference and the live file myself and take responsibility for the change. Thank you for maintaining nodejs.org.


AugustinMauroy (approved):

bmuenzenmeyer:
@nodejs/security-wg should own this file and its changes

bjohansebas (commented):
Hmm, I think we should have some automation to update that automatically. I feel like that field is going to be forgotten, and we don't want to show that it's expired when all the other information is actually up to date.

cc: @nodejs/web-infra

MattIPv4:
👍 Agreed, a once-per-year cron to open a PR to bump it would be nice. That'll prompt folks to actually review the information is still correct before the date is bumped.

bjohansebas:
@kobihikri Something like that could help you build it https://github.com/expressjs/expressjs.com/blob/main/.github/workflows/update-external-docs.yml#L13-L57

kobihikri:
I can take on this PR, but I’m not familiar with your internal rules. Can I take the PR?

Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________

kobihikri:
I’m on it.

kobihikri:
Good call, and thanks for the pointer @bjohansebas — you're right that a static `Expires` will drift and eventually show as expired.

I've pushed a workflow modeled on the `expressjs.com` one you linked: `.github/workflows/security-txt-expiry.yml`. It runs monthly (and on demand), and when `Expires` is within 30 days it bumps it a year out and opens a PR — deliberately a PR, not a direct commit, so a human confirms the `Contact`/`Policy` are still correct before the extension lands (it cc's `@nodejs/security-wg` for that). It uses the same `gr2m/create-or-update-pull-request-action` pin this repo already uses in `create-release-post.yml`, and pins `actions/checkout` to a SHA.

Happy to adjust the cadence, the labels, or hand the workflow off to `@nodejs/web-infra` if you'd prefer it live somewhere else or be shaped differently — whatever fits the repo's conventions.

kobihikri:
Thanks @MattIPv4 — that's exactly the intent. And good catch on the drift (via the bot review): I've reworked it into a once-a-year cron that always sets `Expires` ~13 months out (a >1-year window, so the file can't lapse even if a run is delayed), instead of the monthly "within 30 days → today+365" check that crept the window earlier each cycle.

It still opens a PR rather than committing directly, so a human re-confirms the `Contact`/`Policy` are still valid before the bump lands, and it cc's `@nodejs/security-wg` per @bmuenzenmeyer's point about ownership. Happy to adjust the cadence further or hand the workflow off to the security-wg to own.

flakey5 (approved):

.github/workflows/security-txt-expiry.yml
```diff
@@ -0,0 +1,55 @@
...
+jobs:
+  refresh-expires:
+    runs-on: ubuntu-latest
+    if: github.repository_owner == 'nodejs'
+    permissions:
+      contents: write
+      pull-requests: write
+    steps:
+      - name: Checkout
+        uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
```
avivkeller: Run `pinact -u` on this file, please

.github/workflows/security-txt-expiry.yml
```diff
@@ -0,0 +1,55 @@
+# Refreshes apps/site/public/.well-known/security.txt so its RFC 9116 Expires
+# field never silently lapses. Runs once a year and opens a PR bumping Expires
+# ~13 months out — a human must confirm the contact/policy are still valid
+# before merging. The generous window (>1yr) means the file never expires even
+# if a run is delayed, and anchoring to the fixed yearly run avoids drift.
```
avivkeller:
```suggestion
# Security Notes
# Only selected Actions are allowed within this repository. Please refer to (https://github.com/nodejs/nodejs.org/settings/actions)
# for the full list of available actions. If you want to add a new one, please reach out a maintainer with Admin permissions.
# REVIEWERS, please always double-check security practices before merging a PR that contains Workflow changes!!
# AUTHORS, please only use actions with explicit SHA references, and avoid using `@master` or `@main` references or `@version` tags.
```

.github/workflows/security-txt-expiry.yml
avivkeller, line 1: Let's call this `bump-security-txt-expiry.yml` so it's clear this _bumps_ it

.github/workflows/security-txt-expiry.yml
```diff
@@ -0,0 +1,55 @@
...
+        with:
+          persist-credentials: false
+
+      - name: Bump Expires (~13 months out)
+        env:
+          SECURITY_TXT: apps/site/public/.well-known/security.txt
+        run: |
+          set -euo pipefail
+          new=$(date -u -d '+400 days' +%Y-%m-%dT23:59:59.000Z)
+          sed -i "s/^Expires:.*/Expires: ${new}/I" "$SECURITY_TXT"
```
avivkeller: As Cursor said, we should bump the in-place year, not re-calculate a date, e.g. `Expiry: \d{4}` gets replaced by `(v) => v + 1`

.github/workflows/security-txt-expiry.yml
```diff
@@ -0,0 +1,55 @@
...
+
+      - name: Create Pull Request (only if the date changed)
```
avivkeller:
```suggestion
      - name: Open pull request
```

Conformity with our other workflows. No need to add the `(...)`, this will always bump the date

.github/workflows/security-txt-expiry.yml
```diff
@@ -0,0 +1,55 @@
...
+
+            **Before merging, please confirm the rest of the file is still accurate:**
+            - The `Contact` (HackerOne) is still the correct, actively-monitored channel
+            - The linked `Policy` is still the intended vulnerability-disclosure path
+
+            If anything has changed, update it in this PR before extending the expiration.
+
+            cc: @nodejs/security-wg
+          labels: security
+          branch: automated/security-txt-expiry
```
avivkeller:
```suggestion
         update-pull-request-title-and-body: true
          branch: chore/update-security-txt
          body: |
            Updates the `Expiry` on our [`security.txt`](https://nodejs.org/.well-known/security.txt] from ${{ steps.update-expiry.outputs.old }} to ${{ steps.update-expiry.outputs.new }}

            Before merging, please explicitly confirm that:
            - The security contact listed is still correct and actively monitored
            - The linked security policy is still accurate
            - The project's preferred vulnerability disclosure path has not changed

            cc @nodejs/security-wg

            Check this workflow's logs at ${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}.
          commit-message: 'meta: update security `Expiry` from ${{ steps.update-expiry.outputs.old }} to ${{ steps.update-expiry.outputs.new }}'
          title: 'meta: update security `Expiry` from ${{ steps.update-expiry.outputs.old }} to ${{ steps.update-expiry.outputs.new }}'
          draft: true
```

For conformity

avivkeller:
Bump @kobihikri!

kobihikri:
Sorry for the long silence, @avivkeller — your review landed after my last reply and I simply missed it. That's on me. All six points are addressed in 9aab000:

- Renamed to `bump-security-txt-expiry.yml`.
- Swapped in the standard `# Security Notes` header.
- `pinact -u`: `actions/checkout` was pinned to v7.0.0, now v7.0.1 (`3d3c42e`). `gr2m/create-or-update-pull-request-action` was already at the latest v1.10.1, so it is unchanged.
- The bump now increments the year in place rather than recomputing a date, so the month/day never drift. I ran it three times in a row against the real file to check: `2027-06-30 → 2028-06-30 → 2029-06-30 → 2030-06-30`.
- Step renamed to `Open pull request`.
- Adopted your PR conventions verbatim — `chore/update-security-txt`, `draft: true`, `update-pull-request-title-and-body`, and the `meta: update security \`Expiry\` from … to …` title and commit message.

One thing I should flag rather than quietly decide, since it touches the Bugbot comment about the value exceeding RFC 9116's *"RECOMMENDED … less than a year"*: with a yearly cron **and** an in-place year bump, the value is necessarily more than 12 months out at the moment of the bump. That's arithmetic, not a choice — to land under a year the bump would have to run *after* the old date had already passed. So the two goals can't both hold, and I've kept your in-place bump and left `Expires` where it is.

The knob, if you want one, is the gap between the Jan 1 run and the date itself: today it's June 30, so the bump leaves ~18 months and gives ~6 months for a human to review and merge the automated PR. Moving the date earlier in the year trades merge headroom for a value closer to 12 months. Happy to shift it to whatever you prefer, or leave it as is — your call, and I don't have a strong view.

Disclosure, unchanged from the original post: I used AI assistance (Claude Opus 5) to help draft this workflow. I read and verified every line myself, ran the bump logic against the real file, and take responsibility for it.

avivkeller (approved):

PR #9008, opened 2026-07-14
feat(banner): add `x` for closing

### LINKED ISSUE #6292: Feat(banner): take place on screen, should we add the possibility of hide

avivkeller:
Makes the banner close-able.

**Note**: This just updated `ui-components`, and doesn't touch Next.js + cookie handling
Screenshot 2026-07-14 at 12 15 13 PM

### DIFF
--- a/apps/site/components/withBanner.tsx
+++ b/apps/site/components/withBanner.tsx
@@ -17,28 +17,28 @@ const WithBanner: FC<{ section: string }> = ({ section }) => {
   const banner = siteConfig.websiteBanners[section];
   const t = useTranslations();
 
-  const [dismissed, setDismissed] = useState(false);
+  const [open, setOpen] = useState(false);
 
-  // localStorage is only available on the client, so the banner is always
-  // rendered on the server and hidden after hydration if previously dismissed
   useEffect(() => {
     if (banner) {
-      setDismissed(localStorage.getItem(STORAGE_KEY) === banner.text);
+      // eslint-disable-next-line @eslint-react/set-state-in-effect
+      setOpen(localStorage.getItem(STORAGE_KEY) !== banner.text);
     }
   }, [banner]);
 
-  if (banner && !dismissed && dateIsBetween(banner.startDate, banner.endDate)) {
+  if (banner && open && dateIsBetween(banner.startDate, banner.endDate)) {
     const bannerType = banner.type || 'default';
 
     const onClose = () => {
       localStorage.setItem(STORAGE_KEY, banner.text);
-      setDismissed(true);
+      setOpen(false);
     };
 
     return (
       <Banner
         type={banner.type}
         aria-label={t(`components.banner.${bannerType}`)}
+        closeButtonAriaLabel={t('components.banner.close')}
         onClose={onClose}
       >
         {banner.link ? (
--- a/apps/site/components/withBanner.tsx
+++ b/apps/site/components/withBanner.tsx
@@ -1,6 +1,9 @@
+'use client';
+
 import { ArrowUpRightIcon } from '@heroicons/react/24/outline';
 import Banner from '@node-core/ui-components/Common/Banner';
 import { useTranslations } from 'next-intl';
+import { useEffect, useState } from 'react';
 
 import Link from '#site/components/Link';
 import { siteConfig } from '#site/next.json.mjs';
@@ -8,17 +11,35 @@ import { dateIsBetween } from '#site/util/date';
 
 import type { FC } from 'react';
 
+const STORAGE_KEY = 'banner-dismissal';
+
 const WithBanner: FC<{ section: string }> = ({ section }) => {
   const banner = siteConfig.websiteBanners[section];
   const t = useTranslations();
 
-  if (banner && dateIsBetween(banner.startDate, banner.endDate)) {
+  const [dismissed, setDismissed] = useState(false);
+
+  // localStorage is only available on the client, so the banner is always
+  // rendered on the server and hidden after hydration if previously dismissed
+  useEffect(() => {
+    if (banner) {
+      setDismissed(localStorage.getItem(STORAGE_KEY) === banner.text);
+    }
+  }, [banner]);
+
+  if (banner && !dismissed && dateIsBetween(banner.startDate, banner.endDate)) {
     const bannerType = banner.type || 'default';
+
+    const onClose = () => {
+      localStorage.setItem(STORAGE_KEY, banner.text);
+      setDismissed(true);
+    };
 
     return (
       <Banner
         type={banner.type}
         aria-label={t(`components.banner.${bannerType}`)}
+        onClose={onClose}
       >
         {banner.link ? (
           <Link href={banner.link}>{banner.text}</Link>
--- a/packages/ui-components/src/Common/Banner/index.tsx
+++ b/packages/ui-components/src/Common/Banner/index.tsx
@@ -22,7 +22,9 @@ const Banner: FC<PropsWithChildren<BannerProps>> = ({
   ...props
 }) => {
   const [dismissed, setDismissed] = useState(false);
-  if (dismissed) {return null;}
+  if (dismissed) {
+    return null;
+  }
 
   return (
     <section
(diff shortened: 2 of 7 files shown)


packages/ui-components/src/Common/Banner/index.tsx
```diff
@@ -1,22 +1,48 @@
...
-  <section
-    className={`${styles.banner} ${styles[type] || styles.default}`}
-    {...props}
-  >
-    {children}
-  </section>
-);
+}) => {
+  const [dismissed, setDismissed] = useState(false);
+  if (dismissed) {return null;}
```
MattIPv4: I don't think the banner needs to do its own state management, let the consumer of the component handle that with the `onClose`

packages/ui-components/src/Common/Banner/index.tsx
```diff
@@ -1,22 +1,48 @@
...
+        type="button"
+        className={styles.closeButton}
+        aria-label="Close banner"
+        onClick={e => {
+          setDismissed(true);
+          onClose?.(e);
+        }}
+      >
+        <span aria-hidden="true">&times;</span>
+      </button>
```
MattIPv4: To the above point of leaving this to the consumer, let's only show the button when an `onClose` is passed in

packages/ui-components/src/Common/Banner/index.tsx
avivkeller: We could, however, that would require duplicating the state logic in doc-kit, and here

MattIPv4: I think that's fine, as we're going to need to tie this to a cookie/localStorage anyway

bjohansebas (commented):
I think that if we ever update the banner and a user previously dismissed it, we should show it to them again

apps/site/components/withBanner.tsx
```diff
@@ -1,24 +1,45 @@
...
   const t = useTranslations();
 
-  if (banner && dateIsBetween(banner.startDate, banner.endDate)) {
+  const [dismissed, setDismissed] = useState(false);
+
+  // localStorage is only available on the client, so the banner is always
+  // rendered on the server and hidden after hydration if previously dismissed
+  useEffect(() => {
+    if (banner) {
+      setDismissed(localStorage.getItem(STORAGE_KEY) === banner.text);
```
AugustinMauroy: there are a thing here localstroage is shared with api docs but not with sub domain like `undici.nodejs.org`

AugustinMauroy:
there are a flash humm not fan of it maybe not display it by default. also what will happen on doc-kit since it's already flash when fetching site.json

apps/site/components/withBanner.tsx
bjohansebas: Then I guess a cookie would solve that, right?

trivikr:
To avoid flash, can we dismiss it by default and show only if the value is not set?

<details>
<summary>Screen recording of flash on reload post dismiss</summary>

https://github.com/user-attachments/assets/97175002-2a5b-43e8-97c7-e12ca32c7a5f

</details>

trivikr:
Fixes: https://github.com/nodejs/nodejs.org/issues/6292

apps/site/components/withBanner.tsx
avivkeller: I'm not a fan of accessing the cookie API from the client, IMHO localStorage is meant for client side editing in KV, cookies aren't

avivkeller:
@trivikr either we flash on the way in, or the way out (fwiw we can make it a fade)

The way I see it, if we fade on the way in, we fade for all website viewers.

If we fade on the way out, we only fade for a small subset of viewers who have opted in to hiding this banner

trivikr:
I'm sorry for the git push. I wasn't aware that I have push permissions on upstream.
I removed my experimental commit, and did force push to get back this PR in the previous state.

---

UPDATE: Looks like force-push did not update the website preview to previous version.
I'll wait for the author to respond and won't take any other action.

trivikr:
With assistance from openai:gpt-5.6-sol, I attempted going prehydration in https://github.com/nodejs/nodejs.org/pull/9009. Would that help fix the flashing issue?

avivkeller:
No, unfortunately. While I appreciate the time and tokens spent trying to help, there's several layers to how we render our components currently, including:
1. Server-Side Next.js Rendering
2. Static-Side `rolldown` Bundling (doc-kit)
3. Client-Side Hydration (here and doc-kit, seperately)

Attaching a `script` element is a extremely hacky way (and one we can't accept for a number of reasons) to resolve the issue on one client (here), it wouldn't resolve it on the other clients. Ultimately, we can't predict the value of `localStorage`, so we need to either **fade/flash the banner in** or **fade/flash the banner out**.

In `doc-kit`, we are forced to fade the banner in no matter what (due to the fetching of banner content) so, the solution is already in place there. For consistency, I suppose, we can adjust this PR to also fade the banner in.

Would that satisfy your concern?

avivkeller:
I've manually re-deployed the previous commit to match Vercel's branch preview with the correct commit.

trivikr:
> we can adjust this PR to also fade the banner in.
> 
> Would that satisfy your concern?

Sure.

apps/site/components/withBanner.tsx
MattIPv4: 👍 localStorage seems fine to me, I think it's acceptable (and perhaps desired) that different domains (e.g. undici) will have their own state.

MattIPv4:
> In `doc-kit`, we are forced to fade the banner in no matter what (due to the fetching of banner content) so, the solution is already in place there. For consistency, I suppose, we can adjust this PR to also fade the banner in.

++ I think I would actually prefer that we do fade in everywhere, for consistency and to clear the way for this potentially being fully dynamic in the future instead of needing a site rebuild every time it is changed.

apps/site/components/withBanner.tsx
```diff
@@ -1,24 +1,45 @@
...
 
-  if (banner && dateIsBetween(banner.startDate, banner.endDate)) {
+  const [open, setOpen] = useState(false);
+
+  useEffect(() => {
+    if (banner) {
+      // eslint-disable-next-line @eslint-react/set-state-in-effect
+      setOpen(localStorage.getItem(STORAGE_KEY) !== banner.text);
+    }
+  }, [banner]);
```
MattIPv4, line 27:
```suggestion
  const [open, setOpen] = useState(() => !!banner && localStorage.getItem(STORAGE_KEY) !== banner.text);
```
Can we avoid the effect + lint disable here?

trivikr (approved):

apps/site/components/withBanner.tsx
avivkeller: No, unfortunately, Next.js attempts to evaluate that on the server side, causing an error

MattIPv4: Ah, surely we can just check for window in that then? Or will the state not get re-evaluated on the client?

```suggestion
  const [open, setOpen] = useState(() => !!banner && typeof window !== 'undefined' && window.localStorage.getItem(STORAGE_KEY) !== banner.text);
```

avivkeller: If I understand correctly, that'll cause a hydration mismatch, since on the server, it'll be `false`, and on the client, it'll be either `false` or `true`.

MattIPv4: 🤔 Given this is marked as a client component I would hope not, but maybe? If not, no worries, the effect is fine, just wanted to clean it up if we could given the eslint disable

AugustinMauroy (approved):

bmuenzenmeyer (approved):

PR #9016, opened 2026-07-21
chore(remark-lint): improve for node core

avivkeller:
Improves the `remark-lint` package by:
- Fixing an issue with type validation where it would include comments, code blocks, etc
- Adding link validation, as mentioned by @aduh95


bjohansebas (approved):

avivkeller:
Bump @nodejs/web-infra

packages/remark-lint/src/__tests__/index.test.mjs
```diff
@@ -0,0 +1,40 @@
...
+
+import basePreset from '../index.mjs';
+
+describe('base preset', () => {
+  it('reports broken local links', async () => {
+    const linkValidator = basePreset.plugins.find(
+      plugin => plugin === remarkValidateLinks
+    );
+
+    assert.equal(linkValidator, remarkValidateLinks);
```
bmuenzenmeyer: sorta seems like an odd assertion on the road to the real test

bmuenzenmeyer (commented):
there's a cursor concern here `Link validator lacks needed options` that is perhaps valid. can you validate the real runtime scenarios here? it'd be bad if this resulted in false-positivies

packages/remark-lint/src/__tests__/index.test.mjs
avivkeller: Agreed. I added it to make sure the link validator was working locally, but it's probably not needed for committing.

avivkeller:
Replaced `remark-validate-links` with our own validator to allow for:
- Node.js-style slugs
- Path exclusions

bjohansebas (approved):

PR #9020, opened 2026-07-22
chore(ci): add changesets + ci validation

avivkeller:
As we did in `doc-kit`.

Note that I added a require changesets + conventional commits not in doc-kit. I can remove them, but I feel that as we continously patch the UI, we may (I especially) forget to add changesets so it's good to have a reminder

The check will be non-required, so it's more of a "reminder" than a hard block

### DIFF
--- a/.changeset/README.md
+++ b/.changeset/README.md
@@ -1,0 +1,11 @@
+# Changesets
+
+Changesets in this directory describe user-facing changes to published packages. Add one by running:
+
+```shell
+pnpm changeset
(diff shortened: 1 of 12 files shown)


avivkeller:
cc @bmuenzenmeyer

bmuenzenmeyer (approved):
glad the conventional commits are non-blocking and only in PR titles, we have gone down paths like that before and it absolutely kills casual contribution.

PR #9022, opened 2026-07-27
Blog: announce sec release postponed

RafaelGSS:


MattIPv4 (approved):

RafaelGSS:
We no longer should use `Blog:`?

MattIPv4:
> We no longer should use `Blog:`?

Conventional commits check is non-blocking. Part of #9020, for the versioned packages that are in this repo.

avivkeller (approved):

avivkeller:
I've opened https://github.com/nodejs/nodejs.org/pull/9025 to only run this (still non-blocking) check in `packages/`, to avoid confusion

PR #9028, opened 2026-07-28
chore(ui): account for nested `code` elements

avivkeller:
Ref: https://github.com/nodejs/doc-kit/pull/951


bjohansebas (approved):

AugustinMauroy (approved):

bmuenzenmeyer (approved):
it'd be nice to see the before and after of things like this - i know thats hard when we are cross-repo, unpublished, etc... seems reasonable

PR #9032, opened 2026-07-28
chore(ui): improve width of containers

avivkeller:
Partially fix/es https://github.com/nodejs/nodejs.org/issues/9081 by expanding the main content (if possible) and the change history

### DIFF
--- a/.changeset/public-baths-open.md
+++ b/.changeset/public-baths-open.md
@@ -1,0 +1,5 @@
+---
+'@node-core/ui-components': patch
+---
+
+Improve the width of ChangeHistory
--- a/packages/ui-components/src/Common/ChangeHistory/index.module.css
+++ b/packages/ui-components/src/Common/ChangeHistory/index.module.css
@@ -32,7 +32,8 @@
     z-50
     mt-1
     max-h-80
-    w-52
+    w-[26rem]
+    max-w-[calc(100vw-2rem)]
     overflow-hidden
     rounded-sm
     border
@@ -45,7 +46,7 @@
 
 .dropdownContentInner {
   @apply max-h-80
-    w-52
+    w-full
     overflow-y-auto;
 }
 


packages/ui-components/src/Containers/Article/index.module.css
```diff
@@ -1,14 +1,12 @@
 @reference "../../styles/index.css";
 
 .articleLayout {
-  @apply max-w-10xl
```
bjohansebas, line 4: Text should have a maximum line width of **80ch** for better readability and UX: [https://baymard.com/blog/line-length-readability](https://baymard.com/blog/line-length-readability)

bjohansebas: https://web.dev/articles/min-max-clamp#width

AugustinMauroy:
Can we have stories for that change so if there are future change we will know what happened

avivkeller:
> Can we have stories for that change so if there are future change we will know what happened

I don't know if stories will cover this kind of dynamic change

packages/ui-components/src/Containers/Article/index.module.css
avivkeller: I can either:
1) Revert this part of the change
2) Make prose use up to `80ch`, but tables and code blocks still expand

davidje13: FWIW as the person who requested an increase in width:

- From my own experiments swapping out the styles in the documentation, I find it more readable and more skimmable when wider
- Even just a small increase of `6rem` to the total page width made it significantly more comfortable to read (see the screenshots at the top of the issue I raised: https://github.com/nodejs/doc-kit/issues/954#issue-5003228241)
- The oft-cited recommended width limit was figured out for long-form content of plain text, but the documentation mostly contains relatively short paragraphs and is full of inline code and indented list blocks
- The existing content is already quite a bit wider than 80ch, and reducing it to that size makes it very hard to skim (makes it feel a lot more like reading a book, but this is not a book!)
- This only affects relatively large screens, which are presumably biased towards people who prefer seeing more content at a time (like me!). On regular/smaller screens, the content is limited by the presence of the sidebars
- If people on larger screens prefer narrower content, they can easily increase the page zoom or reduce the window size using normal browser/OS controls. But the reverse is not true when size limits are coded into the page.

I wonder if maybe there are competing preferences due to this style being used for both the API documentation and other pages? I could imagine non-API pages may be more readable when narrow, so it might be worth foregoing some consistency to have each page more tuned to its own content. But as I noted above, even if all pages are allowed to be larger, users can always make the content narrower if desired using the browser and OS controls.

avivkeller: Bump @nodejs/web wdyt?

ovflowd: I believe following well documented and tested practices rather than one single individual preference must be the guidance here, as much as I personally relate to @davidje13's preference.

ovflowd (changes requested):
We should follow standards. Document @bjohansebas's comment explicitly.

packages/ui-components/src/Containers/Article/index.module.css
bjohansebas: Hmm, from my perspective, I think having a width limit is generally a good thing. Even Astro does it, and that's one of the projects I look to the most for inspiration. Limiting the text width also helps reduce how far your eyes have to travel, which in practice makes reading less tiring. While a wider layout may let you read more horizontally, it also makes it easier to lose your place in the content.

avivkeller: Reverted that part of the PR

PR #9049, opened 2026-07-30
chore(deps): change dependabot prefix

avivkeller:
Makes Dependabot follow Conventional Commits

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -10,7 +10,7 @@ updates:
       - 'github_actions:pull-request'
       - 'auto-merge'
     commit-message:
-      prefix: meta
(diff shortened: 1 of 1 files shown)


MattIPv4 (approved):

AugustinMauroy (approved):

bmuenzenmeyer (approved):
i dont looove this cause a chore usually maps to patch

avivkeller:
> i dont looove this cause a chore usually maps to patch

Dependabot doesn't generate changesets anyway, so a dependabot bump would already be a "patch" (well, more accurately, it'll be excluded from the changelogs entirely, and will release with whatever version is scheduled next)

PR #9050, opened 2026-07-30
fix: FunctionSignature return type border

canerakdas:
## Description

The bordered wrapper component should not be rendered in the `FunctionSignature` component when only the return type is present.

Ref: https://openjs-foundation.slack.com/archives/CVAMEJ4UV/p1785348769820369

## Validation

The `FunctionSignature` "Return Type" variant should now be fixed in Chromatic, and the other variants shouldn't have any regressions.

### DIFF
--- a/packages/ui-components/src/Common/Signature/SignatureItem/index.tsx
+++ b/packages/ui-components/src/Common/Signature/SignatureItem/index.tsx
@@ -18,6 +18,7 @@ const SignatureItem: FC<PropsWithChildren<SignatureItemProps>> = ({
   children,
 }) => (
   <div
+    data-kind={kind}
     className={classNames(styles.item, {
       [styles.return]: kind === 'return',
     })}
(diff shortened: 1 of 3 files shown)


packages/ui-components/src/Common/Signature/SignatureItem/index.tsx
```diff
@@ -18,6 +18,7 @@ const SignatureItem: FC<PropsWithChildren<SignatureItemProps>> = ({
   children,
 }) => (
   <div
+    data-kind={kind}
     className={classNames(styles.item, {
       [styles.return]: kind === 'return',
```
avivkeller, line 23: Can we use the `styles.return` rather than a data attribute?

canerakdas: Because we use module.css, we can't target `SignatureItem`'s module class from `SignatureRoot` in CSS (the names get rewritten, so each module's .return ends up with a different name).

avivkeller (approved):

PR #9051, opened 2026-07-30
fix(pnpm): warn > error

avivkeller:
Don't fail CI when Changesets uses `npm publish`.

cc @nodejs/web-infra fast-tracking

### DIFF
--- a/package.json
+++ b/package.json
@@ -65,7 +65,7 @@
     "packageManager": {
       "name": "pnpm",
       "version": "11.9.0",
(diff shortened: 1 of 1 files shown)


bjohansebas (approved):

MattIPv4 (changes requested):
This has significant implications for the DX of folks working with this repo. Let's get changesets using the package manager we choose to use for this repo, rather than changing how we handle package managers for every interaction with the repo.

PR #9084, opened 2026-08-05
chore: remove Codemod from partners list

mcollina:
After a brief check with @rginn, I can confirm there is no signed agreement for including the Codemod logo in our website. This is true for all other vendors listed.

Therefore, I'm removing it.

### DIFF
--- a/apps/site/public/static/partners/constants.json
+++ b/apps/site/public/static/partners/constants.json
@@ -54,12 +54,5 @@
     "href": "https://tuxcare.com",
     "weight": 2,
     "categories": ["esp"]
-  },
-  {
-    "id": "CODEMOD",
-    "name": "Codemod",
-    "href": "https://codemod.com",
-    "weight": 1,
-    "categories": []
   }
 ]


AugustinMauroy:
could you also include https://github.com/nodejs/nodejs.org/tree/main/packages/ui-components/src/Icons/PartnerLogos/Codemod in the removal

ovflowd (approved):
How was this added then back then? Did someone verify they had an agreement with us?

AugustinMauroy:
I think we added after https://openjsf.org/blog/codemod-openjs-foundation-partner but we didn't verify that part

mcollina:
> could you also include https://github.com/nodejs/nodejs.org/tree/main/packages/ui-components/src/Icons/PartnerLogos/Codemod in the removal 

we have a lot of other logos that are not in use

bmuenzenmeyer:
> How was this added then back then? Did someone verify they had an agreement with us?

https://github.com/nodejs/nodejs.org/pull/8266

retrospect (blameless), we should have attained @nodejs/marketing before merge - we have gotten a lot better about that of late (IMO)

bmuenzenmeyer (approved):

avivkeller:
Merging as a foundation request

mcollina:
Can we land this and then do the cleanup in a followup?

PR #9086, opened 2026-08-05
feat(blog): add more codemod to migrations guides

AugustinMauroy:
## Description

As titled

## Related Issues

No related issues 

### Check List


AugustinMauroy:
the should be keep it's update blog post with new covered deprecation

avivkeller:
These blog posts shouldn't change, they are point-in-time

AugustinMauroy:
ideally yes, but it's mean we post article only when we cover 100% of version deprecation which is take time.
and also in same idea we should have all codemod ready before LTS is landed but we have lack of maintainer

avivkeller:
> ideally yes, but it's mean we post article only when we cover 100% of version deprecation which is take time.
and also in same idea we should have all codemod ready before LTS is landed but we have lack of maintainer

In that case, a global "v22-to-v24" codemod should exist and be updated, not the blog.

bmuenzenmeyer:
am i remembering correctly that we want to pivot all of these to something more durable than a blog post?

PR #9089, opened 2026-08-06
fix: hide incompatible download options

### LINKED ISSUE #8638: Remove irrelevant dropdown options for installing Node.js rather than disabling them

On https://nodejs.org/en/download, the irrelevant options for downloading Node.js are currently shown in the dropdowns as disabled items:

Screenshot 1

This creates ambiguity for new users who are not familiar with these technologies. E.g., Chocolatey being disabled for Linux could imply that it's supported but not endorsed by Node.js.

And I believe that AIX needs to be removed from this list:

Screenshot 2

Even though AIX is enabled for Node versions 6+, it seems that all installation options for it are disabled. The AIX option is disabled for Node versions below 6. So, there is no way to download Node.js on AIX with the recommended or community methods.

ghost:
## Description

- hide incompatible installation methods, package managers, platforms, and operating systems instead of rendering them as disabled options
- exclude AIX from the installation-method selector while keeping it available for official prebuilt binaries
- add unit coverage for compatibility filtering and explicit exclusions

AIX remains available in the prebuilt selector because Node.js publishes official `aix-ppc64` binaries.

## Validation

- `git diff --check`
- TypeScript syntax checks for the changed TSX files
- targeted behavior checks for Linux/Chocolatey and AIX install/prebuilt visibility
- full `pnpm format`, `pnpm test`, and `pnpm build` were not run because dependencies are not installed in the bundle environment

## Related Issues

Fixes #8638.

### Check List

### DIFF
--- a/apps/site/components/Downloads/Release/InstallationMethodDropdown.tsx
+++ b/apps/site/components/Downloads/Release/InstallationMethodDropdown.tsx
@@ -5,7 +5,11 @@ import { useTranslations } from 'next-intl';
 import { use, useEffect, useMemo } from 'react';
 
 import { ReleaseContext } from '#site/providers/releaseProvider';
-import { nextItem, INSTALL_METHODS, parseCompat } from '#site/util/download';
+import {
+  filterDropdownItems,
+  nextItem,
+  INSTALL_METHODS,
+} from '#site/util/download';
 
 import type { InstallationMethod } from '#site/types/release';
 import type { FC } from 'react';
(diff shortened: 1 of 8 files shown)


avivkeller (approved):

AugustinMauroy (approved):

ovflowd (changes requested):
Incompatible options are already disabled. I'm -1 on hiding them.

ghost:
Thanks for the feedback. I implemented the behavior proposed in #8638, but I understand the preference to keep incompatible options visible and disabled. Since #8638 is now closed as not planned, should I close this PR, or would a narrower AIX-only change still be useful?

PR #9093, opened 2026-08-09
chore(qol): CSS changes

### LINKED ISSUE #9081: Feedback: request for various minor style changes

I've been poking around the new documentation for a little while and I'd like to request a couple of small quality-of-life improvements to the styling. I think these are all too small to be individual issues in their own right, so here's an amalgamation of all the styling stuff I've found:

- can the maximum page width be expanded a bit? `110rem` seems to work well. The current `104rem` feels rather squashed on a large display, especially compared to the unlimited size of the old docs. (and currently there is slightly more content space if the window is 1919 pixels than if it is wider)

  **current:**
  Current document width

  **proposed:**
  Proposed document width

- the right-hand side-bar makes the page a bit "busy". Removing the bold styling helps make it less visually intrusive (and also makes the text vs. code links more consistent):

  **current:**
  Bold sidebar text

  **proposed:**
  Regular sidebar text

- with the new busier page, I sometimes find myself getting "lost" among the headings (especially after following links, which leave a small margin at the top for the page header, but it does not appear, so a line of text from the previous section is visible). Adding a small amount of top margin to all headings seems to help a lot:

  **current:**
  Page with small gaps between headers

  **proposed:**
  Page with large gaps between headers

- the scrolling index on the left should have more bottom padding, so that the browser's link preview does not obscure the bottom element when scrolled fully to the bottom:

  Bottom link obscured by link preview

  (personally I like being able to over-scroll areas like this by quite some margin, but even an extra 20 pixels would be enough to avoid the obscuring)

- the "history" drop-downs (which I personally use quite a lot in the current documentation, because I support all supported versions of Node.js in my libraries, so need to know how old methods are / if they have changed) is _very_ squashed. Doubling the width makes it much more readable:

  **current:**
  Narrow history

  **proposed:**
  Wide history

  (there's also some minor inconsistency in the highlighting style of history items; links do not go white on hover, so the first item usually appears different compared to the others when hovering)

- it's tracked elsewhere, but a "real" light mode would be very welcome! (#940)

---

Lest this be taken as a purely negative review, there's also a lot I like about the new design. The new display for function parameters and return types is clearer at-a-glance, and the always-available table-of-contents is useful (in fact it might be nice to give it more space by default; personally I don't find any value in things like "reading time" and the other details above the ToC could perhaps be moved beneath it). For the most part the suggestions here are just about small increases to margins & sizes.

avivkeller:
Fixes: #9081 (see it for the before images)

1. Longer `ChangeHistory`
Screenshot 2026-08-09 at 5 08 24 PM

2. Sidebar padding before footer
Screenshot 2026-08-09 at 5 09 19 PM

3. No bold in MetaBar
Screenshot 2026-08-09 at 5 09 44 PM

### DIFF
--- a/packages/ui-components/src/styles/theme.css
+++ b/packages/ui-components/src/styles/theme.css
@@ -7,7 +7,7 @@
 @theme {
   --container-8xl: 88rem;
   --container-9xl: 96rem;
-  --container-10xl: 104rem;
+  --container-10xl: 110rem;
   --container-11xl: 112rem;
   --container-12xl: 120rem;
   --breakpoint-ml: 896px;
--- a/.changeset/large-pandas-refuse.md
+++ b/.changeset/large-pandas-refuse.md
@@ -1,0 +1,5 @@
+---
+'@node-core/ui-components': patch
+---
+
+Quality-of-life style improvements
--- a/packages/ui-components/src/Common/ChangeHistory/index.module.css
+++ b/packages/ui-components/src/Common/ChangeHistory/index.module.css
@@ -32,7 +32,8 @@
     z-50
     mt-1
     max-h-80
-    w-52
+    w-[26rem]
+    max-w-[calc(100vw-2rem)]
     overflow-hidden
     rounded-sm
     border
@@ -45,7 +46,7 @@
 
 .dropdownContentInner {
   @apply max-h-80
-    w-52
+    w-full
     overflow-y-auto;
 }
 
@@ -64,7 +65,7 @@
   &:hover,
   &:focus-visible {
     @apply bg-brand-600
-      text-white;
+      text-white!;
   }
 }
 
--- a/packages/ui-components/src/Common/TableOfContents/index.module.css
+++ b/packages/ui-components/src/Common/TableOfContents/index.module.css
@@ -41,7 +41,6 @@
   .link {
     @apply inline-block
       text-sm
-      font-semibold
       text-neutral-900
       underline
       hover:text-neutral-700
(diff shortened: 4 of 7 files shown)


AugustinMauroy (approved):

AugustinMauroy:
not blocking but can be cool to be sure this pr fix it, is updating the metbar stories to include badge+popover

ovflowd:
@avivkeller could you add screenshots of before and after and describe better what you changed on PR description?

avivkeller:
> @avivkeller could you add screenshots of before and after and describe better what you changed on PR description?

See the issue body of #9081

ovflowd:
> > @avivkeller could you add screenshots of before and after and describe better what you changed on PR description?
> 
> See the issue body of #9081

I understand, but I'm kindly asking an before/after as the issue doesn't show before/after :)

packages/ui-components/src/styles/theme.css
```diff
@@ -7,7 +7,7 @@
 @theme {
   --container-8xl: 88rem;
   --container-9xl: 96rem;
-  --container-10xl: 104rem;
+  --container-10xl: 110rem;
```
ovflowd, line 10: @bjohansebas are you OK with this change?

ovflowd: Also, @avivkeller is container-10xl used anywhere else? And then 11xl becomes very close from 10xl, I wonder if this should be better adjusted as leaving 10xl as is, changing 11xl maybe?

avivkeller:
Fast-tracking to get this out before RenderATL

ovflowd (approved):

bjohansebas (approved):

PR #9107, opened 2026-08-14
feat(mdx): add mermaid diagram support

### LINKED ISSUE #7540: Mermaid diagram support

ashrees:
Adds Mermaid diagram support to the MDX pipeline, as requested in #7540.

## How it works

- `apps/site/mdx/plugins.mjs`: `rehype-mermaid` (strategy `pre-mermaid`) runs in the rehype chain before `@node-core/rehype-shiki`, so ```` ```mermaid ```` fenced blocks become `<pre class="mermaid">` instead of being syntax-highlighted as plain code. This strategy is AST-only and never starts a browser at build time (verified against the dependency source and benchmarked, see below).
- `apps/site/components/MDX/CodeBox` (the `pre` MDX override) routes `<pre class="mermaid">` to the Mermaid component; all other code blocks render as before.
- The Mermaid component lives in `@node-core/ui-components` (`packages/ui-components/src/MDX/Mermaid`) and renders diagrams client-side through a dedicated `useMermaid` hook (`packages/ui-components/src/hooks/useMermaid.ts`) backed by the `@mermanjs/web` WASM renderer. The hook lazy-loads the renderer only when a diagram is on the page, and accepts an injectable `loader` so unit tests can stub the WASM module.
- The site-level wrapper (`apps/site/components/MDX/Mermaid`) maps the `next-themes` resolved theme onto the component (`light`/`dark` host theme).
- Invalid diagram sources fall back to rendering the raw source in a `<pre>`.

## Testing

- `packages/ui-components/src/hooks/__test__/useMermaid.test.jsx`: 2/2 pass (node:test + testing-library, stubbed loader).
- Storybook stories at `packages/ui-components/src/MDX/Mermaid/index.stories.tsx` (flowchart, sequence, dark, invalid source).
- Benchmark: compiled 25 real blog posts through the production rehype chain, with vs without the plugin, 3 warmed runs: 5.62 vs 5.19 ms/doc, no measurable build-time impact.
- eslint/stylelint pass on all changed files.

Fixes #7540

### DIFF
--- a/apps/site/components/MDX/CodeBox/index.tsx
+++ b/apps/site/components/MDX/CodeBox/index.tsx
@@ -1,4 +1,5 @@
 import { getLanguageDisplayName } from '@node-core/rehype-shiki';
+import Mermaid from '@node-core/ui-components/MDX/Mermaid';
 
 import CodeBox from '#site/components/Common/CodeBox';
 
@@ -8,6 +9,12 @@ const MDXCodeBox: FC<HTMLAttributes<HTMLElement>> = ({
   children: code,
   className,
 }) => {
+  // Mermaid diagrams arrive as `<pre class="mermaid">` (see rehype-mermaid),
+  // so we render them as diagrams instead of code boxes
+  if (className?.split(' ').includes('mermaid')) {
+    return <Mermaid>{String(code)}</Mermaid>;
+  }
+
   const matches = className?.match(/language-(?<language>[a-zA-Z]+)/);
   const language = matches?.groups?.language ?? '';
 
--- a/apps/site/components/MDX/Mermaid/index.tsx
+++ b/apps/site/components/MDX/Mermaid/index.tsx
@@ -1,0 +1,22 @@
+'use client';
+
+import Mermaid from '@node-core/ui-components/MDX/Mermaid';
+import { useTheme } from 'next-themes';
+
+import type { FC } from 'react';
+
+/**
+ * Site-level wrapper that maps the next-themes resolved theme onto the
+ * shared Mermaid component from @node-core/ui-components.
+ */
+const MermaidDiagram: FC<{ children: string }> = ({ children }) => {
+  const { resolvedTheme } = useTheme();
+
+  return (
+    <Mermaid theme={resolvedTheme === 'dark' ? 'dark' : 'light'}>
+      {children}
+    </Mermaid>
+  );
+};
+
+export default MermaidDiagram;
--- a/apps/site/mdx/plugins.mjs
+++ b/apps/site/mdx/plugins.mjs
@@ -4,6 +4,7 @@ import { shikiOptions } from '#platform/shiki.mjs';
 import rehypeShikiji from '@node-core/rehype-shiki/plugin';
 import remarkHeadings from '@vcarl/remark-headings';
 import rehypeAutolinkHeadings from 'rehype-autolink-headings';
+import rehypeMermaid from 'rehype-mermaid';
 import rehypeSlug from 'rehype-slug';
 import remarkGfm from 'remark-gfm';
 import readingTime from 'remark-reading-time';
@@ -21,6 +22,9 @@ export const rehypePlugins = [
   rehypeSlug,
   // Automatically add anchor links to headings (H1, ...)
   [rehypeAutolinkHeadings, { behavior: 'wrap' }],
+  // Transforms ```mermaid code blocks into renderable diagrams;
+  // must run before Shiki so they are not highlighted as plain code
+  [rehypeMermaid, { strategy: 'pre-mermaid' }],
   // Transforms sequential code elements into code tabs and
   // adds our syntax highlighter (Shikiji) to Codeboxes
   () => singletonShiki,
--- a/apps/site/package.json
+++ b/apps/site/package.json
@@ -54,6 +54,7 @@
     "react-dom": "^19.2.8",
     "reading-time": "~1.5.0",
     "rehype-autolink-headings": "~7.1.0",
+    "rehype-mermaid": "^3.0.0",
     "rehype-slug": "~6.0.0",
     "remark-gfm": "~4.0.1",
     "remark-reading-time": "~2.1.0",
--- a/packages/ui-components/package.json
+++ b/packages/ui-components/package.json
@@ -50,6 +50,7 @@
   },
   "dependencies": {
     "@heroicons/react": "^2.2.0",
+    "@mermanjs/web": "^0.7.0",
     "@orama/orama": "^3.1.18",
     "@orama/ui": "^1.5.4",
     "@radix-ui/react-avatar": "^1.2.3",
--- a/packages/ui-components/src/MDX/Mermaid/index.module.css
+++ b/packages/ui-components/src/MDX/Mermaid/index.module.css
@@ -1,0 +1,17 @@
+.mermaid {
+  display: flex;
+  justify-content: center;
+  margin: 1rem 0;
+}
+
+.mermaid svg {
+  max-width: 100%;
+  height: auto;
+}
+
+.fallback {
+  overflow-x: auto;
+  padding: 1rem;
+  border-radius: 0.5rem;
+  background: rgb(0 0 0 / 0.05);
+}
--- a/packages/ui-components/src/MDX/Mermaid/index.stories.tsx
+++ b/packages/ui-components/src/MDX/Mermaid/index.stories.tsx
@@ -1,0 +1,39 @@
+
+import type { Meta as MetaObj, StoryObj } from '@storybook/react-webpack5';
+import type { ComponentProps } from 'react';
+
+import Mermaid from './index';
+
+type Story = StoryObj<typeof Mermaid>;
+type Meta = MetaObj<typeof Mermaid>;
+
+const defaultArgs: ComponentProps<typeof Mermaid> = {
+  children: 'graph TD;\n  A[Start] --> B[Process];\n  B --> C[End];',
+};
+
+export const Flowchart: Story = {};
+
+export const SequenceDiagram: Story = {
+  args: {
+    children:
+      'sequenceDiagram\n  Alice->>Bob: Hello Bob\n  Bob-->>Alice: Hi Alice',
+  },
+};
+
+export const Dark: Story = {
+  args: {
+    theme: 'dark',
+  },
+};
+
+export const InvalidSource: Story = {
+  args: {
+    children: 'this is not a diagram',
+  },
+};
+
+export default {
+  title: 'MDX/Mermaid',
+  component: Mermaid,
+  args: defaultArgs,
+} as Meta;
--- a/packages/ui-components/src/MDX/Mermaid/index.tsx
+++ b/packages/ui-components/src/MDX/Mermaid/index.tsx
@@ -1,0 +1,35 @@
+'use client';
+
+import type { FC } from 'react';
+
+import useMermaid from '../../hooks/useMermaid';
+
+import styles from './index.module.css';
+
+
+type MermaidProps = {
+  /** The Mermaid diagram source. */
+  children: string;
+  /** The host appearance used for the rendered diagram. */
+  theme?: 'light' | 'dark';
+};
+
+const Mermaid: FC<MermaidProps> = ({ children, theme = 'light' }) => {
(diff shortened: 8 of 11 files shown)


ashrees:
Addressed the Copilot review on the Mermaid component (head `bc37755`):
- the dynamic `import('mermaid')` now sits inside the `try` block, so a failed chunk load falls back gracefully
- `securityLevel: 'strict'` is set on `mermaid.initialize`
- `bindFunctions` from `mermaid.render()` is invoked, enabling diagram interactions
- the error fallback now renders the source inside a `<pre>` to preserve whitespace

apps/site/mdx/plugins.mjs
```diff
@@ -4,6 +4,7 @@ import { shikiOptions } from '#platform/shiki.mjs';
 import rehypeShikiji from '@node-core/rehype-shiki/plugin';
 import remarkHeadings from '@vcarl/remark-headings';
 import rehypeAutolinkHeadings from 'rehype-autolink-headings';
+import rehypeMermaid from 'rehype-mermaid';
```
avivkeller, line 7: Doesn't this dramatically slow down builds since it requires initializing a whole browser instance?

ashrees: No — with `strategy: 'pre-mermaid'` no browser is ever started at build time. Verified in the dependency source:

- `createMermaidRenderer()` only creates a **lazy** `browserPromise` (`browserPromise ||= getBrowser(...)` inside the returned render function in `mermaid-isomorphic`) — nothing launches at plugin setup.
- `rehype-mermaid` never calls that render function for `pre-mermaid`; it only rewrites the AST (`<pre><code class="language-mermaid">` → `<pre class="mermaid">`). The package comments this exact path as "No need to start a browser in this case."
- Playwright (peer dep) is only exercised by the `inline-svg` / `img-*` strategies.

So build cost is a single AST walk per document, and rendering happens client-side — with the `mermaid` library itself lazy-loaded via dynamic `import()`, so it stays out of the initial bundle too.

Happy to switch to `inline-svg` if the team prefers zero client-side JS — that's the trade-off (build-time Chromium vs. client rendering).

apps/site/mdx/plugins.mjs
```diff
@@ -21,6 +22,9 @@ export const rehypePlugins = [
   rehypeSlug,
   // Automatically add anchor links to headings (H1, ...)
   [rehypeAutolinkHeadings, { behavior: 'wrap' }],
+  // Transforms ```mermaid code blocks into renderable diagrams;
+  // must run before Shiki so they are not highlighted as plain code
+  [rehypeMermaid, { strategy: 'pre-mermaid' }],
```
ovflowd, line 27: If we're using rehype-mermaid, do we need to add mermaid support on code boxes? Or what exactly this rehype plugin does?

apps/site/mdx/plugins.mjs
ovflowd: Can you do some benchmarks and compare building site previously and after? + adding an example storybook with mermaid content so we can render/test?

apps/site/components/MDX/Mermaid/index.tsx
```diff
@@ -0,0 +1,63 @@
...
+type MermaidProps = {
+  children: string;
+};
+
+const Mermaid: FC<MermaidProps> = ({ children }) => {
+  const containerRef = useRef<HTMLDivElement>(null);
+  const { resolvedTheme } = useTheme();
+  const reactId = useId().replace(/:/g, '-');
+
+  useEffect(() => {
```
ovflowd: So we're doing mermaid rendering solely on the client-side? Is this what we'd like to have, @avivkeller?

BTW, @ashrees could you extract this effect into a dedicated hook, so we can unit test the Mermaid specific API within an unit testable hook?

avivkeller: I saw some people made browserless versions of Mermaid, e.g.:
https://github.com/1jehuang/mermaid-rs-renderer

avivkeller: I can write bindings for it?

ovflowd: Define bindings? 👀

avivkeller: Looks like they already exist at https://www.npmjs.com/package/@mermanjs/web

ovflowd: That looks nice, @ashrees can you update your PR?

ashrees:
Updated the PR per the review (head `99db225`):

**@ovflowd's asks:**
1. **Dedicated, unit-testable hook** — the effect now lives in `useMermaid` (`packages/ui-components/src/hooks/useMermaid.ts`) with an injectable `loader` for test stubbing. Unit tests in `__test__/useMermaid.test.jsx` (node:test + testing-library): **2/2 pass**.
2. **Storybook story** — `packages/ui-components/src/MDX/Mermaid/index.stories.tsx` with flowchart, sequence, dark-theme, and invalid-source variants.
3. **Benchmarks** — compiled 25 real blog posts through the production rehype chain (slug → autolink → shiki), with vs. without the plugin, 3 warmed runs: **5.62ms/doc vs 5.19ms/doc** — no measurable build-time impact (delta within run variance), because `pre-mermaid` is AST-only and never starts a browser.

**Renderer switch (@avivkeller):** rendering now uses **`@mermanjs/web`** (Merman WASM) instead of `mermaid.js` — so even the client-side path drops the ~500KB JS dependency. Theme-aware (`light`/`dark` via `next-themes` in the site wrapper), invalid sources fall back to a `<pre>` with the raw source.

**What the rehype plugin does (clarifying question):** it only rewrites ```` ```mermaid ```` fenced blocks into `<pre class="mermaid">` at compile time; the `MDXCodeBox` override then routes those to the Mermaid component. Code boxes are otherwise untouched — no separate codebox support needed.

The component lives in `@node-core/ui-components` (alongside the other MDX components) with a thin `next-themes` wrapper in `apps/site`.

apps/site/mdx/plugins.mjs
ashrees: it's pretty minimal — the plugin just rewrites ```mermaid fenced blocks into <pre class="mermaid"> at compile time. MDXCodeBox routes those to the mermaid component instead of a code box, and regular code blocks are untouched. so no, no extra codebox support needed.

apps/site/mdx/plugins.mjs
ashrees: both done. stories are in packages/ui-components/src/MDX/Mermaid/index.stories.tsx (flowchart, sequence, dark, invalid-source). for the benchmark i compiled 25 real blog posts through the production chain, with vs without the plugin, 3 warmed runs — 5.62 vs 5.19 ms/doc, so no measurable build cost (pre-mermaid is ast-only, no browser involved).

apps/site/components/MDX/Mermaid/index.tsx
ashrees: yes, client-side only. the effect is extracted into useMermaid now (packages/ui-components/src/hooks/useMermaid.ts) with an injectable loader so tests can stub the wasm renderer — unit tests live in __test__/useMermaid.test.jsx, 2/2 passing locally. if you'd rather have build-time rendering (inline-svg) instead, happy to explore that.

ashrees: done — rendering now goes through @mermanjs/web (wasm), so the 500kb mermaid.js is out of the client bundle entirely.

avivkeller (changes requested):
Can't we make a remark plugin to do the transform?

avivkeller:
@ashrees Please familiarize yourself with Node.js's AI Policy, and review the code you write. Your comments do not match the actual code, which makes it seem as though you don't understand what you've written. If that's the case, it's okay, but be transparent.

ashrees:
i've updated the pr body so it matches the current implementation exactly. the mismatch was on me. the pr went through several fast iterations (copilot review, the mermaid.js to wasm switch, then the hook extraction) and i didn't keep the description and comments in sync with the branch.

on the policy point: yes, i used ai assistance while drafting, and i've read the ai policy. i take full responsibility for every line. i reviewed and understand the code, verified it locally (unit tests 2/2, lint clean, benchmarks in the thread above), and i'm glad to walk through any part of it in review.

apps/site/mdx/plugins.mjs
ovflowd: Could you please provide a table with the benchmark results?

apps/site/mdx/plugins.mjs
ovflowd: So why we need a mermaid UI component?

apps/site/components/MDX/Mermaid/index.tsx
ovflowd: So, is the UI component piece still needed for @mermanjs/web? Why we need rehype mermaid and this at the same time?

ovflowd:
@ashrees several CI steps are failing, please give them an eye and fix it.

apps/site/components/MDX/CodeBox/index.tsx
```diff
@@ -8,6 +9,12 @@ const MDXCodeBox: FC<HTMLAttributes<HTMLElement>> = ({
   children: code,
   className,
 }) => {
+  // Mermaid diagrams arrive as `<pre class="mermaid">` (see rehype-mermaid),
+  // so we render them as diagrams instead of code boxes
+  if (className?.split(' ').includes('mermaid')) {
+    return <Mermaid>{String(code)}</Mermaid>;
```
ovflowd, line 15: I don't think our codebox component should have anything to do with Mermaid. If Mermaid renders a `pre` then it shouldn't. It is ambiguous that we have two transformations and I wonder if this is the right path. As asked on other comments could you describe _why_

- We need a rehype plugin that transforms the mermaid markdown code snippets into a pre and then another UI component that then grabs that `pre` element and renders as mermaid? Wouldn't it be more efficient if only one pass is done? Is that possible? Or at the very least it'd be good to remove the dependency of the Mermaid component within the MDX CodeBox component. Point being, that the CodeBox component should have no logic that goes to Mermaid path. And I believe that can be easily fixed on our remark package within this monorepo, so that you early on catch that this is a Mermaid and easily point to the right component as we do for code boxes 🤔

apps/site/components/MDX/Mermaid/index.tsx
ashrees: you're right, the current shape is redundant. it grew that way through the review rounds: the rehype plugin rewrites the fenced block into a pre, and then the component catches that pre and renders it. two passes where one would do. i can collapse it into a single remark transform that emits the Mermaid component directly from the fenced block, which drops the pre step entirely.

apps/site/mdx/plugins.mjs
ashrees: | run set | avg | per doc |
| --- | --- | --- |
| without the plugin | 140ms (3 runs) | 5.62ms |
| with pre-mermaid | 130ms (3 runs) | 5.19ms |

25 real blog posts per run through the production chain (slug, autolink, shiki), warmed shiki singleton, 3 runs each. delta is within run variance, so no measurable build cost.

apps/site/mdx/plugins.mjs
ashrees: the component is the host that runs the renderer on the client. the wasm module needs a dom node to render into, plus theme handling and an error fallback, and the component owns those. it is the piece i want to keep. what should go away is the pre transformation in front of it, which is the part that grew weird.

apps/site/components/MDX/CodeBox/index.tsx
ashrees: agreed on all of it. codebox should not know mermaid exists, and i would not defend the two-transformation flow as a design. it accreted through the review rounds. plan: a small remark plugin in the monorepo that catches ```mermaid nodes early and emits the Mermaid component directly, registered in the mdx component map, with the rehype-mermaid dependency removed and codebox reverted to its original state. single pass, no coupling. that also answers aviv's remark suggestion.

PR #9142, opened 2026-09-04
fix: correct the code style in the hover status

btea:
## Description
sync https://github.com/nodejs/doc-kit/pull/1080

### Check List

### DIFF
--- a/packages/ui-components/.storybook/preview.tsx
+++ b/packages/ui-components/.storybook/preview.tsx
@@ -16,7 +16,7 @@ const preview: Preview = {
 
   decorators: [
     withThemeByDataAttribute<ReactRenderer>({
-      themes: { light: '', dark: 'dark' },
+      themes: { light: 'light', dark: 'dark' },
       defaultTheme: 'light',
       attributeName: 'data-theme',
     }),
--- a/packages/ui-components/src/Common/ChangeHistory/index.module.css
+++ b/packages/ui-components/src/Common/ChangeHistory/index.module.css
@@ -68,9 +68,8 @@
       text-white!;
   }
 
-  :where([data-theme='light'], [data-theme='light'] *) &:hover code {
-    color: var(--color-brand-900);
-  }
+:where([data-theme='light'], [data-theme='light'] *) &:is(:hover, :focus-visible) code {
+  color: var(--color-brand-900);
 }
 
 .dropdownLabel {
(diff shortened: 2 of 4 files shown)


packages/ui-components/src/Common/ChangeHistory/index.module.css
```diff
@@ -67,6 +67,9 @@
     @apply bg-brand-600
       text-white!;
   }
+
+:where([data-theme='light'], [data-theme='light'] *) &:is(:hover, :focus-visible) code {
+  color: var(--color-brand-900);
```
AugustinMauroy: use `@apply`

AugustinMauroy (changes requested):
Could you add a story that allow us to see that this issue is fixed

btea:
> Could you add a story that allow us to see that this issue is fixed

I added a brief note.

AugustinMauroy:
oh nice but I mean a story of storybook you have to update this file to have code in children https://github.com/nodejs/nodejs.org/blob/main/packages/ui-components/src/Common/ChangeHistory/index.stories.tsx

we use chromatic to automatically deploy storybook like vercel do for website

btea:
Thank you for your detailed explanation. 👍 
I will update it.

packages/ui-components/.storybook/preview.tsx
```diff
@@ -16,7 +16,7 @@ const preview: Preview = {
 
   decorators: [
     withThemeByDataAttribute<ReactRenderer>({
-      themes: { light: '', dark: 'dark' },
+      themes: { light: 'light', dark: 'dark' },
```
btea, line 19: Add specific attribute values ​​to ensure that the newly added class can be matched in the storybook.

ovflowd (approved):

packages/ui-components/src/Common/ChangeHistory/index.stories.tsx
```diff
@@ -107,9 +112,13 @@ const LARGE_SAMPLE_CHANGES = [
 
 export const Default: Story = {
   render: args => (
-    <div className="right-0 flex justify-end">
-      <ChangeHistory {...args} />
-    </div>
+    // `main` reproduces the docs page context: markdown.css scopes inline
+    // `code` styles under `main`, which is what the hover fix interacts with /ui-components/src/styles/markdown.css:85
```
mikeesto: Could we avoid the :85 line reference here? It may get stale as the file changes

btea: That's true (I previously thought that specifying a certain line was convenient for quick location); let me remove it.

avivkeller (approved):

AugustinMauroy (approved):

PR #9144, opened 2026-09-04
feat(ui): add css :target deeplinks to CodeTabs

### LINKED ISSUE #9140: feat(ui): add anchor linking support to CodeTabs component
The CodeTabs component lacks deep linking support. When users click on a search result that targets a specific code tab, the page jumps to the top instead of the exact tab.

Adding deep linking to the tabs will allow direct navigation to specific tab using the URL as source of truth.

joebasrawi:
Implements deep linking for `CodeTabs` with pure HTML/CSS, following [@ovflowd](https://github.com/ovflowd)’s direction on #9140. Radix Tabs are no longer used for this component; tab switching is fragment links and `:target`.

Clicking a tab is an `<a href="#id">`, so the URL updates natively. Loading `/path#id` selects that tab via CSS. No `hashchange` listeners.

### How `:target` works

- Each tab label is an `<a id="…" href="#…">`.
- CSS `:target` styles the active trigger.
- The matching panel is shown with `:has(.trigger:nth-child(n):target)`.
- If this `CodeTabs` instance does not contain the current fragment (no hash, unmatched hash, or another group’s tab), `[data-default]` (first tab, or `defaultTab` / `defaultValue`) is shown.

### ID scheme

| Situation | Fragment |
| --- | --- |
| Optional `groupId` | `{slug(groupId)}-{slug(tabKey)}` e.g. `install-js-0` |
| Default (no `groupId`) | `codetabs-{useId}-{slug(tabKey)}` e.g. `codetabs-r1-js-0` |

MDX already uses `tabKey` = `{language}-{index}`. `useId` makes multiple CodeTabs on one page unique. Authors can pass `groupId` for stable permalinks.

`Common/Tabs` is unchanged and still uses Radix.

## Validation

- `@node-core/ui-components` unit tests: initial `#id` selects the right tab, clicks update `location.hash`, hash navigation between tabs, unmatched hash keeps the default, two CodeTabs on one page do not collide.
- `pnpm --filter @node-core/ui-components test:unit` passing (60 tests).
- Package `lint:js`, `lint:css`, and `lint:types` passing.

Fixes #9140

### Check List

- [x] I have run the relevant package tests (`@node-core/ui-components` unit tests).

### DIFF
--- a/.changeset/codetabs-anchor-deeplinks.md
+++ b/.changeset/codetabs-anchor-deeplinks.md
@@ -1,0 +1,5 @@
+---
+'@node-core/ui-components': minor
+---
+
+Add HTML/CSS `:target` deep linking to CodeTabs and replace Radix Tabs for that component so fragments work without JavaScript hash listeners.
--- a/.changeset/codetabs-anchor-deeplinks.md
+++ b/.changeset/codetabs-anchor-deeplinks.md
@@ -1,5 +1,9 @@
 ---
-'@node-core/ui-components': minor
+'@node-core/ui-components': major
 ---
 
-Add HTML/CSS `:target` deep linking to CodeTabs and replace Radix Tabs for that component so fragments work without JavaScript hash listeners.
+Add URL-fragment deep links to CodeTabs. CSS selects the visible panel without JavaScript; a client enhancement keeps keyboard navigation and ARIA state in sync with the fragment.
+
+CodeTabs now expects one raw child per tab, in tab order. Replace Radix `Tabs.Content` children with their contents. Arrays and fragments are supported; components that internally render multiple panels must be expanded at the call site. This replaces the previous Radix context and is a breaking change for direct CodeTabs consumers. The MDX wrapper remains compatible.
+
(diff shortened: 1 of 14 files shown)


avivkeller (changes requested):
This is an extremely complex, likely AI-generated fix for an almost nonexistent issue. If there is a valid resolution to the issue, it should be as simple as a `<a/>` anchor component, or similar, and not rely on a bunch of JavaScript majorly changing our current implementation.

Please note Node.js's AI Policy.

ovflowd:
> This is an extremely complex, likely AI-generated fix for an almost nonexistent issue. If there is a valid resolution to the issue, it should be as simple as a `<a/>` anchor component, or similar, and not rely on a bunch of JavaScript majorly changing our current implementation.
> 
> Please note Node.js's AI Policy.

Yeah, I mean we appreciate the effort, but the PR contents differ from the title, anyways this is not an easy solution for a newcomer unfortunately, now that we're considering the path of simplifying the component as described on https://github.com/nodejs/nodejs.org/issues/9140.

joebasrawi:
Thanks @avivkeller and @ovflowd for the review.

Agreed this went past a simple `<a/>` deep link and shouldn't sit as a major CodeTabs rewrite. Closing this PR so #9140 can follow the simpler component path you described. Happy to help on a focused follow-up once that direction is settled.

PR #9154, opened 2026-09-09
fix(ui): keep active sidebar item visible on navigation

### LINKED ISSUE #8828: Sidebar scroll position resets on navigation, active item can be off-screen
I couldn't push my own branch to make a PR so sharing this as issue instead. The fix belongs in `components/Sidebar/index.jsx`.

**Problem**
Clicking an item in the left sidebar causes a full page reload because navigation uses window.location.href in components/Sidebar/index.jsx. After the reload, the sidebar’s internal scroll position resets to the top.

As a result, when a user clicks a page located lower in the sidebar list, such as an item under TypeScript or Diagnostics, the selected item may still be active but no longer visible. The user then has to scroll through the sidebar again to find their current location.

**Reproduction steps**

- Open https://nodejs.org/learn
- Scroll the left sidebar to a lower item, for example Diagnostics → Memory
- Click the item
- After the page reloads, the sidebar scroll position has returned to the top, and the active item is off-screen

**Expected behavior**

The active sidebar item should remain visible after navigation.

**Proposed fix**

When the sidebar mounts, locate the active link inside the sidebar `<aside>`. 
If that link is outside the visible portion of the sidebar’s own scroll container, adjust `aside.scrollTop` so the active item is centered in view. This only affects the sidebar’s internal scroll position and does not change the main window scroll.

**Example approach**

```js
useEffect(() => {
  const aside = asideRef.current;
  const active = aside?.querySelector(`a[href="${CSS.escape(pathname)}"]`);
  if (!aside || !active) return;

  const offsetTop =
    active.getBoundingClientRect().top -
    aside.getBoundingClientRect().top +
    aside.scrollTop;

  const viewTop = aside.scrollTop;
  const viewBottom = viewTop + aside.clientHeight;

  if (offsetTop >= viewTop && offsetTop + active.offsetHeight <= viewBottom) {
    return;
  }

  aside.scrollTop = Math.max(
    0,
    offsetTop - aside.clientHeight / 2 + active.offsetHeight / 2
  );
}, [pathname]);
```

**Notes**

This approach uses the existing forwardRef exposed by @node-core/ui-components/Containers/Sidebar, so no upstream or vendor changes are required.

**Alternative considered**

Persisting the sidebar’s scrollTop in sessionStorage across reloads was considered, but rejected. That approach adds more complexity, such as saving on unload, restoring at the right time during hydration, and handling stale scroll state when the sidebar structure changes.

Ensuring the active item is visible is simpler and aligns with expected user behavior in most cases.

codebyanas:
## Description

Fixes #8828.

The sidebar's internal scroll position could leave the active navigation item outside the visible area after navigation. This change ensures that the active sidebar link is automatically scrolled into view when the pathname changes.

The implementation:

- Finds the active sidebar link after navigation.
- Checks whether it is inside the sidebar viewport.
- Adjusts only the sidebar's `scrollTop` when the item is off-screen.
- Keeps the existing sidebar scroll-position restoration behavior.
- Supports locale-prefixed routes such as `/en/about/...`.

## Testing

Tested locally by navigating between sidebar pages and verifying that the active item remains visible after navigation.

### DIFF
--- a/apps/site/components/withSidebar.tsx
+++ b/apps/site/components/withSidebar.tsx
@@ -1,7 +1,7 @@
 'use client';
 
 import Sidebar from '@node-core/ui-components/Containers/Sidebar';
-import { useTranslations } from 'next-intl';
+import { useLocale, useTranslations } from 'next-intl';
 import { useEffect, useRef } from 'react';
 
 import Link from '#site/components/Link';
@@ -43,6 +43,7 @@ const mapItem = ([, item]: [string, MappedItem]): SidebarMappedEntry => ({
 const WithSidebar: FC<WithSidebarProps> = ({ navKeys, context, ...props }) => {
   const { getSideNavigation } = useSiteNavigation();
   const pathname = usePathname()!;
+  const locale = useLocale();
   const t = useTranslations();
   const { push } = useRouter();
   const { frontmatter } = useClientContext();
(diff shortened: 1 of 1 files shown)


apps/site/components/withSidebar.tsx
```diff
@@ -52,15 +53,46 @@ const WithSidebar: FC<WithSidebarProps> = ({ navKeys, context, ...props }) => {
...
+
+      const asideBounds = aside.getBoundingClientRect();
+      const activeBounds = active.getBoundingClientRect();
+
+      const offsetTop = activeBounds.top - asideBounds.top + aside.scrollTop;
+      const viewTop = aside.scrollTop;
+      const viewBottom = viewTop + aside.clientHeight;
+
+      if (offsetTop < viewTop || offsetTop + active.offsetHeight > viewBottom) {
+        aside.scrollTop = Math.max(
```
Botato300: Correct me if I'm wrong, but you can use [scrollIntoView](https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollIntoView) here. Just make sure to pass `{ block: 'center'}` as an argument to prevent scrolling in parent containers.

That way, you avoid all those calculations you did!

PR #9157, opened 2026-09-12
fix(ui): keep badge tooltips visible inside scrollable containers

### LINKED ISSUE #938: The tooltips for methods or attributes in the sidebar are not fully displayed.

Image 

Image

As shown in the figure, when the mouse hovers over the `E` mark, the floating information content is covered and not fully displayed.

Similarly, a similar issue occurs when the `E` tag is aligned to the right. It appears to be a problem with the element's layout hierarchy style.

I've encountered a similar issue before; it's likely caused by the container element having a style that hides beyond its horizontal width. It seems like you need to adjust the style or set the tooltip to dynamically adjust its position, moving it closer to the center of the content area.

btea:
## Description
close https://github.com/nodejs/doc-kit/issues/938

### Check List

### DIFF
--- a/packages/ui-components/src/Common/Badge/index.module.css
+++ b/packages/ui-components/src/Common/Badge/index.module.css
@@ -54,3 +54,84 @@
     }
   }
 }
+
+/*
+ * Real tooltip element rendered as the next sibling of badges carrying
+ * `data-tooltip`. It stays hidden unless the browser supports CSS anchor
+ * positioning, in which case it replaces the global `[data-tooltip]`
+ * pseudo-element tooltip so it cannot be clipped by scrolling ancestors
+ * (see nodejs/doc-kit#938).
+ */
+.tooltip {
+  @apply pointer-events-none
+    z-10
+    w-max
+    max-w-64
+    rounded-md
+    border
+    border-neutral-200
+    bg-white
+    px-2.5
+    py-1.5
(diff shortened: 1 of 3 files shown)


ovflowd (changes requested):
I'm unsure if the best approach here is a dedicated HTML element, I worry this will add many many many more HTML elements to this already very inflatted DOM (thinking of doc-kit and API docs)

I I'm -1 for this, I do believe the solution can be CSS-only; Also the usage of "@supports (anchor-name: --doc-kit-badge-tooltip) {" with a `.tooltip` using raw CSS instead of Tailwind also is against our styling guidelines 👀

btea:
Thanks for the review!

As far as I can tell, this is the only solution that meets the requirements while staying at the HTML/CSS level. On my testing, pseudo-elements alone can't satisfy them in a pure-CSS approach: a pseudo-element can't resolve `position-anchor` against its own badge in Chromium (and shared `anchor-name` values resolve to the last matching element in the document, not the nearest preceding one), which forces the tooltip to be a dedicated element per badge.

That said, I might be missing a more reasonable approach — if you have one in mind, or know of a tracked Chromium fix for the pseudo-element cases, I'm happy to rework the PR around it.

Also noted on the styling guidelines — I'll move the raw declarations to `@apply` with arbitrary properties.

PR #9159, opened 2026-09-13
feat(ui): implement deep linking in CodeTabs via URL hash

### LINKED ISSUE #9140: feat(ui): add anchor linking support to CodeTabs component

moshams272:
## Description

Implement deep linking in `CodeTabs` component by using the active tab state via the URL hash.

#### In Details

- Add optional `groupId` prop to generate unique anchor IDs across multiple CodeTabs instances on the same page in format: `tab-{groupId}-{language}-{index}`, falling back to React's `useId()` when groupId is not provided.
- Read the URL hash on mount via a lazy initializer in `useState()` to set the correct initial tab without a double render.
- Update the URL silently on tab click using `history.replaceState` to avoid polluting the browser history

## Validation

After building, I checked the tabs by clicking on them, and then the URL hash changed.

## Related Issues

Fixes #9140 

### Check List

- [X] I have run `pnpm format` to ensure the code follows the style guide.
- [X] I have run `pnpm test` to check if all tests are passing.
- [X] I have run `pnpm build` to check if the website builds without errors.


ovflowd (changes requested):
We want to avoid having the CodeTabs being a client-component in all  cases. We could use a wrapper that uses use client specifically for when these groupIds or whatnot are passed, ands we can do the logic of what gets used on shiki I guess? 

That said per the repo guidelines it is an absolute no-go using winodw APIs from within components, not to mention adding so many event listeners (1 per CodeTab) is going to be an absolute nightmare for the browser.

AugustinMauroy:
Capture d’écran 2026-09-13 à 22 40 52

I dunno how to reproduce what I have done but it's may happened

avivkeller (approved):
LGTM minus Claudio's concerns

moshams272:
> We want to avoid having the CodeTabs being a client-component in all cases. We could use a wrapper that uses use client specifically for when these groupIds or whatnot are passed, ands we can do the logic of what gets used on shiki I guess?
> 
> That said per the repo guidelines it is an absolute no-go using winodw APIs from within components, not to mention adding so many event listeners (1 per CodeTab) is going to be an absolute nightmare for the browser.

Got it, I saw the `use client` used in `BaseCodeBox` first. After your comment, I find the pattern you mean used in `AvatarGroup`.

I think Augustin's bug happened because of the **React Hydration Error** in the `useState`.

ovflowd:
> Got it, I saw the use client used in BaseCodeBox first. After your comment, I find the pattern you mean used in AvatarGroup.

We might need to revisit why BaseCodeBox uses "use client" but still we should thrive for reducing client components as much as possible :)

moshams272:
> We might need to revisit why BaseCodeBox uses "use client" but still we should thrive for reducing client components as much as possible :)

I reached for a fair solution. We use `useMediaQuery.ts` as a custom hook and access the `window` through it. So, could I do the same?! Or we should revisit it too ;)

It's an initial thinking for now.

ovflowd:
> > 
> 
> I reached for a fair solution. We use `useMediaQuery.ts` as a custom hook and access the `window` through it. So, could I do the same?! Or we should revisit it too ;)

In terms of should window APIs be behind safe Hooks, yes. But in terms of, should we have god knows how many Event Listeners because they're tied to one CodeBox? No.

You might want to think of a React Context Provider, with one hook that keeps an eye on Hash changes. Or you might defer it to a HoC, so that on the apps, such as Next.js and doc-kit we defer the implementation of this, for example, Next.js has a Router/Navigation API, you could inject results on the CodeBox...

I guess it's more of a thinking of "who should be responsible for listening to this" _you could argue_ it should be the Component itself, because he's the one interested on this, that said, it's only interested on the scenarios where it should even care abou that. IMO, the Component should subscribe to the changes aka ask to the parent component or context provider "has the hash changed?" ... yet that the responsibility of actually setting up the Event Listener shouldn't be part of the Component's responsibility.... As "Window" or even a Hook that "listens to Window" is something unbeknown to the Component. The Component's "world" should always be anything that is inside said Component. That's a very common mistake some React Components do on, for example, Modals. Where they let the Modal itself be the one observing if anyone clicked outside the Modal, whereas that should _never_ be the Modal's responsibility.

So think a bit of how to approach this, both in a way that is properly designed and that can be reused (ie: if this is a prop/callback to the Component, you defer the implementation details (be it window APIs, be it Next.js APIs...) to the consumer) and how to make it reliable and performant.

I hope my explanation helps!

moshams272:
[Screencast from 2026-09-16 02-04-21.webm](https://github.com/user-attachments/assets/514cbe97-3002-493a-9a29-01dc41f68e80)

I hope I did it right :)

AugustinMauroy:
that seem cool but on safari and vercel preview I cannot having them working

moshams272:
> that seem cool but on safari and vercel preview I cannot having them working

Yeah, I'll summarize what I did:

- In our app:
  - Created a hash hook. This hook works on both sides (server & browser). On server or hydration, it'll do nothing that makes the `CodeTabs` trigger the default tab, so they will not dismatch. On browser, it will take the hash, and I just add one event listener on `window`, and created a set to rerender the components manually as the React new hook [useSyncExternalStore](https://react.dev/reference/react/useSyncExternalStore) gives me a callback to trigger.
  - Created a hash provider to use this hook and pass this result to ui-components.
  
- In ui-components:
  - Created a hash context, why not create it in the app? If we did that, we let our library call the context from the app and we need to make it stanalone library. So who will use this feature should give us the hash through our context.
  - Added an optional `groupId` feature in `CodeTabs`. This component works on the server side. However, If you need to use the `groupId`, the heavy work will still be on the server, and then pass the data to a client component to manage the state and use the context and hash here for sure.


So we haven't been added `groupId` to any MDX files yet, that means there is no hash needed and the component will fallback to the server component, that is the same for the old one we use now before this PR.

To actually see it works, a `<CodeTabs>` component needs the new `groupId` prop. I can push a temporary test snippet to `about/index.mdx` -that's what I did locally, BTW- so you can test the Safari & Vercel preview. We can simply revert it right before merging.

PR #9163, opened 2026-09-18
fix(ui-components): add Search storybook examples and support rich hit descriptions

btea:
## Description
image

The description content in the search results of the search box shown in the figure also includes backtick characters. We can change it to support tag rendering to make the display results more intuitive.

### Check List


AugustinMauroy (approved):
seem good could you provide a screenshot of storybook ? cc @nodejs/web-infra why chromatic didn't work ?

aymen94 (approved):

btea:
> seem good could you provide a screenshot of storybook ?

image

The screenshot is shown above. However, the code here is not nested within the main tag, so the background color style does not match. Even if it did match, the background color of the code would be consistent with the background color of the item.

ovflowd (approved):

avivkeller (approved):

